Issue #51, opened 2025-06-02
feat: Implement `trip-for-vehicle` endpoint


Ahmedhossamdev:
Implement support for the `trip-for-vehicle` endpoint as defined in the OneBusAway API.

📄 **Docs:**  
[https://developer.onebusaway.org/api/where/methods/trip-for-vehicle](https://developer.onebusaway.org/api/where/methods/trip-for-vehicle)

## ✅ Acceptance Criteria

- [x] Implement the `trip-for-vehicle` endpoint in **Maglev**
- [x] Ensure the **response matches** the Java production server’s output
- [x] Add **unit tests** to validate the correctness of the response and prevent regressions

PR #79, opened 2025-07-14
Feat/trip-for-vehicle

Ahmedhossamdev:
Fix: #51


internal/gtfs/gtfs_manager.go
```diff
@@ -273,6 +273,20 @@ func (manager *Manager) GetVehicleForTrip(tripID string) *gtfs.Vehicle {
 	return nil
 }
 
+func (manager *Manager) GetVehicleByID(vehicleID string) (*gtfs.Vehicle, error) {
+
+	manager.realTimeMutex.RLock()
+	defer manager.realTimeMutex.RUnlock()
+
+	for _, v := range manager.realTimeVehicles {
```
aaronbrethorst, line 281: not right now, but at some point we should look at how to build a map out of the realtime vehicles list so we can optimize this from O(n) to O(1).

aaronbrethorst (approved):
looks good!

aaronbrethorst (changes requested):
uh oh, lint and tests are failing now

internal/gtfs/gtfs_manager.go
Ahmedhossamdev: I agree with you <3

Ahmedhossamdev: Added it to my todo list.

aaronbrethorst (approved):
great work 🚢

PR #80, opened 2025-07-18
Update GTFS package import path to OneBusAway/go-gtfs

amrhossamdev:
Migrated the go-gtfs parser to use our fork to support interpolating missing arrival and departure stop times


aaronbrethorst (approved):
Looks good!

PR #110, opened 2025-09-04
refactor: add explicit check for agency existence before processing shape requests

Ahmedhossamdev:
Fix: #102


aaronbrethorst (approved):
👍

PR #119, opened 2025-09-19
Feature/stop direction support 111

### LINKED ISSUE #111: Add Support for Stop Direction Field (W, S, E, D)
Currently, the codebase contains many `TODO`s related to implementing stop directions (e.g., N, E, S, W).

- [ ]  Implement this functionality.
- [ ] Write tests for it

nakulbh:
- [x] Implement this functionality.
- [x] Write tests for it

Fixes #111


internal/gtfs/direction_calculator.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func NewDirectionCalculator(queries *gtfsdb.Queries) *DirectionCalculator {
+	return &DirectionCalculator{
+		queries: queries,
+	}
+}
+
+// CalculateStopDirection determines the compass direction for a stop
+func (dc *DirectionCalculator) CalculateStopDirection(ctx context.Context, stopID string) string {
```
aaronbrethorst, line 20: not right now, but down the road we should add support for caching these stop direction calculations.

internal/gtfs/direction_calculator.go
```diff
@@ -0,0 +1,131 @@
...
+
+	// No direction could be calculated
+	return ""
+}
+
+func (dc *DirectionCalculator) calculateFromShape(ctx context.Context, stopID string) string {
+	// Get trips serving this stop
+	stopTrips, err := dc.queries.GetStopsWithTripContext(ctx, stopID)
+	if err != nil || len(stopTrips) == 0 {
+		return ""
```
aaronbrethorst, line 39: This should probably be `"UNKNOWN"`, as should any other cases where we return early on an error. (e.g. in `calculateFromNextStop()`. Alternatively, these functions that can produce errors should return a string `(string, error)` tuple. I don't actually feel strongly about this one way or another, as long as we're consistent and the API output always reads "UNKNOWN"

internal/restapi/stops_for_route_handler.go
```diff
@@ -315,3 +315,16 @@ func formatStopIDs(agencyID string, stops map[string]bool) []string {
...
 }
+
+func (api *RestAPI) calculateStopDirection(ctx context.Context, stopID string) string {
+	if api.DirectionCalculator == nil {
+		return "Direction" // Fallback for tests
+	}
+
+	direction := api.DirectionCalculator.CalculateStopDirection(ctx, stopID)
+	if direction == "" {
+		return "Direction" // Default when calculation fails
```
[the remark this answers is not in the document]
aaronbrethorst: I'd like to see all of these new, hardcoded "UNKNOWN" values replaced with a single string constant to help make sure that we don't ever accidentally introduce a typo in the string.

aaronbrethorst (changes requested):
this is looking really good! there are a couple minor things I'd like to see changed before merging, but I think with one round of polishing this will be ready to merge! Please don't worry about the caching right now.

aaronbrethorst (approved):
🔥 looks great!

PR #130, opened 2025-11-07
feat: add GetStopForAgency query and update handlers to retrieve stop IDs for a specific agency

Ahmedhossamdev:
Fix: #106


aaronbrethorst (approved):
nice! lgtm

Issue #134, opened 2025-11-20
Integrate gtfstidy GTFS validation/sanitization tool


Ahmedhossamdev:
While comparing Maglev output with the Java server, I noticed that the Java codebase uses **gtfstidy** to validate, sanitize, and minimize GTFS feeds before loading them. This likely explains some differences in stop results and other GTFS-derived data.

We should integrate `gtfstidy` into our GTFS processing pipeline to ensure both systems produce consistent results.

### Tasks
- [ ] Add `gtfstidy` as a preprocessing step before loading GTFS bundles.
- [ ] Run validation + sanitization automatically during GTFS import.
- [ ] Compare sanitized output with Java to ensure parity.
- [ ] Add configuration flag to enable/disable GTFS tidy step.

PR #141, opened 2025-11-26
Feat/gtfs-tidy

Ahmedhossamdev:
Fix: #134

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -50,7 +50,8 @@ Example `config.json`:
   "gtfs-static-feed": {
     "url": "https://example.com/gtfs.zip",
     "auth-header-name": "Authorization",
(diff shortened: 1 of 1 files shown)


aaronbrethorst (changes requested):
Very cool! I appreciate that you make this feature an optional dependency on a binary, instead of a dependency on the source code since gtfstidy uses the viral GPL 2 license.

I'd like to see a note added to the README about the gtfs tidy flag in `example config.json` (https://github.com/OneBusAway/maglev/?tab=readme-ov-file#json-configuration-file), but otherwise I think this is ready to go!

aaronbrethorst (approved):
lgtm! 👍

Issue #144, opened 2025-11-28
Goroutine Leak in RateLimitMiddleware


Bhup-GitHUB:
# 🛠️ Rate Limiter Cleanup Goroutine Never Stops During Graceful Shutdown

## ❗ Problem Summary

The `RateLimitMiddleware` starts a background cleanup goroutine inside `NewRateLimitMiddleware()` (line **49** in `internal/restapi/rate_limit_middleware.go`). However, this goroutine **never stops**, even during a graceful shutdown, causing a **goroutine leak**.

---

### 1. Goroutine Starts but Never Ends

```go
// internal/restapi/rate_limit_middleware.go:49
go middleware.cleanup() // Runs forever
```

This launches a cleanup loop that continues running even after the server shuts down.

---

### 2. A Stop Method Exists (But Is Never Called)

```go
// internal/restapi/rate_limit_middleware.go:174-179
func (rl *RateLimitMiddleware) Stop() {
    if rl.cleanupTick != nil {
        rl.cleanupTick.Stop()
    }
}
```

The cleanup ticker can be stopped — but nothing ever calls this method.

---

### 3. Root Cause

The `RestAPI` struct only stores middleware *handlers*, not the actual `RateLimitMiddleware` instance.

As a result, `cmd/api/app.go` has **no reference** to the middleware, so it cannot call `Stop()` during shutdown.

This prevents graceful cleanup of the goroutine.

---

## 🚨 Impact

* Goroutine leaks
* Cleanup ticker continues running after shutdown
* Incomplete graceful shutdown
* Unnecessary resource usage

---

## ✅ Proposed Fix

1. Modify `RestAPI` to store a reference to the `RateLimitMiddleware` instance.
2. During shutdown, call `rateLimiter.Stop()` (similar to `gtfsManager.Shutdown()`).
3. Ensure the cleanup goroutine exits when the ticker stops.

This ensures a clean, leak-free shutdown.

---

## 🙋 I Would Love to Work on This!

I'm interested in implementing this fix and improving the shutdown behavior of the rate-limiter middleware.

Ahmedhossamdev:
> Also, even after wiring the Stop() call, the cleanup goroutine still blocks because time.Ticker.Stop() doesn’t close the channel. Would you like me to add a shutdown signal (context/done channel) so cleanup() can exit, or do you have a different plan in mind?

Great idea you could try it.

PR #147, opened 2025-12-04
Goroutine Leak in RateLimitMiddleware

Bhup-GitHUB:
# Fix Rate Limiter Cleanup Goroutine Leak

## Problem #144 

The `RateLimitMiddleware` starts a background cleanup goroutine in `NewRateLimitMiddleware()` that never stops during graceful shutdown, causing a goroutine leak. This leads to:

- Goroutine leaks on every server shutdown
- Cleanup ticker continues running after shutdown
- Incomplete graceful shutdown
- Unnecessary resource usage

### Root Cause

1. `NewRateLimitMiddleware()` returned only `func(http.Handler) http.Handler`, discarding the middleware instance
2. `RestAPI` struct only stored the handler function, not the middleware reference
3. `Stop()` method existed but was never called during shutdown
4. Even if `Stop()` were called, the cleanup goroutine would still block indefinitely because `ticker.Stop()` doesn't close the channel, and the `for range` loop would continue waiting

## Solution

This fix ensures the cleanup goroutine properly exits during graceful shutdown by:

1. **Retaining middleware instance**: Changed `NewRateLimitMiddleware()` to return `*RateLimitMiddleware` instead of just the handler function
2. **Adding stop channel**: Added `stopChan chan struct{}` to signal the goroutine to exit
3. **Making Stop() idempotent**: Used `sync.Once` to ensure `Stop()` can be safely called multiple times
4. **Updating cleanup loop**: Modified `cleanup()` to use a `select` statement that listens for both ticker events and stop signals
5. **Integrating with shutdown**: Updated server shutdown flow to call `api.Shutdown()` which stops the rate limiter

### Core Changes

- **`internal/restapi/rate_limit_middleware.go`**:
  - Added `stopChan chan struct{}` and `stopOnce sync.Once` fields
  - Changed `NewRateLimitMiddleware()` return type to `*RateLimitMiddleware`
  - Added `Handler()` method to return the middleware handler function
  - Updated `cleanup()` to use `select` with stop channel
  - Updated `Stop()` to close stop channel using `sync.Once` for idempotency

- **`internal/restapi/rest_api.go`**:
  - Changed `rateLimiter` field type from `func(http.Handler) http.Handler` to `*RateLimitMiddleware`
  - Added `Shutdown()` method to stop the rate limiter

- **`internal/restapi/routes.go`**:
  - Updated to call `api.rateLimiter.Handler()` instead of calling the function directly

- **`cmd/api/app.go`**:
  - Updated `CreateServer()` to return both `*http.Server` and `*restapi.RestAPI`
  - Updated `Run()` to accept `*restapi.RestAPI` parameter
  - Added `api.Shutdown()` call during graceful shutdown

- **`cmd/api/main.go`**:
  - Updated to capture and pass API instance to `Run()`

### Test Changes

- **`internal/restapi/rate_limit_shutdown_test.go`** (NEW):
  - Added tests for proper shutdown behavior
  - Added tests for idempotent shutdown calls

- **`internal/restapi/rate_limit_middleware_test.go`**:
  - Updated all tests to use `.Handler()` method
  - Added `defer middleware.Stop()` to prevent test-time goroutine leaks

- **`cmd/api/app_test.go`**:
  - Updated all `CreateServer()` calls to handle new return value
  - Added `defer api.Shutdown()` to prevent test-time goroutine leaks

## Testing

All existing tests pass, and new tests verify:

- ✅ Shutdown completes within reasonable time
- ✅ Multiple `Stop()` calls are safe (idempotent)
- ✅ No goroutine leaks during shutdown
- ✅ All existing functionality remains intact

### Test Commands

Run these commands to verify the fix:

```bash
# Run all rate limiter tests
go test ./internal/restapi -run TestRateLimitMiddleware -v

# Run shutdown-specific tests
go test ./internal/restapi -run "TestRateLimitMiddleware_Shutdown|TestRestAPI_Shutdown" -v

# Run app tests
go test ./cmd/api -v

# Run full test suite
go test ./...
```

## Impact

- **Low Risk**: Changes are isolated to rate limiter and server lifecycle
- **Backward Compatible**: External API unchanged
- **Pattern Consistency**: Follows existing GTFS manager shutdown pattern
- **No Breaking Changes**: All existing functionality preserved

## Verification

The fix has been verified to:
- ✅ Stop cleanup goroutine during graceful shutdown
- ✅ Prevent goroutine leaks
- ✅ Maintain all existing rate limiting functionality
- ✅ Pass all existing tests
- ✅ Follow the same shutdown pattern as `gtfsManager.Shutdown()`

### DIFF
--- a/cmd/api/app.go
+++ b/cmd/api/app.go
@@ -5,18 +5,19 @@ import (
 	"encoding/json"
 	"fmt"
 	"log/slog"
-	"maglev.onebusaway.org/internal/app"
-	"maglev.onebusaway.org/internal/appconf"
-	"maglev.onebusaway.org/internal/gtfs"
-	"maglev.onebusaway.org/internal/logging"
-	"maglev.onebusaway.org/internal/restapi"
-	"maglev.onebusaway.org/internal/webui"
 	"net/http"
 	"os"
 	"os/signal"
 	"strings"
 	"syscall"
 	"time"
+
+	"maglev.onebusaway.org/internal/app"
+	"maglev.onebusaway.org/internal/appconf"
+	"maglev.onebusaway.org/internal/gtfs"
+	"maglev.onebusaway.org/internal/logging"
+	"maglev.onebusaway.org/internal/restapi"
+	"maglev.onebusaway.org/internal/webui"
 )
 
 // ParseAPIKeys splits a comma-separated string of API keys and trims whitespace from each key.
@@ -133,13 +134,14 @@ func Run(srv *http.Server, gtfsManager *gtfs.Manager, api *restapi.RestAPI, logg
 		return fmt.Errorf("server forced to shutdown: %w", err)
 	}
 
-	// Shutdown GTFS manager
-	if gtfsManager != nil {
-		gtfsManager.Shutdown()
-	}
-
+	// Shutdown API rate limiter first (stops background goroutines for request handling)
 	if api != nil {
 		api.Shutdown()
+	}
+
+	// Then shutdown GTFS manager (stops data fetching - the lowest-level dependency)
+	if gtfsManager != nil {
+		gtfsManager.Shutdown()
 	}
 
 	logger.Info("server exited")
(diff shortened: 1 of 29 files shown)


aaronbrethorst (approved):
great work, @Bhup-GitHUB

Issue #161, opened 2025-12-24
Fix: arrival and departure times in trip details endpoint it should be in seconds, not milliseconds.


Ahmedhossamdev:
- arrivalTime - time, in seconds since the start of the service date, when the trip arrives at the specified stop.

- departureTime - time, in seconds since the start of the service date, when the trip arrives at the specified stop.

For more details: https://developer.onebusaway.org/api/where/elements/trip-details

Arbaaz123676:
@MohamedKamal000 Assign me for this project

userjaggu:
Hi  @Ahmedhossamdev ,  I'm working on Maglev locally. Is this issue still available? I'd like to help with correcting the arrival/departure time units and can add tests + documentation if needed.

Ahmedhossamdev:
Hey @userjaggu thanks for your interest, but the issue's resolved now, so it'll be closed.

PR #168, opened 2025-12-28
fix: change arrival and departure time to seconds only from trips_helper

MohamedKamal000:
#161 
changed back the insertion of the time in nanoseconds in the database since the values are always treated as nanoseconds in the codebase 
converted the arrival and departure time to seconds from the trips_helper instead

### DIFF
--- a/internal/restapi/trips_helper.go
+++ b/internal/restapi/trips_helper.go
@@ -170,8 +170,8 @@ func (api *RestAPI) BuildTripSchedule(ctx context.Context, agencyID string, serv
 		}
 
 		stopTimesVals[i] = models.StopTime{
-			ArrivalTime:         int(st.ArrivalTime),
-			DepartureTime:       int(st.DepartureTime),
+			ArrivalTime:         int(st.ArrivalTime / 1e9),
+			DepartureTime:       int(st.DepartureTime / 1e9),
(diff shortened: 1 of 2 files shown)


internal/restapi/trips_helper.go
```diff
@@ -170,8 +170,8 @@ func (api *RestAPI) BuildTripSchedule(ctx context.Context, agencyID string, serv
 		}
 
 		stopTimesVals[i] = models.StopTime{
-			ArrivalTime:         int(st.ArrivalTime),
```
Ahmedhossamdev: Nice!

Ahmedhossamdev:
@aaronbrethorst  I think it's ready for merge.

internal/restapi/trips_helper.go
```diff
@@ -170,8 +170,8 @@ func (api *RestAPI) BuildTripSchedule(ctx context.Context, agencyID string, serv
...
-			DepartureTime:       int(st.DepartureTime),
+			ArrivalTime:         int(st.ArrivalTime / 1e9),
+			DepartureTime:       int(st.DepartureTime / 1e9),
```
Ahmedhossamdev: Nice

aaronbrethorst (approved):
Good work 👍

PR #169, opened 2025-12-30
Add Clock Abstraction for Deterministic Testing

Satyam709:
# Add Clock Abstraction for Deterministic Testing

Introduces a new Clock API for deterministic logic testing and validation.

---

## Problem

Several endpoints and calculations in the codebase are **time-dependent** e.g. active stops filtering, arrivals/departures, schedule lookups, trip details, etc.

The challenge comes in **testing this logic deterministically**. Tests that rely on GTFS schedule data behave differently depending on when they run passing on weekdays but failing on weekends or holidays when transit services don't operate.

### Example

`TestStopsForLocationHandlerEndToEnd` fails on several days (e.g., Dec 7, 14, 21, 25) while passing on others. The test queries stops near a location, but stops are filtered based on whether they have active service **on the current date**:

```go
// internal/gtfs/gtfs_manager.go
if !queryTime.IsZero() {
    currentDate = queryTime.Format("20060102")
} else {
    currentDate = time.Now().Format("20060102")  // ← uses system time
}

activeServiceIDs, _ := manager.GtfsDB.Queries.GetActiveServiceIDsForDate(ctx, currentDate)
```

This PR fixes this for `TestStopsForLocationEndToEnd` and provides the foundation to fix similar issues across the codebase.

---

### Clock Interface

```go
type Clock interface {
    Now() time.Time
    NowUnixMilli() int64
}
```

### Implementations

| Type | Purpose |
|------|---------|
| `RealClock` | Production — uses `time.Now()` |
| `MockClock` | Unit tests — returns fixed/controlled time |
| `EnvironmentClock` | Integration tests — reads time from env var or file |

---

### RealClock (Production)

```go
import "maglev.onebusaway.org/internal/clock"

// Default for production use
c := clock.RealClock{}
now := c.Now()           // Returns time.Now()
millis := c.NowUnixMilli() // Returns current time in milliseconds
```

### MockClock (Unit Tests)

```go
import "maglev.onebusaway.org/internal/clock"

// Create with a fixed time
fixedTime := time.Date(2025, 12, 26, 14, 39, 18, 0, time.UTC)
c := clock.NewMockClock(fixedTime)

// Get the current (fixed) time
now := c.Now()  // Always returns 2025-12-26 14:39:18

// Change the time
c.Set(time.Date(2025, 6, 15, 10, 0, 0, 0, time.UTC))

// Advance time by duration
c.Advance(1 * time.Hour)      // Move forward 1 hour
c.Advance(-30 * time.Minute)  // Move backward 30 minutes
```

**Test Example:**

```go
func TestStopsForLocationHandlerEndToEnd(t *testing.T) {
    // Fixed date ensures stop is active
    clock := clock.NewMockClock(time.Date(2025, 12, 26, 14, 39, 18, 0, time.UTC))
    api := createTestApiWithClock(t, clock)
    
    resp, model := serveApiAndRetrieveEndpoint(t, api, 
        "/api/where/stops-for-location.json?key=TEST&lat=40.583321&lon=-122.426966")
    
    assert.Equal(t, http.StatusOK, resp.StatusCode)
}
```

### EnvironmentClock (Integration Tests)

Reads time from external sources, useful for integration tests where time can be controlled by environment without code changes.

```go
import "maglev.onebusaway.org/internal/clock"

// Priority: env var → file → system time (fallback)
c := clock.NewEnvironmentClock(
    "MAGLEV_TIME",              // Environment variable name
    "/tmp/maglev-time.txt",     // File path
    time.UTC,                   // Location for parsing
)

now := c.Now()  // Reads from configured sources
```

### In Handlers

```go
func (api *Application) stopsForLocationHandler(w http.ResponseWriter, r *http.Request) {
    // Use injected clock instead of time.Now()
    queryTime := api.Clock.Now()
    
    stops := api.GtfsManager.GetStopsForLocation(ctx, lat, lon, radius, 
        latSpan, lonSpan, query, maxCount, false, routeTypes, queryTime)
}
```

### In Response Models

```go
// All response builders accept Clock for consistent currentTime field
response := models.NewListResponseWithRange(results, references, true, api.Clock)
response := models.NewEntryResponse(entry, references, api.Clock)
response := models.NewOKResponse(data, api.Clock)
```
## Advantages
- Enables reliable testing of GTFS service calendars, schedule boundaries, and date-specific logic by making time an explicit dependency.
- Improves architectural clarity by removing implicit reliance on time.Now() and centralizing time access behind a well-defined abstraction.
- Provides controlled time manipulation in unit and integration tests without impacting production behavior, and can be extended to support api-validator tool for more exhaustive integration testing.

### DIFF
--- a/internal/restapi/rate_limit_shutdown_test.go
+++ b/internal/restapi/rate_limit_shutdown_test.go
@@ -6,10 +6,11 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"maglev.onebusaway.org/internal/clock"
 )
 
 func TestRateLimitMiddleware_Shutdown(t *testing.T) {
-	middleware := NewRateLimitMiddleware(10, time.Second)
+	middleware := NewRateLimitMiddleware(10, time.Second, clock.RealClock{})
 
 	assert.NotNil(t, middleware)
 	assert.NotNil(t, middleware.Handler())
@@ -28,7 +29,7 @@ func TestRateLimitMiddleware_Shutdown(t *testing.T) {
 }
 
 func TestRateLimitMiddleware_ShutdownIdempotent(t *testing.T) {
-	middleware := NewRateLimitMiddleware(10, time.Second)
+	middleware := NewRateLimitMiddleware(10, time.Second, clock.RealClock{})
 
 	middleware.Stop()
 	middleware.Stop()
@@ -68,7 +69,7 @@ func TestRateLimitMiddleware_GoroutineActuallyExits(t *testing.T) {
 	// Get baseline goroutine count
 	initial := runtime.NumGoroutine()
 
-	middleware := NewRateLimitMiddleware(10, time.Second)
+	middleware := NewRateLimitMiddleware(10, time.Second, clock.RealClock{})
 	time.Sleep(50 * time.Millisecond) // Give goroutine time to start
 
 	afterCreate := runtime.NumGoroutine()
--- a/internal/restapi/responses.go
+++ b/internal/restapi/responses.go
@@ -71,7 +71,7 @@ func (api *RestAPI) sendError(w http.ResponseWriter, r *http.Request, code int, 
 
 	response := models.ResponseModel{
 		Code:        code,
-		CurrentTime: models.ResponseCurrentTime(),
+		CurrentTime: models.ResponseCurrentTime(api.Clock),
 		Text:        message,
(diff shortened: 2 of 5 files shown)


Satyam709:
I’ve fixed a few more tests failing for the same reason
I think this should be good to merge for now, and the remaining tests can be fixed going forward.
Happy to make further changes if needed.

PR #170, opened 2025-12-30
perf: optimize stops-for-route endpoint and add shape caching

### LINKED ISSUE #165: Performance: 'stops-for-route' times out (>30s) due to N+1 query pattern
I noticed high latency in the stops-for-route endpoint when testing with larger routes (e.g., Sound Transit "1 Line"). Locally, the response time exceeds 30 seconds, which leads to timeouts in clients like Postman.

While some of this delay might be exacerbated by local SQLite locking, profiling suggests there is an N+1 query pattern that could be optimized to improve performance for larger agencies.

**Observed Behavior:**

Client side(Postman):
Image

Server logs:
Image

**Root Cause:**
Upon profiling internal/restapi/stops_for_route_handler.go, I identified an N+1 query bottleneck in the buildStopsList function.

The code iterates through every stop ID associated with the route and executes synchronous database queries inside the loop. For a route with N stops, this results in 2N database round-trips per request.

Additional Finding: In stopsForRouteHandler, the error for time.LoadLocation is currently ignored in the same file.

**Proposed Solution:**

1. Refactoring buildStopsList: Implement a bulk SQL query (e.g., GetStopsByIDs) to fetch all stop metadata in a single database call.
2. Fixing Error Handling: Add a nil check for time.LoadLocation to default to UTC if the system timezone is missing, preventing server panics.

I have a local branch ready to work on this. Please assign this issue to me.

vedanthnyk25:
This PR addresses the severe latency issues on the stops-for-route endpoint (Issue #165), where large routes were frequently timing out (30s+). By implementing bulk-loading, localized caching, and database indexing, the latency has been reduced to ~1.5s .

Technical changes:

- Added a CREATE INDEX on shapes(shape_id) to prevent full table scans during geometry lookups.

- Replaced individual stop-by-stop database queries with a bulk-loading pattern in stopsForRouteHandler.

- Implemented SetContextCache and SetShapeCache in AdvancedDirectionCalculator to allow for in-memory  calculations.

-  Updated CalculateStopDirection to accept existing GTFS directions (sql.NullString), bypassing calculations when data is already present in the database.

Initial:

terminal_log1

postman_log1

Final:

Screenshot from 2025-12-28 21-50-15

Screenshot from 2025-12-28 21-49-18

Closes #165


aaronbrethorst (approved):
this looks great, Vedanth. Thanks for incorporating my feedback and getting this ready for merge. This is a huge win! 🚀

PR #171, opened 2025-12-30
Verifiy trip details 108

MohamedKamal000:
solves #108 
Verification & Testing Checklist for trip-details Endpoint:

i tested the endpoint against pugetsound server with agency id 1 and 40

[x] Verify the endpoint is reachable and responding with correct status codes (e.g., 200 OK for valid requests).

[x] Compare the response from testing against production to ensure consistency in data and structure.

[x] Validate response schema (fields, types, required properties).
 currently there is only one type change against the server
 in the trips field inside the references field, the directionId is integer in the development
 but it is string in the server response
 So I iam not sure should i change or not. i didn't want to because it would result in more refactoring across the codebase

[x] Confirm trip details are accurate for the given trip ID.

[x] Test with different trip IDs (valid, invalid, non-existent).

[x] Test edge cases (e.g., trips with missing fields).

[] Check response times and ensure performance is acceptable. (the direction calculation causes request time out, see issue #165 )

[x] Ensure proper error handling (400, 404, 500 cases).

[] Write/strengthen unit tests if coverage is missing or weak.

[x] Confirm no regressions compared to production behavior.

Notes:
- in order for me to test the end point correclty, i had to use the DirectionCalculator in the
precomputation step, so that i don't wait for a long times and maybe don't even get a response when testing
when i did that, i noticed difference in the direction values compared to the server response

- i didn't test the status field much, since it depends on real time data

### DIFF
--- a/internal/models/constants.go
+++ b/internal/models/constants.go
@@ -3,7 +3,9 @@ package models
 // Common constants used across the application
 const (
 	// UnknownValue is the fallback value when data is unavailable or calculation fails
-	UnknownValue  = "UNKNOWN"
-	Accessible    = "ACCESSIBLE"
+	UnknownValue = "UNKNOWN"
+	// Accessible indicates wheelchair boarding is possible (GTFS wheelchair_boarding = 1)
+	Accessible = "ACCESSIBLE"
+	// NotAccessible indicates wheelchair boarding is not possible (GTFS wheelchair_boarding = 2)
 	NotAccessible = "NOT_ACCESSIBLE"
 )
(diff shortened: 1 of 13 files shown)


aaronbrethorst (approved):
Thanks for the thorough work addressing all the feedback from my previous review. The changes look solid and the verification checklist in the PR description demonstrates good attention to detail.

### Important Items

**Potential test flakiness at midnight boundary**

**Files:** `internal/restapi/trip_details_handler_test.go:47-53`, `internal/restapi/trip_for_vehicle_handler_test.go:155-163`

The service date tests calculate the expected value using `time.Now()` after the HTTP request has already been made:

```go
currentTimeInLoc := time.Now().In(loc)
y, m, d := currentTimeInLoc.Date()
expectedServiceDate := time.Date(y, m, d, 0, 0, 0, 0, loc)
```

If the test runs exactly at midnight and the request happens just before midnight while the assertion happens just after, the dates will differ and the test will fail. This is a rare edge case but worth noting.

I don't think it's worth delaying merging this pull request for that very unlikely occurrence.

MohamedKamal000:
will try to open a pr for that test as well once iam done with other prs

PR #172, opened 2026-01-05
fix: limit Exceeded now using maxCount to find its value, OutOfBounds…

### LINKED ISSUE #99: Verification & Testing Checklist for `routes-for-location` Endpoint
## Verification & Testing Checklist for `routes-for-location` Endpoint

- [] Verify the endpoint is reachable and responding with correct status codes (e.g., `200 OK` for valid requests).
- [] Compare the response from **testing** against **production** to ensure consistency in data and structure.
- [] Validate response schema (fields, types, required properties).
- [] Confirm returned routes near the location match expected results (including optional route name filtering).
- [] Test with different location coordinates (valid, invalid, edge cases).
- [] Test with and without route name filter.
- [] Test edge cases (e.g., locations with no nearby routes).
- [] Check response times and ensure performance is acceptable.
- [] Ensure proper error handling (`400`, `404`, `500` cases).
- [] Write/strengthen unit tests if coverage is missing or weak.
- [] Confirm no regressions compared to production behavior.
- [] If issues are found, create a PR with fixes and updated tests.

MohamedKamal000:
Closes #99 
Verification & Testing Checklist for routes-for-location Endpoint

- [x] Verify the endpoint is reachable and responding with correct status codes (e.g., 200 OK for valid requests).

- [x] Compare the response from testing against production to ensure consistency in data and structure.

- [x] Validate response schema (fields, types, required properties).

- [x] Confirm returned routes near the location match expected results (including optional route name filtering).

- [x] Test with different location coordinates (valid, invalid, edge cases).

- [x] Test with and without route name filter.

- [x] Test edge cases (e.g., locations with no nearby routes).

- [x] Check response times and ensure performance is acceptable.

- [x] Ensure proper error handling (400, 404, 500 cases).

- [x] Write/strengthen unit tests if coverage is missing or weak. ( will make a commit later for tests ) 

- [x] Confirm no regressions compared to production behavior.

Notes and questions:
- The maxCount in production uses 50 as the maximum default value and always uses it if the maxCount exceeds it, but it returns an error once the field overflows. In development, it returns an error once the count exceeds 250, so I am not sure if I should do the same for development or not.

- The list response from production sorts the results by route ID. Should I do the same for development or not?

- Should I make a constant maxCount value and use it for every other endpoint that accepts a maxCount?

### DIFF
--- a/internal/models/response_test.go
+++ b/internal/models/response_test.go
@@ -87,10 +87,11 @@ func TestNewListResponseWithRange(t *testing.T) {
 	itemList := []string{"item1", "item2", "item3"}
 	references := NewEmptyReferences()
 	outOfRange := true
+	isLimitExceeded := true
 
 	clock := clock.RealClock{}
 
-	response := NewListResponseWithRange(itemList, references, outOfRange, clock, false)
+	response := NewListResponseWithRange(itemList, references, outOfRange, clock, isLimitExceeded)
 
 	assert.Equal(t, http.StatusOK, response.Code)
 	assert.Equal(t, "OK", response.Text)
@@ -101,7 +102,7 @@ func TestNewListResponseWithRange(t *testing.T) {
 	assert.True(t, ok, "Response data should be a map")
 	assert.Equal(t, itemList, responseData["list"], "List in response data should match input list")
 	assert.Equal(t, references, responseData["references"], "References in response data should match input references")
-	assert.False(t, responseData["limitExceeded"].(bool), "limitExceeded should be false (hardcoded)")
+	assert.True(t, responseData["limitExceeded"].(bool), "limitExceeded should be true")
 	assert.True(t, responseData["outOfRange"].(bool), "outOfRange should be true")
 }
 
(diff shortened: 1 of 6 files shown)


mann-patwa:
Hey @MohamedKamal000,

I don’t think you need to acquire a lock inside `checkIfOutOfBounds`. This function is already called from within the handler, which holds the required lock. Adding another lock here could potentially lead to a deadlock with the `gtfsData` writer goroutines.

Instead, I’d suggest documenting the locking contract explicitly. Adding a comment like this above the method should be sufficient:
```go
```

That way, the responsibility is clear and we avoid redundant locking.

@aaronbrethorst  could you confirm if this approach looks good?

aaronbrethorst (approved):
Hey Mohamed, you've done an excellent job iterating on this PR through three review cycles. You addressed all 12 items from the first review and all 6 items from the second review, and the code is in great shape. The switch to `strconv.Atoi`, the break-early `limitExceeded` pattern, the `checkIfOutOfBounds` helper with proper doc comments, the `ParseMaxCount` utility returning `defaultCount` on validation failure, and the comprehensive test coverage across `ParseMaxCount`, `IsOutOfBounds`, and the handler integration tests all demonstrate real care and attention to detail.

There's nothing left to change--this PR is ready to merge. The extracted utilities (`ParseMaxCount`, `IsOutOfBounds`, `checkIfOutOfBounds`) are well-designed for reuse, the `outOfRange` calculation is now semantically correct across all three location-based endpoints, and the test suite gives good confidence in the behavior. Great contribution!

PR #142, opened 2025-11-27
Add Makefile target coverage-report for CI-friendly coverage output (fixes #137)

Santosh69:
## Summary

This PR adds a new Makefile target `coverage-report` that prints a CI-friendly coverage summary directly to the terminal.

- New target: `coverage-report`
- Purpose: provide a concise coverage summary that CI tools or scripts can easily consume.
- The existing `coverage` target (HTML output) is left unchanged.

## What I changed

Added this target to `Makefile`:

```makefile
coverage-report:
	@go test -coverprofile=coverage.out ./...
	@go tool cover -func=coverage.out | grep total | awk '{print "Coverage:", $$3}'
```

Updated `.PHONY` to include `coverage-report`.

## Why

Issue #137 requests a terminal-friendly coverage output.  
The current Makefile only generates an HTML report.  
This new target provides a simple, script-friendly summary without replacing the existing workflow.

## How I tested (local)

- Ran `make coverage-report` → confirmed coverage output is printed for all packages.
- Verified `coverage.out` is generated.
- Ran `make coverage` → confirmed the HTML report is still produced successfully.

## Screenshots

1. **Terminal output from `make coverage-report`:**  
    WhatsApp Image 2025-11-27 at 12 06 23 PM (2)

2. **Makefile change (`coverage-report` target):**   
     WhatsApp Image 2025-11-27 at 12 06 23 PM (1)

3. **HTML coverage report generated by `make coverage`:**  
   WhatsApp Image 2025-11-27 at 12 30 14 PM

Fixes #137

Issue #137, opened 2025-11-25
Add Makefile Target for Test Coverage Report


Ahmedhossamdev:
## Description:

The current make coverage command generates HTML output. Add a new target that outputs coverage percentage to the terminal for CI/CD integration.

Implementation Guidance:

- [ ]  Add a new coverage-report target to the Makefile
- [ ]  Use go test -cover with appropriate flags
- [ ]  Format output for easy parsing by CI tools
- [ ]  Keep the existing coverage target for HTML output

PR #179, opened 2026-01-12
Added code coverage target + formatting (Fix for #137)

theKimSingh:
Hey, 

I noticed the previous PR (#142) to solve this issue (#137) had many mistakes + had code from a separate issue + didn't even solve the actual issue properly. 

I've submitted a PR with a simple target that runs code coverage on demand and parses it into a JSON for CI parsing.
Here is an example of what it looks like when `make coverage-report` is run:  

```
{
  "pkg": "maglev.onebusaway.org/internal/utils",
  "coverage": "83.6%"
}
```

The only thing I would like to note is that while testing, I noticed that test results can be cached by the CLI tool, which replaces the time that is outputted, so this: 
`ok      maglev.onebusaway.org/cmd/api   6.664s  coverage: 20.8% of statements`
is replaced by 
`ok      maglev.onebusaway.org/cmd/api   (cached)        coverage: 20.8% of statements`

I could update this target to add runtime/cached to the json format if it is preferred.


theKimSingh:
Apologies, most of these errors have been due to merge conflicts which is why the changes get lost in the process (the 2 errors you just mentioned + the build tags)

aaronbrethorst (approved):
Hey Kim, you stuck with it and addressed every piece of feedback -- the error handling, the `grep '^ok'` filter, `$(SET_ENV)` with build tags, `.PHONY` declarations for both targets, the `check-jq` prerequisite, the CLAUDE.md documentation, and the rebase. All the substantive issues from the previous four rounds are resolved.

There's nothing left to change -- this PR is ready to merge. Solid addition for CI-parseable coverage output.

PR #180, opened 2026-01-15
fix: Guard against vehicle updates with nil IDs

hufman:
The OVapi service publishes vehicle update information where some rows have nil fields, and causes a crash during import. This skips importing those rows to protect against the crash.

### DIFF
--- a/internal/gtfs/realtime_test.go
+++ b/internal/gtfs/realtime_test.go
@@ -144,14 +144,19 @@ func TestRebuildRealTimeVehicleLookupByVehicle(t *testing.T) {
 	assert.Equal(t, 1, manager.realTimeVehicleLookupByVehicle["vehicle2"])
 }
 
-func TestRebuildRealTimeVehicleLookupByVehicle_WithNilID(t *testing.T) {
+func TestRebuildRealTimeVehicleLookupByVehicle_WithInvalidIDs(t *testing.T) {
 	manager := &Manager{
 		realTimeVehicles: []gtfs.Vehicle{
 			{
 				ID: &gtfs.VehicleID{ID: "vehicle1"},
 			},
 			{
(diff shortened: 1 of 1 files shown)


aaronbrethorst (changes requested):
Thank you for addressing all the feedback from my previous review. The implementation now filters vehicles with nil IDs at the source, which is the cleanest approach. I have one last, minor request before merging your PR. Can you add one more test case?

**1. Test should also cover empty string ID case**

**File:** `internal/gtfs/realtime_test.go`

The new test covers the nil ID case, which is great. However, since `rebuildRealTimeVehicleLookupByVehicle` now has an explicit check for empty string IDs (`if vehicle.ID.ID != ""`), it would be good to add a test case that verifies this behavior as well.

Add a test case with an empty string ID:

```go
func TestRebuildRealTimeVehicleLookupByVehicle_WithEmptyStringID(t *testing.T) {
    manager := &Manager{
        realTimeVehicles: []gtfs.Vehicle{
            {
                ID: &gtfs.VehicleID{ID: "vehicle1"},
            },
            {
                ID: &gtfs.VehicleID{ID: ""}, // Vehicle with empty string ID - should be skipped in lookup
            },
            {
                ID: &gtfs.VehicleID{ID: "vehicle3"},
            },
        },
    }

    rebuildRealTimeVehicleLookupByVehicle(manager)

    assert.NotNil(t, manager.realTimeVehicleLookupByVehicle)
    assert.Len(t, manager.realTimeVehicleLookupByVehicle, 2)
    assert.Equal(t, 0, manager.realTimeVehicleLookupByVehicle["vehicle1"])
    assert.Equal(t, 2, manager.realTimeVehicleLookupByVehicle["vehicle3"])
}
```

aaronbrethorst (approved):
awesome, thank you so much!

PR #195, opened 2026-01-22
perf(api): optimize trip distance calculation from O(NxM) to O(N+M)

### LINKED ISSUE #86: Potential Performance Bottleneck in calculatePreciseDistanceAlongTrip function - trip_helper.go
The calculatePreciseDistanceAlongTrip function performs distance calculations for every stop time in a trip. This results in a potential time complexity of O(n * m), where:

    n = number of stops in the trip

    m = number of shape points in the route

For trips with many stops and highly detailed shape data, this could lead to significant performance issues.

## Recommendation

Consider optimizing this function to avoid redundant distance calculations, or explore approximations that can maintain acceptable accuracy while improving performance.

vedanthnyk25:
This PR refactors the trip distance calculation logic in `trip_helper.go` to address a significant performance bottleneck. By switching from a brute-force linear scan to a monotonic (sliding window) search, we achieved a ~46x speedup for typical routes.

The previous implementation (calculatePreciseDistanceAlongTripWithCoords) performed a full linear scan of the route shape points (size $M$) for every single stop (size $N$).
- Complexity: $O(N \times M)$
- Impact: For a route with 10,000 shape points and 100 stops, this resulted in ~1 million distance calculations per request, causing noticeable latency (~140 ms).

I introduced a new batch-processing function, calculateBatchStopDistances, that leverages the sequential nature of bus routes: 
Monotonic Search $O(N+M)$: Since stops appear in order, the search for Stop $B$ resumes from the index where Stop $A$ was found (lastMatchedIndex), rather than resetting to zero. 

Early Exit Heuristic: The search loop breaks early if the distance to the shape segments begins to increase significantly (currently set to a threshold of minDistance + 100m).
**Note: This implies the local minimum has been passed. This heuristic is conservative but effective; it can be further tuned based on specific route curvature data if needed.** 
image

**Benchmark Results**
Benchmarks run on a standard Linux environment (AMD Ryzen 5) with a test dataset of 10,000 shape points and 100 stops:
benchmarks

**Verification**
- Added TestCalculateBatchStopDistances to verify correctness of the monotonic logic against a known straight-line geometry.
- Verified comprehensive benchmarks (BenchmarkLegacy vs BenchmarkOptimized) to validate performance gains.

image

Closes #86


aaronbrethorst (approved):
Great work, Vedanth. This is ready for merging! 🔥

PR #209, opened 2026-01-27
fix: prevent panic on nil stop coordinates in gtfs import

### LINKED ISSUE #208: Fix potential panic: nil pointer dereference for Stop Lat/Lon
**Description:**
I noticed a potential panic risk in `processAndStoreGTFSDataWithSource` inside `helpers.go`.
Currently, the code directly dereferences `s.Latitude` and `s.Longitude` without checking if they are nil:

```go
Lat: *s.Latitude, // Panic if nil
Lon: *s.Longitude,
```

Reasoning:

The go-gtfs library defines these fields as pointers (*float64).

According to GTFS specs, stop_lat and stop_lon are optional for certain location_type values (like entrances or generic nodes).

Impact: If a GTFS feed contains a stop without coordinates, the application will panic and crash during import.

Proposed Solution: I will create a PR to add nil checks and use default values (0.0) if the pointers are nil.

3rabiii:
### Description
This PR fixes a potential panic during GTFS data import. The previous implementation directly dereferenced `s.Latitude` and `s.Longitude` pointers without checking if they were `nil`.

According to the GTFS spec, stops of certain types (like entrances or generic nodes) may not have coordinates. Without this fix, importing such feeds causes the application to crash.

### Changes
- Added nil checks for `Latitude` and `Longitude` in `helpers.go`.
- Used a default value of `0.0` if the coordinates are missing, ensuring safe execution.

### Related Issue
Fixes #208

### DIFF
--- a/gtfsdb/helpers.go
+++ b/gtfsdb/helpers.go
@@ -179,20 +179,16 @@ func (c *Client) processAndStoreGTFSDataWithSource(b []byte, source string) erro
 
 	var allStopParams []CreateStopParams
 	for _, s := range staticData.Stops {
-		var lat, lon float64
-		if s.Latitude != nil {
-			lat = *s.Latitude
-		}
-		if s.Longitude != nil {
-			lon = *s.Longitude
+		if s.Latitude == nil || s.Longitude == nil {
+			continue
 		}
 		params := CreateStopParams{
 			ID:                 s.Id,
 			Code:               toNullString(s.Code),
 			Name:               toNullString(s.Name),
 			Desc:               toNullString(s.Description),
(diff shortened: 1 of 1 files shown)


3rabiii:
Thanks for the great feedback! 

I've updated the logic to **skip** stops that lack coordinates, as suggested. This safely prevents the panic while maintaining data integrity.

Ready for review! 

@aaronbrethorst 
closes : #208

aaronbrethorst (approved):
looks good, Adel. I'm going to merge this and then follow up shortly with a comment explaining why we're short-circuiting the code.

Issue #210, opened 2026-01-28
Refactor: FindAgency returns pointer to loop variable


3rabiii:
**Description:**
The `FindAgency` method in `gtfs_manager.go` currently returns a pointer to the loop variable `agency`.

```go
for _, agency := range manager.gtfsData.Agencies {
    if agency.Id == id {
        return &agency // Returns pointer to loop variable
    }
}
```
Reasoning: Taking the address of a loop variable is considered a bad practice in Go. While it might work in the current context because of the immediate return, it is fragile.

1. Linter Warnings: Tools like exportloopref flag this as an error.

2. Future Safety: If the code is modified later to store these pointers instead of returning them immediately, it could lead to bugs where all pointers point to the same updated value.

Proposed Solution: I will refactor the code to create an explicit local copy of the struct before returning its address

PR #211, opened 2026-01-28
refactor: return explicit copy in FindAgency to avoid pointer export

3rabiii:
### Description
This PR refactors `FindAgency` to return a pointer to an explicit copy of the agency struct, rather than a pointer to the loop variable.

### Reasoning
Taking the address of a loop variable (`&agency`) can lead to subtle bugs or be flagged by linters (like `exportloopref`) as unsafe in older Go versions or specific contexts.
By creating `agencyCopy := agency` and returning `&agencyCopy`, we ensure:
1. **Safety:** The returned pointer is distinct and safe from loop iteration mutations.
2. **Clarity:** The intent to return a copy is explicit.
3. **Clean Code:** Adheres to Go best practices and linter standards.

### Changes
- Updated `FindAgency` in `gtfs_manager.go` to create a local copy before returning its address.

fixes : #210

### DIFF
--- a/internal/gtfs/gtfs_manager.go
+++ b/internal/gtfs/gtfs_manager.go
@@ -139,7 +139,8 @@ func (manager *Manager) FindAgency(id string) *gtfs.Agency {
 	defer manager.staticMutex.RUnlock()
 	for _, agency := range manager.gtfsData.Agencies {
 		if agency.Id == id {
-			return &agency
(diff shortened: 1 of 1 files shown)


aaronbrethorst (commented):
oops, I didn't mean to merge this without reviewing it first 😆 this looks good, though. thanks for taking care of this.

Issue #215, opened 2026-01-28
Log error when closing GTFS database during shutdown


3rabiii:
### Description
In `internal/gtfs/gtfs_manager.go`, the `Shutdown` method currently ignores any error returned when closing the GTFS database:

```go
if manager.GtfsDB != nil {
    _ = manager.GtfsDB.Close()
}
```
Problem: Silently discarding this error prevents operators from knowing if the database failed to close cleanly. This could hide issues like pending transactions, file locks, or resource leaks.

Proposed Solution: I will update the Shutdown method to capture the error and log it using slog.Error to ensure visibility into shutdown problems.

PR #216, opened 2026-01-28
fix: log error if GTFS database fails to close on shutdown

3rabiii:
### Description
This PR addresses an issue where errors occurring during the closure of the GTFS database were being silently ignored.

Previously:
```go
_ = manager.GtfsDB.Close()
```

Now, any error returned by Close() is logged using slog.Error, ensuring that potential resource leaks or database locking issues are visible to operators during shutdown.

Changes
 * Imported log/slog in gtfs_manager.go.

 * Added error checking and logging to the Shutdown method.

Related Issue
Fixes : #215


3rabiii:
Thanks for the review!
I've updated the code to use `logging.LogError` with a component logger, matching the pattern used in `realtime.go` and `direction_precomputer.go`.

Ready for review!
@aaronbrethorst 
closes : #215

Issue #219, opened 2026-01-28
Propagate error when direction precomputation fails during import


3rabiii:
### Description
In `internal/gtfs/static.go`, the `buildGtfsDB` function catches errors from `PrecomputeAllDirections` but only logs them and continues:

```go
if err := precomputer.PrecomputeAllDirections(ctx); err != nil {
    logging.LogError(logger, "Failed to precompute stop directions", err)
}
```
Problem: This leads to an application state where the GTFS data is partially loaded (missing directions), but the system appears healthy. This violates the "Fail Fast" principle and can confuse downstream consumers of the data.

Proposed Solution: Update buildGtfsDB to return the error immediately if precomputation fails, ensuring the application does not start with incomplete data.

PR #220, opened 2026-01-28
fix: enforce hard failure when direction precomputation fails

3rabiii:
### Description
This PR updates `buildGtfsDB` in `static.go` to return an error immediately if `PrecomputeAllDirections` fails.

### Reasoning
Previously, the code only logged the error and allowed the application to start.
I have changed this to a **hard failure** because missing direction data compromises the integrity of the application's core functionality. It is safer to crash on startup than to serve incomplete data.

### Changes
- Removed the "log and continue" logic.
- Added `return nil, fmt.Errorf(...)` to propagate the error.

### Related Issue
Fixes : #219

### DIFF
--- a/internal/gtfs/static.go
+++ b/internal/gtfs/static.go
@@ -87,7 +87,8 @@ func buildGtfsDB(config Config, isLocalFile bool) (*gtfsdb.Client, error) {
 	// Precompute stop directions after GTFS data is loaded
 	precomputer := NewDirectionPrecomputer(client.Queries, client.DB)
 	if err := precomputer.PrecomputeAllDirections(ctx); err != nil {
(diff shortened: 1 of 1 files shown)


3rabiii:
Thanks for the detailed context regarding the fallback strategies in `CalculateStopDirection`.

I wasn't fully aware that the system handles missing precomputed directions gracefully by calculating them on-demand (shape-based or stop-to-stop). Given that reliability, I agree that hard failing here is too aggressive, especially since a transient DB issue could prevent the entire application from starting.

**Changes made:**
- Reverted to the "Graceful Degradation" approach.
- Updated the log message to explicitly state that the system is falling back to on-demand calculation, ensuring operators are aware of the potential performance impact.

@aaronbrethorst 
closes : #219

Issue #221, opened 2026-01-28
Fix incomplete FTS5 query sanitization in searchStopsHandler


3rabiii:
### Description
The `sanitizeFTS5Query` function in `search_stops_handler.go` currently misses several FTS5 operators (like `NEAR`) and special characters (like `:`, `^`, `{`, `}`).

**Problem:**
This allows users to inadvertently trigger SQLite syntax errors by including these characters in their search queries, potentially leading to 500 Internal Server Errors.

**Proposed Solution:**
Update the sanitization logic to strip ALL FTS5 control characters and operators to ensure query stability.

PR #222, opened 2026-01-28
Fix incomplete FTS5 query sanitization

3rabiii:
### Description
This PR updates the `sanitizeFTS5Query` function to include a more comprehensive list of FTS5 special characters and operators.

### Changes
- Added `NEAR` to the list of stripped operators.
- Added missing special characters (e.g., `:`, `^`, `{`, `}`, `[`, `]`, etc.) to the regex.
- Used `regexp` for broader character matching instead of simple string replacement.

### Reasoning
Preventing SQLite FTS5 syntax errors is critical for API stability. The previous sanitizer allowed characters that could cause the search query to fail with a 500 error.

### Related Issue
Fixes  : #221
@aaronbrethorst

### DIFF
--- a/internal/restapi/search_stops_handler_test.go
+++ b/internal/restapi/search_stops_handler_test.go
@@ -325,6 +325,51 @@ func TestSanitizeFTS5Query(t *testing.T) {
 			input:    "中央駅 テスト",
 			expected: "中央駅 テスト",
 		},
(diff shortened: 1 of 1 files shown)


aaronbrethorst (approved):
terrific, thanks Adel. 🚀

Issue #223, opened 2026-01-29
Fix context handling in blockHandler to prevent zombie queries


3rabiii:
### Description
The `blockHandler` currently creates a fresh `context.Background()` for database operations instead of using the incoming request context `r.Context()`.

**Problem:**
If a client cancels the request (or disconnects) while the database query is running, the server continues to process the query unnecessarily. This leads to resource leaks (zombie queries) and wasted database connections.

**Proposed Solution:**
Use `r.Context()` to propagate cancellation signals correctly to the database layer.

PR #224, opened 2026-01-29
fix: use request context in blockHandler to prevent zombie queries

3rabiii:
### Description
This PR fixes a potential resource leak (zombie query) in the `blockHandler` by correctly using the HTTP request context instead of a detached `context.Background()`.

### Problem
Previously, `context.Background()` was used for database operations. This meant that if a client cancelled the request or disconnected, the database query would continue running to completion, wasting server resources and database connections.

### Changes
- Replaced `context.Background()` with `r.Context()` to properly propagate cancellation signals.
- Added checks for `ctx.Err()` to handle client disconnections gracefully.
- Prevented sending 404 responses or log errors if the operation failed due to a client disconnect.

### Related Issue
Fixes : #223
@aaronbrethorst

### DIFF
--- a/internal/restapi/block_handler.go
+++ b/internal/restapi/block_handler.go
@@ -31,6 +31,7 @@ func (api *RestAPI) blockHandler(w http.ResponseWriter, r *http.Request) {
 	block, err := api.GtfsManager.GtfsDB.Queries.GetBlockDetails(ctx, sql.NullString{String: blockID, Valid: true})
 	if err != nil {
 		if ctx.Err() != nil {
+			api.serverErrorResponse(w, r, ctx.Err())
 			return
 		}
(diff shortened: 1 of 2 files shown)


3rabiii:
Thanks for the encouraging feedback and the detailed review!

I've addressed both points:
1.  **Silent Return:** I updated the error handling logic in `block_handler.go`. Now, if `ctx.Err()` is present, it explicitly calls `api.serverErrorResponse` instead of silently returning, ensuring the client receives a proper response.
2.  **Tests:** I added `TestBlockHandlerContextCancellation` to verify that the handler behaves correctly under context cancellation/timeout scenarios.

Verification
Screenshot From 2026-01-30 11-24-59

Ready for review!
@aaronbrethorst 
closes : #223

Issue #237, opened 2026-01-29
Refactor hardcoded magic numbers in routesForLocationHandler


3rabiii:
### Problem
The `routesForLocationHandler` function currently relies on hardcoded integer literals (`600` and `10000`) to define default search radii. These "magic numbers" obscure the code's intent and make future maintenance difficult, as the units and reasoning behind these values are not explicit.

### Proposed Solution
Replace these raw numbers with named constants (e.g., `DefaultSearchRadiusInMeters` and `QuerySearchRadiusInMeters`) defined at the top of the file. This will improve code readability and facilitate easier configuration changes in the future.

PR #238, opened 2026-01-29
refactor: replace magic numbers with named constants in routes handler

3rabiii:
### Description
Replaces hardcoded numeric literals (magic numbers) in `routes_for_location_handler.go` with named constants to improve readability and maintainability.

### Changes
- Defined `DefaultSearchRadiusInMeters` (600).
- Defined `QuerySearchRadiusInMeters` (10000).
- Updated logic to use these constants instead of raw numbers.

### Related Issue
Fixes : #237
@aaronbrethorst


MohamedKamal000:
there is a file for constants you might consider to use rather than placing it at the top

PR #242, opened 2026-01-29
fix: return compound route IDs for routes-for-agency

### LINKED ISSUE #231: `route_for_agency` returns inconsistent `data.list.id` compared to Java production server
## Description:
While validating responses between the Java production server and Maglev, I noticed a mismatch in the `route_for_agency` endpoint response.

The production server `route_for_agency` endpoint the `data.list[].id` field returns the combined id {agencyId}_{routeId}
but the currently maglev returns just {routeId}, this is a mismatch.

Image

## Expected behavior:
For consistency with the production Java server and existing clients, data.list[].id should match the compound ID format:
`{agencyId}_{routeId}`

**Note:**
Please ignore other mismatch warnings related to ordering, there is some sort misordering present, but that is unrelated to this issue. This report is only about the id field format.

I am working on this

uloamaka:
### Changes
- Updated the route_for_agency endpoint to return compound route IDs:
data.list[].id == "{agencyId}_{routeId}"
This aligns Maglev behavior with the Java production server.
- Added a test asserting the compound route ID format to prevent future regressions.

fixes #231

### DIFF
--- a/internal/restapi/routes_for_agency_handler.go
+++ b/internal/restapi/routes_for_agency_handler.go
@@ -19,7 +19,7 @@ func (api *RestAPI) routesForAgencyHandler(w http.ResponseWriter, r *http.Reques
 	routesList := make([]models.Route, 0, len(routesForAgency))
 	for _, route := range routesForAgency {
 		routesList = append(routesList, models.NewRoute(
-			route.Id, route.Agency.Id, route.ShortName, route.LongName,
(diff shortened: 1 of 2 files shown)


aaronbrethorst (approved):
Thanks for fixing the route ID format in the routes-for-agency endpoint. Using `utils.FormCombinedID` to prefix the agency ID matches how every other handler in the codebase constructs route IDs, so this brings the endpoint in line with production behavior. The test you added is a nice touch for preventing regressions.

There's nothing left to change—this PR is ready to merge. Solid contribution!

PR #246, opened 2026-01-30
 fix: Resolve memory leak and performance issues in RateLimitMiddleware cleanup

### LINKED ISSUE #245: Critical: Ineffective cleanup in `RateLimitMiddleware` causes memory leaks and performance degradation
### **Issue Description**

I have identified a critical logical flaw in the garbage collection routine of the `RateLimitMiddleware`. The current implementation incorrectly identifies "active" users as candidates for deletion while potentially preserving "abusive" users, leading to both performance degradation and memory leaks.

### **Location**

* **File:** `internal/restapi/rate_limit_middleware.go`
* **Function:** `cleanup`

### **Analysis (Root Cause)**

The cleanup loop currently uses the following condition to decide if a limiter should be removed:

```go
// internal/restapi/rate_limit_middleware.go

if limiter := rl.limiters[key]; limiter.Tokens() > 0 {
    delete(rl.limiters, key)
}

```

This logic is inverted and dangerous:

1. **Legitimate Users (Active):** A normal user making requests within their limit will almost always have `Tokens() > 0`. The cleanup routine **deletes** their limiters every 5 minutes. This forces the server to re-allocate a new limiter and acquire a write lock on the next request, causing unnecessary CPU overhead and lock contention.
2. **Abusive Users (Attackers):** An attacker flooding the API will exhaust their tokens, leaving `Tokens() == 0`. The cleanup routine **skips** these limiters. If an attacker rotates through random API keys (high cardinality), the `limiters` map will grow indefinitely, resulting in a memory leak.

### **Impact**

* **Stability (Memory Leak):** High risk of OOM if the server is subjected to a distributed attack with random API keys.
* **Performance:** Constant deletion and re-creation of limiters for active users increases GC pressure and lock contention on the `sync.RWMutex`.

### **Problem Checklist**

* [ ] Cleanup routine deletes active limiters (Tokens > 0).
* [ ] Cleanup routine retains exhausted limiters (Tokens == 0).
* [ ] Map grows indefinitely under high-cardinality key attacks.
* [ ] No timestamp tracking is used to determine actual "idleness."

### **Proposed Solution**

Refactor the middleware to use a **Least Recently Used (LRU)** or **Time-Based Eviction** strategy:

1. Wrap the `rate.Limiter` in a struct that tracks a `lastSeen` timestamp.
2. Update `lastSeen` on every request.
3. Modify `cleanup` to delete only those limiters where `time.Since(lastSeen) > Threshold` (e.g., 10 minutes).

ARCoder181105:
**Description:**
This PR addresses a critical logical flaw in the `RateLimitMiddleware` garbage collection where active users were being aggressively evicted while exhausted limiters (potential attacks) were retained.

Closes #245 

### The Fix

* **Refactored Cleanup Strategy:** Switched from a token-based check (`Tokens() > 0`) to a Time-Based Eviction (LRU) strategy.
* **Internal Changes:**
* Introduced `rateLimitClient` struct to wrap `rate.Limiter` with a `lastSeen` timestamp.
* Updated `getLimiter` to refresh `lastSeen` on every request.
* Updated `cleanup` to only delete limiters that have been idle for more than **10 minutes**.

### Resolution Checklist

* [x] **Active Limiters:** No longer deletes active users (who typically have tokens > 0), reducing lock contention and reallocation overhead.
* [x] **Exhausted Limiters:** Now correctly deletes limiters with 0 tokens if they become inactive, removing the memory leak vector.
* [x] **High-Cardinality Attacks:** Prevents unbounded map growth by ensuring all keys eventually expire.
* [x] **Idleness Tracking:** Implemented precise timestamp tracking to determine actual inactivity.

### Verification

* Added new regression tests in `internal/restapi/rate_limit_cleanup_test.go` covering active users, inactive users, and exhausted limiters.
* Verified `make build` and `make test` pass.

### DIFF
--- a/internal/restapi/rate_limit_cleanup_test.go
+++ b/internal/restapi/rate_limit_cleanup_test.go
@@ -1,0 +1,298 @@
+package restapi
+
+import (
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"maglev.onebusaway.org/internal/clock"
+)
+
+// TestRateLimitMiddleware_CleanupKeepsActiveClients verifies active users are not deleted.
+func TestRateLimitMiddleware_CleanupKeepsActiveClients(t *testing.T) {
+	mockClock := clock.NewMockClock(time.Now())
+	middleware := NewRateLimitMiddleware(10, time.Second, mockClock)
+	defer middleware.Stop()
+
+	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	})
+	limitedHandler := middleware.Handler()(handler)
+
+	// Make a request from a user (creates a limiter)
+	req := httptest.NewRequest("GET", "/test?key=active-user", nil)
+	w := httptest.NewRecorder()
+	limitedHandler.ServeHTTP(w, req)
+	assert.Equal(t, http.StatusOK, w.Code)
+
+	// Verify limiter exists
+	middleware.mu.RLock()
+	client, exists := middleware.limiters["active-user"]
+	middleware.mu.RUnlock()
+	require.True(t, exists, "Limiter should exist after first request")
+	require.NotNil(t, client)
+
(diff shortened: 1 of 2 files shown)


ARCoder181105:
@aaronbrethorst 
test passing perfectly, the PR is ready for review

image

Issue #251, opened 2026-01-31
refactor: Replace silent validation failures with 400 Bad Request in API handlers


3rabiii:
**Problem Description**
Currently, several API handlers (`arrival-and-departure`, `trip-details`, etc.) silently swallow parsing errors for numeric parameters like `time`, `serviceDate`, or `minutesAfter`. If a user provides an invalid value (e.g., `?minutesAfter=abc`), the API ignores it and falls back to default values without notifying the client.

**Proposed Solution**
Refactor the parameter parsing logic in the REST API handlers to explicitly check for parsing errors. If an error occurs, the API should immediately return a `400 Bad Request` response with a descriptive error message indicating which field is invalid.

**Affected Handlers**
1. `arrival-and-departure-for-stop`
2. `arrivals-and-departures-for-stop`
3. `trip-details`
4. `trip-for-vehicle`

**Acceptance Criteria**
- Invalid parameters result in HTTP 400.
- Valid parameters continue to work as expected.
- Unit tests are added to cover invalid input scenarios.

PR #252, opened 2026-01-31
feat: enforce validation for time and date parameters across handlers

3rabiii:
Summary
This PR addresses the "Fit and Finish" feedback regarding silent validation failures. It enforces strict validation for numeric and date parameters across the API handlers. Instead of silently ignoring invalid inputs (e.g., ?time=invalid) and falling back to defaults, the API now returns a 400 Bad Request with a descriptive error message.

Changes
* Refactored Parameter Parsing: Updated the following handlers to return an error when parsing fails:
    * arrival-and-departure-for-stop
    * arrivals-and-departures-for-stop
    * trip-details
    * trip-for-vehicle

* Expanded Validation: Applied validation logic to time, serviceDate, minutesAfter, minutesBefore, and stopSequence.

* Test Coverage: Added comprehensive unit tests for each handler to verify that invalid parameters correctly trigger a 400 Bad Request.

* Test Fixes: Resolved timezone-related test failures in trip_for_vehicle by dynamically loading the agency's timezone instead of hardcoding.

Verification
Screenshot From 2026-01-31 20-42-24

Related
Follow-up to feedback from PR #235.

@aaronbrethorst 
closes : #251

### DIFF
--- a/internal/restapi/arrival_and_departure_for_stop_handler.go
+++ b/internal/restapi/arrival_and_departure_for_stop_handler.go
@@ -265,6 +265,10 @@ func (api *RestAPI) arrivalAndDepartureForStopHandler(w http.ResponseWriter, r *
 			if err == nil && v != nil && v.Trip != nil && v.Trip.ID.ID == tripID {
 				vehicle = v
 			}
+		} else {
+			api.Logger.Warn("malformed vehicleId provided",
+				"vehicleId", params.VehicleID,
+				"error", err)
 		}
 	} else {
 		// If vehicleId is not provided, get the vehicle for the trip
@@ -404,6 +408,7 @@ func (api *RestAPI) arrivalAndDepartureForStopHandler(w http.ResponseWriter, r *
 
 			if err != nil {
 				api.serverErrorResponse(w, r, err)
+				return
 			}
 
 			stopIDSet[nextStopID] = true
@@ -413,6 +418,7 @@ func (api *RestAPI) arrivalAndDepartureForStopHandler(w http.ResponseWriter, r *
 
 			if err != nil {
 				api.serverErrorResponse(w, r, err)
+				return
 			}
 			stopIDSet[closestStopID] = true
 		}
--- a/internal/restapi/arrival_and_departure_for_stop_handler_test.go
+++ b/internal/restapi/arrival_and_departure_for_stop_handler_test.go
@@ -662,7 +662,6 @@ func TestArrivalsAndDeparturesForStopHandlerInvalidTime(t *testing.T) {
 	endpoint := "/api/where/arrival-and-departure-for-stop/1_75403.json?key=TEST&time=invalid_time"
 
 	resp, _ := serveApiAndRetrieveEndpoint(t, api, endpoint)
-	defer resp.Body.Close()
 
 	assert.Equal(t, http.StatusBadRequest, resp.StatusCode)
 }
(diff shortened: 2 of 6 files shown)


aaronbrethorst (changes requested):
Hey Adel, thanks for taking the time to address the "Fit and Finish" feedback from PR #235. Adding strict validation for parameters that were previously silently ignored is exactly the kind of improvement that makes an API more predictable and developer-friendly. Before we can merge this, I will need you to make a couple changes:

## Critical

1. **Test function signature issue in `trip_details_handler_test.go`**: The new test `TestTripDetailsHandlerWithInvalidParams` creates an API with `createTestApi(t)` and then calls `serveAndRetrieveEndpoint(t, endpoint)`, which creates a *second* API internally. This means your test is getting the trip ID from one API instance but making the request against a different one. While this happens to work because both use the same test data, the returned API from `serveAndRetrieveEndpoint` is not shut down properly (you're shutting down the first API, not the one that actually handled the request).

   **Fix**: Either:
   - Use `serveApiAndRetrieveEndpoint(t, api, endpoint)` to use your existing API instance, or
   - Capture and shut down the API returned by `serveAndRetrieveEndpoint`:
   ```go
   testApi, resp, _ := serveAndRetrieveEndpoint(t, endpoint)
   defer testApi.Shutdown()
   ```

   The same pattern issue exists in the second call to `serveAndRetrieveEndpoint` in that test.

## Important

2. **Inconsistent error messages**: The existing validation in `arrivalsAndDeparturesForStopHandler` (on main) returned a more descriptive error message:
   ```go
   "time": {"must be a valid Unix timestamp in milliseconds"}
   ```

   Your changes now return `"invalid value"` for all parameters. This is a behavioral change that makes error responses less helpful for API consumers debugging their requests. Consider using more descriptive messages like `"must be a valid integer"` or `"must be a valid Unix timestamp in milliseconds"`.

3. **Missing unit tests for new parsing functions**: You added a new `ArrivalsAndDeparturesParams` struct and `parseArrivalsAndDeparturesParams` function, but only test it through the HTTP handler. The existing `parseArrivalAndDepartureParams` function has dedicated unit tests (`TestParseArrivalAndDepartureParams_AllParameters`, `TestParseArrivalAndDepartureParams_DefaultValues`, `TestParseArrivalAndDepartureParams_InvalidValues`). Consider adding equivalent unit tests for the new parsing function to maintain consistency and catch regressions at a more granular level.

4. **Missing unit tests for `parseTripIdDetailsParams` and `parseTripForVehicleParams`**: Similarly, these two parsing functions were modified to return errors but don't have direct unit tests - only handler-level integration tests. Adding unit tests for these would provide faster feedback and clearer error isolation.

## Fit and Finish

5. **Resource leak in `TestTripForVehicleHandlerWithInvalidParams`**: The test creates an `httptest.Server` with `defer server.Close()`, which is good. However, it captures `resp.Body` but only closes it with `resp.Body.Close()` after the assertions. If an assertion fails (calling `t.FailNow()` internally), the response body won't be closed. Consider using:
   ```go
   defer resp.Body.Close()
   ```
   immediately after checking `require.NoError(t, err)`.

6. **Numbered comments in parsing functions**: The comments like `// 1. Validate minutesAfter` are helpful for readability, but they're a bit mechanical. The numbered sequence doesn't add much since the order of validation isn't significant—any invalid parameter should fail regardless of which is checked first.

Thanks again, and I look forward to merging this change once these items are addressed.

aaronbrethorst (approved):
Hey Adel, excellent work iterating through the feedback! You've addressed every item from both of my previous reviews. The persistence and attention to detail really shows.



Great job seeing this through! Ready to merge.

PR #254, opened 2026-01-31
Fix: Real-time predictions and nearby stops logic in arrivals endpoint

### LINKED ISSUE #253: Real-time predictions missing in `/arrivals-and-departures-for-stop`
### Description

The endpoint for listing arrivals at a stop (`/arrivals-and-departures-for-stop`) fails to include real-time arrival and departure predictions. It currently returns scheduled times even when real-time trip updates (delays) are available in the system.

### Root Cause

The issue stems from a mismatch in how the application retrieves real-time data for this specific endpoint. The code attempts to read delay information directly from the "Vehicle" object. However, in this system's data model, delay information is stored in "Trip Update" feeds, which are separate from "Vehicle Position" feeds.

Because the system stores these two types of real-time data separately, the "Vehicle" object used in this handler does not contain the necessary delay information. Consequently, the application assumes there is no delay and falls back to displaying the static scheduled times.

### Impact

* **Missing Data:** The arrivals board displays all buses as "on time" (scheduled) regardless of actual delays, potentially causing users to miss their bus or wait unnecessarily.
* **Inconsistency:** This creates a confusing user experience where the main list view shows a bus as on time, but clicking into the specific trip details (which uses a different logic) correctly shows the delay.

### Proposed Solution

The logic for this endpoint needs to be updated to explicitly fetch the "Trip Update" associated with each trip, rather than relying on the "Vehicle" object to contain it. This will allow the application to correctly calculate and display real-time predictions in the arrivals list.

ARCoder181105:
#### Summary

Closes #253 

This PR resolves two critical bugs in the `/arrivals-and-departures-for-stop` endpoint:

1. **Missing Real-Time Predictions:** The endpoint was failing to show arrival/departure delays because it attempted to read `StopTimeUpdates` from the `Vehicle` object (which does not contain them in this data model).
2. **Incomplete Nearby Stops:** The `getNearbyStopIDs` helper function contained a logic error that caused it to return only a single nearby stop instead of the full list.

#### Changes

* **File:** `internal/restapi/arrivals_and_departure_for_stop.go`
* **Fix 1:** Updated `arrivalsAndDeparturesForStopHandler` to explicitly call `api.GtfsManager.GetTripUpdateByID(st.TripID)`. The handler now correctly uses the fetched `TripUpdate` object to calculate predicted arrival and departure times.
* **Fix 2:** Updated the loop in `getNearbyStopIDs` to use `append` instead of overwriting the slice, ensuring all found nearby stops are returned.

#### Impact

* **Accuracy:** The arrivals board will now correctly display "X min late" or "X min early" based on real-time GTFS-RT data.
* **Completeness:** The "Nearby Stops" field in the response will now correctly populate with up to 5 stops (or the configured limit) instead of just one.

### DIFF
--- a/internal/restapi/arrivals_and_departure_for_stop.go
+++ b/internal/restapi/arrivals_and_departure_for_stop.go
@@ -190,10 +190,13 @@ func (api *RestAPI) arrivalsAndDeparturesForStopHandler(w http.ResponseWriter, r
 		if vehicle != nil && vehicle.Trip != nil {
 			vehicleID = vehicle.ID.ID
 
-			// Check if we have stop time updates for this trip
-			if len(vehicle.Trip.StopTimeUpdates) > 0 {
+			// Fetch the Trip Update separately
+			tripUpdate, _ := api.GtfsManager.GetTripUpdateByID(st.TripID)
+
+			// Use the tripUpdate for predictions
+			if tripUpdate != nil && len(tripUpdate.StopTimeUpdates) > 0 {
 				// Look for StopTimeUpdate that matches this stop
-				for _, stopTimeUpdate := range vehicle.Trip.StopTimeUpdates {
+				for _, stopTimeUpdate := range tripUpdate.StopTimeUpdates {
 					// Match by stop sequence or stop ID
 					if (stopTimeUpdate.StopSequence != nil && int64(*stopTimeUpdate.StopSequence) == st.StopSequence) ||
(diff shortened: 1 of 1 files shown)


aaronbrethorst (approved):
Hey Aditya! Thanks for tackling this bug fix. Your issue write-up was thorough and clearly explained the root cause - it's great when contributors take time to understand the data model before diving into fixes. This is ready to merge.

Issue #256, opened 2026-02-01
[Security] Potential SQL Injection in TableCounts Debug Function


3rabiii:
### Description
A potential SQL injection vulnerability exists in the `TableCounts` function within `gtfsdb/debugging.go`.

The function currently iterates over table names and uses `fmt.Sprintf` to interpolate the table name directly into the SQL query string:
```go
query := fmt.Sprintf("SELECT COUNT(*) FROM %s", table)
```
Impact
While the table variable currently originates from database metadata (which limits immediate exploitability), this pattern violates defense-in-depth principles. If the code is refactored in the future to accept user input or external parameters for the table name, this becomes a critical vulnerability.

Suggested Fix
Implement a strict allowlist (whitelist) of valid GTFS table names. The function should verify that the table variable matches a known, safe table name before constructing the query.

PR #257, opened 2026-02-01
fix(security): prevent SQL injection in TableCounts via allowlist

3rabiii:
### Description
This PR addresses a potential SQL Injection vulnerability flagged in `gtfsdb/debugging.go`.
Although the table names currently originate from database metadata, passing them directly to `fmt.Sprintf` violates **defense-in-depth** principles.

### Changes
* Implemented a strict **allowlist (map)** containing valid GTFS schema table names.
* Added a validation check inside the loop to ensure only allowed tables are queried.
* Optimized the map declaration to sit outside the loop for better performance.

### Impact
Ensures that even if the source of table names changes in the future, the query construction remains secure against injection attacks.

@aaronbrethorst 
fix : #256

### DIFF
--- a/gtfsdb/debugging.go
+++ b/gtfsdb/debugging.go
@@ -3,11 +3,12 @@ package gtfsdb
 import (
 	"database/sql"
 	"fmt"
-	"github.com/OneBusAway/go-gtfs"
 	"log"
 	"log/slog"
-	"maglev.onebusaway.org/internal/logging"
 	"strings"
+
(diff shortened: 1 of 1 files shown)


aaronbrethorst (approved):
Hey Adel! Good eye on this defense-in-depth improvement. Even though the practical attack surface here is minimal (table names come from sqlite_master), adding an allowlist is the right approach for secure coding practices.

PR #264, opened 2026-02-01
Refactor: Remove duplicate GTFS source field from Manager

sreenivasivbieb:
## Summary
This PR removes the duplicate `gtfsSource` field from the `Manager` struct and uses `config.GtfsURL` as the single source of truth.

## Problem
We currently define the GTFS source in two places:
- `Manager.gtfsSource` (string field)
- `Manager.config.GtfsURL` (Config field)

Both represent the same concept, creating ambiguity about which is authoritative. This duplication leads to:
- Potential inconsistency when updating the source
- Extra memory usage
- Confusion about which field to use
- Tests needing to update both fields

## Changes
- ✅ Removed `gtfsSource` field from `Manager` struct
- ✅ Updated all references to use `config.GtfsURL` consistently
- ✅ Updated initialization to not set duplicate field
- ✅ Fixed all logging statements to use `config.GtfsURL`
- ✅ Updated all tests to modify `config.GtfsURL` directly

## Impact
- Single source of truth for GTFS source URL
- No synchronization issues between fields
- Simpler, cleaner code
- Easier to maintain and understand
- Reduced memory footprint

## Testing
- ✅ All existing tests pass
- ✅ Hot swap tests updated and working
- ✅ ForceUpdate functionality unchanged

### DIFF
--- a/internal/gtfs/gtfs_manager.go
+++ b/internal/gtfs/gtfs_manager.go
@@ -23,7 +23,7 @@ const NoRadiusLimit = -1
 
 // Manager manages the GTFS data and provides methods to access it
 type Manager struct {
-	gtfsSource                     string
+	
 	gtfsData                       *gtfs.Static
 	GtfsDB                         *gtfsdb.Client
 	lastUpdated                    time.Time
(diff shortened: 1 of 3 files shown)


aaronbrethorst:
this was fixed in #263

Issue #273, opened 2026-02-01
[Security] Path Traversal Vulnerability in Marketing Handler


3rabiii:
### Description
A potential path traversal vulnerability exists in the `marketingHandler` function within `internal/webui/static_handler.go`.

The current implementation relies solely on `filepath.Base(r.URL.Path)`, which is insufficient to prevent directory traversal attacks in all environments. Additionally, there is no validation of file extensions, potentially allowing access to non-static files.

### Impact
Attackers could potentially access sensitive files outside the intended `./marketing` directory by manipulating the URL path (e.g., using `..` or encoded characters).

### Suggested Fix
*  **Whitelist** allowed file extensions (e.g., .html, .css, .js, .png).
*  **Sanitize** the filename to explicitly reject `..` and path separators.
*  **Resolve** the absolute path and verify it remains within the intended directory using `filepath.Abs` and `strings.HasPrefix`.

PR #274, opened 2026-02-01
fix(security): prevent path traversal in marketing handler

3rabiii:
### Description
This PR mitigates a potential **Path Traversal** vulnerability in the `marketingHandler`.

### The Issue
The previous implementation relied solely on `filepath.Base`, which might be bypassed to access files outside the intended directory. It also lacked validation for file types.

### The Fix
Implemented a 3-layer defense strategy:
1.  **Extension Whitelisting:** Only allows specific static assets (.html, .css, .js, images).
2.  **Input Sanitization:** Rejects filenames containing `..` or path separators.
3.  **Path Resolution:** Resolves the absolute path and verifies it stays within the `./marketing` directory using `filepath.Abs` and `strings.HasPrefix`.

@aaronbrethorst 
fixes : #273

### DIFF
--- a/internal/webui/static_handler_test.go
+++ b/internal/webui/static_handler_test.go
@@ -70,7 +70,7 @@ func TestMarketingHandler_PathTraversal(t *testing.T) {
 		{
 			name:       "backslash traversal",
 			path:       "/marketing/..\\marketing-secret\\secret.html",
-			wantStatus: http.StatusBadRequest,
+			wantStatus: http.StatusNotFound,
 		},
 		{
 			name:       "disallowed extension",
@@ -91,20 +91,25 @@ func TestMarketingHandler_PathTraversal(t *testing.T) {
 
 			webUI.marketingHandler(rr, req)
 
-			currentStatus := rr.Code
+			got := rr.Code
 
-			if tt.wantStatus == http.StatusOK && currentStatus == http.StatusMovedPermanently {
-				currentStatus = http.StatusOK
+			if got == tt.wantStatus {
+				return
 			}
 
-			if tt.name == "null byte injection" && currentStatus == http.StatusInternalServerError {
-				currentStatus = tt.wantStatus
+			if tt.name == "valid file access" && got == http.StatusMovedPermanently {
+				return
 			}
 
-			if currentStatus != tt.wantStatus {
-				t.Errorf("handler returned wrong status code: got %v want %v",
-					rr.Code, tt.wantStatus)
+			if tt.name == "backslash traversal" && (got == http.StatusBadRequest || got == http.StatusNotFound) {
+				return
 			}
+
+			if tt.name == "null byte injection" && got == http.StatusInternalServerError {
+				return
+			}
+
+			t.Errorf("handler returned wrong status code: got %v want %v", got, tt.wantStatus)
 		})
 	}
 }


aaronbrethorst (changes requested):
Nice work on implementing defense-in-depth for the path traversal vulnerability, Adel. The 3-layer approach (extension whitelist, input sanitization, path resolution) shows good security thinking. However, there's one critical issue with the path containment check that needs to be fixed before this can be merged.

## Critical Issues

### 1. `strings.HasPrefix` is Unsafe for Path Containment Checks

**File:** `internal/webui/static_handler.go:45`

The current check:
```go
if !strings.HasPrefix(absPath, marketingDir) {
```

This can be bypassed if a sibling directory exists with a name that starts with "marketing". For example:
- `marketingDir` resolves to `/app/marketing`
- An attacker could potentially access `/app/marketing-secret/passwords.html`
- `strings.HasPrefix("/app/marketing-secret/passwords.html", "/app/marketing")` returns `true`

**Fix Option 1** - Add path separator:
```go
if !strings.HasPrefix(absPath, marketingDir+string(filepath.Separator)) {
```

**Fix Option 2** - Use `filepath.Rel` (canonical approach):
```go
relPath, err := filepath.Rel(marketingDir, absPath)
if err != nil || strings.HasPrefix(relPath, "..") {
    http.Error(w, "File not found", http.StatusNotFound)
    return
}
```

## Important Issues

### 2. Inconsistent Path References

**File:** `internal/webui/static_handler.go:50-55`

The code computes `absPath` from `filePath` but then uses `filePath` (the relative path) for `os.Stat` and `http.ServeFile`. While this should work correctly, it would be cleaner and safer to use `absPath` consistently after validating it:

```go
// Current:
if _, err := os.Stat(filePath); os.IsNotExist(err) { ... }
http.ServeFile(w, r, filePath)

// Recommended - use the validated path:
if _, err := os.Stat(absPath); os.IsNotExist(err) { ... }
http.ServeFile(w, r, absPath)
```

This ensures the validated path is exactly the path used for file operations.

## Fit and Finish

### 3. Missing Unit Tests

Security-critical code should include tests that verify the protection works. Consider adding:

```go
func TestMarketingHandler_PathTraversal(t *testing.T) {
    tests := []struct {
        name       string
        path       string
        wantStatus int
    }{
        {"valid file", "/marketing/index.html", http.StatusOK},
        {"path traversal", "/marketing/../../../etc/passwd", http.StatusNotFound},
        {"invalid extension", "/marketing/config.json", http.StatusNotFound},
        {"double dots in name", "/marketing/..file.html", http.StatusBadRequest},
    }
    // ... test implementation
}
```

## What's Done Well

The implementation has several strong points:

1. **`filepath.Base()` usage** - Correctly extracts only the filename, stripping directory components. Strong first line of defense.

2. **Extension whitelist** - Properly restricts to known safe static file types with case-insensitive matching.

3. **Defense-in-depth** - Multiple layers of validation, so a bypass of one layer doesn't compromise security.

4. **Route pattern** - `GET /marketing/{file}` only matches single path segments, providing router-level protection.

## Attack Vectors Analyzed

| Attack Vector | Protected? | How |
|--------------|------------|-----|
| `../../../etc/passwd` | Yes | `filepath.Base` strips to `passwd`, fails extension check |
| URL-encoded traversal | Yes | Go decodes first, `filepath.Base` strips |
| `/marketing/subdir/file.html` | Yes | Router pattern `{file}` only matches single segment |
| Null byte injection | Yes | Go handles safely, extension check still applies |
| `marketing-private/secret.html` | **No** | `strings.HasPrefix` bypass (Critical issue #1) |

## Required Actions

1. Fix the `strings.HasPrefix` vulnerability (Critical)
2. Consider using `absPath` consistently for file operations
3. Optional: Add unit tests for path traversal protection

## Checklist

- [x] `make lint` - passes
- [x] `make test` - passes
- [ ] Security fix complete - needs `strings.HasPrefix` fix

aaronbrethorst (approved):
Hey Adel! Excellent work addressing all the security feedback. The implementation is now solid.

## Previous Issues - Status

### ✅ Critical: `strings.HasPrefix` Sibling Directory Bypass - FIXED

You correctly switched to using `filepath.Rel`:

```go
rel, err := filepath.Rel(marketingDir, absPath)
if err != nil || strings.HasPrefix(rel, "..") {
    slog.Warn("potential path traversal attempt blocked", "path", absPath)
    http.Error(w, "File not found", http.StatusNotFound)
    return
}
```

This properly prevents the `marketing-secret` sibling directory attack because:
- `filepath.Rel("/app/marketing", "/app/marketing-secret/file.html")` returns `"../marketing-secret/file.html"`
- The `strings.HasPrefix(rel, "..")` check catches this

### ✅ Important: Inconsistent Path References - FIXED

Now correctly uses `absPath` for all file operations:
- `os.Stat(absPath)` at line 53
- `http.ServeFile(w, r, absPath)` at line 59

### ✅ Missing Unit Tests - FIXED

Added comprehensive security tests in `static_handler_test.go`:

| Test Case | Attack Vector | Verified |
|-----------|--------------|----------|
| `valid file access` | Legitimate request | ✅ |
| `path traversal attempt` | `../../../etc/passwd` | ✅ |
| `sibling directory bypass (Critical)` | `../marketing-secret/secret.html` | ✅ |
| `double encoded traversal` | `%2e%2e/` | ✅ |
| `backslash traversal` | `..\` | ✅ |
| `disallowed extension` | `.json` file | ✅ |
| `null byte injection` | `file.html%00.png` | ✅ |

Good handling of platform-specific behavior (Windows backslash) in the test assertions.

## Security Analysis

The final implementation has robust defense-in-depth:

1. **`filepath.Base()`** - Strips directory components from input
2. **Extension whitelist** - Only allows known-safe static file types
3. **Input validation** - Rejects `..` and path separators in filename
4. **`filepath.Rel()` check** - Canonical containment verification
5. **Directory rejection** - `stat.IsDir()` check prevents directory listing
6. **Security logging** - `slog.Warn` for blocked traversal attempts

---

Great security work, Adel!

Issue #275, opened 2026-02-01
[Security] Error Message Leakage in Debug Handler


3rabiii:
### Description
A potential information disclosure vulnerability exists in the `writeDebugData` function within `internal/webui/debug_index_handler.go`.

The code directly returns raw `err.Error()` messages to the HTTP response when template parsing or execution fails. This behavior can leak internal server details (e.g., file paths, stack traces) to the client.

### Impact
Leaking internal error details can help attackers map the application structure or identify other vulnerabilities.

### Suggested Fix
1.  Replace raw error output with a generic "Internal Server Error" message for the client.
2.  Log the actual error details server-side using `slog` for debugging purposes.

PR #276, opened 2026-02-01
fix(security): prevent error detail leakage in debug handler

3rabiii:
### Description
This PR mitigates a potential **Information Disclosure** issue in `writeDebugData`.

### The Issue
The code was returning raw `err.Error()` messages directly to the HTTP response. This could leak internal implementation details if an error occurs.

### The Fix
* Replaced raw error output with a generic "Internal Server Error" message.
* Added server-side structured logging (`slog.Error`) to capture the actual error details for administrators.

@aaronbrethorst 
fixes : #275

### DIFF
--- a/internal/webui/debug_index_handler.go
+++ b/internal/webui/debug_index_handler.go
@@ -3,6 +3,7 @@ package webui
 import (
 	"embed"
 	"html/template"
+	"log/slog"
 	"net/http"
 
 	"github.com/davecgh/go-spew/spew"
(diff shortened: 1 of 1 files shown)


aaronbrethorst (approved):
Clean, focused security fix, Adel. Replacing raw `err.Error()` output with generic messages while logging the actual errors server-side is the right approach. This prevents potential information disclosure without losing visibility into what went wrong.

## What Changed

The PR modifies two error handling paths in `writeDebugData()`:

1. **Template parse error (line 25-29):** Now logs with `slog.Error` and returns generic message
2. **Template execute error (line 38-40):** Same pattern

Both changes follow the same secure pattern:
```go
slog.Error("descriptive message", "error", err)
http.Error(w, "Internal Server Error", http.StatusInternalServerError)
```

## Fit and Finish

### Minor: Inconsistent Capitalization

**File:** `internal/webui/debug_index_handler.go`

The two error messages have slightly different capitalization:
- Line 28: `"Internal server error"` (lowercase 's')
- Line 40: `"Internal Server Error"` (uppercase 'S')

This is cosmetic, but consistency would be nice. The standard HTTP status text is "Internal Server Error" (title case), so line 28 could be updated to match.

----

This is ready to merge. The capitalization inconsistency is minor and doesn't block the PR - it can be fixed in a follow-up if desired.

PR #278, opened 2026-02-01
Fix: Allow Future Date Queries in API

### LINKED ISSUE #277: API Incorrectly Rejects Future Date Queries
## Summary
The API currently blocks all queries for future dates (tomorrow and beyond), preventing users from planning trips in advance. This violates expected OneBusAway API behavior where future scheduling is a core feature.

## Impact
- ❌ Users cannot query schedules for tomorrow or any future date
- ❌ Trip planning functionality is effectively broken
- ❌ Non-compliant with standard OneBusAway API expectations

## Root Cause
**File**: `internal/utils/api.go`  
**Lines**: 85-91  
**Function**: `ParseTimeParameter`

The function contains validation logic that explicitly rejects any date after today's midnight.

## Reproduction
```bash
# Query for tomorrow's schedule - this will fail
curl "http://localhost:4000/api/where/schedule-for-stop/{id}?key=test&date=2026-02-02"

# Returns 400 error: "Invalid field value for field \"time\"."
```

## Proposed Fix
Remove lines 85-91 from `internal/utils/api.go` that check `if parsedTimeMidnight.After(todayMidnight)` and return an error.

The function should accept future dates and let the GTFS data availability determine what can be returned.

## Test Updates Required
**File**: `internal/utils/api_test.go`

1. Update the test case named "Future date (should fail)" to expect success instead of failure
2. Add a new test case `TestParseTimeParameter_FutureDate` to verify future dates are accepted and prevent regression

## Verification
```bash
make test
go test ./internal/utils -run TestParseTimeParameter
```

ARCoder181105:
## Description
Removes the validation that prevented users from querying transit schedules for future dates. This fix enables proper trip planning functionality by allowing queries for tomorrow and beyond.

## Problem
The API was incorrectly rejecting all date queries for any date after today, blocking a core use case: planning trips in advance. This behavior was inconsistent with standard OneBusAway API expectations where future scheduling is essential.

### Example of Previous Behavior
```bash
# Query for tomorrow's schedule - would fail with 400 error
curl "http://localhost:4000/api/where/schedule-for-stop/25_123?key=test&date=2026-02-02"

# Response: 400 Bad Request - "Invalid field value for field 'time'"
```

### `internal/utils/api.go`
- Removed lines 85-91 in `ParseTimeParameter` function that validated dates against today's midnight
- Function now accepts any valid date format (past, present, or future)
- Date availability is determined by GTFS data, not arbitrary validation

### `internal/utils/api_test.go`
- Updated test case "Future date (should fail)" → "Future date" to expect success
- Updated test case "Future epoch (should fail)" → "Future epoch" to expect success
- Added `validateParsedTime` functions to both tests to verify correct date parsing
- Tests now verify that future dates are properly accepted and parsed

## Testing
- ✅ All existing tests pass
- ✅ Future date tests now verify correct behavior
- ✅ Edge cases (today, yesterday, invalid formats) still work correctly
- ✅ `make test` passes
- ✅ `make lint` passes

## Impact
- ✅ Users can now query schedules for future dates
- ✅ Trip planning functionality is restored
- ✅ API now complies with OneBusAway standard behavior
- ✅ No breaking changes to existing functionality

## Related Issue
Fixes #277

### DIFF
--- a/internal/utils/api.go
+++ b/internal/utils/api.go
@@ -115,18 +115,6 @@ func ParseTimeParameter(timeParam string, currentLocation *time.Location) (strin
 		return "", time.Time{}, fieldErrors, false
 	}
 
-	// Set time to midnight for accurate comparison
-	currentTime := time.Now().In(currentLocation)
-	todayMidnight := time.Date(currentTime.Year(), currentTime.Month(), currentTime.Day(), 0, 0, 0, 0, currentLocation)
-	parsedTimeMidnight := time.Date(parsedTime.Year(), parsedTime.Month(), parsedTime.Day(), 0, 0, 0, 0, currentLocation)
-
-	if parsedTimeMidnight.After(todayMidnight) {
-		fieldErrors := map[string][]string{
-			"time": {"Invalid field value for field \"time\"."},
-		}
-		return "", time.Time{}, fieldErrors, false
-	}
-
 	// Valid date, use it
 	return parsedTime.Format("20060102"), parsedTime, nil, true
 }
--- a/internal/utils/api_test.go
+++ b/internal/utils/api_test.go
@@ -444,18 +444,28 @@ func TestParseTimeParameter(t *testing.T) {
 			expectedErrorKey: "time",
 		},
 		{
-			name:             "Future date (should fail)",
-			timeParam:        now.AddDate(0, 0, 1).Format("2006-01-02"),
-			expectedDate:     "",
-			expectError:      true,
-			expectedErrorKey: "time",
+			name:         "Future date",
+			timeParam:    now.AddDate(0, 0, 1).Format("2006-01-02"),
+			expectedDate: now.AddDate(0, 0, 1).Format("20060102"),
+			expectError:  false,
+			validateParsedTime: func(t *testing.T, parsedTime time.Time) {
+				tomorrow := now.AddDate(0, 0, 1)
+				assert.Equal(t, tomorrow.Year(), parsedTime.Year())
+				assert.Equal(t, tomorrow.Month(), parsedTime.Month())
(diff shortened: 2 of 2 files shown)


aaronbrethorst (changes requested):
Great catch, Aditya! The future date restriction was indeed blocking a core transit API use case - nobody wants to check bus schedules only for today when they're planning tomorrow's commute. Removing this overly-restrictive validation while keeping format validation intact is the right approach.

## Required Fix

### Test Case Names Still Say "(should fail)"

**File:** `internal/utils/api_test.go:447, 458`

The test names weren't updated to reflect the new expected behavior:
```go
name: "Future date (should fail)",   // But now expects success
name: "Future epoch (should fail)",  // But now expects success
```

Should be:
```go
name: "Future date",
name: "Future epoch",
```

This is a small fix - just remove the "(should fail)" suffix from both test names.

-----

Please update the two test case names to remove "(should fail)" before merging. The core fix is correct and well-tested - just need this small cleanup.

ARCoder181105:
@aaronbrethorst 
I updated the branch, PR is ready for review

aaronbrethorst (changes requested):
Hey Aditya, you've addressed the test-name rename I asked for — both "Future date" and "Future epoch" no longer carry the misleading "(should fail)" suffix, and the new `validateParsedTime` callbacks are a nice addition that verify the parsed year/month/day round-trips correctly. Before we can merge this, I need you to make one more change:

## Important

### 1. Squash and rebase

The branch currently has four commits (two feature commits and two merge commits) and is six commits behind `main`. Please squash everything into a single atomic commit and rebase against `main`. If rebasing feels intimidating, I wrote a short guide that might help: https://www.brethorsting.com/blog/2026/01/git-rebase-for-the-terrified/

Issue #279, opened 2026-02-01
feat: implement arrivals-and-departures-for-stop handler with performance optimizations


3rabiii:
### Description
This issue tracks the implementation of the `arrivalsAndDeparturesForStop` REST API endpoint. This endpoint provides clients with scheduled and real-time arrival data for a specific stop.

### Implemented Features
* **Core Logic:** Retrieves static stop times and merges them with real-time GTFS-RT data (VehiclePositions and TripUpdates).
* **Filtering:** Supports `minutesBefore` and `minutesAfter` parameters to define the time window.
* **Nearby Stops:** Calculates and returns a list of nearby stops based on geospatial location.

### Performance Optimizations 
* **N+1 Query Fix:** Implemented in-memory caching (`map`) for `Route` and `Trip` lookups within the processing loop. This drastically reduces database round-trips for high-traffic stops.
* **Active Service Filtering:** Logic added to filter trips based on `activeServiceIDs` to ensure only valid trips for the current day are processed.

### Fixes
* Corrected the function signature call for `calc.CalculateStopDirection` by passing the missing `direction` argument.

Issue #260, opened 2026-02-01
Add debug logging for silent error handling in arrivals endpoint


aaronbrethorst:
The `/arrivals-and-departures-for-stop` endpoint silently discards errors in several places, which can make debugging production issues difficult. When database queries fail or data integrity issues exist, arrivals silently disappear from results without any indication why.

### Affected Code

**File:** `internal/restapi/arrivals_and_departure_for_stop.go`

**1. Route/Trip lookup loop (lines 152-160)**
```go
route, err := api.GtfsManager.GtfsDB.Queries.GetRoute(ctx, st.RouteID)
if err != nil {
    continue  // Silent skip
}

trip, err := api.GtfsManager.GtfsDB.Queries.GetTrip(ctx, st.TripID)
if err != nil {
    continue  // Silent skip
}
```

**2. Routes for stop lookup (line 340)**
```go
routesForThisStop, _ := api.GtfsManager.GtfsDB.Queries.GetRoutesForStops(ctx, []string{stopID})
```

**3. Stop reference lookup (lines 334-338)**
```go
stopData, err := api.GtfsManager.GtfsDB.Queries.GetStop(ctx, stopID)
if err != nil {
    continue  // Silent skip
}
```

### Impact

- Arrivals may silently disappear from API responses
- Data integrity issues (orphaned stop_times) go undetected
- Operators have no visibility into why data might be missing

### Proposed Solution

Add debug-level logging to make failures visible:

```go
route, err := api.GtfsManager.GtfsDB.Queries.GetRoute(ctx, st.RouteID)
if err != nil {
    api.Logger.Debug("skipping stop time: route not found",
        slog.String("routeID", st.RouteID),
        slog.String("tripID", st.TripID),
        slog.Any("error", err))
    continue
}
```

This preserves the current behavior (continue processing other arrivals) while providing observability for debugging.

### Context

Identified during review of PR #254.

PR #280, opened 2026-02-01
feat: implement optimized arrivals-and-departures handler

3rabiii:
### Description
This PR implements the `arrivalsAndDeparturesForStopHandler` endpoint, providing scheduled and real-time arrival data for a specific stop.

### Key Features
* **Time Window Logic:** Filters stop times based on `minutesBefore` and `minutesAfter` parameters.
* **Real-time Integration:** Merges GTFS-RT data (VehiclePositions & TripUpdates) with static schedules.
* **Nearby Stops:** Returns a list of nearby stops using geospatial queries.

### Performance Optimizations 
* **N+1 Query Fix:** Implemented in-memory caching (`map`) for `Route` and `Trip` lookups inside the processing loop, significantly reducing database hits.
* **Active Service Filtering:** Added logic to filter trips by `activeServiceIDs` to ensure only valid trips for the current day are processed.

### Fixes
* Corrected the `calc.CalculateStopDirection` function call by passing the missing `direction` argument.

@aaronbrethorst 
fixes : #279

### DIFF
--- a/internal/restapi/arrivals_and_departure_for_stop.go
+++ b/internal/restapi/arrivals_and_departure_for_stop.go
@@ -184,6 +184,15 @@ func (api *RestAPI) arrivalsAndDeparturesForStopHandler(w http.ResponseWriter, r
 	// Add the current stop
 	stopIDSet[stop.ID] = true
 
+	serviceDateMillis := params.Time.UnixMilli()
+	serviceMidnight := time.Date(
+		params.Time.Year(),
+		params.Time.Month(),
+		params.Time.Day(),
+		0, 0, 0, 0,
+		loc,
+	)
+
 	for _, st := range stopTimes {
 		var route gtfsdb.Route
 		if cachedRoute, exists := routeIDSet[st.RouteID]; exists {
@@ -216,16 +225,6 @@ func (api *RestAPI) arrivalsAndDeparturesForStopHandler(w http.ResponseWriter, r
 			trip = fetchedTrip
 			tripIDSet[trip.ID] = &trip
 		}
-
-		serviceDateMillis := params.Time.UnixMilli()
-
-		serviceMidnight := time.Date(
-			params.Time.Year(),
-			params.Time.Month(),
-			params.Time.Day(),
-			0, 0, 0, 0,
-			loc,
-		)
 
 		scheduledArrivalTime := serviceMidnight.Add(time.Duration(st.ArrivalTime)).UnixMilli()
 		scheduledDepartureTime := serviceMidnight.Add(time.Duration(st.DepartureTime)).UnixMilli()
@@ -318,7 +317,12 @@ func (api *RestAPI) arrivalsAndDeparturesForStopHandler(w http.ResponseWriter, r
 						// Check cache for active trip
 						if _, exists := tripIDSet[activeTripID]; !exists {
 							activeTrip, err := api.GtfsManager.GtfsDB.Queries.GetTrip(ctx, activeTripID)
-							if err == nil {
+							if err != nil {
+								api.Logger.Debug("skipping active trip reference: trip not found",
+									slog.String("activeTripID", activeTripID),
+									slog.String("scheduledTripID", st.TripID),
+									slog.Any("error", err))
+							} else {
 								tripIDSet[activeTrip.ID] = &activeTrip
 							}
 						}
@@ -335,6 +339,9 @@ func (api *RestAPI) arrivalsAndDeparturesForStopHandler(w http.ResponseWriter, r
 		tripStopTimes, err := api.GtfsManager.GtfsDB.Queries.GetStopTimesForTrip(ctx, st.TripID)
 		var totalStopsInTrip int
 		if err != nil {
+			api.Logger.Debug("failed to get stop times for trip",
+				slog.String("tripID", st.TripID),
+				slog.Any("error", err))
 			totalStopsInTrip = 0
 		} else {
 			totalStopsInTrip = len(tripStopTimes)
--- a/internal/restapi/http_test.go
+++ b/internal/restapi/http_test.go
@@ -78,6 +78,7 @@ func createTestApiWithClock(t testing.TB, c clock.Clock) *RestAPI {
 	}
 
 	api := NewRestAPI(application)
+	api.Logger = slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{Level: slog.LevelDebug}))
 
 	return api
 }


3rabiii:
Resolved merge conflicts with main. The PR is ready for review.
@aaronbrethorst

aaronbrethorst (changes requested):
Hey Adel, you're making great progress on the optimized arrivals-and-departures handler. The duplicate parameter parsing issue from my previous review is properly resolved, and the N+1 query caching is a solid performance improvement. Before we can merge this, I need you to make a couple more changes:

## Critical

### 1. Restore debug logging removed from route/trip cache-miss paths

**File:** `internal/restapi/arrivals_and_departure_for_stop.go`, lines 197-199 and 209-211

PR #260 (commit `babd90b`) specifically added debug logging for these silent `continue` statements to aid in diagnosing missing arrivals. This PR removes that logging as part of the caching refactor:

```go
// Current (PR #280) — no logging
r, err := api.GtfsManager.GtfsDB.Queries.GetRoute(ctx, st.RouteID)
if err != nil {
    continue
}
```

Please restore the debug logging in both the route and trip cache-miss error paths:

```go
fetchedRoute, err := api.GtfsManager.GtfsDB.Queries.GetRoute(ctx, st.RouteID)
if err != nil {
    api.Logger.Debug("skipping stop time: route not found",
        slog.String("routeID", st.RouteID),
        slog.String("tripID", st.TripID),
        slog.Any("error", err))
    continue
}
route = fetchedRoute
routeIDSet[route.ID] = &route
```

(Same pattern for the trip lookup.)

### 2. Variable `r` shadows the HTTP request parameter

**File:** `internal/restapi/arrivals_and_departure_for_stop.go`, line 197

The route query result is assigned to `r`, which shadows the handler's `r *http.Request` parameter. While it doesn't cause a bug today, it's a maintenance hazard — anyone adding code in that block that references `r` expecting the HTTP request will silently get a `gtfsdb.Route` instead. Note that you correctly used `t` for the trip lookup to avoid this issue.

Fix: rename `r` to `fetchedRoute` or similar (as shown in the logging fix above).

## Important

### 3. Handle zero/negative values for `minutesAfter` and `minutesBefore`

**File:** `internal/restapi/arrivals_and_departure_for_stop.go`, lines 46-51 and 58-63

When a user passes `minutesAfter=0` or `minutesBefore=-5`, the value parses as a valid integer but falls through both conditions (`> maxMinutesAfter` and `> 0`), silently keeping the default (35 and 5 respectively). This means `?minutesAfter=0` returns 35 minutes of data — the opposite of what was requested.

Either allow zero as valid (`>= 0` instead of `> 0`), or return a validation error for non-positive values:

```go
if minutes > maxMinutesAfter {
    params.MinutesAfter = maxMinutesAfter
} else if minutes >= 0 {
    params.MinutesAfter = minutes
} else {
    addError("minutesAfter", "must be a non-negative integer")
}
```

### 4. Fix error-before-use pattern in `GetStopTimesForTrip`

**File:** `internal/restapi/arrivals_and_departure_for_stop.go`, lines 332-336

```go
tripStopTimes, err := api.GtfsManager.GtfsDB.Queries.GetStopTimesForTrip(ctx, st.TripID)
totalStopsInTrip := len(tripStopTimes)
if err != nil {
    totalStopsInTrip = 0
}
```

The result is used before the error is checked. While `len(nil)` happens to return 0 in Go, this is a Go anti-pattern and obscures any DB errors. Please restructure:

```go
tripStopTimes, err := api.GtfsManager.GtfsDB.Queries.GetStopTimesForTrip(ctx, st.TripID)
if err != nil {
    totalStopsInTrip = 0
} else {
    totalStopsInTrip = len(tripStopTimes)
}
```

## Fit and Finish

### 5. Simplify the stop time filtering loop

**File:** `internal/restapi/arrivals_and_departure_for_stop.go`, lines 168-182

The filtering loop copies each `GetStopTimesForStopInWindowRow` field-by-field. Since you're copying the entire struct, just append the value directly:

```go
var stopTimes []gtfsdb.GetStopTimesForStopInWindowRow
for _, st := range allStopTimes {
    if activeTripIDs[st.TripID] {
        stopTimes = append(stopTimes, st)
    }
}
```

This is functionally identical but much more concise and won't break if fields are added to the struct in the future.

aaronbrethorst (changes requested):
Hey Adel, great work addressing all the feedback from the previous two rounds of review. The duplicate parameter parsing is properly consolidated, the debug logging is restored, the variable shadowing is fixed, the zero/negative handling works correctly, and the error-before-use pattern is resolved. The caching implementation continues to look solid. Just a couple small things left and we'l be ready to merge:

## Important

### 1. Add debug logging for active trip lookup failure

**File:** `internal/restapi/arrivals_and_departure_for_stop.go`, lines 319-324

When the active trip (from a block reassignment) fails to load, the error is silently swallowed with no logging. This is inconsistent with the route and trip cache-miss paths ~100 lines earlier, which both log at Debug level before skipping. Missing trip references can cause downstream rendering issues in mobile apps, and without logging, this will be very hard to diagnose.

```go
// Current
if _, exists := tripIDSet[activeTripID]; !exists {
    activeTrip, err := api.GtfsManager.GtfsDB.Queries.GetTrip(ctx, activeTripID)
    if err == nil {
        tripIDSet[activeTrip.ID] = &activeTrip
    }
}
```

Please add debug logging on the error path to match the established pattern:

```go
if _, exists := tripIDSet[activeTripID]; !exists {
    activeTrip, err := api.GtfsManager.GtfsDB.Queries.GetTrip(ctx, activeTripID)
    if err != nil {
        api.Logger.Debug("skipping active trip reference: trip not found",
            slog.String("activeTripID", activeTripID),
            slog.String("scheduledTripID", st.TripID),
            slog.Any("error", err))
    } else {
        tripIDSet[activeTrip.ID] = &activeTrip
    }
}
```

### 2. Add debug logging for `GetStopTimesForTrip` failure

**File:** `internal/restapi/arrivals_and_departure_for_stop.go`, lines 335-341

The restructured error handling correctly checks the error before using the result (thank you for fixing that), but the error path has no logging. Since we already know this trip has stop times (we found it via `GetStopTimesForStopInWindow`), a failure here likely indicates a database issue worth surfacing:

```go
tripStopTimes, err := api.GtfsManager.GtfsDB.Queries.GetStopTimesForTrip(ctx, st.TripID)
var totalStopsInTrip int
if err != nil {
    api.Logger.Debug("failed to get stop times for trip",
        slog.String("tripID", st.TripID),
        slog.Any("error", err))
    totalStopsInTrip = 0
} else {
    totalStopsInTrip = len(tripStopTimes)
}
```

## Fit and Finish

### 3. Hoist loop-invariant computations

**File:** `internal/restapi/arrivals_and_departure_for_stop.go`, lines 220-228

`serviceDateMillis` and `serviceMidnight` depend only on `params.Time` and `loc`, which don't change within the loop. Move them before the `for _, st := range stopTimes` loop to avoid recomputing on every iteration:

```go
serviceDateMillis := params.Time.UnixMilli()
serviceMidnight := time.Date(
    params.Time.Year(), params.Time.Month(), params.Time.Day(),
    0, 0, 0, 0, loc,
)

for _, st := range stopTimes {
    // ... use serviceDateMillis and serviceMidnight ...
}
```

PR #282, opened 2026-02-01
perf: hoist DB calls out of realTimeMutex in GetAlertsForTrip

### LINKED ISSUE #281: Performance: `GetAlertsForTrip` holds `realTimeMutex` during DB I/O, blocking updates
**Description**
The `GetAlertsForTrip` function acquires a global read lock (`RLock`) on `realTimeMutex` *before* fetching data from the database. Because the real-time updater requires a Write lock on this same mutex, any latency in database operations (e.g., `GetTrip`) directly blocks the ingestion of fresh real-time data.

**Location**

* **File:** `internal/gtfs/realtime.go`
* **Function:** `GetAlertsForTrip`

**Current Behavior**

1. Acquire `realTimeMutex.RLock()`.
2. Perform blocking DB calls (`GetTrip`, `GetRoute`) while holding the lock.
3. Filter in-memory alerts.
4. Release lock.

**Expected Behavior**

1. Perform blocking DB calls (`GetTrip`, `GetRoute`) *without* the lock.
2. Acquire `realTimeMutex.RLock()`.
3. Filter in-memory alerts.
4. Release lock.

**Impact**
High read traffic or slow database queries cause the real-time data feed to stall, preventing vehicle positions and trip updates from being processed.

**Suggested Fix**
Refactor `GetAlertsForTrip` to fetch the Trip, Route, and Agency IDs from the database before acquiring the mutex.

ARCoder181105:
**Description:**
This PR fixes a concurrency bug where blocking database I/O was performed while holding the global `realTimeMutex`.

Closes #281 

**Motivation:**
Previously, `GetAlertsForTrip` acquired a read lock before querying the database. This caused the real-time updater (which requires a write lock) to starve during high read traffic or slow database queries, leading to stale vehicle data.

**Changes:**
Refactored `GetAlertsForTrip` in `internal/gtfs/realtime.go` to:

1. Fetch static data (Trip, Route, Agency) from the database *before* acquiring the lock.
2. Hold the `RLock` only during the in-memory filtering of alerts.

**Verification:**

* Logic verified to ensure correct alert filtering with pre-fetched IDs.
* Existing tests pass.

### DIFF
--- a/internal/gtfs/gtfs_manager.go
+++ b/internal/gtfs/gtfs_manager.go
@@ -23,7 +23,6 @@ const NoRadiusLimit = -1
 
 // Manager manages the GTFS data and provides methods to access it
 type Manager struct {
-	gtfsSource                     string
 	gtfsData                       *gtfs.Static
 	GtfsDB                         *gtfsdb.Client
 	lastUpdated                    time.Time
@@ -56,7 +55,6 @@ func InitGTFSManager(config Config) (*Manager, error) {
 	}
 
 	manager := &Manager{
-		gtfsSource:                     config.GtfsURL,
 		isLocalFile:                    isLocalFile,
 		config:                         config,
 		shutdownChan:                   make(chan struct{}),
@@ -430,7 +428,7 @@ func (manager *Manager) GetAllTripUpdates() []gtfs.Trip {
 
 // IMPORTANT: Caller must hold manager.RLock() before calling this method.
 func (manager *Manager) PrintStatistics() {
-	fmt.Printf("Source: %s (Local File: %v)\n", manager.gtfsSource, manager.isLocalFile)
+	fmt.Printf("Source: %s (Local File: %v)\n", manager.config.GtfsURL, manager.isLocalFile)
 	fmt.Printf("Last Updated: %s\n", manager.lastUpdated)
 	fmt.Println("Stops Count: ", len(manager.gtfsData.Stops))
 	fmt.Println("Routes Count: ", len(manager.gtfsData.Routes))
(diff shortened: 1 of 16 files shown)


ARCoder181105:
@aaronbrethorst 
The test are passing successfully, this PR is ready for review

image

aaronbrethorst (changes requested):
Smart optimization, Aditya. Moving blocking database I/O outside the `realTimeMutex` is the right call - holding a lock during slow operations like database queries is a classic recipe for contention and stale data. The change itself is correct and safe, but there's a misleading comment that should be fixed.

### Incorrect/Misleading Comment

**File:** `internal/gtfs/realtime.go:73`

The comment says:
```go
```

This comment is **incorrect** for this function:

1. `manager.RLock()` refers to `staticMutex` (protects `gtfsData`, `lastUpdated`, etc.)
2. `GetAlertsForTrip` doesn't access any data protected by `staticMutex`
3. It only accesses:
   - `manager.GtfsDB` - thread-safe (Go's `database/sql` guarantees)
   - `manager.realTimeAlerts` - protected by `realTimeMutex`, which the function acquires internally
4. Sibling functions `GetAlertsForRoute` and `GetAlertsForStop` have no such comment and follow the same pattern

**Fix:** Remove or correct the comment:
```go
// GetAlertsForTrip returns alerts matching the trip, its route, or agency.
func (manager *Manager) GetAlertsForTrip(ctx context.Context, tripID string) []gtfs.Alert {
```

----

Good optimization. Just fix the misleading comment and this is ready to merge.

Issue #283, opened 2026-02-01
[Code Quality] Ignored Error in Rate Limit Middleware


3rabiii:
### Description
In `internal/restapi/rate_limit_middleware.go`, the error returned by `json.NewEncoder(w).Encode()` is assigned to the blank identifier `_`, silently ignoring potential encoding or network failures.

### Impact
Silencing these errors makes debugging difficult, as issues like client disconnections or serialization failures go unnoticed in the logs.

### Suggested Fix
Check the error returned by `Encode()` and log it using `slog.Error` to ensure visibility.

PR #284, opened 2026-02-01
fix: log ignored json encoding errors in rate limit middleware

3rabiii:
### Description
The `rateLimitHandler` was silently ignoring errors returned by `json.NewEncoder(w).Encode()`.

### The Fix
Added error checking and structured logging (`slog.Error`) to capture any encoding failures or network issues during response writing. This improves system observability.

@aaronbrethorst 
fixes : #283

### DIFF
--- a/internal/restapi/rate_limit_middleware.go
+++ b/internal/restapi/rate_limit_middleware.go
@@ -2,6 +2,7 @@ package restapi
 
 import (
 	"encoding/json"
+	"log/slog"
 	"net/http"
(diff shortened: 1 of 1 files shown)


aaronbrethorst (approved):
Good catch, Adel. Silently ignoring errors with `_ =` is exactly the kind of thing that makes debugging production issues painful. Adding structured logging here means you'll actually know when something goes wrong with response encoding.

This is a small but meaningful improvement to observability.

Issue #285, opened 2026-02-01
[Bug] Malformed BlockID and ShapeID in Trip Response


3rabiii:
### Description
In `internal/restapi/trip_handler.go`, `utils.FormCombinedID` is called on `BlockID` and `ShapeID` fields without checking if the underlying `sql.NullString` is valid.

### The Issue
If the field is `NULL` in the database, `.String` returns an empty string. `FormCombinedID` then concatenates the agency ID with this empty string, resulting in malformed IDs like `"agency_"` instead of an empty string or null.

### Suggested Fix
Check `.Valid` property before formatting the ID.

PR #286, opened 2026-02-01
fix(logic): validate NullString fields in trip handler before formatting

3rabiii:
### Description
This PR fixes a logic issue in `tripHandler` where `BlockID` and `ShapeID` were being formatted even when they were NULL in the database.

### The Fix
Added explicit checks for `.Valid` on `sql.NullString` fields.
* If valid: Format the ID normally.
* If invalid: Keep it as an empty string.

This prevents malformed IDs (e.g., `agency_`) from appearing in the API response.

@aaronbrethorst 
fixes : #285

### DIFF
--- a/internal/restapi/trip_handler.go
+++ b/internal/restapi/trip_handler.go
@@ -56,13 +56,21 @@ func (api *RestAPI) tripHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	var blockID, shapeID string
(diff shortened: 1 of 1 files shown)


aaronbrethorst (approved):
Good fix, Adel. Without the `.Valid` check, NULL database values were producing malformed IDs like `"25_"` instead of empty strings. This is exactly how `sql.NullString` is meant to be used.

PR #288, opened 2026-02-02
Fix: populate missing direction value in stop response

### LINKED ISSUE #287: Missing value for 'direction' field in stop endpoint response
### Description:
The `direction` field in the stop endpoint response is currently returning an empty string, even when the underlying GTFS data contains a valid `direction` (e.g., "N", "S", "E"...). This results in a mismatch with the production OneBusAway server, which correctly populates this value.

### Expected behavior:
The response should include a correct value for the `direction` field that matches the production OneBusAway.

Image

uloamaka:
### Changes:
- Populated missing direction value: The stop endpoint now returns direction data, matching the OneBusAway production server implementation.
- Normalized nullable GTFS fields: Ensured consistent API responses by handling null values, aligning Maglev's behavior with the Java OBA production environment.
- Code Formatting: Applied go fmt to the project to maintain standard styling.

fixes: #287

### DIFF
--- a/internal/restapi/stop_handler.go
+++ b/internal/restapi/stop_handler.go
@@ -52,11 +52,11 @@ func (api *RestAPI) stopHandler(w http.ResponseWriter, r *http.Request) {
 
 	stopData := &models.Stop{
 		ID:                 utils.FormCombinedID(agencyID, stop.ID),
-		Name:               stop.Name.String,
+		Name:               utils.NullStringOrEmpty(stop.Name),
 		Lat:                stop.Lat,
 		Lon:                stop.Lon,
-		Code:               stop.Code.String,
-		Direction:          "",
+		Code:               utils.NullStringOrEmpty(stop.Code),
+		Direction:          utils.NullStringOrEmpty(stop.Direction),
 		LocationType:       int(stop.LocationType.Int64),
 		WheelchairBoarding: utils.MapWheelchairBoarding(utils.NullWheelchairBoardingOrUnknown(stop.WheelchairBoarding)),
 		RouteIDs:           combinedRouteIDs,
--- a/internal/restapi/stop_handler_test.go
+++ b/internal/restapi/stop_handler_test.go
@@ -57,6 +57,8 @@ func TestStopHandlerEndToEnd(t *testing.T) {
 	assert.Equal(t, models.UnknownValue, entry["wheelchairBoarding"])
 	assert.Equal(t, *stops[0].Latitude, entry["lat"])
 	assert.Equal(t, *stops[0].Longitude, entry["lon"])
+	
+	assert.Contains(t, entry, "direction", "direction field should exist")
 
 	routeIds, ok := entry["routeIds"].([]interface{})
 	assert.True(t, ok, "routeIds should exist and be an array")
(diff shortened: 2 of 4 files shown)


aaronbrethorst (changes requested):
Hey Godsgift! Thanks for tackling issue #287 - getting the direction field populated correctly in the stop endpoint is a valuable fix for API consistency.

## Important Issues

### 1. Inconsistent direction handling pattern

**Location:** `internal/restapi/stop_handler.go:59`

The stop handler now uses `utils.NullStringOrEmpty(stop.Direction)` which returns an empty string `""` when the direction is null. However, other handlers in the codebase use different patterns:

- `stops_for_agency_handler.go:101-104` uses `models.UnknownValue` ("UNKNOWN") as the fallback
- `stops_for_route_handler.go:247` uses `calc.CalculateStopDirection()` which can infer direction from shape geometry

This creates inconsistency: the same stop might return `""` from one endpoint and `"UNKNOWN"` or a calculated direction from another.

**Options to consider:**
1. **Keep as-is** - The OBA production server may return empty strings for stops without explicit direction data. If you've verified this matches production behavior, that's valid.
2. **Use `models.UnknownValue`** - Fall back to "UNKNOWN" when direction is null (matching `stops_for_agency_handler.go`)
3. **Use the direction calculator** - Leverage `calc.CalculateStopDirection()` to infer direction from shape geometry when available (matching `stops_for_route_handler.go`)

Could you clarify what the OBA production server returns for stops without direction data? That will help determine the right approach.

## Fit and Finish

### 1. Missing test coverage for direction field

**Location:** `internal/restapi/stop_handler_test.go`

The existing tests verify many fields but don't assert on the `direction` field. Consider adding a test assertion to validate the direction is returned correctly. For example:

```go
assert.Contains(t, entry, "direction", "direction field should exist")
```

### 2. Code formatting changes look good

The test file formatting changes (removing extra newlines before closing braces) are appropriate cleanup per `go fmt` standards.

## Strengths

- Good use of the existing `utils.NullStringOrEmpty()` utility rather than inline null checks
- Consistent application of the pattern to both `Name` and `Code` fields
- Clean, minimal diff that's easy to review

----

Please clarify the expected behavior for null directions (empty string vs "UNKNOWN" vs calculated) to ensure consistency across endpoints. Once that's resolved, this is ready to merge.

uloamaka:
Hello @aaronbrethorst 
I’d like to align `stops_for_agency_handler.go` with `stop_handler.go` so that when stop direction data is unavailable, the API returns an empty string ("") rather than "UNKNOWN". This matches the behavior of the OBA production server.

`stop_handler.go` will return the precomputed direction when present and fall back to "" if it’s missing.

`stops_for_route_handler.go` will continue to use CalculateStopDirection(), since route context makes calculated direction meaningful there.

In both cases, I plan to use utils.NullStringOrEmpty() for consistency.

I’ll also add a test case in stop_handler_test.go to assert that the direction field exists.

Does this approach look good?

aaronbrethorst (approved):
Hey Godsgift, nice work following through on the direction consistency fix. You investigated the OBA production server behavior, confirmed that `""` (not `"UNKNOWN"`) is the correct fallback for stops without direction data, and then aligned both `stop_handler.go` and `stops_for_agency_handler.go` to match. That's exactly the kind of diligence I was hoping for.

The `NullStringOrEmpty` normalization on the `Name` and `Code` fields in the stop handler is a nice touch for consistency, and the cleanup in `stops_for_agency_handler.go`—replacing the 5-line direction block with a single `utils.NullStringOrEmpty()` call—is cleaner and easier to follow.

There's nothing left to change—this PR is ready to merge. Thanks for your patience with the back-and-forth on direction handling; the end result is solid.

PR #290, opened 2026-02-02
perf: fix N+1 query issue in stops-for-agency handler

### LINKED ISSUE #289: N+1 Query Problem in stops-for-agency Endpoint
### Summary
The `stops-for-agency` endpoint executes two database queries inside a loop for every stop, causing severe performance degradation for agencies with large numbers of stops.

### Impact
- For an agency with 5,000 stops, this endpoint executes **10,001 database queries** instead of 3
- Response times scale linearly with stop count, making the endpoint unusable for large agencies
- High database lock contention that can block other operations

### Location
**File:** `internal/restapi/stops_for_agency_handler.go`  
**Function:** `buildStopsListForAgency` (lines 60-100)

### Problem Code
```go
for _, stopID := range stopIDs {
    // Query executed N times
    stop, err := api.GtfsManager.GtfsDB.Queries.GetStop(ctx, stopID)
    if err != nil {
        continue
    }

    // Query executed N times
    routeIds, err := api.GtfsManager.GtfsDB.Queries.GetRouteIDsForStop(ctx, stop.ID)
    if err != nil {
        continue
    }
    // ...
}
```

### Suggested Fix
Replace the loop with batch queries that are already available in the codebase:

1. Use `GetStopsByIDs(ctx, stopIDs)` to fetch all stops in one query
2. Use `GetRouteIDsForStops(ctx, stopIDs)` to fetch all route associations in one query
3. Build a map of route IDs by stop ID in memory
4. Construct the response list by iterating over the batched results

This pattern is already used correctly in `stops_for_location_handler.go` and `trip_details_handler.go`.

### Expected Outcome
Query count reduced from `2N + 1` to `3` total queries, regardless of the number of stops.

ARCoder181105:
**Description:**
This PR optimizes the `stops-for-agency` endpoint by replacing iterative database queries with batch operations, significantly improving performance for agencies with large numbers of stops.

Closes #289 

**Motivation:**
Previously, `buildStopsListForAgency` suffered from an N+1 query problem. It iterated through every stop ID and executed two separate SQL queries (`GetStop` and `GetRouteIDsForStop`) for each one. For an agency with 5,000 stops, this resulted in over 10,000 database queries for a single API request.

**Changes:**
Refactored `buildStopsListForAgency` in `internal/restapi/stops_for_agency_handler.go` to:

1. **Batch Fetch:** Use `GetStopsByIDs` and `GetRouteIDsForStops` to fetch all data in just 2 queries, regardless of the number of stops.
2. **In-Memory Mapping:** Map route IDs to stops in memory to avoid repeated DB lookups.
3. **Safety Check:** Added an early return `if len(stopIDs) == 0` to prevent unnecessary DB calls for empty lists.

**Verification:**

* Validated that the response structure remains identical to the previous implementation.
* Verified that the logic follows the batching patterns used in other handlers (e.g., `stops_for_location_handler.go`).
* `make test` passes.

### DIFF
--- a/internal/restapi/report_problem_with_stop_handler.go
+++ b/internal/restapi/report_problem_with_stop_handler.go
@@ -39,7 +39,7 @@ func (api *RestAPI) reportProblemWithStopHandler(w http.ResponseWriter, r *http.
 		http.Error(w, `{"code":500, "text":"internal server error"}`, http.StatusInternalServerError)
 		return
 	}
-	
+
 	query := r.URL.Query()
 
 	code := query.Get("code")
--- a/internal/restapi/stops_for_agency_handler.go
+++ b/internal/restapi/stops_for_agency_handler.go
@@ -80,10 +80,10 @@ func (api *RestAPI) stopsForAgencyHandler(w http.ResponseWriter, r *http.Request
 
 // IMPORTANT: Caller must hold manager.RLock() before calling this method.
 func (api *RestAPI) buildStopsListForAgency(ctx context.Context, agencyID string, stopIDs []string) ([]models.Stop, error) {
-	// If no stops, return empty list 
-    if len(stopIDs) == 0 {
-        return []models.Stop{}, nil
-    }
+	// If no stops, return empty list
+	if len(stopIDs) == 0 {
+		return []models.Stop{}, nil
+	}
 
 	// Batch fetch all stops in one query
 	stops, err := api.GtfsManager.GtfsDB.Queries.GetStopsByIDs(ctx, stopIDs)
@@ -101,7 +101,7 @@ func (api *RestAPI) buildStopsListForAgency(ctx context.Context, agencyID string
 	routesByStop := make(map[string][]string)
 	for _, row := range routeIDsRows {
 		if rid, ok := row.RouteID.(string); ok {
-			routesByStop[row.StopID] = append(routesByStop[row.StopID], utils.FormCombinedID(agencyID, rid))
+			routesByStop[row.StopID] = append(routesByStop[row.StopID], rid)
 		}
 	}
 
--- a/internal/restapi/test_helper_test.go
+++ b/internal/restapi/test_helper_test.go
@@ -22,10 +22,8 @@ func (m *mockTestingFatalf) Fatalf(format string, args ...any) {
 
 func TestCollectAllNestedIdsFromObjects(t *testing.T) {
 	data := []interface{}{
-		map[string]interface{}{"routes": []interface{}{"234", "235"},
-		},
-		map[string]interface{}{"routes": []interface{}{"345"},
-		},
+		map[string]interface{}{"routes": []interface{}{"234", "235"}},
+		map[string]interface{}{"routes": []interface{}{"345"}},
 	}
 	expected := []string{"234", "235", "345"}
 	actual := collectAllNestedIdsFromObjects(t, data, "routes")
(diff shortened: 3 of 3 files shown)


aaronbrethorst (changes requested):
Hey Aditya! Excellent work on tackling the N+1 query problem. Reducing from 10,000+ queries to 3 queries is a massive performance win.

## Summary

This PR refactors `buildStopsListForAgency` to use batch queries (`GetStopsByIDs` and `GetRouteIDsForStops`) instead of iterating and querying each stop individually. This follows the same pattern established in `stops_for_location_handler.go`.

## Critical Issues

### 1. Double-prefixed route IDs (pre-existing bug, reproduced in new code)

**Location:** `internal/restapi/stops_for_agency_handler.go:102-105`

```go
for _, row := range routeIDsRows {
    if rid, ok := row.RouteID.(string); ok {
        routesByStop[row.StopID] = append(routesByStop[row.StopID], utils.FormCombinedID(agencyID, rid))
    }
}
```

The `GetRouteIDsForStops` query already returns combined IDs (see `gtfsdb/query.sql:534`):
```sql
routes.agency_id || '_' || routes.id AS route_id
```

So `rid` is already `"agencyID_routeID"`. Calling `FormCombinedID(agencyID, rid)` produces `"agencyID_agencyID_routeID"`.

**Note:** This bug existed in the old code too (`FormCombinedID(agencyID, id.(string))`), so the PR preserves existing behavior. However, since you're refactoring this code anyway, now is an ideal time to fix it.

**Fix:** Use `rid` directly, like `stops_for_location_handler.go` does:
```go
for _, row := range routeIDsRows {
    if rid, ok := row.RouteID.(string); ok {
        routesByStop[row.StopID] = append(routesByStop[row.StopID], rid)  // Already combined
    }
}
```

## Important Issues

### 1. Inconsistent indentation in early return

**Location:** `internal/restapi/stops_for_agency_handler.go:83-86`

```go
	// If no stops, return empty list
    if len(stopIDs) == 0 {
        return []models.Stop{}, nil
    }
```

The first line uses tabs, but lines 84-86 use spaces. This should be consistent (use tabs per Go convention).

### 2. Behavior change: errors now propagate instead of being silently skipped

**Location:** `internal/restapi/stops_for_agency_handler.go:89-98`

The old code silently skipped stops that failed to fetch:
```go
stop, err := api.GtfsManager.GtfsDB.Queries.GetStop(ctx, stopID)
if err != nil {
    continue  // Silent skip
}
```

The new code returns errors to the caller:
```go
stops, err := api.GtfsManager.GtfsDB.Queries.GetStopsByIDs(ctx, stopIDs)
if err != nil {
    return nil, err  // Propagates error
}
```

This is actually a **better** behavior—failing fast is preferable to silently degraded responses. Just noting it as a behavioral change that's intentional and correct.

## Fit and Finish

### 1. Trailing space in comment

**Location:** `internal/restapi/stops_for_agency_handler.go:83`

```go
// If no stops, return empty list
```

Remove the trailing space after "list".

## Strengths

- **Significant performance improvement**: 2N+1 queries → 3 queries is a major win
- **Follows established patterns**: Matches the approach in `stops_for_location_handler.go`
- **Clean logic flow**: Fetch all data upfront, then process in memory
- **Good error handling**: Propagates errors instead of silently degrading

## Recommendation

Please fix the double-prefixed route IDs (use `rid` directly instead of wrapping in `FormCombinedID`) and the indentation issue. The route ID fix actually improves correctness beyond what the old code did.

After those fixes, this is ready to merge.

---

Great performance optimization work!

ARCoder181105:
@aaronbrethorst 
I addressed all the requested changes and also ran `make fmt` for making sure there are not any indentation problem ,
This PR is ready for review

aaronbrethorst (approved):
Hey Aditya, you've done a great job addressing all the feedback from the first round. The double-prefixed route ID fix is correct, the indentation is clean, and the code now follows the exact same batch query pattern used in `stops_for_location_handler.go` and `stops_for_route_handler.go`.

There's nothing left to change—this PR is ready to merge. Reducing 10,000+ individual queries down to 3 batch queries is a big performance win, and the fail-fast error handling is a nice improvement over the old silent `continue` behavior. Well done!

PR #294, opened 2026-02-02
Optimize Regex Compilation in Search Handler

### LINKED ISSUE #293: Regex Compiled on Every Search Request
## Problem
The `sanitizeFTS5Query` function in `internal/restapi/search_stops_handler.go` compiles two regular expressions on every search request, causing unnecessary CPU overhead.

## Impact
- 100x slower than necessary
- Regex compilation takes ~100,000 ns per request
- Regex execution takes ~1,000 ns per request
- At 1,000 searches/sec: wastes approximately 10% of a CPU core

## Solution
Move regex compilation to package-level variables so they are compiled once at startup and reused for all requests.

## Benefits
- 100x performance improvement
- Reduced CPU usage and latency
- Less GC pressure
- No behavioral changes
- Thread-safe (regexp.Regexp is safe for concurrent use)

ARCoder181105:
## Description
This PR fixes a performance issue where regular expressions were being compiled on every search request instead of once at startup.

Closes #293 

## Problem
The `sanitizeFTS5Query` function in `internal/restapi/search_stops_handler.go` was calling `regexp.MustCompile` twice per request, causing unnecessary CPU overhead and memory allocations.

## Solution
Moved regex compilation to package-level variables so patterns are compiled once at application startup and reused for all subsequent requests.

## Changes
- Added two package-level variables: `fts5SpecialCharsRegex` and `fts5OperatorsRegex`
- Updated `sanitizeFTS5Query` function to use pre-compiled regex patterns
- No behavioral changes to search functionality

## Performance Impact
- 100x faster regex operations (from ~101µs to ~1µs per request)
- Reduced CPU usage by approximately 10% at 1,000 requests/second
- Reduced memory allocations and GC pressure
- Lower latency for search queries under load

## Testing
- Existing unit tests pass without modification
- Search functionality verified to work correctly
- No breaking changes to API or behavior

## Thread Safety
The `regexp.Regexp` type is explicitly documented as safe for concurrent use by multiple goroutines, making this change safe in a concurrent environment.

### DIFF
--- a/internal/restapi/search_stops_handler.go
+++ b/internal/restapi/search_stops_handler.go
@@ -13,16 +13,20 @@ import (
 	"maglev.onebusaway.org/internal/utils"
 )
 
+// Pre-compiled regex patterns for FTS5 query sanitization
+var (
+	fts5SpecialCharsRegex = regexp.MustCompile(`[*"():^$@#~<>{}[\]\\|&!]`)
+	fts5OperatorsRegex    = regexp.MustCompile(`(?i)\b(AND|OR|NOT|NEAR)\b`)
+)
+
 // sanitizeFTS5Query removes special FTS5 characters by replacing them with spaces
 // to prevent query syntax errors. Does not preserve the original characters.
 func sanitizeFTS5Query(input string) string {
-	// 1. Remove ALL FTS5 special characters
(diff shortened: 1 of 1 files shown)


aaronbrethorst (approved):
great work Aditya.

PR #297, opened 2026-02-02
Fix broken polyline rendering on looping routes and optimize shape processing

### LINKED ISSUE #296: Fix broken polyline rendering on looping routes and optimize shape processing
**File:** `internal/restapi/shapes_handler.go`

**Description:**
I found a logic bug that causes visual artifacts (lines "shooting off" the map) for routes that loop back on themselves, along with a performance inefficiency.

**The Issue:**

1. **Broken Rendering:** The current implementation splits a shape into multiple segments if it detects a repeated edge (e.g., a round trip). It then concatenates the encoded polyline strings. Since polyline encoding relies on **delta offsets** from the previous point, concatenating separate strings resets this offset. The decoder interprets the start of the second segment (an absolute coordinate) as a delta, causing massive coordinate jumps.
2. **Performance:** The loop appends points to `currentLine` one by one without pre-allocation, causing unnecessary memory re-allocations for large shapes.

**Proposed Solution:**
Standard GTFS shapes are defined as a single continuous sequence. We should remove the edge-splitting logic and encode the entire shape in one pass. This fixes the rendering bug and allows us to pre-allocate memory.

**Suggested Fix:**
Refactor the loop to pre-allocate the slice and remove the `edges` map check:

```go
// Optimization: Pre-allocate slice to avoid resizing
lineCoords := make([][]float64, 0, len(shapes))

for _, point := range shapes {
    lineCoords = append(lineCoords, []float64{point.Lat, point.Lon})
}

// Encode as a single continuous polyline to ensure valid delta offsets
encodedPoints := string(polyline.EncodeCoords(lineCoords))

```

ARCoder181105:
**Description:**
This PR fixes a rendering bug where routes that loop back on themselves would display visual artifacts (lines "shooting off" the map). It also improves performance by reducing memory allocations during shape processing.

Closes #296 

**The Problem:**

1. **Visual Bug:** The previous implementation split shapes into multiple segments if a duplicate edge was detected (e.g., a round trip) and concatenated the encoded strings. Because polyline encoding relies on **delta offsets**, concatenating strings reset the offset, causing the decoder to interpret the start of the next segment as a massive coordinate jump.
2. **Performance:** The loop appended points to a slice without pre-allocation, causing repeated memory resizing for large shapes.

**The Solution:**

1. **Fix:** Removed the edge-splitting logic. The entire shape is now encoded as a single continuous polyline sequence, ensuring valid delta encoding throughout.
2. **Optimization:** Pre-allocated the `lineCoords` slice using `make` with the known capacity `len(shapes)` to eliminate unnecessary re-allocations.

**Changes:**

* `internal/restapi/shapes_handler.go`: Refactored loop to pre-allocate memory and encode shape in a single pass.


ARCoder181105:
@aaronbrethorst 
The test are passing PR is ready for review

image

aaronbrethorst (changes requested):
Hey Aditya, you've addressed all five items from my first review — the `ORDER BY shape_pt_sequence` is in place, the consecutive-duplicate filter is restored, `geo.go` and `geo_test.go` are deleted, the `assert.NotEmpty` is now an early guard, and the out-and-back test (`A → B → C → B → A`) is a solid addition. Before we can merge this, I need you to make a couple more changes:

## Important

### 1. Revert the `RoutesFt` struct addition in `gtfsdb/models.go`

Thanks for flagging the `routes_fts` build issue with `make models` — that's a real problem with sqlc and FTS5 virtual tables, and I appreciate you working through it. However, the workaround has left a spurious `RoutesFt` struct in the generated `models.go` that doesn't exist on `main`. Since this struct has no consumers, please revert that file so it matches `main` exactly. I'll track the underlying sqlc/FTS5 issue separately — it shouldn't block your PR.

### 2. Squash and rebase

The branch has four commits (including a merge commit) and is seven commits behind `main`. Please squash into a single atomic commit and rebase against `main`. If rebasing feels unfamiliar, here's a short guide that might help: https://www.brethorsting.com/blog/2026/01/git-rebase-for-the-terrified/

ARCoder181105:
@aaronbrethorst 
Done with the changes, please also consider to look at the anomality in query.sql I commented above

aaronbrethorst (approved):
Hey Aditya, great work seeing this through four rounds of review — the persistence and attention to detail really shows. Every single item from the prior three rounds has been addressed: the `ORDER BY`, the duplicate filter, the dead code cleanup, the safe type assertions, the longitude checks, the consecutive-duplicate test, and the squash + rebase. The branch is a single clean commit, CI is green across the board, and the code is tight.

The fix is correct — single-pass delta encoding is the mathematically sound approach, and the three new tests lock in the behavior for looping routes, out-and-back routes, and consecutive duplicate filtering. Well done.

The `RoutesFt` issue has been resolved in main, but I'm not going to hold up your PR to have you re-run `make models`. I'll take care of that in a followup PR.

Issue #302, opened 2026-02-02
Bug: GTFS downloader does not validate HTTP status codes


3rabiii:
### Description
The GTFS downloader in `internal/gtfs/static.go` currently reads the response body without checking the HTTP status code.

If the remote server returns a 404 (Not Found) or 500 (Internal Server Error), the application attempts to process the error page as a GTFS file, leading to confusing downstream errors.

### Goal
Implement a check to ensure `resp.StatusCode` is `http.StatusOK` (200) before attempting to read and process the GTFS data.

PR #303, opened 2026-02-02
fix(gtfs): validate HTTP status code when downloading data

3rabiii:
### Description
This PR addresses the missing HTTP status code validation in the GTFS downloader (`internal/gtfs/static.go`). 

Previously, the code would attempt to read and process the response body regardless of the status code. This meant that 404 (Not Found) or 500 (Server Error) HTML responses were being passed to the GTFS parser, causing confusing downstream errors.

### Changes
- Added a check to ensure `resp.StatusCode` is `http.StatusOK` (200).
- Returns a descriptive error if the status code is not 200.

@aaronbrethorst 
fixes : #302

### DIFF
--- a/internal/gtfs/static.go
+++ b/internal/gtfs/static.go
@@ -53,6 +53,10 @@ func rawGtfsData(source string, isLocalFile bool, config Config) ([]byte, error)
 			slog.Default().With(slog.String("component", "gtfs_downloader")),
 			"http_response_body")
 
+		if resp.StatusCode != http.StatusOK {
+			return nil, fmt.Errorf("failed to download GTFS data: received HTTP status %s", resp.Status)
+		}
+
 		b, err = io.ReadAll(resp.Body)
 		if err != nil {
 			return nil, fmt.Errorf("error reading GTFS data: %w", err)


aaronbrethorst (changes requested):
Hey Adel, thanks for tackling this HTTP status code validation issue. Catching non-200 responses before they reach the GTFS parser is exactly the right fix — nobody wants to debug "invalid zip format" errors when the real problem is a 404. Before we can merge this, I will need you to make a couple changes:

## Critical

### 1. Redundant status code in error message

**Location:** `internal/gtfs/static.go:57`

```go
return nil, fmt.Errorf("failed to download GTFS data: HTTP %d %s", resp.StatusCode, resp.Status)
```

In Go, `resp.Status` already includes the numeric code (e.g., `"404 Not Found"`), so this produces `"HTTP 404 404 Not Found"`. Use just one:

```go
return nil, fmt.Errorf("failed to download GTFS data: received HTTP status %s", resp.Status)
```

## Important

### 1. Rebase against main

The branch is 9 commits behind `main`. Please rebase against `main` before we merge. Here's a blog post I wrote that might help if you're not comfortable with rebasing: https://www.brethorsting.com/blog/2026/01/git-rebase-for-the-terrified/

### 2. Clean up "trigger ci" commit

The PR has a `trigger ci` commit (`3ee64d8`) that should be squashed into the main commit during your rebase.

3rabiii:
Thanks for the guidance on rebasing! I've updated the PR:

1.  **Fixed Error Message:** Removed the redundant HTTP status code from the error message.
2.  **Cleaned History:** Rebased against `main` and squashed the `trigger ci` and fix commits into a single atomic commit.

Ready for merge!
@aaronbrethorst 
closes : #302

PR #324, opened 2026-02-04
Integraton tests

### LINKED ISSUE #249: Integration test to verify handler locking during ForceUpdate
### Problem

We have tests that verify `ForceUpdate` succeeds, but none that ensure HTTP handlers correctly hold locks while a GTFS static data swap is in progress. This behavior is currently untested and could regress silently.

### Proposal

Add an **integration test** that:
1. Starts the HTTP server with GTFS data loaded.
2. Triggers `ForceUpdate` in a goroutine.
3. Issues concurrent HTTP requests during the swap.
4. Verifies requests complete safely (no panics, errors, or races).

The goal is to validate handler locking and concurrency safety, not GTFS correctness.

### Acceptance Criteria

- Requests are served correctly during `ForceUpdate`
- Test passes under `-race`
- Locking regressions cause test failures

mann-patwa:
Fixes #249 
# Ensure Handler Lock Safety for Static GTFS Data

This PR ensures that HTTP handlers reliably acquire read locks when accessing static GTFS data, preventing potential race conditions during runtime data updates (`ForceUpdate`).

### 1. Verification Test (`cmd/api/lock_safety_test.go`)
Added a new integration test `TestHandlerLockSafety` that verifies thread safety under load:
- **Concurrent Readers**: Simulates clients querying the API (`/api/where/stop/{id}`) concurrently.
- **Atomic Swap**: Triggers a `ForceUpdate` to hot-swap the GTFS dataset (from `raba.zip` to `gtfs.zip`) while readers are active.
- **Validation**: Confirms that no requests fail or deadlock during the swap and verifies the data update via the `/api/where/agencies-with-coverage` endpoint.

### 2. Refactors for Testability
- **Export `Manager.Config`**: Renamed `config` to `Config` in `internal/gtfs/gtfs_manager.go`. This enables tests to inspect and modify the GTFS source URL dynamically.
- **Context-Aware `Run`**: Updated the `Run` function in `cmd/api/app.go` to accept a `context.Context`. This allows integration tests to control the server lifecycle (start/stop) programmatically, avoiding Reliance on OS signals.

### DIFF
--- a/cmd/api/app.go
+++ b/cmd/api/app.go
@@ -113,14 +113,14 @@ func CreateServer(coreApp *app.Application, cfg appconf.Config) (*http.Server, *
 }
 
 // Run manages the server lifecycle with graceful shutdown.
-// Starts the server in a goroutine, waits for shutdown signals (SIGINT, SIGTERM),
+// Starts the server in a goroutine, waits for shutdown signals (SIGINT, SIGTERM) or context cancellation,
 // and performs graceful shutdown with a 30-second timeout.
 // Returns an error if the server fails to start or shutdown fails.
-func Run(srv *http.Server, gtfsManager *gtfs.Manager, api *restapi.RestAPI, logger *slog.Logger) error {
+func Run(ctx context.Context, srv *http.Server, gtfsManager *gtfs.Manager, api *restapi.RestAPI, logger *slog.Logger) error {
 	logger.Info("starting server", "addr", srv.Addr)
 
-	// Set up signal handling for graceful shutdown
-	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+	// Set up signal handling for graceful shutdown, merging with provided context
+	ctx, stop := signal.NotifyContext(ctx, os.Interrupt, syscall.SIGTERM)
 	defer stop()
 
 	// Channel to capture server errors
@@ -133,7 +133,7 @@ func Run(srv *http.Server, gtfsManager *gtfs.Manager, api *restapi.RestAPI, logg
 		}
 	}()
 
-	// Wait for either shutdown signal or server error
+	// Wait for either shutdown signal/context cancellation or server error
 	select {
 	case err := <-serverErrors:
 		return fmt.Errorf("server failed to start: %w", err)
--- a/cmd/api/lock_safety_test.go
+++ b/cmd/api/lock_safety_test.go
@@ -1,0 +1,197 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net"
+	"net/http"
+	"runtime"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"maglev.onebusaway.org/internal/appconf"
+	"maglev.onebusaway.org/internal/gtfs"
+	"maglev.onebusaway.org/internal/models"
+
+	_ "github.com/mattn/go-sqlite3"
+)
+
+func TestHandlerLockSafety(t *testing.T) {
+	if runtime.GOOS == "windows" {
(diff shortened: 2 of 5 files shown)


aaronbrethorst (changes requested):
Hey Mann, thanks for taking the time to add this lock safety integration test — verifying that HTTP handlers remain safe during a concurrent GTFS hot-swap is a genuinely important production scenario, and the `Run` function's new `context.Context` parameter is a clean, well-motivated improvement for testability. Before we can merge this, I will need you to make a couple changes:

## Critical

### 1. Hardcoded port 50001 will break in CI and parallel runs

**File:** `cmd/api/lock_safety_test.go:38`

The test binds to port 50001, which will fail if another process uses that port or if tests run in parallel (`go test -count=2`). The existing tests in this very file already demonstrate the right pattern — `TestRunServerStartsAndStopsCleanly` and `TestRunWithPortZeroAndImmediateShutdown` both use `Port: 0` to let the OS pick a free port.

The challenge is that your test needs to make HTTP requests to the server, so you need to discover the actual bound address. The cleanest approach is to use `net.Listen("tcp", ":0")` to get a listener with a random port, extract the port from `listener.Addr()`, close the listener, then configure the server with that port. Alternatively, restructure the test to use `httptest.NewServer(srv.Handler)` like `TestCreateServerHandlerResponds` does — though that approach bypasses `Run()`, which you want to exercise.

### 2. Exporting `config` → `Config` breaks encapsulation

**File:** `internal/gtfs/gtfs_manager.go:39`

Renaming the private `config` field to the exported `Config` means any package can now mutate the manager's entire configuration at runtime — the GTFS URL, the database path, auth headers, and more. The only field the test actually needs to change is `GtfsURL`. Please add a targeted setter instead:

```go
func (manager *Manager) SetGtfsURL(url string) {
    manager.staticUpdateMutex.Lock()
    defer manager.staticUpdateMutex.Unlock()
    manager.config.GtfsURL = url
}
```

This keeps `config` private, exposes only what tests need, and adds synchronization that protects against a data race — right now, the test mutates `Config.GtfsURL` on one goroutine while `updateStaticGTFS` may be reading `Config.GtfsURL` on another without holding `staticUpdateMutex`.

The existing `hot_swap_test.go` references (`manager.config.GtfsURL = ...`) work because they're in the same package. Those tests can keep using the private field directly, so no changes needed there.

## Important

### 3. No assertion that readers actually ran during the hot-swap

**File:** `cmd/api/lock_safety_test.go:113–155`

The test launches 3 reader goroutines, sleeps 100ms, then calls `ForceUpdate`. But there's no verification that any reader issued a request during the critical window. On a slow CI machine, the readers could be too slow to start, and the test would pass vacuously — "zero errors out of zero requests" is trivially true. Add an atomic counter for successful iterations and assert it's nonzero after `wg.Wait()`.

### 4. HTTP errors silently swallowed

**File:** `cmd/api/lock_safety_test.go:137–139`

```go
if err != nil {
    // Ignore errors during cancel
}
```

This swallows *all* HTTP errors unconditionally, not just cancellation errors. A genuine connection reset or handler panic during the hot-swap would go undetected. Check `readerCtx.Err()` first — only ignore errors if the context was cancelled:

```go
if err != nil {
    if readerCtx.Err() == nil {
        errChan <- fmt.Errorf("reader %d HTTP error: %w", id, err)
        return
    }
    return
}
```

### 5. Test doesn't wait for `Run` goroutine to finish

**File:** `cmd/api/lock_safety_test.go:56–59`

The goroutine running `Run()` may still be performing shutdown (closing SQLite connections, cleaning up) when the test function returns. Since `t.TempDir()` deletes the temp directory on test completion, this could race with database file cleanup. Capture the `Run` goroutine's error in a channel and wait for it after `serverCancel()`.

## Fit and Finish

### 6. Squash and rebase

The branch has three commits and is seven commits behind `main`. Please squash into a single atomic commit and rebase against `main`. If rebasing feels unfamiliar, here's a short guide that might help: https://www.brethorsting.com/blog/2026/01/git-rebase-for-the-terrified/

Issue #325, opened 2026-02-04
Security: Harden HTTP clients against hanging and OOM risks


3rabiii:
### Description
A security review identified potential resilience issues in how the application fetches remote data in `internal/gtfs/realtime.go` and `gtfsdb/client.go`.

Currently:
1. **Unbounded Timeouts:** `realtime.go` uses `http.DefaultClient`, which has no timeout. If a remote server hangs, the connection will block indefinitely, potentially leading to resource exhaustion (goroutine leaks).
2. **Unbounded Reads:** Both `realtime.go` and `client.go` use `io.ReadAll` to read response bodies into memory without a size limit. A malicious or misconfigured server could send an extremely large response, causing an Out-Of-Memory (OOM) crash.

### Impact
- **Denial of Service (DoS):** The service can be taken down by slowloris attacks (hanging connections) or large-payload attacks (memory exhaustion).
- **Stability:** Network glitches could cause permanent hangs in data fetching routines.

### Goal
Refactor the HTTP fetching logic to:
1. Use a custom `http.Client` with an explicit `Timeout` (e.g., 10 seconds for realtime).
2. Wrap response bodies with `io.LimitReader` to enforce strict upper bounds on memory usage (e.g., 25MB for realtime, 200MB for static GTFS).

PR #326, opened 2026-02-04
fix(security): harden http clients with timeouts and size limits

3rabiii:
### Description
This PR improves the resilience and security of the application's network clients by addressing potential resource exhaustion risks identified in `realtime.go` and `client.go`.

Previously, the application used `http.DefaultClient` (which has no timeout) and unbounded `io.ReadAll`, making it vulnerable to:
- **Hanging connections:** If a remote server stalls, the application could hang indefinitely.
- **Out-of-Memory (OOM) crashes:** A malicious or misconfigured server could send an excessively large response, consuming all available RAM.

### Changes
- **`internal/gtfs/realtime.go`**:
  - Replaced `http.DefaultClient` with a custom `http.Client` enforcing a **10-second timeout**.
  - Wrapped the response body with `io.LimitReader` capped at **25MB** (sufficient for GTFS-RT feeds but prevents OOM).

- **`gtfsdb/client.go`**:
  - Added `io.LimitReader` to the static GTFS download, capped at **200MB** (accommodates large feeds while preventing infinite stream attacks).

Screenshot From 2026-02-05 01-47-19

@aaronbrethorst 
fixes : #325

### DIFF
--- a/gtfsdb/client.go
+++ b/gtfsdb/client.go
@@ -73,12 +73,17 @@ func (c *Client) DownloadAndStore(ctx context.Context, url, authHeaderKey, authH
 	}
 	defer func() { _ = resp.Body.Close() }()
 
-	b, err := io.ReadAll(resp.Body)
+	const maxBodySize = 200 * 1024 * 1024
+	body, err := io.ReadAll(io.LimitReader(resp.Body, maxBodySize+1))
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to read response body: %w", err)
 	}
 
-	err = c.processAndStoreGTFSDataWithSource(b, url)
+	if int64(len(body)) > maxBodySize {
+		return fmt.Errorf("static GTFS response exceeds size limit of %d bytes", maxBodySize)
+	}
+
+	err = c.processAndStoreGTFSDataWithSource(body, url)
 
 	return err
 }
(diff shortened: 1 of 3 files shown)


aaronbrethorst (changes requested):
Adel, good progress on this one — the `limit + 1` truncation detection pattern is correctly applied across all three files now, the size limits are right (25MB for real-time, 200MB for static), and the intermediate variable and redundant timeout are cleaned up. Almost there, but I spotted one regression that needs fixing before we merge.

## Critical

### HTTP status code check removed from `static.go`

**File:** `internal/gtfs/static.go` (between lines 54 and 56)

The diff shows this check was removed during the rebase/edit:

```go
if resp.StatusCode != http.StatusOK {
    return nil, fmt.Errorf("failed to download GTFS data: received HTTP status %s", resp.Status)
}
```

This existed on `main` and was removed by this PR. Without it, a 404 or 500 response from the GTFS server will be silently read and passed to the GTFS parser, which will fail with a confusing parsing error (e.g., `"zip: not a valid zip file"`) instead of the clear `"received HTTP status 404 Not Found"` message.

Please restore the status check before the `io.ReadAll` call:

```go
if resp.StatusCode != http.StatusOK {
    return nil, fmt.Errorf("failed to download GTFS data: received HTTP status %s", resp.Status)
}

const maxStaticSize = 200 * 1024 * 1024
b, err = io.ReadAll(io.LimitReader(resp.Body, maxStaticSize+1))
```

thank you! I'm looking forward to getting this merged!

Issue #327, opened 2026-02-05
Feature: Support environment variable overrides for sensitive configuration


3rabiii:
### Description
The current configuration system (`json_config.go`) relies solely on `config.json` for all settings, including sensitive secrets like API keys and authentication headers.

This presents two major issues:
1. **Security Risk:** Secrets might be accidentally committed to version control if they are hardcoded in the JSON file.
2. **Deployment Friction:** Containerized environments (Docker, Kubernetes) and PaaS platforms typically inject secrets via environment variables, not by editing files on disk.

### Goal
Update the application configuration loader to support the **12-Factor App** methodology by allowing environment variables to override sensitive fields defined in `config.json`.

### Proposed Changes
Modify `internal/appconf/json_config.go` to check for the following environment variables after loading the JSON file. If these variables are present, they should override the file-based values:

- `GTFS_API_KEYS` -> Overrides the list of valid API keys (comma-separated).
- `GTFS_STATIC_AUTH_VALUE` -> Overrides the static GTFS feed authentication token.
- `GTFS_REALTIME_AUTH_VALUE` -> Overrides the Realtime GTFS feed authentication token.

PR #328, opened 2026-02-05
feat: add env var overrides for secrets in config

3rabiii:
### Description
This PR implements **Environment Variable Overrides** for the configuration system, aligning the application with the **12-Factor App** methodology regarding config management (Factor III).

Previously, sensitive secrets (like API keys and auth headers) had to be hardcoded in `config.json`, posing a security risk if committed to version control. This change allows these values to be injected securely via environment variables at runtime, which is standard practice for containerized deployments (Docker/Kubernetes).

### Changes
Modified `internal/appconf/json_config.go` to prioritize environment variables over file-based config for the following fields:
- `GTFS_API_KEYS`: Overrides the list of API keys.
- `GTFS_STATIC_AUTH_VALUE`: Overrides the static feed authentication token.
- `GTFS_REALTIME_AUTH_VALUE`: Overrides the realtime feed authentication token.

### Verification
Screenshot From 2026-02-05 16-24-01

@aaronbrethorst 
closes : #327

### DIFF
--- a/internal/appconf/json_config_test.go
+++ b/internal/appconf/json_config_test.go
@@ -500,11 +500,16 @@ func TestLoadFromFile_EnvVarOverrides(t *testing.T) {
 
 	tmpFile, err := os.CreateTemp("", "config_test_*.json")
 	require.NoError(t, err)
-	defer os.Remove(tmpFile.Name())
+
+	t.Cleanup(func() {
+		_ = os.Remove(tmpFile.Name())
+	})
 
 	_, err = tmpFile.WriteString(content)
 	require.NoError(t, err)
-	tmpFile.Close()
+
+	err = tmpFile.Close()
+	require.NoError(t, err)
 
 	t.Run("Happy Path - Override All Secrets", func(t *testing.T) {
 		t.Setenv("GTFS_API_KEYS", "env-key-1,env-key-2")


aaronbrethorst (changes requested):
Hey Adel! Supporting env var overrides for secrets is a great addition — it's exactly the right approach for containerized deployments and keeps sensitive values out of version control. The implementation logic is solid and correctly placed between `setDefaults()` and `validate()`. Before we can merge this, I will need you to make a couple changes:

## Critical

1. **Please add tests for the env var override functionality.** This PR adds 26 lines of new branching logic (comma splitting, whitespace trimming, empty filtering, conditional overrides across three config fields) with zero test coverage. The existing test suite for file-based config is excellent — let's bring the env var overrides up to the same standard.

   In Go, use `t.Setenv()` which automatically restores the original value after the test. Here are the key scenarios to cover:

   - **Happy paths**: `GTFS_API_KEYS` overrides file keys, `GTFS_STATIC_AUTH_VALUE` overrides static auth, `GTFS_REALTIME_AUTH_VALUE` overrides RT auth
   - **Parsing edge cases**: `GTFS_API_KEYS` with trailing commas, extra spaces, empty segments (e.g., `"key1 , key2 , , key3,"` should produce `["key1", "key2", "key3"]`)
   - **Empty string doesn't override**: `t.Setenv("GTFS_API_KEYS", "")` should preserve file-based keys
   - **Validation still fires**: `GTFS_API_KEYS=key1,key1` should fail with duplicate key error
   - **Auth header pairing** (see item 2 below)

   The existing `config_full.json` fixture already has API keys, static auth, and RT auth, so you can load from that in most tests.

## Important

2. **`GTFS_STATIC_AUTH_VALUE` creates a validation trap when used without `auth-header-name` in the config file.** The existing validation at `json_config.go:111-114` requires `AuthHeaderName` and `AuthHeaderValue` to both be set or both be empty. If someone sets `GTFS_STATIC_AUTH_VALUE` via env var but their config file has no `auth-header-name`, `LoadFromFile` will fail with:

   > "both auth-header-name and auth-header-value must be provided together for gtfs-static-feed"

   This is technically correct, but a user following the 12-Factor approach would reasonably expect to configure auth entirely via env vars. Please either:
   - **(a)** Add `GTFS_STATIC_AUTH_NAME` and `GTFS_REALTIME_AUTH_NAME` env vars so users can fully configure auth without editing the config file, or
   - **(b)** Add a test that documents this behavior and a code comment explaining that `auth-header-name` must still be set in the config file

   I'd recommend option (a) for completeness, but (b) is acceptable if you want to keep this PR focused.

## Fit and Finish

3. **Consider logging a warning when `GTFS_REALTIME_AUTH_VALUE` is set but no RT feeds are configured.** The `if len(config.GtfsRtFeeds) > 0` guard correctly prevents a panic, but the env var is silently dropped. A `slog.Warn` message would help users diagnose misconfiguration in production.

4. **Add a comment noting that only the first RT feed is overridden.** The env var name `GTFS_REALTIME_AUTH_VALUE` doesn't indicate it applies only to `GtfsRtFeeds[0]`. A one-line comment would prevent future confusion. (This matches the existing `ToGtfsConfigData()` pattern, which also only uses the first feed.)

aaronbrethorst (approved):
Hey Adel, the linter fix looks exactly right -- `t.Cleanup` with `_ = os.Remove(...)` and the explicit `require.NoError` on `tmpFile.Close()` are both idiomatic. `make lint` passes clean and all tests are green. We're there! Thanks for your patience across these rounds of reviews.

PR #339, opened 2026-02-07
Implement pagination for agency, route, and vehicle lists

### LINKED ISSUE #338: Feature: Implement Pagination for List Endpoints
## Goal
Address potential scalability and performance bottlenecks by introducing consistent `offset` and `limit` pagination logic across key list-based endpoints. This ensures the API remains responsive even when handling large datasets (e.g., thousands of routes or vehicles).

### 1. Shared Utility Implementation
**File:** `internal/utils/pagination.go`
- **Create `ParsePaginationParams`:** Implement a helper function to standardize query parameter parsing.
- **Constraints:**
    - **Default Limit:** `100` (if not specified).
    - **Max Limit:** `1000` (to prevent abuse).
    - **Offset:** Defaults to `0`.

### 2. Handler Refactoring
Refactor the following handlers to apply **slice-based pagination** on the retrieved datasets using the new helper:
- **Agencies:** `agencies-with-coverage`
- **Routes:** `routes-for-agency`
- **Vehicles:** `vehicles-for-agency`

### 3. Bug Fixes
- **File:** `internal/restapi/routes_for_agency_handler.go`
- **Fix:** Resolve a detected type mismatch preventing the handler from compiling or returning correct types.

## Endpoints Affected
- `GET /api/where/agencies-with-coverage.json`
- `GET /api/where/routes-for-agency.json`
- `GET /api/where/vehicles-for-agency.json`

## Verification Plan
- [ ] **Unit Tests:** Verify `ParsePaginationParams` handles edge cases (e.g., negative offsets, limits > 1000).
- [ ] **Integration Tests:** Call `agencies-with-coverage` with `limit=1` and `offset=0`, then `offset=1`, and ensure different records are returned.

AhmedAlian7:
## Description
This PR introduces consistent pagination support across three list-based API endpoints: `agencies-with-coverage`, `routes-for-agency`, and `vehicles-for-agency`. This ensures better scalability when handling large datasets.
## Changes
- **Added `ParsePaginationParams`:** A utility to extract `offset` (default 0) and `limit` (default 100, max 1000).
- **Added `PaginateSlice`:** A generic helper to safely slice result lists.
- **Updated Handlers:** Applied pagination logic to the following handlers using the new helpers.
- **Added Tests:** Comprehensive unit tests for parameter parsing and boundary conditions.

## Endpoints Modified
- `GET /api/where/agencies-with-coverage.json`
- `GET /api/where/routes-for-agency/{id}.json`
- `GET /api/where/vehicles-for-agency/{id}.json`

### DIFF
--- a/internal/models/response.go
+++ b/internal/models/response.go
@@ -18,9 +18,9 @@ func NewOKResponse(data interface{}, c clock.Clock) ResponseModel {
 	return NewResponse(200, data, "OK", c)
 }
 
-func NewListResponse(list interface{}, references ReferencesModel, c clock.Clock) ResponseModel {
+func NewListResponse(list interface{}, references ReferencesModel, limitExceeded bool, c clock.Clock) ResponseModel {
 	data := map[string]interface{}{
-		"limitExceeded": false,
+		"limitExceeded": limitExceeded,
 		"list":          list,
 		"references":    references,
 	}
--- a/internal/models/response_test.go
+++ b/internal/models/response_test.go
@@ -69,7 +69,7 @@ func TestNewListResponse(t *testing.T) {
 
 	clock := clock.RealClock{}
 
-	response := NewListResponse(itemList, references, clock)
+	response := NewListResponse(itemList, references, false, clock)
 
 	assert.Equal(t, http.StatusOK, response.Code)
 	assert.Equal(t, "OK", response.Text)
--- a/internal/restapi/agencies_with_coverage_handler.go
+++ b/internal/restapi/agencies_with_coverage_handler.go
@@ -4,6 +4,7 @@ import (
 	"net/http"
 
 	"maglev.onebusaway.org/internal/models"
+	"maglev.onebusaway.org/internal/utils"
 )
 
 func (api *RestAPI) agenciesWithCoverageHandler(w http.ResponseWriter, r *http.Request) {
@@ -24,9 +25,13 @@ func (api *RestAPI) agenciesWithCoverageHandler(w http.ResponseWriter, r *http.R
 		return
 	}
 
+	// Apply pagination
+	offset, limit := utils.ParsePaginationParams(r)
+	agencies, limitExceeded := utils.PaginateSlice(agencies, offset, limit)
+
 	lat, lon, latSpan, lonSpan := api.GtfsManager.GetRegionBounds()
-	var agenciesWithCoverage []models.AgencyCoverage
-	var agencyReferences []models.AgencyReference
+	agenciesWithCoverage := make([]models.AgencyCoverage, 0)
+	agencyReferences := make([]models.AgencyReference, 0)
 
 	for _, a := range agencies {
 		agenciesWithCoverage = append(
@@ -61,6 +66,6 @@ func (api *RestAPI) agenciesWithCoverageHandler(w http.ResponseWriter, r *http.R
 		Trips:      []interface{}{},
 	}
 
-	response := models.NewListResponse(agenciesWithCoverage, references, api.Clock)
+	response := models.NewListResponse(agenciesWithCoverage, references, limitExceeded, api.Clock)
 	api.sendResponse(w, r, response)
 }
--- a/internal/restapi/agencies_with_coverage_handler_test.go
+++ b/internal/restapi/agencies_with_coverage_handler_test.go
@@ -65,3 +65,40 @@ func TestAgenciesWithCoverageHandlerEndToEnd(t *testing.T) {
 	assert.Empty(t, refs["stops"])
 	assert.Empty(t, refs["trips"])
 }
+
+func TestAgenciesWithCoverageHandlerPagination(t *testing.T) {
+	// Test data (raba.zip) has 1 agency
+
+	// Case 1: Default (Offset 0, Limit -1) -> Should return 1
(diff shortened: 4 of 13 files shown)


aaronbrethorst (changes requested):
Hey Ahmed, thanks for taking the initiative to think about scalability for list endpoints. The utility code itself — `ParsePaginationParams` and `PaginateSlice` — is clean, well-bounded, and has strong unit tests for the parsing logic. Before we can merge this, I need you to address some design concerns:

## Critical

1. **Silent data truncation breaks existing clients** (`internal/utils/api.go:159`): The default limit of 100 means every existing OBA client that calls these endpoints without passing `limit` will silently receive truncated results. For `vehicles-for-agency`, a typical large transit agency has 1,000–5,000+ active vehicles — existing apps would suddenly get only 100 with no indication that data was cut off. This is a breaking change to the API contract.

   I checked the [OBA OpenAPI specification](https://github.com/OneBusAway/sdk-config/blob/main/openapi.yml) and confirmed that **none of these three endpoints define `offset` or `limit` parameters**. They are designed to return complete lists:
   - `agencies-with-coverage` — all agencies
   - `routes-for-agency` — all routes for the agency
   - `vehicles-for-agency` — all current vehicles for the agency

   We need to preserve this behavior. If you want to add pagination as an opt-in extension, only apply it when the client explicitly passes pagination parameters:

   ```go
   if r.URL.Query().Has("limit") || r.URL.Query().Has("offset") {
       offset, limit := utils.ParsePaginationParams(r)
       items = utils.PaginateSlice(items, offset, limit)
   }
   ```

2. **No `limitExceeded` signal in the response**: Other OBA list endpoints that support result limiting include `limitExceeded: true/false` so clients know whether more data exists. This PR truncates silently. If pagination is kept (as opt-in), the response must indicate when truncation occurred.

## Important

3. **Parameter naming contradicts OBA conventions**: The existing codebase uses `maxCount` for result limiting (`stops_for_location_handler.go:27`, `route_search_handler.go:38`, `search_stops_handler.go:52`). Introducing `limit`/`offset` creates an inconsistent API surface. If we add pagination, we should follow the established convention.

4. **Missing `PaginateSlice` unit tests** (`internal/utils/api.go:174-183`): This generic function is called in three handlers but has no dedicated unit tests. It's only tested indirectly via the agencies integration test with 1 item. Please add a table-driven test covering: empty/nil slices, offset at/beyond length, limit larger than remaining items, and normal multi-page pagination (e.g., 10 items, offset=3, limit=3 → items [3:6]).

5. **Four reference assertions removed without justification** (`internal/restapi/agencies_with_coverage_handler_test.go`): The `TestAgenciesWithCoverageHandlerEndToEnd` test previously asserted that `refs["routes"]`, `refs["situations"]`, `refs["stopTimes"]`, and `refs["stops"]` were empty. These were removed but the handler still produces those fields. Please restore them — they guard against structural regressions in the response.

6. **Please squash into a single commit and rebase against `main`.** The branch currently has 6 commits including a merge commit. A single atomic commit keeps the history clean. If you're unfamiliar with rebasing, here's a guide I wrote: https://www.brethorsting.com/blog/2026/01/git-rebase-for-the-terrified/

## Fit and Finish

7. **Pagination integration tests need a multi-item dataset**: The agencies pagination test uses RABA data with 1 agency, so every case is either "return everything" or "return nothing." Consider adding a pagination test for `routes-for-agency` instead — RABA has 13 routes, so you can meaningfully test `limit=5` returning 5 routes and `offset=5&limit=5` returning the next 5.

8. **Unsafe type assertions in pagination test cases 2–4** (`agencies_with_coverage_handler_test.go:78-92`): Case 1 uses `require.True(t, ok)` for safety, but cases 2–4 use bare type assertions (`data2 := model2.Data.(map[string]interface{})`) that will panic on failure instead of producing a test error.

aaronbrethorst (approved):
Hey Ahmed, great work addressing all the feedback from the first round. I've verified each item and everything looks solid. Here's the rundown:

## Previous Feedback Status

| # | Item | Status |
|---|------|--------|
| 1 | Silent data truncation (default limit) | Fixed — default is now `-1` (return all) |
| 2 | Missing `limitExceeded` in response | Fixed — `NewListResponse` and `PaginateSlice` both propagate it |
| 3 | Parameter naming (`maxCount` convention) | Fixed — `maxCount` checked first, `limit` as fallback |
| 4 | Missing `PaginateSlice` unit tests | Fixed — 10-case table-driven test |
| 5 | Reference assertions removed | Fixed — all 5 assertions restored |
| 6 | Squash into single commit | Fixed — single atomic commit |
| 7 | Multi-item pagination test | Fixed — routes-for-agency test uses 13 RABA routes |
| 8 | Unsafe type assertions in tests | Fixed — `require.True(t, ok)` guards throughout |

## Fit and Finish

Two minor observations (non-blocking):

1. **Bare type assertions in routes pagination test** (`routes_for_agency_handler_test.go:140-141`): `firstItem1 := list[0].(map[string]interface{})` would panic on failure instead of producing a test error. Safe in practice since the list length is already verified, but using `require.True` would be more consistent with the rest of the test.

2. **Agencies pagination test doesn't assert `limitExceeded`**: The routes pagination test covers this thoroughly, so it's not a gap — just noting it.

PR #342, opened 2026-02-08
Refactor: Centralize Global Cache & Fix Test Dependency Injection

vedanthnyk25:
This pull request introduces a global cache warmup for GTFS data and refactors how the direction calculator is used throughout the codebase. It also adds a new query for retrieving all stop IDs, updates the GTFS database query preparation and teardown logic, and makes minor improvements to the GTFS models. The most significant changes are grouped below.

**Global Cache Initialization and Direction Calculator Refactor:**

* Added a new `InitializeGlobalCache` function in `internal/gtfs/global_cache.go` to preload stop and shape data into memory at application startup, improving performance for direction calculations. The application now calls this function during initialization and sets up the advanced direction calculator with pre-cached data.
* Refactored REST API handlers to use the application's shared `DirectionCalculator` instance instead of creating new instances, ensuring cache utilization and consistency. 

**GTFS Database Query Enhancements:**

* Added a new `GetAllStopIDs` query and its associated statement preparation, closing, and struct wiring in `gtfsdb/db.go`, `gtfsdb/query.sql`, and generated code, enabling efficient retrieval of all stop IDs for cache warmup. 
* Fixed the ordering of statement preparation and closing for `searchRoutesByFullTextStmt` to ensure correct resource management. 

**Other Improvements:**

* Updated the `RoutesFt` struct in `gtfsdb/models.go` to better represent full-text search results on routes.
* Minor import cleanups and test improvements.

**Known Issues:**

-  SQLite FTS5 & SQLC Conflict There is a known conflict between sqlc validation and SQLite runtime behavior regarding the Full Text Search (FTS5) virtual table:

- The Issue: sqlc requires the column routes_fts to be explicitly defined in the schema to validate MATCH queries. However, explicitly defining this column causes SQLite to crash at runtime with "vtable constructor failed" (as routes_fts is a reserved shadow table name).

- Current State: The schema is currently set to the Runtime Safe version (the explicit column is removed).

- Impact: The application runs correctly and tests pass (with -tags sqlite_fts5), but running make models will fail.

- Workaround: To regenerate models, temporarily add the column to the schema, run make models, and then revert the change before running the app.

Please let me know if any improvements or corrections are required and what should be done regarding the SQLC & FTS5 conflict.

### DIFF
--- a/cmd/api/app.go
+++ b/cmd/api/app.go
@@ -78,6 +78,11 @@ func BuildApplication(cfg appconf.Config, gtfsCfg gtfs.Config) (*app.Application
 	var directionCalculator *gtfs.AdvancedDirectionCalculator
 	if gtfsManager != nil {
 		directionCalculator = gtfs.NewAdvancedDirectionCalculator(gtfsManager.GtfsDB.Queries)
+
+		err = gtfs.InitializeGlobalCache(context.Background(), gtfsManager.GtfsDB.Queries, directionCalculator)
+		if err != nil {
+			return nil, fmt.Errorf("failed to initialize global cache: %w", err)
+		}
 	}
 
 	// Select clock implementation based on environment
--- a/gtfsdb/db.go
+++ b/gtfsdb/db.go
@@ -117,6 +117,9 @@ func Prepare(ctx context.Context, db DBTX) (*Queries, error) {
 	if q.getAllShapesStmt, err = db.PrepareContext(ctx, getAllShapes); err != nil {
 		return nil, fmt.Errorf("error preparing query GetAllShapes: %w", err)
 	}
+	if q.getAllStopIDsStmt, err = db.PrepareContext(ctx, getAllStopIDs); err != nil {
+		return nil, fmt.Errorf("error preparing query GetAllStopIDs: %w", err)
+	}
 	if q.getAllTripsForRouteStmt, err = db.PrepareContext(ctx, getAllTripsForRoute); err != nil {
 		return nil, fmt.Errorf("error preparing query GetAllTripsForRoute: %w", err)
 	}
@@ -458,6 +461,11 @@ func (q *Queries) Close() error {
 	if q.getAllShapesStmt != nil {
 		if cerr := q.getAllShapesStmt.Close(); cerr != nil {
 			err = fmt.Errorf("error closing getAllShapesStmt: %w", cerr)
+		}
+	}
+	if q.getAllStopIDsStmt != nil {
+		if cerr := q.getAllStopIDsStmt.Close(); cerr != nil {
+			err = fmt.Errorf("error closing getAllStopIDsStmt: %w", cerr)
 		}
 	}
 	if q.getAllTripsForRouteStmt != nil {
@@ -835,6 +843,7 @@ type Queries struct {
 	getAgencyStmt                             *sql.Stmt
 	getAgencyForStopStmt                      *sql.Stmt
 	getAllShapesStmt                          *sql.Stmt
+	getAllStopIDsStmt                         *sql.Stmt
 	getAllTripsForRouteStmt                   *sql.Stmt
 	getArrivalsAndDeparturesForStopStmt       *sql.Stmt
 	getBlockDetailsStmt                       *sql.Stmt
@@ -933,6 +942,7 @@ func (q *Queries) WithTx(tx *sql.Tx) *Queries {
 		getAgencyStmt:                             q.getAgencyStmt,
 		getAgencyForStopStmt:                      q.getAgencyForStopStmt,
 		getAllShapesStmt:                          q.getAllShapesStmt,
+		getAllStopIDsStmt:                         q.getAllStopIDsStmt,
 		getAllTripsForRouteStmt:                   q.getAllTripsForRouteStmt,
 		getArrivalsAndDeparturesForStopStmt:       q.getArrivalsAndDeparturesForStopStmt,
 		getBlockDetailsStmt:                       q.getBlockDetailsStmt,
--- a/gtfsdb/query.sql
+++ b/gtfsdb/query.sql
@@ -194,6 +194,12 @@ ORDER BY
 LIMIT
     1;
 
+-- name: GetAllStopIDs :many
+SELECT
+    id
+FROM
+    stops;    
+
 -- name: GetStopIDsForAgency :many
 SELECT DISTINCT
     s.id
--- a/gtfsdb/query.sql.go
+++ b/gtfsdb/query.sql.go
@@ -1136,6 +1136,36 @@ func (q *Queries) GetAllShapes(ctx context.Context) ([]Shape, error) {
 	return items, nil
 }
 
+const getAllStopIDs = `-- name: GetAllStopIDs :many
+SELECT
+    id
+FROM
+    stops
+`
+
+func (q *Queries) GetAllStopIDs(ctx context.Context) ([]string, error) {
+	rows, err := q.query(ctx, q.getAllStopIDsStmt, getAllStopIDs)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+	var items []string
+	for rows.Next() {
+		var id string
+		if err := rows.Scan(&id); err != nil {
+			return nil, err
+		}
+		items = append(items, id)
+	}
+	if err := rows.Close(); err != nil {
+		return nil, err
+	}
+	if err := rows.Err(); err != nil {
+		return nil, err
+	}
+	return items, nil
+}
+
 const getAllTripsForRoute = `-- name: GetAllTripsForRoute :many
 SELECT DISTINCT id, route_id, service_id, trip_headsign, trip_short_name, direction_id, block_id, shape_id, wheelchair_accessible, bikes_allowed
 FROM trips t
--- a/internal/gtfs/advanced_direction_calculator_test.go
+++ b/internal/gtfs/advanced_direction_calculator_test.go
@@ -4,12 +4,52 @@ import (
 	"context"
 	"database/sql"
 	"math"
+	"os"
+	"sync"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"maglev.onebusaway.org/gtfsdb"
 	"maglev.onebusaway.org/internal/models"
 )
+
+// This uses a Singleton pattern to load the DB and Warm the Cache exactly ONCE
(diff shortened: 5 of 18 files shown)


aaronbrethorst (changes requested):
Hey Vedanth -- Excellent work addressing the feedback from the first round. All 12 items have been properly resolved, and the architectural approach is clean. The handler refactoring is consistent, the `InitializeGlobalCache` error handling is solid with proper `fmt.Errorf` wrapping, and the test singleton pattern is thoughtful. This is looking very close to merge-ready. Just a few remaining items to tidy up.

---

## Prior Feedback Status

All 12 items from the first review have been addressed:

| # | Item | Status |
|---|------|--------|
| 1 | Nil dereference if `gtfsManager` is nil | **Fixed** -- `InitializeGlobalCache` now inside `if gtfsManager != nil` |
| 2 | Generated code diverges from source SQL | **Resolved** -- `RoutesFt` struct exists on `main` already; FTS queries are in `fts_queries.go` |
| 3 | `InitializeGlobalCache` error silently discarded in tests | **Fixed** -- `panic()` in `advanced_direction_calculator_test.go`, `require.NoError` in `http_test.go` |
| 4 | Dead `dataPath` fields in table-driven tests | **Fixed** -- fields removed from test case structs |
| 5 | `TestManager_GetVehicleForTrip` mutates shared state | **Fixed** -- still creates its own isolated manager |
| 6 | `processRouteStops` unused `adc` parameter | **Fixed** -- parameter removed from `processRouteStops`, `buildStopsList`, and `BuildStopReferencesAndRouteIDsForStops` |
| 7 | `ShapePtSequence` dropped in shape cache | **Fixed** -- now included in `global_cache.go:62` |
| 8 | `DISTINCT` unnecessary in `GetAllStopIDs` | **Fixed** -- `SELECT id FROM stops` without `DISTINCT` |
| 9 | Misleading `slog.Warn` before `return err` | **Fixed** -- clean `fmt.Errorf` returns, no misleading logs |
| 10 | No `Shutdown()` on shared test manager | **Fixed** -- `TestMain` calls `sharedManager.Shutdown()` |
| 11 | `RoutesFt` struct unused | **Pre-existing** -- exists on `main`, not introduced by this PR |
| 12 | Error paths lack wrapping context | **Fixed** -- all three error returns use `fmt.Errorf("failed to ...: %w", err)` |

---

## Important

### 1. `createTestApiWithRealTimeData` missing `DirectionCalculator` (`vehicles_for_agency_handler_test.go:343-352`)

This function constructs an `app.Application` without setting `DirectionCalculator`. Since all handlers now use `api.DirectionCalculator.CalculateStopDirection(...)`, any test path that reaches a direction calculation will panic with a nil pointer dereference. Tests pass today because the real-time test paths don't happen to hit the direction calculation code, but this is a latent bug -- one new test or data change away from crashing.

**Fix:** Add `DirectionCalculator` initialization to `createTestApiWithRealTimeData`:

```go
dirCalc := gtfs.NewAdvancedDirectionCalculator(gtfsManager.GtfsDB.Queries)
err = gtfs.InitializeGlobalCache(context.Background(), gtfsManager.GtfsDB.Queries, dirCalc)
require.NoError(t, err)

application := &app.Application{
    // ...existing fields...
    DirectionCalculator: dirCalc,
}
```

### 2. `TestManager_GetVehicleForTrip` missing `defer manager.Shutdown()` (`gtfs_manager_test.go:301`)

This test correctly creates its own isolated manager (since it mutates `realTimeVehicles`), but doesn't clean it up. Add `defer manager.Shutdown()` after the nil check. Also worth adding a brief comment explaining why this test doesn't use `getSharedTestComponents` -- future contributors might wonder.

### 3. Vestigial single-element table-driven tests (`gtfs_manager_test.go`)

`TestManager_GetAgencies` and `TestManager_RoutesForAgencyID` still use table-driven structure with a single test case whose only field is `name`. The `dataPath` field that justified the table was removed, but the scaffolding remains. Simplify to direct tests:

```go
func TestManager_GetAgencies(t *testing.T) {
    manager, _ := getSharedTestComponents(t)
    assert.NotNil(t, manager)

    agencies := manager.GetAgencies()
    assert.Equal(t, 1, len(agencies))
    // ...
}
```

---

## Fit and Finish

### 4. `tc := tc` unnecessary in Go 1.24.2 (`gtfs_manager_test.go:94,280`)

Since Go 1.22, loop variables are scoped per-iteration. These captures are dead code. Remove them.

### 5. No-op `int64()` cast (`global_cache.go:62`)

`GetShapePointsByIDsRow.ShapePtSequence` is already `int64`, so `int64(p.ShapePtSequence)` is a no-op cast. The comment `// Added ShapePtSequence` is also unclear. Simplify to just `ShapePtSequence: p.ShapePtSequence`.

### 6. Noise comments in `global_cache.go`

Comments like `// Map the DB row to the Cache row struct`, `// Set Cache`, and `// Fetch ALL Stop IDs` describe what the next line obviously does. The code is clear without them. Consider removing or replacing with comments that explain *why*.

---

## Pre-existing (not blocking, but noted)

- **`schedule_for_route_handler.go:193,219`**: Two `if err == nil` patterns for `GetAgency` and `GetTripsByIDs` silently swallow errors. The PR correctly fixed the `BuildStopReferencesAndRouteIDsForStops` call on line 244, but these two remain. Worth addressing in a follow-up.

---

Nice job systematically addressing all the first-round feedback, Vedanth. The important items above are straightforward fixes. Once those are in, this is good to go.

aaronbrethorst (changes requested):
Hey Vedanth -- Systematic and thorough work addressing every item from both prior rounds. All 6 items from the second review are properly resolved, and the `schedule_for_route_handler.go` error handling fix goes beyond what was required by also fixing the two pre-existing `if err == nil` patterns I noted as "not blocking." That's great initiative. Just one handler that was missed in the refactor and a couple of smaller items.

---

## Prior Feedback Status

All 6 items from the second review have been addressed:

| # | Item | Status |
|---|------|--------|
| 1 | `createTestApiWithRealTimeData` missing `DirectionCalculator` | **Fixed** -- `dirCalc` created, cache warmed, `DirectionCalculator` set on application struct |
| 2 | `TestManager_GetVehicleForTrip` missing `defer manager.Shutdown()` | **Fixed** -- `defer manager.Shutdown()` added with explanatory comment |
| 3 | Vestigial single-element table-driven tests | **Fixed** -- `TestManager_GetAgencies` and `TestManager_RoutesForAgencyID` are now flat tests |
| 4 | `tc := tc` unnecessary in Go 1.24.2 | **Fixed** -- removed |
| 5 | No-op `int64()` cast and comment | **Fixed** -- `ShapePtSequence: p.ShapePtSequence` with no cast |
| 6 | Noise comments in `global_cache.go` | **Fixed** -- noise comments removed, useful comments retained |

Bonus: The pre-existing `if err == nil` patterns in `schedule_for_route_handler.go` for `GetAgency` (line 184) and `GetTripsByIDs` (line 213) have also been fixed with proper `if err != nil` error handling. Nice catch.

---

## Important

### 1. `stops_for_location_handler.go` not migrated to shared `DirectionCalculator` (`stops_for_location_handler.go:203`)

This handler still creates a per-request calculator:

```go
calc := gtfs.NewAdvancedDirectionCalculator(api.GtfsManager.GtfsDB.Queries)
```

Every other handler was refactored to use `api.DirectionCalculator`, but this one was missed. The fresh calculator won't have the global cache populated, so it falls back to per-stop database queries on every request -- defeating the purpose of the global cache warmup for this endpoint.

**Fix:** Replace with `api.DirectionCalculator.CalculateStopDirection(...)`, following the same pattern as the other handlers.

### 2. No dedicated unit tests for `InitializeGlobalCache` (`global_cache.go`)

The function is exercised indirectly via `getSharedTestComponents()` and `createTestApiWithClock()`, but there's no `global_cache_test.go` that directly tests:
- Error paths (what happens when `GetAllStopIDs`, `GetStopsWithShapeContextByIDs`, or `GetShapePointsByIDs` fail)
- Empty database (0 stops)
- Stops with no associated shapes (`len(uniqueShapeIDs) == 0` branch)

Since this function is the sole mechanism for populating the direction caches in production, it deserves direct test coverage.

---

## Fit and Finish

### 3. `Fetch Context (Stop -> Shape mappings)` comment (`global_cache.go:19`)

This comment is borderline -- it explains what "context rows" means, which is useful. But the `// Fetch Shape Points (Geometry)` comment on line 48 follows the same "what, not why" pattern from the previous round. Consider removing line 48 or rephrasing to explain *why* shape points are needed separately from context rows.

### 4. Blank line at top of `TestManager_GetAgencies` and `TestManager_RoutesForAgencyID` (`gtfs_manager_test.go:18,38`)

Both functions have a leading blank line before the first statement. Minor formatting nit.

---

## Pre-existing (not blocking, but noted)

- **`schedule_for_route_handler.go:138-139,146-147`**: Silent `continue` statements in the inner loop for `GetOrderedStopIDsForTrip` and `GetStopTimesForTrip` -- database errors are silently swallowed with no logging. Worth a follow-up.
- **`stops_for_route_handler.go:288-289,303-304`**: Same silent `continue` pattern in `processTripGroups`.

---

The missed handler in item 1 is the key remaining fix. Once that's in and `global_cache_test.go` is added, this is good to merge.

aaronbrethorst (approved):
Vedanth, everything looks great. Clean rebase, single squashed commit, and all outstanding items resolved.

## Prior Feedback Status

All 4 items from the fifth review have been addressed:

| # | Item | Status |
|---|------|--------|
| 1 | Rebase and squash before merge | **Fixed** — single commit: `1afe5a9 Refactor: centralize direction calculator with global cache warmup` |
| 2 | Blank line at top of `TestManager_GetVehicleForTrip` | **Fixed** — no leading blank line at line 264 |
| 3 | Missing space after `//` in comment | **Fixed** — proper `// We use isolated...` at line 270 |
| 4 | `TestManager_FindRoute_UsesMap` leading blank line | **Fixed** — no leading blank line at line 347 |

Bonus: The `Error` → `Warn` suggestion from round 4 (item 3, noted as "not blocking") was also addressed — both `schedule_for_route_handler.go:157,166` and `stops_for_route_handler.go:289,305` now correctly use `Warn` for per-trip data failures.

## Critical Issues (0 found)

None.

## Important Issues (0 found)

None.

## Checklist

- [x] Tests pass
- [x] Lint clean
- [x] Single squashed commit
- [x] All prior feedback addressed (rounds 1–5)

---

Ready for merge. Nice work.

PR #347, opened 2026-02-08
fix(build): remove debug compiler flags from production build

### LINKED ISSUE #345: fix(build): remove debug compiler flags from production build target
The [Makefile](vscode-file://vscode-app/c:/Users/HP/AppData/Local/Programs/Microsoft%20VS%20Code/bdd88df003/resources/app/out/vs/code/electron-browser/workbench/workbench.html) build target currently uses -gcflags "all=-N -l", which disables all compiler optimizations and function inlining. This is a debugger flag that should never be in a production build it degrades performance significantly and increases binary size.

Current behavior:
build:
    go build -gcflags "all=-N -l" -o bin/maglev ./cmd/api

**Proposed fix:**
- Remove -gcflags from the default build target
- Add a separate build-debug target with -gcflags "all=-N -l" for Delve debugging
- Optionally inject version info via -ldflags for production builds

**Impact:** Direct performance improvement for all deployed instances with zero risk.

GaneshPatil7517:
Removes debug compiler flags from production builds

Introduces a dedicated build-debug target for Delve debugging

This improves runtime performance and reduces binary size with no behavior
change for developers.

Resolves #345.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ else
     SET_ENV := CGO_ENABLED=1 CGO_CFLAGS="-DSQLITE_ENABLE_FTS5"
 endif
 
-.PHONY: build clean coverage test run lint watch fmt \
+.PHONY: build build-debug clean coverage test run lint watch fmt \
 	gtfstidy models check-golangci-lint \
 	docker-build docker-run docker-stop docker-compose-up docker-compose-down docker-compose-dev docker-clean docker-clean-all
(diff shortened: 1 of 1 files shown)


aaronbrethorst (approved):
Hey Ganesh, nice catch on the debug flags in the production build! This is a clean, focused fix, and ready to go. I'm merging this.

PR #350, opened 2026-02-08
fix: Correct route agency ID generation in arrival-and-departure-for-stop handler

### LINKED ISSUE #349: Bug: Incorrect Route ID prefix and missing agency reference in arrival/departure handlers (Multi-Agency)
**Description:**
The `arrival-and-departure-for-stop` (singular) and `arrivals-and-departures-for-stop` (plural) handlers currently assume that a Route's Agency ID always matches the Stop's Agency ID.

Similar to issue #334

In multi-agency environments where a route from "Agency B" serves a stop from "Agency A", this causes two issues:

1. **Invalid Route IDs:** The response generates the `routeId` using the Stop's agency (e.g., `AgencyA_RouteID`) instead of the Route's actual agency (`AgencyB_RouteID`).
2. **Missing Metadata:** The `references.agencies` section only includes the Stop's agency. The client receives a route reference but has no corresponding agency metadata to look up "Agency B".

**Location:**

* `internal/restapi/arrival_and_departure_for_stop_handler.go`
* `internal/restapi/arrivals_and_departure_for_stop.go`

**Impact:**
Clients receive broken route IDs that lead to 404s when queried. Cross-agency routes fail to display agency details (name, URL, timezone) because the reference is missing.

**Proposed Fix:**

1. Update `NewArrivalAndDeparture` and `NewTripReference` in both handlers to explicitly use `route.AgencyID` when generating the combined Route ID.
2. Add a check to inject the Route's Agency into `references.agencies` if it differs from the Stop's Agency.
3. Add regression tests in `arrival_and_departure_for_stop_handler_test.go` and `arrivals_and_departures_for_stop_handler_test.go` to verify multi-agency ID generation.

ARCoder181105:
**New Title**
fix: Correct route agency ID generation in arrival/departure handlers

**Problem**
The `arrival-and-departure-for-stop` (singular) and `arrivals-and-departures-for-stop` (plural) handlers incorrectly assume that a Route's Agency ID always matches the Stop's Agency ID.

In multi-agency setups (e.g., Agency B route serving Agency A stop), this results in:

closes #349

1. **Invalid Route IDs:** The `routeId` is prefixed with the stop's agency (e.g., `AgencyA_RouteID`) instead of the route's actual agency (`AgencyB_RouteID`), causing broken links.
2. **Missing Metadata:** The `references` response fails to include agency metadata for the route if it differs from the stop.

**Solution**

* Updated `arrivalAndDepartureForStopHandler` and `arrivalsAndDeparturesForStopHandler` to explicitly use `route.AgencyID` when generating combined Route IDs and Trip References.
* Added logic to both handlers to inject the Route's Agency metadata into the `references.agencies` list if it differs from the Stop's Agency.
* Ensured active trips (real-time) also fetch and register their route metadata correctly in the plural handler.

**Testing**

* Added regression test `TestArrivalAndDepartureForStopHandler_MultiAgency_Regression` to `arrival_and_departure_for_stop_handler_test.go`.
* Added regression test `TestArrivalsAndDeparturesForStopHandler_MultiAgency_Regression` to `arrivals_and_departures_for_stop_handler_test.go`.
* Both tests verify that the returned `routeId` is prefixed with the correct agency and that the corresponding agency metadata is present in the references.

### DIFF
--- a/internal/restapi/arrival_and_departure_for_stop_handler.go
+++ b/internal/restapi/arrival_and_departure_for_stop_handler.go
@@ -101,7 +101,7 @@ func (api *RestAPI) parseArrivalAndDepartureParams(r *http.Request) (ArrivalAndD
 
 func (api *RestAPI) arrivalAndDepartureForStopHandler(w http.ResponseWriter, r *http.Request) {
 	parsed, _ := utils.GetParsedIDFromContext(r.Context())
-	agencyID := parsed.AgencyID
+	stopAgencyID := parsed.AgencyID
 	stopCode := parsed.CodeID
 	stopID := parsed.CombinedID
 
@@ -148,7 +148,7 @@ func (api *RestAPI) arrivalAndDepartureForStopHandler(w http.ResponseWriter, r *
 		return
 	}
 
-	agency, err := api.GtfsManager.GtfsDB.Queries.GetAgency(ctx, agencyID)
+	stopAgency, err := api.GtfsManager.GtfsDB.Queries.GetAgency(ctx, stopAgencyID)
 	if err != nil {
 		api.serverErrorResponse(w, r, err)
 		return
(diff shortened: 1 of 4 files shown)


ARCoder181105:
@aaronbrethorst 
The test are passing perfectly, PR is ready for review

image

ARCoder181105:
Made the requested changes , add some extra comments also, the PR is ready for review
@aaronbrethorst

PR #352, opened 2026-02-08
Implement Prometheus Instrumentation for Maglev Service to Track HTTP Latency and Database Pool Statistics

### LINKED ISSUE #351: Feature: Implement Prometheus Instrumentation for Maglev Service to Track HTTP Latency and Database Pool Statistics
## Description
The Maglev service currently lacks observability into its runtime performance and internal state. There are no exposed metrics for critical operational data such as HTTP request latency, error rates, throughput, or database connection pool statistics.

Without these metrics, it is impossible to:
1. Configure automated alerting for high error rates or latency spikes.
2. Visualize system health in Grafana.
3. Diagnose performance bottlenecks (e.g., differentiating between slow handlers and database exhaustion).

Currently, the only way to inspect system behavior is via raw logs, which is insufficient for real-time monitoring and incident response.

## Proposed Solution
Integrate the Prometheus client library (`github.com/prometheus/client_golang`) to instrument the application.

### 1. HTTP Middleware
Implement a middleware function to wrap all API handlers to track:
* **Request Count:** `maglev_http_requests_total` (labels: `method`, `path`, `status`).
* **Latency:** `maglev_http_request_duration_seconds` (labels: `method`, `path`).
* **Cardinality Control:** Ensure path normalization is implemented (e.g., `/users/:id` instead of `/users/123`) to prevent high-cardinality metrics.

### 2. Database Instrumentation
Implement a background collector to query `database/sql` stats and expose the following gauges:
* `maglev_db_connections_open`
* `maglev_db_connections_in_use`
* `maglev_db_connections_idle`
* `maglev_db_wait_seconds_total`

### 3. Endpoint Exposure
* Register a custom Prometheus registry to avoid global state conflicts.
* Expose the `/metrics` endpoint (supporting standard Prometheus exposition format).
* Ensure the solution supports graceful shutdown.

AhmedAlian7:
## Description
This PR adds Prometheus metrics to the maglev application to enable real-time monitoring, alerting, and visualization via Grafana.
Previously, there was no observability into runtime performance—the only way to inspect system behavior was through raw logs, which is insufficient for real-time monitoring and incident response.

**closes #351**

### New Files
* `internal/metrics/metrics.go`: Metric definitions and registry.
* `internal/restapi/metrics_middleware.go`: HTTP request instrumentation middleware.

### Modified Files
* `internal/app/application.go`: Added `Metrics` and `MetricsCancel` fields to Application struct.
* `cmd/api/app.go`: Integrated metrics initialization, middleware chain, and `/metrics` endpoint.
* `cmd/api/main.go`: Updated `Run` function signature to pass Application for proper shutdown.
* `go.mod` / `go.sum`: Added `github.com/prometheus/client_golang` dependency.

## Metrics Exposed

| Metric Name | Type | Description | Labels |
| :--- | :--- | :--- | :--- |
| `maglev_http_requests_total` | Counter | Total HTTP requests | `method`, `path`, `status` |
| `maglev_http_request_duration_seconds` | Histogram | Request latency distribution | `method`, `path` |
| `maglev_db_connections_open` | Gauge | Open database connections | - |
| `maglev_db_connections_in_use` | Gauge | Active database connections | - |
| `maglev_db_connections_idle` | Gauge | Idle database connections | - |
| `maglev_db_wait_seconds_total` | Counter | Time blocked waiting for DB connection | - |

## Implementation Details
* **Custom Registry:** Uses a custom Prometheus registry to avoid global state conflicts in tests.
* **Background Collection:** DB stats collector runs on a separate background goroutine with proper lifecycle management.
* **Streaming Support:** HTTP middleware supports streaming (`Flusher`, `Hijacker`) for SSE/WebSocket compatibility.
* **Cardinality Control:** Path normalization is implemented to prevent high-cardinality metrics by using route patterns instead of actual IDs.
* **Graceful Shutdown:** Implemented via context cancellation and `sync.Once` for safe multi-call Shutdown.

## Testing
* All existing tests pass.
* Manual verification performed on endpoint: `GET /metrics`.
* Validated that the response adheres to the Prometheus exposition format.

## Usage
```bash
# Start server
make run

# View metrics
curl http://localhost:4000/metrics

### CODE REVIEW

**Review by aaronbrethorst — CHANGES_REQUESTED** [review]
Hey Ahmed, thanks for taking the time to add Prometheus instrumentation to Maglev. Having observability into HTTP latency and DB connection pool health is genuinely important for running this in production, and I appreciate the thought you put into cardinality control and the custom registry. Before we can merge this, I will need you to make a couple changes:

## Critical

1. **Use `r.Pattern` instead of manual path normalization** (`metrics_middleware.go:91-165`). The `staticSegments` map approach has an unbounded cardinality vulnerability: any path segment that isn't in the map *and* doesn't match the numeric ID regex passes through verbatim. An attacker or web crawler hitting arbitrary URLs (`/api/where/stop/some-random-string`, `/wp-admin/login`, etc.) would each create a unique Prometheus time series, which can cause OOM in Prometheus. Since we target Go 1.24, you can use `r.Pattern` from Go 1.22+ which gives you the registered route pattern directly:

   ```go
   path := r.Pattern
   if path == "" {
       path = "unmatched"
   }
   ```

   This eliminates the entire `staticSegments` map, `stripExtension`, `idPattern`, and `normalizePath` function, and it can never drift out of sync with `routes.go` when new endpoints are added.

2. **The `maglev_db_wait_seconds_total` metric tracks count, not seconds** (`metrics.go:75-78,131-135`). The metric name and help text say "seconds" / "Total time blocked waiting for a database connection", but the code increments by `stats.WaitCount` (number of waits), not `stats.WaitDuration` (actual time). Anyone building Grafana dashboards on this metric will misinterpret the data. Either rename to `maglev_db_wait_count_total` or change the code to use `stats.WaitDuration.Seconds()`.

3. **Add tests for the new code**. The `internal/metrics` package has `[no test files]` and `metrics_middleware.go` has no corresponding test file either. Every other middleware in this project has tests (`rate_limit_middleware_test.go`, `security_middleware_test.go`, `request_logging_test.go`). At minimum, please add:
   - `metrics_middleware_test.go`: table-driven tests for path normalization (or whatever replaces it), status code capture in `metricsResponseWriter`, and an integration test confirming metrics are actually recorded after a request.
   - `metrics_test.go`: tests for `New()`, `StartDBStatsCollector` idempotency, nil DB handling, `Shutdown` safety, and the goroutine actually stopping on shutdown.

## Important

4. **Remove the `status` label from the histogram** (`metrics.go:57`, `metrics_middleware.go:50`). Including `status` as a label on `HTTPRequestDuration` is a well-known anti-pattern -- each unique `(method, path, status)` combination creates 14 time series (11 buckets + sum + count + inf). The counter `HTTPRequestsTotal` already tracks status codes, which is the right place for it. For the histogram, `method` and `path` are sufficient.

5. **Eliminate the dual shutdown mechanism** (`metrics.go:30-36,138-141`, `app/application.go:24`). The DB stats collector goroutine listens on both `ctx.Done()` and `stopChan`, and the caller must invoke both `MetricsCancel()` and `Metrics.Shutdown()`. Pick one. I'd suggest keeping only the context: have `StartDBStatsCollector` create its own child context internally, and have `Shutdown()` cancel it. This removes `stopChan`, `closeOnce`, and the `MetricsCancel` field from `Application` (which leaks a lifecycle implementation detail into the DI container).

6. **Add panic recovery to the DB stats collector goroutine** (`metrics.go:119`). If `db.Stats()` panics (e.g., on a closed database), the goroutine dies silently and DB metrics freeze at their last values. Stale metrics masquerading as current data is worse than no metrics. Add a `defer recover()` with error logging.

7. **Confirm the goroutine has stopped before proceeding with DB shutdown** (`metrics.go:149-153`, `app.go:188-191`). `Shutdown()` closes the stop channel and returns immediately, but the goroutine may still be mid-execution of `db.Stats()` when `GtfsManager.Shutdown()` closes the database on the next line. Add a `sync.WaitGroup` or done channel so `Shutdown()` blocks until the goroutine has confirmed exit.

8. **Route `promhttp` errors through the structured logger** (`app.go:110`). The empty `promhttp.HandlerOpts{}` sends metric collection errors to stderr, bypassing the app's structured logging. Set `ErrorLog`:
   ```go
   promhttp.HandlerOpts{
       ErrorLog: slog.NewLogLogger(coreApp.Logger.Handler(), slog.LevelError),
   }
   ```

## Fit and Finish

9. **Remove `Flusher`/`Hijacker` implementations from `metricsResponseWriter`** (`metrics_middleware.go:75-88`). This app has no WebSocket or SSE endpoints, and no other response writer wrapper in the codebase implements these. The `Flush()` implementation is particularly problematic -- it silently swallows the operation when the underlying writer doesn't support it, which means callers that type-assert to `http.Flusher` get `true` but the flush is a no-op. Remove both and add them back if/when the app actually needs them. (If you keep `Flush` for gzip compatibility, at least remove `Hijack`.)

10. **Initialize `metricsResponseWriter.statusCode` to `http.StatusOK`** (`metrics_middleware.go:34`). The current approach defaults to 0 and then checks for 0 in two separate places (`Write` on line 69 and `Handler` on line 41). Setting the initial value to 200 at construction eliminates both conditional checks and the comment explaining the behavior.

11. **Add a nil guard to `NewMetricsMiddleware`** (`metrics_middleware.go:21-23`). If someone passes nil (which the test helper in `http_test.go` would, since it doesn't set `Metrics`), every HTTP request will panic with a nil pointer dereference on line 49. Either panic at construction time (it's a programmer error) or return a pass-through no-op middleware.

12. **Consider simplifying `MetricsMiddleware` to a plain function** (`metrics_middleware.go:16-52`). The struct holds a single field and has one method. Unlike `RateLimitMiddleware` which has genuine state (limiter maps, cleanup goroutines), this has none. A simple `func MetricsHandler(m *metrics.Metrics) func(http.Handler) http.Handler` would be cleaner. Not a blocker, just a suggestion.

**Review by aaronbrethorst — CHANGES_REQUESTED** [review]
Ahmed, you are *so* close. The rebase landed cleanly — lint passes, all tests pass, the code is exactly as I approved it. Just one thing crept in during the rebase.

---

## Critical

### 1. A 23MB compiled binary is committed to the repository

**File:** `api` (repository root)

The PR includes a Linux ELF executable at the root of the repository:

```
api: ELF 64-bit LSB executable, x86-64, 23MB
```

This is a Go build artifact — likely produced by running `go build` from `cmd/api/` which placed the output in the working directory. The `.gitignore` has `/cmd/api/api` but not `/api` at the root, so git didn't catch it.

**Fix:** Remove the file from the commit:

```bash
git rm api
git commit --amend --no-edit
git push --force-with-lease
```

That's it. Everything else is exactly as approved. Once the binary is removed, this is ready to merge.

Issue #353, opened 2026-02-08
Critical: Resolve race condition in GetAlertsForTrip


3rabiii:
### Description
I identified a critical concurrency issue (Time-of-Check to Time-of-Use race condition) in the `GetAlertsForTrip` function within `internal/gtfs/realtime.go`.

### The Issue 
The function currently checks if the database is initialized without holding the necessary read lock:

```go
// Current code
if manager.GtfsDB != nil {
    trip, err := manager.GtfsDB.Queries.GetTrip(ctx, tripID)
}
```
If a concurrent ForceUpdate (or data reload) occurs immediately after the nil check but before the query execution, the GtfsDB instance could be closed or set to nil, leading to a panic or undefined behavior.

Proposed Solution 
 * Protect the static data access by acquiring the read lock, consistent with other methods in gtfs_manager.go.

Plan:
 * Acquire manager.RLock() before checking manager.GtfsDB.
 * Release manager.RUnlock() immediately after the static database operations are complete and before proceeding to the real-time alerts logic (to          avoid holding the lock longer than necessary).

Impact
 * This fix prevents potential server crashes during GTFS data updates.

PR #354, opened 2026-02-08
fix: resolve race condition in GetAlertsForTrip by adding read lock

3rabiii:
### Description
This PR resolves the critical race condition identified in #353

The `GetAlertsForTrip` function was performing a check on `GtfsDB` without holding the necessary read lock. This created a **TOCTOU (Time-of-Check to Time-of-Use)** vulnerability where a concurrent `ForceUpdate` could set `GtfsDB` to `nil` immediately after the check but before the query execution, leading to a potential panic.

### Changes 
1.  **Concurrency Safety:** Added `manager.RLock()` before accessing the static `GtfsDB` instance.
2.  **Performance:** Explicitly called `manager.RUnlock()` immediately after the database queries are complete and **before** entering the real-time processing block. This ensures the static lock is not held longer than necessary, preventing lock contention.

@aaronbrethorst 
fixes : #353

### DIFF
--- a/internal/gtfs/realtime.go
+++ b/internal/gtfs/realtime.go
@@ -71,9 +71,11 @@ func (manager *Manager) GetAlertsForRoute(routeID string) []gtfs.Alert {
 }
 
 // GetAlertsForTrip returns alerts matching the trip, its route, or agency.
+// IMPORTANT: Caller must hold manager.RLock() before calling this method.
 func (manager *Manager) GetAlertsForTrip(ctx context.Context, tripID string) []gtfs.Alert {
 	var routeID string
 	var agencyID string
+
 	if manager.GtfsDB != nil {
 		trip, err := manager.GtfsDB.Queries.GetTrip(ctx, tripID)
 		if err == nil {


aaronbrethorst (changes requested):
Hey Adel -- nice catch identifying the TOCTOU race condition in `GetAlertsForTrip`. You're absolutely right that accessing `manager.GtfsDB` without holding the static mutex is unsafe when `ForceUpdate` can nil it out concurrently. That said, the fix as implemented will introduce a deadlock, so I need to request changes.

## Critical

### Deadlock: Double `RLock()` from the same goroutine

The PR adds `manager.RLock()` / `manager.RUnlock()` inside `GetAlertsForTrip`, but **all three callers already hold `staticMutex.RLock()`**:

1. **`tripForVehicleHandler`** (`trip_for_vehicle_handler.go:98-99`):
   ```go
   api.GtfsManager.RLock()
   defer api.GtfsManager.RUnlock()
   // ... later at line 198:
   alerts := api.GtfsManager.GetAlertsForTrip(r.Context(), vehicle.Trip.ID.ID)
   ```

2. **`arrivalAndDepartureForStopHandler`** (`arrival_and_departure_for_stop_handler.go:116-117`):
   ```go
   api.GtfsManager.RLock()
   defer api.GtfsManager.RUnlock()
   // ... later at line 484:
   alerts := api.GtfsManager.GetAlertsForTrip(r.Context(), tripID)
   ```

3. **`GetSituationIDsForTrip`** (`trips_helper.go:769-771`):
   ```go
   func (api *RestAPI) GetSituationIDsForTrip(...) []string {
       alerts := api.GtfsManager.GetAlertsForTrip(ctx, tripID)
   ```

Go's `sync.RWMutex` is **not reentrant**. The deadlock scenario:

1. **Goroutine A** (HTTP handler): Holds first `RLock()` from the handler
2. **Goroutine B** (`ForceUpdate`): Calls `Lock()` at `static.go:251` -- blocks, waiting for A to release
3. **Goroutine A**: Enters `GetAlertsForTrip`, calls `RLock()` again -- **blocks** because B's pending `Lock()` takes priority over new readers
4. **Deadlock**: A waits for B, B waits for A

This is documented in Go's [`sync.RWMutex` docs](https://pkg.go.dev/sync#RWMutex):

> If any goroutine calls Lock while the lock is already held by one or more readers, concurrent calls to RLock will block until the writer has acquired (and released) the lock.

### Recommended fix

Follow the existing codebase pattern. Instead of locking inside `GetAlertsForTrip`, add the documentation comment that callers must hold the lock (they already do):

```go
// GetAlertsForTrip returns alerts matching the trip, its route, or agency.
func (manager *Manager) GetAlertsForTrip(ctx context.Context, tripID string) []gtfs.Alert {
```

No `RLock()`/`RUnlock()` calls should be added inside the method. The callers already hold the lock, so the race condition is already protected -- the only thing missing is the documentation comment making this contract explicit.

## Important

### `GetSituationIDsForTrip` duplicates work

While reviewing this, I noticed that `GetSituationIDsForTrip` (`trips_helper.go:770-781`) calls `GetAlertsForTrip` and then immediately re-queries `GetTrip` and `GetRoute` to get the `agencyID` -- the exact same DB queries that `GetAlertsForTrip` already performs internally. This isn't introduced by this PR, but it's worth noting as a follow-up optimization.

3rabiii:
Thanks for the detailed explanation on the deadlock! 
I've removed the internal `manager.RLock()` calls from `GetAlertsForTrip` and added the documentation comment enforcing the lock contract.
Regarding the `GetSituationIDsForTrip` optimization: I'm working on a follow-up PR for that right now! 

@aaronbrethorst 
closes : #353

aaronbrethorst (approved):
Adel, perfect — this is exactly the right fix. The deadlock-inducing `RLock()`/`RUnlock()` calls are gone, replaced by the documentation comment making the caller-holds-lock contract explicit. Clean and correct.

PR #356, opened 2026-02-08
Optimize `stop_times` lookups with composite index

### LINKED ISSUE #355: Optimization: Add composite index to `stop_times` for faster schedule lookups
### Problem

The `stop_times` table currently has a single-column index on `stop_id`.
However, frequently used queries like `GetStopTimesForStopInWindow` and `GetArrivalsAndDeparturesForStop` filter by `stop_id` but also **sort or filter by** `arrival_time`.

This causes the database to:

1. Fetch all rows for a stop (which can be thousands for major hubs).
2. Perform a full scan and in-memory sort operation to order them by time.

### Proposed Solution

Replace the existing single-column index with a composite index on `(stop_id, arrival_time)`. This allows SQLite to retrieve rows that are already sorted, significantly reducing CPU usage and latency for these high-traffic endpoints.

### Implementation Details

Modify `gtfsdb/schema.sql`:

```sql
-- Remove the old index
-- CREATE INDEX IF NOT EXISTS idx_stop_times_stop_id ON stop_times (stop_id);

-- Add the new composite index
CREATE INDEX IF NOT EXISTS idx_stop_times_stop_arrival ON stop_times (stop_id, arrival_time);

```

### Verification

* The existing queries in `gtfsdb/query.sql` already use `ORDER BY arrival_time`, so they will automatically utilize this new index.
* Run `make models` to regenerate the schema artifacts.

ARCoder181105:
**Description:**
This PR optimizes database performance for schedule-related queries.

closes #355 

Currently, queries like `GetStopTimesForStopInWindow` and `GetArrivalsAndDeparturesForStop` use an index on `stop_id` but must perform an expensive scan and sort operation to order results by time.

**Changes:**

* Replaced the single-column index `idx_stop_times_stop_id` with a composite index `idx_stop_times_stop_arrival` on `(stop_id, arrival_time)`.
* Ran `make models` to update schema artifacts.

**Impact:**

* Significantly reduces CPU usage and latency for high-traffic stop schedule endpoints by allowing SQLite to retrieve pre-sorted rows directly from the index.
* The new index still satisfies pure `stop_id` lookups, making the old index redundant.

**Verification:**

* Verified via `sqlite_master` that the new composite index is created and the old redundant index is removed.
* Existing queries automatically utilize the new index structure without modification.

### DIFF
--- a/gtfsdb/schema.sql
+++ b/gtfsdb/schema.sql
@@ -330,7 +330,10 @@ CREATE INDEX IF NOT EXISTS idx_trips_service_id ON trips (service_id);
 CREATE INDEX IF NOT EXISTS idx_stop_times_trip_id ON stop_times (trip_id);
 
 -- migrate
-CREATE INDEX IF NOT EXISTS idx_stop_times_stop_id ON stop_times (stop_id);
+DROP INDEX IF EXISTS idx_stop_times_stop_id;
+
+-- migrate
+CREATE INDEX IF NOT EXISTS idx_stop_times_stop_arrival ON stop_times (stop_id, arrival_time);
 
 -- migrate
 CREATE INDEX IF NOT EXISTS idx_stop_times_stop_id_trip_id ON stop_times (stop_id, trip_id);


ARCoder181105:
@aaronbrethorst 
The test are passing perfectly, the PR is ready for review

image

aaronbrethorst (changes requested):
Hey Aditya, the `DROP INDEX` migration is exactly what I asked for — existing databases will now cleanly replace the old single-column index with the new composite one. Before we can merge this, I need one more thing:

## Important

### 1. Squash and rebase

The branch has 3 commits (including a merge commit) and should be squashed into a single atomic commit and rebased against `main`. Here's the guide if you need it: https://www.brethorsting.com/blog/2026/01/git-rebase-for-the-terrified/

PR #364, opened 2026-02-09
fix: Add missing service calendar in TestStopHandlerMultiAgencyScenario

ARCoder181105:
## Description

Fixes a `FOREIGN KEY constraint failed` error in `TestStopHandlerMultiAgencyScenario` by adding the missing calendar/service entry that trips reference.

## Problem

The test creates trips with `ServiceID: "service1"` but never creates the corresponding calendar entry in the database, causing the test to fail with:
```
FOREIGN KEY constraint failed
Error Trace: /internal/restapi/stop_handler_test.go:214
```

## Solution

Added `CreateCalendar` call to create the `service1` calendar entry before creating trips that reference it.

## Changes
- Added calendar creation in `TestStopHandlerMultiAgencyScenario` (line ~205)
- Ensures foreign key constraint `trips.service_id` → `calendar.id` is satisfied

## Testing
```bash
go test -v ./internal/restapi -run TestStopHandlerMultiAgencyScenario
```

✅ Test now passes

## Related
- Discovered while investigating CI failures in PR #350
- This is a pre-existing test infrastructure bug, unrelated to any feature changes

### DIFF
--- a/internal/restapi/stop_handler_test.go
+++ b/internal/restapi/stop_handler_test.go
@@ -203,6 +203,20 @@ func TestStopHandlerMultiAgencyScenario(t *testing.T) {
 	})
 	require.NoError(t, err)
 
+	_, err = queries.CreateCalendar(ctx, gtfsdb.CreateCalendarParams{
+		ID:        "service1",
+		Monday:    1,
+		Tuesday:   1,
+		Wednesday: 1,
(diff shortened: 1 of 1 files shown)


aaronbrethorst (approved):
nice work, thanks for tracking this down and fixing it.

Issue #365, opened 2026-02-09
Security: Missing input sanitization in problem report handlers


3rabiii:
### Description
The `report-problem-with-trip` and `report-problem-with-stop` handlers currently accept user inputs without adequate validation or sanitization. This exposes the system to potential security risks.

### Identified Issues
* **Unbounded Input:** `userComment` has no length limit, which could lead to log flooding or DoS via massive payloads.
* **Type Safety:** `userLat` and `userLon` accept non-numeric strings, potentially causing data corruption or application errors.
* **XSS Risk:** Storing raw user input poses a Cross-Site Scripting (XSS) risk if these fields are displayed in admin dashboards.

### Proposed Fix
- Truncate `userComment` to a maximum of 500 characters.
- Validate `userLat` and `userLon` as proper floats using `strconv.ParseFloat`; discard invalid values.
- Ensure all inputs are sanitized before logging.

PR #366, opened 2026-02-09
fix(security): sanitize user inputs in problem report handlers

3rabiii:
### Description
This PR addresses security vulnerabilities in the problem reporting endpoints by implementing strict input sanitization and validation.

### Changes
- **Input Truncation:** `userComment` is now strictly truncated to 500 characters to prevent log flooding and potential DoS payloads.
- **Numeric Validation:** `userLat` and `userLon` are explicitly parsed as floats using `strconv.ParseFloat`. Invalid non-numeric inputs are discarded (defaulting to empty strings) instead of being logged or processed raw.
- **Testing:** Added comprehensive unit tests in `report_problem_with_trip_handler_test.go` and `report_problem_with_stop_handler_test.go` to verify:
    - Graceful handling of invalid non-numeric coordinates.
    - Correct truncation of excessively long comments.
### Verifications
Screenshot From 2026-02-09 17-17-12

@aaronbrethorst 
fixes : #365

### DIFF
--- a/internal/restapi/report_problem_with_stop_handler.go
+++ b/internal/restapi/report_problem_with_stop_handler.go
@@ -43,11 +43,10 @@ func (api *RestAPI) reportProblemWithStopHandler(w http.ResponseWriter, r *http.
 	}
 
 	query := r.URL.Query()
-
 	code := query.Get("code")
-	userComment := query.Get("userComment")
-	userLat := query.Get("userLat")
-	userLon := query.Get("userLon")
+	userComment := utils.TruncateComment(query.Get("userComment"))
+	userLatStr := utils.ValidateNumericParam(query.Get("userLat"))
+	userLonStr := utils.ValidateNumericParam(query.Get("userLon"))
 	userLocationAccuracy := query.Get("userLocationAccuracy")
 
 	// TODO: Add storage logic for the problem report, I leave it as a log statement for now
@@ -56,8 +55,8 @@ func (api *RestAPI) reportProblemWithStopHandler(w http.ResponseWriter, r *http.
 		slog.String("stop_id", stopID),
 		slog.String("code", code),
 		slog.String("user_comment", userComment),
-		slog.String("user_lat", userLat),
-		slog.String("user_lon", userLon),
+		slog.String("user_lat", userLatStr),
+		slog.String("user_lon", userLonStr),
 		slog.String("user_location_accuracy", userLocationAccuracy))
 
 	api.sendResponse(w, r, models.NewOKResponse(struct{}{}, api.Clock))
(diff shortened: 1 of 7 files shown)


aaronbrethorst (changes requested):
Hey Adel, nice work hardening the problem report endpoints. The lat/lon numeric validation, the comment truncation, and the tests are all well-motivated — and the WaitGroup race fix in `test_helper_test.go` (moving `Add(1)` before the goroutine launch) is a sharp catch that fixes a real concurrent test bug. Two things to address before we can merge:

## Important

1. **Comment truncation breaks multi-byte UTF-8 characters** (`report_problem_with_stop_handler.go:50`, `report_problem_with_trip_handler.go:53`): `rawComment[:500]` slices at **byte** 500, not character 500. If a user writes a comment with emoji, CJK characters, or accented text, this can split a multi-byte UTF-8 sequence in half, producing an invalid string. Today this only affects log output, but the handler has a TODO to add storage — and truncated UTF-8 in a database can cause real problems.

   Fix by truncating on rune boundaries:

   ```go
   const maxCommentLength = 500

   userComment := rawComment
   if len([]rune(rawComment)) > maxCommentLength {
       userComment = string([]rune(rawComment)[:maxCommentLength])
   }
   ```

   This also pulls out the magic number `500` into a named constant, which makes both handlers self-documenting.

2. **Duplicated sanitization logic across both handlers** (`report_problem_with_stop_handler.go:47-65`, `report_problem_with_trip_handler.go:50-70`): The comment truncation + lat/lon validation block is copied identically in both handlers. The codebase already has a pattern of shared utilities in `internal/utils/api.go` (e.g., `ParseFloatParam`, `ParseMaxCount`). Please extract the repeated logic into helpers there:

   ```go
   // In internal/utils/api.go

   const MaxCommentLength = 500

   // TruncateComment safely truncates a comment to MaxCommentLength runes.
   func TruncateComment(s string) string {
       runes := []rune(s)
       if len(runes) > MaxCommentLength {
           return string(runes[:MaxCommentLength])
       }
       return s
   }

   // ValidateNumericParam returns the string if it's a valid float, empty string otherwise.
   func ValidateNumericParam(s string) string {
       if s == "" {
           return ""
       }
       if _, err := strconv.ParseFloat(s, 64); err != nil {
           return ""
       }
       return s
   }
   ```

   Then both handlers become:

   ```go
   userComment := utils.TruncateComment(query.Get("userComment"))
   userLatStr := utils.ValidateNumericParam(query.Get("userLat"))
   userLonStr := utils.ValidateNumericParam(query.Get("userLon"))
   ```

   This collapses ~18 lines per handler into 3, eliminates the duplication, and puts the validation logic where other parameter helpers already live. The debug logging for invalid lat/lon can be dropped — silently discarding bad input for a fire-and-forget report endpoint is the right behavior, and debug logs for expected invalid input are noise.

## Strengths

- Good test coverage: both handlers tested for invalid coordinates and oversized comments.
- The WaitGroup race fix in `test_helper_test.go` is a correct and valuable bonus fix.
- Single clean commit.

## Recommended Action

1. Fix UTF-8 truncation (item 1)
2. Extract shared helpers (item 2)
3. Add unit tests for the new helpers (table-driven, covering: empty string, under limit, at limit, over limit, multi-byte characters for truncation; empty, valid float, invalid non-numeric, valid negative for validation)

Issue #367, opened 2026-02-09
Security: Potential SQL Injection in TableCounts


3rabiii:
### Problem Description
This issue identified a High Severity vulnerability in `gtfsdb/debugging.go`.
The function `TableCounts` constructs SQL queries using `fmt.Sprintf` with table names. Although there is a map check, relying on string interpolation for SQL queries is considered insecure and fragile ("SQL Injection via table name interpolation").

### Proposed Solution
Replace the dynamic string formatting with a strict `switch-case` allowlist. This ensures that the SQL query strings are constants (hardcoded) and eliminates any risk of injection, adhering to secure coding best practices.

PR #368, opened 2026-02-09
fix(security): prevent potential SQL injection in TableCounts 

3rabiii:
### Description
Addresses the **High Severity** security vulnerability identified in `gtfsdb/debugging.go`.

The original implementation of `TableCounts` constructed SQL queries using string interpolation (`fmt.Sprintf`), which is flagged as a potential SQL Injection vector despite the map-based check.

**Changes:**
- **Security Fix:** Replaced dynamic string formatting with a strict `switch-case` allowlist. This ensures the executed SQL queries are constant strings, eliminating any risk of injection.
- **Refactoring:** Removed the redundant `allowedTables` map.
- **Testing:** Added `gtfsdb/debugging_test.go` with a new unit test `TestTableCounts` to verify:
    - Correct counting of whitelisted tables.
    - Proper ignoring of unknown/non-whitelisted tables.

### Verification
Screenshot From 2026-02-09 22-46-29

@aaronbrethorst 
fixes : #367

### DIFF
--- a/gtfsdb/debugging.go
+++ b/gtfsdb/debugging.go
@@ -63,8 +63,8 @@ func (c *Client) TableCounts() (map[string]int, error) {
 	defer logging.SafeCloseWithLogging(rows,
 		slog.Default().With(slog.String("component", "debugging")),
 		"database_rows")
-	var tables []string
 
+	var tables []string
 	for rows.Next() {
 		var tableName string
 		if err := rows.Scan(&tableName); err != nil {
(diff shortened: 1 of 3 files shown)


aaronbrethorst (approved):
Nice work on this security hardening, Adel. The prebuilt query map is a clean, idiomatic solution that eliminates the SQL injection vector entirely.

## Prior Feedback Status

All items from the previous review have been addressed:

| Feedback | Status |
|----------|--------|
| Fix errcheck lint violation on `db.Close()` | Addressed — `defer func() { _ = db.Close() }()` |
| Replace switch with map of prebuilt queries | Addressed — `tableCountQueries` map with string literals |
| Rebase onto main | Addressed |

## Summary

No issues found. The security fix correctly eliminates dynamic SQL construction by replacing it with a map of prebuilt query literals. The test coverage validates the allowlist behavior, and the WaitGroup race fix in the test helper is a nice bonus.

PR #372, opened 2026-02-10
perf: Optimize RoutesForAgencyID lookup to O(1) and fix concurrency issues

### LINKED ISSUE #371: Optimize `RoutesForAgencyID` lookup from O(N) to O(1)
### Description

The current implementation of `RoutesForAgencyID` in `internal/gtfs/gtfs_manager.go` performs a linear scan O(N) over the entire `manager.gtfsData.Routes` slice every time it is called.

For large regional feeds (e.g., 20+ agencies, 5,000+ routes), this creates significant CPU overhead, especially for endpoints like `/agencies-with-coverage` which calls this method in a loop for every agency.

### Impact

* **Current Complexity:** O(R) where R is the total number of routes.
* **Proposed Complexity:** O(1) constant time.
* **Performance:** Benchmarks show this change reduces lookup time from **~50µs** to **~0.1µs** (approx 500x speedup) for large datasets.

### Proposed Solution

1. Add a lookup map `routesByAgencyID map[string][]*gtfs.Route` to the `Manager` struct.
2. Pre-populate this map during initialization and `ForceUpdate` (hot-swap).
3. Update `RoutesForAgencyID` to return the slice directly from the map.

### Implementation Plan

* **`internal/gtfs/gtfs_manager.go`**: Add field to struct and update accessor method.
* **`internal/gtfs/static.go`**: Populate the map in `setStaticGTFS` and `ForceUpdate` to ensure thread safety and correct hot-swap behavior.

ARCoder181105:
## Description

This PR optimizes the `RoutesForAgencyID` method in the GTFS Manager. Previously, this method performed a linear O(N) scan over all routes in the system to filter by agency ID. This created a performance bottleneck for large regional feeds and high-traffic endpoints.

Additionally, this PR fixes a data race condition discovered in the `AdvancedDirectionCalculator` during testing.

closes #371

### Performance Optimization

* **`internal/gtfs/gtfs_manager.go`**:
* Added `routesByAgencyID` map to the `Manager` struct for O(1) lookups.
* Refactored `RoutesForAgencyID` to look up routes directly from the map instead of iterating.

* **`internal/gtfs/static.go`**:
* Implemented `buildRouteIndex` helper to pre-compute the route map.
* Updated `setStaticGTFS` (initialization) and `ForceUpdate` (hot-swap) to ensure the index is rebuilt whenever GTFS data changes.

### Bug Fixes

* **`internal/gtfs/advanced_direction_calculator.go`**:
* Added a `sync.RWMutex` to protect the internal `contextCache`.
* Fixed a data race where `SetContextCache` (write) and `computeFromShapes` (read) could occur simultaneously.

### Testing

* **`internal/gtfs/gtfs_manager_test.go`**:
* Added `TestRoutesForAgencyID_MapOptimization` to verify correct map initialization.
* Added `TestRoutesForAgencyID_ConcurrentAccess` to ensure thread safety.

## Checklist

* [x] I have added tests to cover my changes.
* [x] All new and existing tests passed.
* [x] Verified that data hot-swaps (`ForceUpdate`) maintain data integrity.

### DIFF
--- a/internal/gtfs/advanced_direction_calculator.go
+++ b/internal/gtfs/advanced_direction_calculator.go
@@ -8,6 +8,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"sync/atomic"
 
 	"maglev.onebusaway.org/gtfsdb"
@@ -26,6 +27,7 @@ type AdvancedDirectionCalculator struct {
 	contextCache      map[string][]gtfsdb.GetStopsWithShapeContextRow   // Cache of stop shape context data
 	shapeCache        map[string][]gtfsdb.GetShapePointsWithDistanceRow // Cache of all shape data for bulk operations
 	initialized       atomic.Bool                                       // Tracks whether concurrent operations have started
+	cacheMutex        sync.RWMutex                                      // Protects map access
 }
 
 // NewAdvancedDirectionCalculator creates a new advanced direction calculator
@@ -51,6 +53,9 @@ func (adc *AdvancedDirectionCalculator) SetVarianceThreshold(threshold float64) 
 // IMPORTANT: This must be called before any concurrent operations begin.
 // Panics if called after CalculateStopDirection has been invoked.
 func (adc *AdvancedDirectionCalculator) SetShapeCache(cache map[string][]gtfsdb.GetShapePointsWithDistanceRow) {
+	adc.cacheMutex.Lock()
+	defer adc.cacheMutex.Unlock()
+
 	if adc.initialized.Load() {
 		panic("SetShapeCache called after concurrent operations have started")
 	}
@@ -62,6 +67,9 @@ func (adc *AdvancedDirectionCalculator) SetShapeCache(cache map[string][]gtfsdb.
 // Panics if called after internal state has been initialized (i.e., after the first
 // fallback to shape-based calculation).
 func (adc *AdvancedDirectionCalculator) SetContextCache(cache map[string][]gtfsdb.GetStopsWithShapeContextRow) {
+	adc.cacheMutex.Lock()
+	defer adc.cacheMutex.Unlock()
+
 	if adc.initialized.Load() {
 		panic("SetContextCache called after concurrent operations have started")
 	}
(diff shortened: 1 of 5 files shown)


aaronbrethorst (changes requested):
Hey Aditya, the core optimization and the data race fix both look solid — the `buildRouteIndex` approach is clean, both `setStaticGTFS` and `ForceUpdate` rebuild the index correctly, and the concurrent access test is a nice addition. The squash and rebase is done too. One item from the first round still needs a quick pass, though.

## Previous Feedback Status

| # | Item | Status |
|---|------|--------|
| 1 | Squash into a single commit | Addressed |
| 2 | Changelog-style comments on mutex additions | **Not addressed** |

## Important

1. **Changelog-style comments are still present — and multiplied** (`advanced_direction_calculator.go`). The first round asked you to rephrase or remove `// Mutex protection added for race safety` because it describes *why you changed the code*, not *what the code does*. That exact comment is still on lines 56 and 72 in the diff. Two new instances of the same pattern were also added:

   - `// Mutex protection: Use RLock (Read Lock) for checking the cache` (in `computeFromShapes`)
   - `// Mutex protection: Use RLock for checking the cache` (in `calculateOrientationAtStop`)

   All four should either be removed (the `Lock()`/`RLock()` calls are self-documenting) or reworded to describe *what* the code does. For example:

   ```go
   // Protect cache from concurrent readers during assignment
   adc.cacheMutex.Lock()
   ```

   Or simply delete all four comments — `cacheMutex.Lock()` and `cacheMutex.RLock()` speak for themselves.

## Fit and Finish

2. **Self-explanatory comment in `RoutesForAgencyID`** (`gtfs_manager.go`): `// Return empty slice if no routes found for this agency` restates the `return []*gtfs.Route{}` line. Safe to remove.

3. **Comments in `buildRouteIndex`** (`static.go`): `// Pre-allocate based on agency count to avoid resizing` and `// Use pointer to avoid duplication` are accurate but slightly over-explain idiomatic Go patterns. Consider removing for brevity.

## Strengths

- The `buildRouteIndex` function follows the established `buildLookupMaps` pattern perfectly — same style, same location, easy to maintain.
- Both data paths (`setStaticGTFS` for initial load and `ForceUpdate` for hot-swap) correctly rebuild the index.
- The logging addition (`slog.Int("route_index_agencies", ...)`) is a nice operational touch.
- `TestRoutesForAgencyID_ConcurrentAccess` properly exercises reader/writer contention with `setStaticGTFS`.
- Adding `defer manager.Shutdown()` across existing tests is good housekeeping.
- The benchmark is a welcome addition for tracking regression.

## Recommended Action

1. Remove or reword the four changelog-style comments (item 1)
2. Consider the fit-and-finish items (items 2–3)
3. Squash again after changes

PR #388, opened 2026-02-11
fix: serverErrorResponse silently discards error parameter

### LINKED ISSUE #375: fix: serverErrorResponse silently discards error parameter
## Context

Identified during review of PR #374 (fix: move FTS5 queries out of sqlc to fix make models).

## Problem

`serverErrorResponse` in `internal/restapi/errors.go` (lines 34-54) accepts an `err` parameter but **never logs it**. It only sends a generic `"internal server error"` HTTP response to the client. Every caller in the codebase constructs a detailed `fmt.Errorf(...)` with context, and that error is silently discarded.

This means database connection failures, SQLite corruption, FTS5 index corruption, context cancellation — all produce HTTP 500 responses with zero server-side diagnostic logging.

## Recommendation

Add error logging to the function:

```go
func (api *RestAPI) serverErrorResponse(w http.ResponseWriter, r *http.Request, err error) {
    api.Logger.Error("server error", "error", err, "path", r.URL.Path)
    // ... rest of response sending
}
```

## Impact

Every endpoint that calls `serverErrorResponse` is affected. Operators debugging production issues will see 500 responses in request logs but have no record of what actually failed.

MohamedKamal000:
closes #375

### DIFF
--- a/internal/restapi/errors.go
+++ b/internal/restapi/errors.go
@@ -32,7 +32,7 @@ func (api *RestAPI) invalidAPIKeyResponse(w http.ResponseWriter, r *http.Request
 }
 
 func (api *RestAPI) serverErrorResponse(w http.ResponseWriter, r *http.Request, err error) {
-	api.Logger.Error("internal server error", "error Message:", err.Error(), "path:", r.URL.Path)
+	api.Logger.Error("internal server error", "error", err, "path", r.URL.Path)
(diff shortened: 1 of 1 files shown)


aaronbrethorst (approved):
Nice work addressing the feedback, Mohamed — the logging line now matches the codebase conventions perfectly, and dropping the WaitGroup change on rebase was the right call.

PR #403, opened 2026-02-13
feat: Implement readiness probe and startup sequencing

### LINKED ISSUE #402: feat: Implement readiness probe and startup sequencing to prevent cold traffic routing
**The Problem**
Currently, the application starts the HTTP listener and reports itself as "Healthy" (HTTP 200) immediately upon process startup. However, critical background initialization tasks specifically building the SQLite database, constructing the R*Tree spatial index, and fetching the initial GTFS-RT data take time to complete.

During this initialization window, the API accepts requests but may return errors, incomplete data, or empty responses because the underlying data structures are not yet populated.

**Proposed Solution**
We should implement a distinction between **Liveness** (is the process running?) and **Readiness** (is the application ready to serve traffic?).

1. **GTFS Manager State:** Introduce an atomic `isReady` flag in the `Manager` struct.
2. **Startup Sequencing:** Update `InitGTFSManager` to perform the Stop Spatial Index construction and the first Real-time Data fetch *before* marking the manager as ready.
3. **Health Endpoint Update:** Modify `healthHandler` to check the `isReady` status.
* If `!isReady`: Return `503 Service Unavailable`.
* If `isReady`: Return `200 OK`.

**Benefits**

* **Reliability:** Prevents the API from serving requests before it has data.
* **Production Readiness:** Allows load balancers (e.g., Kubernetes, AWS ALB) to wait until the instance is fully "warm" before routing traffic to it.
* **Consistency:** Ensures that if a user gets a `200 OK` from `/healthz`, the spatial index and real-time cache are guaranteed to be populated.

**Implementation Plan**

* [ ] Add `isReady atomic.Bool` to `gtfs.Manager`.
* [ ] Refactor `InitGTFSManager` to wait for initial data loads.
* [ ] Update `internal/restapi/health_handler.go` to return 503 during startup.

ARCoder181105:
**Summary**
This PR implements a proper readiness probe for the application to prevent traffic from being routed to "cold" instances during startup. Previously, the application reported itself as healthy immediately upon process start, even while critical background tasks (database build, spatial indexing, real-time fetch) were still running. This caused early requests to fail or return incomplete data.

### Key Changes

1. **GTFS Manager (`internal/gtfs/gtfs_manager.go`)**:
* Added an `isReady atomic.Bool` flag to the `Manager` struct for thread-safe state tracking.
* Updated `InitGTFSManager` to perform the **initial Real-time Data fetch synchronously** (with a timeout).
* Ensured `MarkReady()` is only called *after* the Database, Spatial Index, and Real-time cache are fully populated.

2. **Health Handler (`internal/restapi/health_handler.go`)**:
* Updated `healthHandler` to check `api.GtfsManager.IsReady()`.
* Returns `503 Service Unavailable` with `status: "starting"` during initialization.
* Returns `200 OK` with `status: "ok"` only when the manager is ready.

I have verified the behavior locally:

1. **Positive Case:** Started the server and waited for initialization to complete.
* `curl http://localhost:4000/api/health` -> Returns `200 OK` `{"status":"ok"}`.

2. **Log Verification:** Confirmed via logs that the "GTFS Realtime updated" message appears *before* the server marks itself as ready.

### Checklist:

* [x] Add `isReady atomic.Bool` to `gtfs.Manager`.
* [x] Refactor `InitGTFSManager` to wait for initial data loads.
* [x] Update `internal/restapi/health_handler.go` to return 503 during startup.

closes #402

### DIFF
--- a/internal/gtfs/gtfs_manager.go
+++ b/internal/gtfs/gtfs_manager.go
@@ -7,6 +7,7 @@ import (
 	"sort"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"maglev.onebusaway.org/gtfsdb"
@@ -53,6 +54,17 @@ type Manager struct {
 	blockLayoverIndices            map[string][]*BlockLayoverIndex
 	regionBounds                   *RegionBounds
 	isHealthy                      bool
+	isReady                        atomic.Bool // Tracks whether initial data loading is complete
+}
+
+// IsReady returns true if the GTFS data is fully initialized and indexed.
+func (manager *Manager) IsReady() bool {
+	return manager.isReady.Load()
+}
+
+// MarkReady sets the manager status to ready.
+func (manager *Manager) MarkReady() {
+	manager.isReady.Store(true)
 }
 
 // InitGTFSManager initializes the Manager with the GTFS data from the given source
@@ -89,15 +101,26 @@ func InitGTFSManager(config Config) (*Manager, error) {
 	}
 	manager.stopSpatialIndex = spatialIndex
 
+	// STARTUP SEQUENCING:
+	// If realtime is enabled, perform the first fetch synchronously to "warm" the cache
+	// before marking the manager as ready.
+	if config.realTimeDataEnabled() {
+		initCtx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+		defer cancel()
+		manager.updateGTFSRealtime(initCtx, config)
+	}
+
+	// Everything is now warm and ready for traffic
+	manager.MarkReady()
+	manager.MarkHealthy()
+
 	if !isLocalFile {
 		manager.wg.Add(1)
 		go manager.updateStaticGTFS()
 	}
 
+	// Start the periodic background updates only after the initial synchronous fetch is done
 	if config.realTimeDataEnabled() {
-		ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
-		defer cancel() // Ensure the context is canceled when done
-		manager.updateGTFSRealtime(ctx, config)
 		manager.wg.Add(1)
 		go manager.updateGTFSRealtimePeriodically(config)
 	}
--- a/internal/restapi/health_handler.go
+++ b/internal/restapi/health_handler.go
@@ -7,26 +7,39 @@ import (
 	"maglev.onebusaway.org/internal/logging"
 )
 
-// JSON response from the health endpoint.
+// HealthResponse represents the JSON response from the health endpoint.
 type HealthResponse struct {
 	Status string `json:"status"`
 	Detail string `json:"detail,omitempty"`
 }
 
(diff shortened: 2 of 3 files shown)


aaronbrethorst:
@ARCoder181105 I just noticed that there's a merge conflict that needs to be resolved before this can be committed. Please rebase and squash into a single commit.

aaronbrethorst (approved):
Hey Aditya, nice work seeing this through five review rounds -- the persistence has paid off. The readiness probe implementation is clean, all prior feedback is addressed, and CI is green across the board.

## Prior Feedback Status

All items from all four prior review rounds are resolved:

| # | Item | Status |
|---|------|--------|
| 1 | Don't expose internal error details on `/healthz` | **Fixed** -- static string, `fmt` import removed |
| 2 | Rebase and squash into single commit | **Fixed** -- single commit `9013be4` |
| 3 | Use `defer cancel()` for context | **Fixed** |
| 4 | Remove "Added:" prefix from field comment | **Fixed** -- now reads "Tracks whether initial data loading is complete" |
| 5 | Revert unrelated `shapes_handler_test.go` changes | **Fixed** -- file no longer in diff |
| 6 | `gofmt` trailing whitespace in `health_handler_test.go` | **Fixed** -- no trailing whitespace detected |

## New Items

No blocking issues. The code is correct, CI passes, and the PR is mergeable.

### Suggestions (non-blocking)

**1. Unit tests for `IsReady()`/`MarkReady()`** (`internal/gtfs/health_test.go`)

The existing `health_test.go` has thorough tests for `IsHealthy()`/`MarkHealthy()`/`MarkUnhealthy()` (state transitions, idempotency, concurrency). A parallel test for the new `IsReady()`/`MarkReady()` pair would be a nice addition for symmetry and to document the zero-value contract. This could be a follow-up PR.

**2. Transition test at the HTTP level** (`internal/restapi/health_handler_test.go`)

`TestHealthHandlerStarting` and `TestHealthHandlerReturnsOK` each test the health endpoint in a single state. A test that verifies 503 -> 200 transition on the *same* server after `MarkReady()` is called would exercise the exact scenario load balancers will see. Also a good follow-up PR candidate.

## Strengths

- **Clean single commit**: The rebase and squash is exactly right -- one logical change, one commit.
- **`atomic.Bool` for readiness**: Thread-safe without mutex overhead for a monotonic one-way flag. Good choice.
- **Synchronous first fetch with timeout**: The 15-second timeout on initial real-time fetch is a practical balance between warming the cache and not blocking startup indefinitely.
- **Clear three-phase health check**: Infrastructure -> readiness -> connectivity is easy to reason about.
- **Good test coverage**: `TestHealthHandlerStarting` directly validates the 503 behavior, and the updated `TestHealthHandlerReturnsOK` correctly calls `MarkReady()` before asserting 200.
- **All CI checks pass**: Lint, tests on Ubuntu and Windows, CLA signed.

PR #408, opened 2026-02-14
Verify trip status

Ahmedhossamdev:
Fix: #108 

In progress..

### DIFF
--- a/internal/restapi/trips_helper_test.go
+++ b/internal/restapi/trips_helper_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"maglev.onebusaway.org/gtfsdb"
+	internalgtfs "maglev.onebusaway.org/internal/gtfs"
 	"maglev.onebusaway.org/internal/models"
 	"maglev.onebusaway.org/internal/utils"
 )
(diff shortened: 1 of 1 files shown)


aaronbrethorst (changes requested):
Ahmed, this is a seriously ambitious piece of work -- tackling the core trip status engine is no small feat. You've fixed a real bug (the nanoseconds-vs-seconds schedule deviation issue), added stale vehicle detection, improved the block trip sequence query from N+1 to a single JOIN, and built position-projection logic. The scope is impressive, and the engineering thinking behind the new stop-finding strategies shows real depth of understanding of the GTFS-RT data model.

That said, there are several behavioral regressions and a critical test coverage gap that need to be addressed before this can merge. I've organized the feedback by severity.

---

## Critical Issues (3 found)

### 1. Missing Status/Phase when vehicle is nil -- behavioral regression
**Files**: `internal/restapi/trips_helper.go:30-37`, `internal/restapi/trip_details_handler.go:169`

In the new `BuildTripStatus`, when `vehicle == nil`, `BuildVehicleStatus` is never called. This means `status.Status` and `status.Phase` remain empty strings. Previously, the old code always called `BuildVehicleStatus(nil)`, which set them to `"SCHEDULED"` and `"scheduled"`.

Compounding this, `trip_details_handler.go` now gates inclusion of the status block on `status.VehicleID != ""`:
```go
if status != nil && status.VehicleID != "" {
    tripDetails.Status = status
}
```

This means for any trip without real-time vehicle data, the entire `status` block is now **omitted** from the API response. Previously it was always included. API consumers relying on the status block being present for scheduled trips will break.

**Suggested fix**: Either call `BuildVehicleStatus` unconditionally (as before), or explicitly set `Status`/`Phase` for nil vehicles. Also reconsider the `VehicleID != ""` gate.

### 2. OccupancyCapacity data dropped -- silent data loss
**File**: `internal/restapi/trips_helper.go:22-27`

The old code populated `OccupancyCapacity` from the vehicle's `OccupancyPercentage`:
```go
if vehicle != nil && vehicle.OccupancyPercentage != nil {
    status.OccupancyCapacity = int(*vehicle.OccupancyPercentage)
}
```

The new code removes this entirely and always sets `OccupancyCapacity: -1`. If any vehicles report `OccupancyPercentage`, this data is now silently discarded.

**Suggested fix**: Restore the `OccupancyPercentage` -> `OccupancyCapacity` mapping within the `if vehicle != nil` block.

### 3. Inconsistent tripID for GetStopDelaysFromTripUpdates
**File**: `internal/restapi/trips_helper.go:87`

`scheduleDeviation` and `stopTimes` are both fetched using `activeTripRawID` (which may differ from `tripID` if `BuildVehicleStatus` changed `status.ActiveTripID`). However, `GetStopDelaysFromTripUpdates` uses the original `tripID`:
```go
stopDelays := api.GetStopDelaysFromTripUpdates(tripID)
```

If the active trip differs from the requested trip (e.g., vehicle has moved to the next block trip), delays from trip A would be applied against stop times from trip B.

**Suggested fix**: Use `activeTripRawID` consistently:
```go
stopDelays := api.GetStopDelaysFromTripUpdates(activeTripRawID)
```

---

## Important Issues (5 found)

### 4. Near-zero test coverage for 808 lines of new code
**File**: `internal/restapi/trips_helper_test.go`

This PR adds ~20 new functions across `trip_updates_helper.go`, `trips_helper.go`, and `vehicles_helper.go`, including `StaleDetector`, `GetScheduleDeviation`, `GetStopDelaysFromTripUpdates`, `findClosestStopBySequence`, `findNextStopBySequence`, `findStopsByScheduleDeviation`, `fillStopsFromSchedule`, `projectPositionOntoRoute`, `interpolateDistanceAtScheduledTime`, and `scheduleRelationshipStatus`. The only test change is modifying one assertion (VehicleID format).

Existing integration tests pass because they primarily exercise code paths where no real-time vehicle data is present, creating a false sense of safety. The behavioral changes to `GetVehicleStatusAndPhase` and stale detection are completely untested.

**Priority test gaps** (ranked by risk):
1. `GetVehicleStatusAndPhase` -- nil vehicle, SCHEDULED/CANCELED/ADDED trips
2. `StaleDetector.Check` -- nil vehicle, nil timestamp, fresh/stale/threshold
3. `GetScheduleDeviation` / `GetStopDelaysFromTripUpdates` -- no trip updates, with delays
4. `BuildVehicleStatus` Predicted/Scheduled flag logic
5. `findClosestStopByTimeWithDelays` / `findNextStopByTimeWithDelays`

### 5. `GetVehicleStatusAndPhase` returns "default" instead of "SCHEDULED"
**File**: `internal/restapi/vehicles_helper.go:79-82`

The function now returns `("default", "")` when the vehicle is nil. The status `"default"` is an internal Java implementation detail (from `TripStatusBeanServiceImpl`), not a documented OBA API value. The documented values are `"SCHEDULED"`, `"CANCELED"`, `"ADDED"`, `"DUPLICATED"`. Additionally, `vehicles_for_agency_handler_test.go:452` still asserts against the old valid statuses -- these assertions pass only because the test path has no real-time vehicles loaded.

### 6. StaleDetector uses `time.Now()` instead of `api.Clock`
**File**: `internal/restapi/vehicles_helper.go:106`

`BuildVehicleStatus` calls `defaultStaleDetector.Check(vehicle, time.Now())` but the codebase has a `Clock` abstraction for testability. Also, vehicles with nil timestamps (valid in GTFS-RT -- `timestamp` is optional in `VehiclePosition`) are treated as stale, discarding all their data including potentially fresh position data.

### 7. SituationIDs computed twice
**File**: `internal/restapi/trip_details_handler.go:160`

`GetSituationIDsForTrip` is called once inside `BuildTripStatus` (which now initializes `status.SituationIDs`) and again separately in the handler. Each call performs multiple DB queries (GetTrip, GetRoute, GetAlertsByIDs).

**Suggested fix**: Reuse `status.SituationIDs` when available.

### 8. `GetScheduleDeviation` returns 0 for both "no data" and "on time"
**File**: `internal/restapi/trip_updates_helper.go:10-11`

When no trip updates exist, the function returns `0`. The caller checks `scheduleDeviation != 0` before setting `Predicted = true`. This means a trip that is genuinely exactly on time (0 seconds deviation) will never show `Predicted = true`, making it look like there's no real-time data.

**Suggested fix**: Return `(int, bool)` to distinguish "no data" from "zero deviation".

---

## Suggestions (5 found)

### 9. Extract `st.ArrivalTime / 1e9` to a helper (32 occurrences)
**File**: `internal/restapi/trips_helper.go`

The nanoseconds-to-seconds conversion `st.ArrivalTime / 1e9` appears 32 times. A `StopTimeSeconds(nanoseconds int64) int64` helper in `internal/utils/api.go` would make the code self-documenting and centralize the conversion factor.

### 10. Extract "arrival, fallback to departure" pattern (10 occurrences)
The 3-line pattern appears 10 times:
```go
stopTimeSeconds := st.ArrivalTime / 1e9
if stopTimeSeconds == 0 {
    stopTimeSeconds = st.DepartureTime / 1e9
}
```
An `EffectiveStopTimeSeconds(st *gtfsdb.StopTime) int64` helper would eliminate ~20 lines.

### 11. Unify `distanceToLineSegment` and `projectPointToSegment`
**Files**: `trips_helper.go:537`, `vehicles_helper.go:201`

These implement identical math with different return types. Could share a single implementation.

### 12. Unify `TripDetailsParams` / `TripForVehicleParams`
**Files**: `trip_details_handler.go:15-82`, `trip_for_vehicle_handler.go:17-84`

Structurally identical -- only `IncludeSchedule` default differs. A single struct with a configurable default would remove ~70 duplicate lines.

### 13. Extract shape row conversion helper (7 occurrences)
The `shapeRows` to `[]gtfs.ShapePoint` conversion loop appears 7 times across 5 files.

---

## Strengths

- **Fixed the nanoseconds-vs-seconds schedule deviation bug**: The old `calculateScheduleDeviationFromTripUpdates` did `int64(*delay)` on a `time.Duration`, giving nanoseconds instead of seconds. The new `GetScheduleDeviation` correctly uses `.Seconds()`.
- **Block trip sequence query optimization**: Replaced N+1 queries with a single `GetTripsByBlockIDOrdered` JOIN query.
- **Stale vehicle detection**: A good concept -- filtering out vehicles with stale data prevents showing outdated positions.
- **Route position projection**: `projectPositionOntoRoute` properly snaps vehicle GPS positions to route shapes.
- **Clean separation of concerns**: `trip_updates_helper.go` properly separates GTFS-RT processing from trip status building.
- **Well-structured stop-finding hierarchy**: The cascading approach (StopID -> StopSequence -> ScheduleDeviation -> Schedule fallback) is well thought out.

---

## Recommended Action

1. Fix Critical issues 1-3 (behavioral regressions)
2. Add unit tests for `GetVehicleStatusAndPhase`, `StaleDetector`, `GetScheduleDeviation`, `BuildVehicleStatus` flags
3. Update stale assertions in `vehicles_for_agency_handler_test.go`
4. Consider the suggestions for code simplification
5. Fix PR title typo ("Verifiy" -> "Verify")
6. Re-run review after fixes

aaronbrethorst (changes requested):
Ahmed, this is great progress -- you've systematically addressed the large majority of feedback from both prior review rounds. The `GetScheduleDeviation` return type fix, consistent time sourcing for stale detection, `BlockID.Valid` guard, `WithThreshold` copy-on-write, `NanosToSeconds`/`EffectiveStopTimeSeconds` helpers, bearing conversion tests, and unified `TripParams` are all cleanly implemented. The ~1,200 lines of new tests across three files show real rigor, and the mock infrastructure (`MockAddVehicleWithOptions`, `MockAddTripUpdate`, `MockResetRealTimeData`) is well-designed for future test authoring.

Before we can merge, I need you to fix two critical issues and address a few important items.

---

## Critical

### 1. SituationIDs in `trip_for_vehicle_handler` use raw alert IDs instead of combined IDs

**File:** `internal/restapi/trip_for_vehicle_handler.go:110-118`

The handler builds `situationIDs` by appending raw `alert.ID` without the agency prefix:

```go
situationIDs = append(situationIDs, alert.ID)
```

Compare with `GetSituationIDsForTrip` (trips_helper.go:630-631) which correctly uses:

```go
situationIDs = append(situationIDs, utils.FormCombinedID(agencyID, alert.ID))
```

This means `trip-for-vehicle` responses will have unformatted IDs like `"alert123"` while `trip-details` responses will have `"25_alert123"`. API consumers cross-referencing situations between endpoints will find no matches.

Additionally, this handler doesn't reuse `status.SituationIDs` (which `BuildTripStatus` already populates at line 24 of trips_helper.go). The `trip_details_handler` was updated to do this (lines 158-162), but `trip_for_vehicle_handler` was not. This causes duplicate DB queries AND the formatting inconsistency.

**Suggested fix:** Reuse `status.SituationIDs` like `trip_details_handler` does:

```go
var situationIDs []string
if status != nil && len(status.SituationIDs) > 0 {
    situationIDs = status.SituationIDs
} else {
    situationIDs = api.GetSituationIDsForTrip(r.Context(), tripID)
}
```

### 2. `BuildTripSchedule` in `trip_for_vehicle_handler` uses `time.Local` instead of agency timezone

**File:** `internal/restapi/trip_for_vehicle_handler.go:101`

The handler passes `time.Local` (the server's local timezone) to `BuildTripSchedule`:

```go
schedule, scheduleErr = api.BuildTripSchedule(ctx, agencyID, serviceDate, &trip, time.Local)
```

Compare with `trip_details_handler.go:150` which correctly passes the agency's timezone:

```go
schedule, err = api.BuildTripSchedule(ctx, agencyID, serviceDate, &trip, loc)
```

On any cloud deployment where the server timezone differs from the agency timezone (which is almost always the case — servers default to UTC), all schedule times in `trip-for-vehicle` responses will be offset by the timezone difference. The `loc` variable is already computed on line 57 of the same handler.

**Fix:** Change `time.Local` to `loc` on line 101.

---

## Important

### 3. `BuildTripStatus` silently swallows `GetStopTimesForTrip` and `GetShapePointsByTripID` errors

**File:** `internal/restapi/trips_helper.go:61-62, 114-115`

Both database queries use the pattern `if err == nil && len(results) > 0`, which treats connection failures, context cancellation, and SQL errors identically to "no data." When `GetStopTimesForTrip` fails, the entire stop-finding logic is skipped, resulting in empty `ClosestStop`/`NextStop` with no indication to the consumer. When `GetShapePointsByTripID` fails, all distance fields are silently zero.

You already added good `slog.Warn` logging for `fillStopsFromSchedule` errors (line 243). Apply the same pattern here:

```go
stopTimes, err := api.GtfsManager.GtfsDB.Queries.GetStopTimesForTrip(ctx, activeTripRawID)
if err != nil {
    slog.Warn("BuildTripStatus: failed to get stop times",
        slog.String("trip_id", activeTripRawID),
        slog.String("error", err.Error()))
}
if err == nil && len(stopTimes) > 0 {
    // ... existing logic
}
```

Same for `shapeErr`.

### 4. `calculateBlockTripSequence` silently swallows three DB errors with no logging

**File:** `internal/restapi/trips_helper.go:453-483`

Three database calls (`GetTrip`, `GetActiveServiceIDsForDate`, `GetTripsByBlockIDOrdered`) all return `0` on error without any logging. Since `0` is also a valid result (first trip in block), there's no way to distinguish failure from success. Add `slog.Warn` for each error path.

### 5. Test coverage: `calculateOffsetForStop` and `findNextStopAfter` have no unit tests

**File:** `internal/restapi/trips_helper.go:640-686`

These two functions handle the most common real-world GTFS-RT scenario (vehicle reporting its `StopID`). The `BuildTripStatus` integration tests only exercise the fallback paths (no `StopID`, only `Position`). Add unit tests for:
- `calculateOffsetForStop` with: matching stop, non-matching stop, deviation applied
- `findNextStopAfter` with: mid-trip stop, last stop (empty result), deviation applied

### 6. Test coverage: `BuildTripStatus` with vehicle `StopID` set

**File:** `internal/restapi/trips_helper.go:72-81`

`TestBuildTripStatus_VehicleWithPosition_FindsStops` creates a vehicle with `Position` but no `StopID`, so it only exercises the `findStopsByScheduleDeviation` fallback. Add a test using `MockAddVehicleWithOptions` with `StopID` set to a valid stop from the trip's stop times.

---

## Fit and Finish

### 7. `findClosestStopByTimeWithDelays` / `findNextStopByTimeWithDelays` prefer departure over arrival, opposite of `EffectiveStopTimeSeconds`

**File:** `internal/restapi/trips_helper.go:283-291, 323-331`

Both functions manually check `DepartureTime > 0` first, then `ArrivalTime`. But `EffectiveStopTimeSeconds` (used in all other stop-finding functions) prefers arrival first. If this is intentional for the "with delays" functions (delays typically apply to departures), add a brief comment explaining the inverted preference. If not, use `EffectiveStopTimeSeconds` for consistency.

### 8. PR title typo

"Verifiy" should be "Verify" -- fix before merge.

---

## Prior Review Items -- All Addressed

For the record, here's the status of every item from both prior reviews:

| # | Issue | Status |
|---|---|---|
| R1-1 | Missing Status/Phase when vehicle is nil | Fixed -- `BuildVehicleStatus` called unconditionally |
| R1-2 | OccupancyCapacity dropped | Fixed -- intentionally -1 with documented Java parity rationale |
| R1-3 | Inconsistent tripID for stop delays | Fixed -- uses `activeTripRawID` |
| R1-4 | Near-zero test coverage | Fixed -- ~1,200 lines of new tests |
| R1-5 | "default" status for nil vehicle | Addressed -- documented Java parity |
| R1-6 | StaleDetector uses time.Now() | Fixed -- accepts `currentTime` parameter |
| R1-7 | SituationIDs computed twice | Fixed in trip_details_handler (not in trip_for_vehicle -- see #1 above) |
| R1-8 | GetScheduleDeviation ambiguous return | Fixed -- returns `(int, bool)` |
| R2-1 | GetScheduleDeviation (0, true) with no delay data | Fixed -- returns (0, false) |
| R2-2 | Inconsistent time source | Fixed -- consistent `currentTime` |
| R2-3 | Missing BlockID.Valid check | Fixed |
| R2-4 | Error handling in handler | Fixed -- `status = nil` on error |
| R2-5 | Loop route StopID matching | Pre-existing pattern, not a regression |
| R2-6 | TripID formatting inconsistency | Fixed |
| R2-7 | WithThreshold mutates receiver | Fixed -- returns new instance |
| R2-8 | Integration test coverage gap | Substantially fixed |
| R2-9 | Bearing conversion untested | Fixed -- thorough test table |
| R2-10 | fillStopsFromSchedule swallows errors | Fixed -- logs slog.Warn |
| R2-11 | Stale CLAUDE.md docs | Fixed |
| R2-12 | Unused vehicle parameter | Removed |
| R2-13 | RUnlock comment accuracy | Fixed |

---

## Strengths

- **Thorough response to feedback:** Nearly every item from two detailed reviews was addressed, including the nuanced ones like Java parity documentation and copy-on-write semantics.
- **Well-designed mock infrastructure:** `MockAddVehicleWithOptions`, `MockAddTripUpdate`, and `MockResetRealTimeData` make it easy to write expressive tests for complex real-time scenarios.
- **Good helper extraction:** `NanosToSeconds`, `EffectiveStopTimeSeconds`, `shapeRowsToPoints`, `ServiceDateMillis`, and `CalculateSecondsSinceServiceDate` all reduce duplication and improve readability.
- **Unified TripParams:** Clean consolidation of two nearly-identical param structs with a configurable default.
- **Shared geometry implementation:** `projectOntoSegment` properly unifies `distanceToLineSegment` and `projectPointToSegment`.

---

## Recommended Action

1. Fix Critical #1 (SituationIDs formatting in trip_for_vehicle_handler)
2. Fix Critical #2 (`time.Local` → `loc` in trip_for_vehicle_handler)
3. Add warning logs for Issues #3 and #4 (silent DB error swallowing)
4. Add unit tests for Issues #5 and #6 (calculateOffsetForStop, findNextStopAfter, StopID branch)
5. Fix PR title typo

Issue #409, opened 2026-02-14
Inconsistent ETA & Distance Calculation (Java Parity)


3rabiii:
### Bug Description
The current Go implementation drifts from the original Java Core logic in two areas:
* **Distance:** It snaps to the nearest vertex instead of using geometric projection.
* **ETA:** It misses delay propagation for stops without explicit updates.

### Expected Behavior
- Distance calculation should use **line segment projection** for accuracy.
- ETA prediction should **propagate delays** from the closest prior stop if no update exists for the target stop.

### Proposed Solution
Update `getDistanceAlongShape` and `getPredictedTimes` to match OBA Java logic.

PR #410, opened 2026-02-14
fix(api): implement shape projection and ETA delay propagation

3rabiii:
### Summary
This PR addresses inconsistencies regarding distance calculation and ETA prediction compared to the original OneBusAway Java implementation.

### Key Changes

**1. Distance Calculation (Geometry Fix):**
* **Before:** The logic naively "snapped" the vehicle to the nearest shape vertex (point), causing distance jumps and errors when the bus was between points.
* **After:** Implemented **Line Segment Projection** in `shape_distance_helpers.go`. The code now projects the vehicle's GPS coordinates onto the shape segment to calculate the precise `distanceAlongShape`.

**2. ETA Prediction (Delay Propagation):**
* **Before:** `getPredictedTimes` only applied delays if an explicit update existed for the exact target stop.
* **After:** Implemented **Delay Propagation** in `arrival_and_departure_for_stop_handler.go`. If the target stop has no update, the system now looks for the closest prior stop with real-time data and propagates its delay.

### Testing
* **Refactored Test Suite:** Separated math/geometry tests from API logic tests for cleaner architecture.
* **New Tests:**
    * `TestGetDistanceAlongShape_Projection`: Verifies geometric projection accuracy.
    * `TestGetDistanceAlongShape_LoopingRoute`: Verifies correct handling of route loops.
    * `TestGetPredictedTimes_DelayPropagationLogic`: Verifies ETA delay propagation.
* **Verification:** 
image

@aaronbrethorst 
fixes : #409

### DIFF
--- a/internal/models/constants.go
+++ b/internal/models/constants.go
@@ -28,4 +28,5 @@ const (
 	MaxAllowedCount          = 250
 )
 
+// RangeSearchBufferMeters provides a 50m tolerance for GPS inaccuracy and curve approximation.
 const RangeSearchBufferMeters = 50.0
--- a/internal/restapi/arrival_and_departure_for_stop_handler_test.go
+++ b/internal/restapi/arrival_and_departure_for_stop_handler_test.go
@@ -11,7 +11,6 @@ import (
 	"time"
 
 	"github.com/OneBusAway/go-gtfs"
-	go_gtfs "github.com/OneBusAway/go-gtfs"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"maglev.onebusaway.org/internal/models"
@@ -669,6 +668,7 @@ func TestArrivalsAndDeparturesForStopHandlerInvalidTime(t *testing.T) {
 
 func TestGetPredictedTimes_DelayPropagationLogic(t *testing.T) {
 	api := createTestApi(t)
+	defer api.Shutdown()
 
 	tripID := "test_trip"
 	targetStopSequence := int64(5)
@@ -677,19 +677,19 @@ func TestGetPredictedTimes_DelayPropagationLogic(t *testing.T) {
 
 	uint32Ptr := func(v uint32) *uint32 { return &v }
 
-	mockTrip := go_gtfs.Trip{
-		ID: go_gtfs.TripID{ID: tripID},
-		StopTimeUpdates: []go_gtfs.StopTimeUpdate{
+	mockTrip := gtfs.Trip{
+		ID: gtfs.TripID{ID: tripID},
+		StopTimeUpdates: []gtfs.StopTimeUpdate{
 			{
 				StopSequence: uint32Ptr(1),
-				Departure: &go_gtfs.StopTimeEvent{
+				Departure: &gtfs.StopTimeEvent{
 					Delay: &delayDuration,
 				},
 			},
 		},
 	}
 
-	api.GtfsManager.SetRealTimeTripsForTest([]go_gtfs.Trip{mockTrip})
+	api.GtfsManager.SetRealTimeTripsForTest([]gtfs.Trip{mockTrip})
 
 	scheduledTime := time.Now()
 	predArrival, predDeparture := api.getPredictedTimes(tripID, "test_stop", targetStopSequence, scheduledTime, scheduledTime)


aaronbrethorst (changes requested):
Adel, the line segment projection upgrade for distance calculation is a real improvement — projecting onto segments instead of snapping to vertices eliminates the "jump" artifacts you described. The shape geometry tests are well-designed too. The delay propagation concept is correct, but the test and some of the logic need work.

## Review Items

### Critical

**1. `TestGetPredictedTimes_DelayPropagationLogic` doesn't test delay propagation** (`arrival_and_departure_for_stop_handler_test.go:667-679`)

This test calls `getPredictedTimes("test_trip", "test_stop", 5, ...)` but `"test_trip"` doesn't exist in the real-time data. The function hits the early return on line 513:

```go
realTimeTrip, _ := api.GtfsManager.GetTripUpdateByID(tripID)
if realTimeTrip == nil || len(realTimeTrip.StopTimeUpdates) == 0 {
    return 0, 0  // <-- test exits here, never reaching propagation code
}
```

The test verifies `(0, 0)` — the exact same behavior as the existing `TestGetPredictedTimes_NoRealTimeData`. The delay propagation code (lines 517-595) has **zero test coverage**.

A meaningful test needs mock real-time data with a trip that has stop time updates. For example:
- A trip with a stop time update at sequence 3 with a 120-second departure delay
- A query for sequence 5 (no update for this stop)
- Assert that the 120-second delay propagates to the predicted times at sequence 5

Use `createTestApiWithRealTimeData` or inject real-time trip updates to exercise the actual propagation logic.

### Important

**2. Rule 1 logic has vestigial nil checks that obscure intent** (`arrival_and_departure_for_stop_handler.go:572-588`)

After lines 565-570, both `arrivalOffset` and `departureOffset` are guaranteed non-nil (they fall back to `&propagatedDelay`). But Rule 1 still checks for `nil`:

```go
// Current (confusing):
if arrivalOffset != nil {          // always true after line 565
    offset = *arrivalOffset
}
if arrivalOffset == nil || (departureOffset != nil && *departureOffset != propagatedDelay) {
    // arrivalOffset == nil: always false
    // departureOffset != nil: always true
    if departureOffset != nil {    // always true
        offset = *departureOffset
    }
}
```

This entire block simplifies to:

```go
// Suggested (clearer):
offset := *arrivalOffset
if *departureOffset != propagatedDelay {
    // Prefer departure offset if it came from an actual GTFS-RT update
    // (not just the propagated fallback)
    offset = *departureOffset
}
```

The logic is sound — prefer a real departure delay over a propagated one — but the nil checks make it hard to reason about.

**3. Magic number 50.0 for range tolerance** (`trips_helper.go:449`)

The range filter tolerance was changed from 10 to 50 meters without explanation:

```go
if segmentEnd < minDistTraveled-50.0 || segmentStart > maxDistTraveled+50.0 {
    continue
}
```

This should be a named constant with a comment explaining why 50 meters was chosen. Is this GPS accuracy? Buffer for curve approximation?

### Fit and Finish

**4. Duplicated post-loop distance interpolation** (`trips_helper.go:418-425` and `473-480`)

Both `getDistanceAlongShape` and `getDistanceAlongShapeInRange` end with identical 7-line blocks:

```go
cumulativeDistance := cumulativeDistances[closestSegmentIndex]
if closestSegmentIndex < len(shape)-1 {
    segmentDistance := utils.Distance(...)
    cumulativeDistance += segmentDistance * projectionRatio
}
return cumulativeDistance
```

Consider extracting this into a small helper like `interpolateDistance(cumulativeDistances, shape, segmentIndex, ratio)`.

**5. Unrelated formatting changes** (`fts_queries_test.go`, `shapes_handler_test.go`, `stop_handler_test.go`)

Same `go fmt` whitespace changes that have appeared in several recent PRs. Harmless but add noise — please keep unrelated formatting in a separate commit.

## Strengths

- **Line segment projection is the right approach**: Projecting onto segments instead of snapping to vertices eliminates distance jumps when the vehicle is between shape points. The `distanceToLineSegment` with clamped `t` parameter is clean and correct.
- **Shape projection tests are well-designed**: `TestGetDistanceAlongShape_Projection` verifies the core improvement (projection vs. vertex snapping), and `TestGetDistanceAlongShape_LoopingRoute` tests the range-based disambiguation.
- **Delay propagation concept is correct**: Looking for the closest prior stop's delay and applying it forward matches the Java implementation's behavior.
- **GTFS-RT `Delay` field support**: Handling both `Time` and `Delay` fields in stop time updates broadens compatibility with different feed providers.
- **Smart fallback in range search**: Falling back to full shape search when no segments are found in range handles GPS drift gracefully.

## Verdict: Request Changes

The delay propagation test (#1) must be fixed before merge — it provides false confidence by appearing to test propagation while only testing the "no data" early return. Items #2 and #3 are important for code clarity but can be addressed in a follow-up. Also, there's a code conflict. Please rebase against `main`.

3rabiii:
@aaronbrethorst 
Ready for review!

aaronbrethorst (changes requested):
Adel, really thorough work on this round of revisions. All five items from the first review are addressed — the delay propagation test now injects real mock data via `SetRealTimeTripsForTest`, Rule 1 is simplified, the magic number has a name, the distance interpolation is properly extracted, and the unrelated formatting is gone. The shape projection upgrade continues to be a genuine improvement.

Two small items need fixing before we can merge:

## Critical

1. **Lint failure: duplicate import** (`arrival_and_departure_for_stop_handler_test.go:13-14`)

   The test file imports `go-gtfs` twice — once as `gtfs` and once as `go_gtfs`:

   ```go
   "github.com/OneBusAway/go-gtfs"
   go_gtfs "github.com/OneBusAway/go-gtfs"
   ```

   This triggers staticcheck ST1019 and causes `make lint` to fail. The fix is to remove the `go_gtfs` alias and use the existing `gtfs` alias for the mock types in `TestGetPredictedTimes_DelayPropagationLogic`:

   ```go
   mockTrip := gtfs.Trip{
       ID: gtfs.TripID{ID: tripID},
       StopTimeUpdates: []gtfs.StopTimeUpdate{
           // ...
       },
   }
   api.GtfsManager.SetRealTimeTripsForTest([]gtfs.Trip{mockTrip})
   ```

2. **Missing `defer api.Shutdown()`** (`arrival_and_departure_for_stop_handler_test.go:671`)

   Every other test in this file calls `defer api.Shutdown()` after `createTestApi(t)`. The new delay propagation test omits it, which can leak background goroutines and database connections. Add it after line 671:

   ```go
   func TestGetPredictedTimes_DelayPropagationLogic(t *testing.T) {
       api := createTestApi(t)
       defer api.Shutdown()
       // ...
   ```

## Suggestion

3. **Add a comment on `RangeSearchBufferMeters`** (`constants.go:31`)

   The constant is defined but has no comment explaining why 50 meters was chosen. A brief note about the rationale (GPS accuracy buffer, curve approximation tolerance, etc.) would help future readers understand the choice.

PR #421, opened 2026-02-18
perf: Cache GetRegionBounds() calculation for O(N) to O(1) performance

### LINKED ISSUE #420: Cache Region Bounds to Eliminate Repeated O(N) Computation in `GetRegionBounds()`
## Description

`GetRegionBounds()` currently recalculates the region bounding box on every invocation by iterating through all GTFS shape points. Because this data is deterministic and only changes when GTFS data reloads, the repeated traversal is unnecessary and introduces avoidable overhead.

## Problem
- Performs an O(N) iteration across all shapes and points per call.
- No caching despite bounds only changing on data reload.
- Likely used by coverage and region boundary queries.

## Proposed Fix
- Compute region bounds once during GTFS initialization.
- Store the result in `manager.regionBounds`.
- Return cached values for subsequent calls.
- Invalidate and recompute the cache when GTFS data reloads.

## Scope
**Files affected:**
- `internal/gtfs/gtfs_manager.go` — add `RegionBounds` struct and cache field.
- `internal/gtfs/shapes.go` — extract `ComputeRegionBounds()` and update `GetRegionBounds()`.
- `internal/gtfs/static.go` — populate cache during initialization and reload.

### Expected Impact
- Reduce lookup complexity from **O(N) → O(1)**.

AhmedAlian7:
## Summary
Add caching to `GetRegionBounds()` to prevent repeated traversal of GTFS shape points and enable constant-time region bounds lookup.

**Closes #420**

## Description
Previously, `GetRegionBounds()` performed an O(N) iteration across all shape points on every call. This created unnecessary overhead, especially for endpoints that invoke the function frequently.

This change caches the computed bounds, ensuring O(1) lookups while maintaining correctness when GTFS data is reloaded.

## Changes

**Files modified (4):**

  - Introduced `RegionBounds` struct.  
  - Added `regionBounds` field to `Manager` for cached storage.

- `internal/gtfs/shapes.go`  
  - Extracted calculation logic into `ComputeRegionBounds()`.  
  - Updated `GetRegionBounds()` to return cached values.

- `internal/gtfs/static.go`  
  - Populate cache during initial load via `setStaticGTFS()`.  
  - Refresh cache during reloads in `ForceUpdate()`.

- `internal/gtfs/shapes_test.go`  
  - Refactored `TestGetRegionBounds` to call `ComputeRegionBounds()` directly instead of using `InitGTFSManager`.  
  - Removes unnecessary GTFS fixture loading and validates correctness with cached bounds.

### Performance Impact
- **Before:** O(N) iteration through all shape points per call.  
- **After:** O(1) cached lookup.

## Design Notes
- Cache is initialized with GTFS data to avoid lazy computation.
- Cache automatically refreshes on data reload.
- Reuses existing `staticMutex`; no new synchronization required.
- Function signature unchanged for backward compatibility.
- Returns zero-value bounds when no shapes are present.

## Testing
- `make lint` — no issues reported.  
- `go build` — successful compilation.  
- REST API tests for the `agencies-with-coverage` endpoint pass, exercising `GetRegionBounds()`.  
- `TestGetRegionBounds` — all 5 cases pass (single shape, multiple shapes, no shapes, empty points, real-world coordinates).

### DIFF
--- a/internal/gtfs/shapes_test.go
+++ b/internal/gtfs/shapes_test.go
@@ -1,10 +1,10 @@
 package gtfs
 
 import (
+	"testing"
+
 	"github.com/OneBusAway/go-gtfs"
 	"github.com/stretchr/testify/assert"
-	"maglev.onebusaway.org/internal/models"
-	"testing"
 )
 
 func TestGetRegionBounds(t *testing.T) {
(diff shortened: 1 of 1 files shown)


aaronbrethorst (approved):
Hey Ahmed, this is a clean and well-designed performance improvement. I like the approach of eager computation during data load rather than lazy caching — it keeps the locking simple and avoids cold-start overhead. The `ComputeRegionBounds` extraction as a pure function is a nice touch that makes the logic independently testable, and the test refactoring to drop the heavyweight `InitGTFSManager` dependency is a welcome simplification.

There's nothing left to change — this PR is ready to merge. The cache is correctly populated in both `setStaticGTFS` and `ForceUpdate`, the existing `staticMutex` protection carries over cleanly, and the function signature stays backward-compatible for all callers.

PR #427, opened 2026-02-19
perf: Resolve O(N²) bottleneck in active trip SQL queries

### LINKED ISSUE #426: Performance: O(N²) bottleneck in active trip SQL queries
**The Problem**
In `gtfsdb/query.sql`, the queries `GetActiveTripForRouteAtTime` and `GetActiveTripInBlockAtTime` use correlated subqueries inside their `JOIN` conditions to find the first and last stop sequences.

Because the subquery references the outer query, SQLite is forced to execute a `MIN()`/`MAX()` scan on the massive `stop_times` table for **every single candidate trip**. This creates an  performance bottleneck that severely degrades API response times at scale.

**The Solution**
We can eliminate the  loop entirely by rewriting these queries into a single-pass operation:

1. `JOIN` the `stop_times` table directly to the candidate trips.
2. Use `GROUP BY t.id` to aggregate the stop times.
3. Use a `HAVING` clause to filter the `MIN(departure_time)` and `MAX(arrival_time)` bounds.

This allows SQLite to execute the aggregations in one highly optimized pass. Additionally, using `GROUP BY` and `HAVING` (instead of CTEs or subqueries) cleanly bypasses `sqlc`'s known SQLite parser bugs, ensuring the Go models generate perfectly.

ARCoder181105:
### The Problem

While reviewing the database access patterns, I identified a severe  performance bottleneck in `gtfsdb/query.sql`, specifically within the `GetActiveTripForRouteAtTime` and `GetActiveTripInBlockAtTime` queries.

Previously, these queries used correlated subqueries within their `JOIN` conditions:

```sql
JOIN stop_times st_first ON t.id = st_first.trip_id AND st_first.stop_sequence = (
    SELECT MIN(stop_sequence) FROM stop_times WHERE trip_id = t.id
)

```

Because the subquery referenced `t.id` from the outer query, the SQLite engine was forced to execute a `MIN()`/`MAX()` table scan on the `stop_times` table for **every single candidate trip** evaluated during the join. Since `stop_times` is typically the largest table in a GTFS dataset (often millions of rows), this degrades real-time API response times exponentially at scale.

### The Solution

I rewrote both queries to execute as a single-pass operation by eliminating the correlated subqueries entirely.

The new queries:

1. `JOIN` the `stop_times` table directly to the pre-filtered candidate trips.
2. Use `GROUP BY t.id` to aggregate the stop times.
3. Use a `HAVING` clause to filter by the `MIN(departure_time)` and `MAX(arrival_time)` bounds.

*Note on implementation:* Typically, a Common Table Expression (CTE) would be used here. However, `sqlc` has known bugs parsing SQLite CTEs with complex joins. Structuring this as a flat `GROUP BY` with a `HAVING` clause completely bypasses the `sqlc` parser issues while allowing SQLite's query optimizer to process the bounds check in one highly optimized pass.

### Impact

* Eliminates the  loop during candidate trip evaluation.
* Significantly reduces CPU utilization and database latency for real-time endpoints.
* Maintains perfect type mapping for the generated Go models.

closes #426

### DIFF
--- a/gtfsdb/query.sql
+++ b/gtfsdb/query.sql
@@ -927,7 +927,7 @@ SELECT t.id
 FROM trips t
 JOIN stop_times st ON t.id = st.trip_id
 WHERE t.block_id = sqlc.arg('block_id')
-  AND t.service_id IN (sqlc.slice(' service_ids'))
+  AND t.service_id IN (sqlc.slice('service_ids'))
 GROUP BY t.id
 HAVING MIN(st.departure_time) <= sqlc.arg('current_time')
    AND MAX(st.arrival_time) >= sqlc.arg('current_time')


Ahmedhossamdev:
Hey @ARCoder181105 

Does this refactor return the same results?

ARCoder181105:
Hi, @Ahmedhossamdev 
Yes, for any valid GTFS feed the results are identical. Both approaches find trips where the earliest departure is before
`current_time` and the latest arrival is after `from_time`. The refactor assumes stop times are ordered chronologically, which the GTFS spec requires. The only case where results could differ is malformed data with out-of-order times, which existing queries also wouldn't handle correctly anyway.

ARCoder181105:
There is also a small bug I noticed now only I will be making a minor fix commit for that...

aaronbrethorst (approved):
Aditya, this is a well-reasoned query optimization. The `GROUP BY`/`HAVING` approach is simpler, avoids the double subquery per candidate trip, and — importantly — already has precedent in the codebase (`GetActiveTripsForRouteInTimeWindow` at line ~880 of `query.sql` uses the exact same pattern). Making these two queries consistent with that existing approach is a clear win.

## Fit and Finish

1. **Clarify the O(N²) claim in the PR description.** The `stop_times` table has an index on `trip_id` (`idx_stop_times_trip_id`), so the correlated subqueries weren't doing full table scans — they were doing indexed lookups. The real improvement is eliminating two separate index traversals + row lookups per candidate trip in favor of a single aggregation pass. It's still a meaningful optimization (fewer I/O operations, simpler query plan), but framing it as "O(N²) → O(1)" overstates it. Not a code change, just a nit for accuracy.

## Strengths

- The transformation is semantically correct for GTFS-compliant data. `MIN(departure_time)` matches the departure_time at `MIN(stop_sequence)` when times are monotonically increasing (as the GTFS spec requires).
- Consistent with the existing `GROUP BY`/`HAVING` pattern already in the codebase.
- The `SELECT t.* ... GROUP BY t.id` is correct in SQLite since `t.id` is the primary key and all other columns are functionally dependent.
- The parameter syntax standardization (`:block_id` → `sqlc.arg('block_id')`) is a nice cleanup.
- Generated Go code (`query.sql.go`) is consistent with the SQL changes, and function signatures are unchanged — no caller impact.

----

In the future, I'd prefer it if you would squash your commits, but I'm happy to do that through the GitHub UI here.

PR #429, opened 2026-02-19
perf(utils): optimize distance calculations with Equirectangular fast-path

### LINKED ISSUE #428: Performance: Optimize distance calculations with Equirectangular fast-path
**The Problem**
The current `Distance` function in `internal/utils/geometry.go` relies exclusively on the Haversine formula. While accurate, it is computationally expensive due to multiple heavy trigonometric operations (`math.Sin`, `math.Cos`, `math.Atan2`, and `math.Pow`).

In high-load scenarios—such as sorting hundreds of vehicles or stops by proximity to a user—these calculations are executed thousands of times per second, creating a significant CPU bottleneck.

**Current Bottleneck:**

```go
// Current implementation in internal/utils/geometry.go
func Distance(lat1, lon1, lat2, lon2 float64) float64 {
    // ... radians conversion ...
    y := math.Sqrt(math.Pow(math.Cos(lat2Rad)*math.Sin(deltaLon), 2) +
        math.Pow(math.Cos(lat1Rad)*math.Sin(lat2Rad)-math.Sin(lat1Rad)*math.Cos(lat2Rad)*math.Cos(deltaLon), 2))
    x := math.Sin(lat1Rad)*math.Sin(lat2Rad) + math.Cos(lat1Rad)*math.Cos(lat2Rad)*math.Cos(deltaLon)

    return RadiusOfEarthInMeters * math.Atan2(y, x)
}

```

**The Solution**
For the vast majority of transit API queries (finding nearby stops or vehicles), the distance is short enough that the Earth's curvature is negligible. I propose adding an **Equirectangular approximation** fast-path for distances under ~22km (0.2 degrees).

**Proposed Optimization:**

```go
// Fast-path approximation for nearby points
if math.Abs(lat2-lat1) < 0.2 && math.Abs(lon2-lon1) < 0.2 {
    x := dLonRad * math.Cos((lat1Rad+lat2Rad)/2)
    y := dLatRad
    return RadiusOfEarthInMeters * math.Sqrt(x*x + y*y)
}
// Fallback to exact Haversine for long distances...

```

**Impact**

1. **Performance:** Replaces heavy trigonometry with simple Pythagorean arithmetic for 99% of intra-city transit queries.
2. **Accuracy:** The difference is measured in centimeters, which is well within the error margin of consumer GPS hardware.
3. **Stability:** Automatically falls back to the exact formula for long-distance calculations.

ARCoder181105:
### The Problem

The current `Distance` function in `internal/utils/geometry.go` relies exclusively on the Haversine formula for all distance calculations. While highly accurate, the Haversine formula is computationally expensive due to multiple trigonometric operations, including `math.Sin`, `math.Cos`, `math.Atan2`, and `math.Pow`.

In high-load scenarios—such as sorting hundreds of vehicles or bus stops by proximity to a user—these calculations are executed thousands of times per second, creating a silent but significant CPU bottleneck in spatial queries.

### The Solution

For the vast majority of transit API queries (finding nearby stops or vehicles), the distance is short enough (typically under 20km) that the Earth's curvature is negligible. I have implemented an **Equirectangular approximation** fast-path for distances where the coordinate difference is less than 0.2 degrees (~22km).

**Key Changes:**

* **Fast-Path Logic:** Added a threshold check to use Pythagorean arithmetic for nearby coordinates, which replaces heavy trigonometry with simple multiplication and square roots.
* **Fallback Mechanism:** Maintained the exact Haversine formula as a fallback for long-distance calculations to ensure global accuracy is never compromised.
* **Test Updates:** Updated `internal/utils/geometry_test.go` to slightly increase the tolerance for the "Very close points" test case (from 1.0m to 2.0m) to account for the approximation.
* **New Coverage:** Added a dedicated test `TestDistance_EquirectangularFastPath` to verify the accuracy and symmetry of the new fast-path logic.

### Benchmarks Added

I added the benchmarks to `geometry_test.go` and ran them locally. The results show exactly why this optimization is needed—the fast-path is **over 12x faster** (a ~92% reduction in CPU time per operation).

**Benchmark on `main` (Exact formula for everything):**

```text
BenchmarkDistance_ShortRange-16          8405379               140.6 ns/op        0 B/op           0 allocs/op
BenchmarkDistance_LongRange-16           8232061               144.7 ns/op        0 B/op           0 allocs/op

```

**Benchmark on this PR (`perf/optimize-distance-calc`):**

```text
BenchmarkDistance_ShortRange-16        100000000               11.68 ns/op        0 B/op           0 allocs/op
BenchmarkDistance_LongRange-16           8198871               144.8 ns/op        0 B/op           0 allocs/op

```

### Impact

* **Performance:** Significant reduction in CPU cycles for the most common spatial query paths (finding nearby transit assets).
* **Accuracy:** The precision trade-off is measured in centimeters, which is well within the standard error margin of consumer GPS hardware.

closes #428

### DIFF
--- a/internal/utils/geometry.go
+++ b/internal/utils/geometry.go
@@ -3,7 +3,7 @@ package utils
 import "math"
 
 const (
-	// RADIUS_OF_EARTH_IN_METERS RADIUS_OF_EARTH_IN_KM * 1000
+	// RadiusOfEarthInMeters is RADIUS_OF_EARTH_IN_KM * 1000
 	RadiusOfEarthInMeters = 6371010.0
 )
 
@@ -15,8 +15,25 @@ type CoordinateBounds struct {
 	MaxLon float64
 }
 
-// Distance calculates the distance between two points on the Earth
+// Distance calculates the distance between two points on the Earth.
+// For short distances (under ~22km), it uses a highly optimized Equirectangular
+// approximation to save CPU cycles. For longer distances, it falls back to the exact formula.
 func Distance(lat1, lon1, lat2, lon2 float64) float64 {
+	// Fast-path for short distances: coordinate differences less than 0.2 degrees (~22km)
+	// Bypasses expensive Atan2, Pow, and multiple Sin/Cos calls for 99% of transit queries.
+	if math.Abs(lat2-lat1) < 0.2 && math.Abs(lon2-lon1) < 0.2 {
+		lat1Rad := lat1 * (math.Pi / 180)
+		lat2Rad := lat2 * (math.Pi / 180)
+		dLatRad := (lat2 - lat1) * (math.Pi / 180)
+		dLonRad := (lon2 - lon1) * (math.Pi / 180)
+
+		// Equirectangular approximation
+		x := dLonRad * math.Cos((lat1Rad+lat2Rad)/2)
+		y := dLatRad
+		return RadiusOfEarthInMeters * math.Sqrt(x*x+y*y)
+	}
+
+	// Exact calculation fallback for longer distances
 	lat1Rad := lat1 * (math.Pi / 180)
 	lon1Rad := lon1 * (math.Pi / 180)
 	lat2Rad := lat2 * (math.Pi / 180)
--- a/internal/utils/geometry_test.go
+++ b/internal/utils/geometry_test.go
@@ -428,3 +428,34 @@ func TestIsOutOfBounds(t *testing.T) {
 		})
 	}
 }
+
+func TestDistance_EquirectangularFastPath(t *testing.T) {
+	// Test coordinates that are less than 0.2 degrees apart
+	// to ensure the fast-path approximation triggers and is accurate
+	lat1, lon1 := 47.6062, -122.3321 // Seattle downtown
+	lat2, lon2 := 47.6101, -122.3421 // Pike Place Market
+
+	// The fast-path should calculate this correctly
+	distAB := Distance(lat1, lon1, lat2, lon2)
+
+	// Ensure it falls within the expected physical distance (~860 meters)
+	assert.InDelta(t, 860.0, distAB, 20.0, "Fast-path distance should be highly accurate")
+
+	// Ensure the fast-path is symmetric
+	distBA := Distance(lat2, lon2, lat1, lon1)
+	assert.InDelta(t, distAB, distBA, 0.0001, "Fast-path distance should be symmetric")
+}
+
+func BenchmarkDistance_ShortRange(b *testing.B) {
+	// Typical transit query: ~860m apart (fast-path)
+	for i := 0; i < b.N; i++ {
+		Distance(47.6062, -122.3321, 47.6101, -122.3421)
+	}
+}
+
+func BenchmarkDistance_LongRange(b *testing.B) {
+	// Cross-country: NYC to LA (Exact fallback)
+	for i := 0; i < b.N; i++ {
+		Distance(40.7128, -74.0060, 34.0522, -118.2437)
+	}
+}


aaronbrethorst (changes requested):
Hey Aditya, the equirectangular fast-path is a well-chosen optimization. The threshold at 0.2 degrees and the `cos(mean_latitude)` correction are both textbook. Before we can merge this, I will need you to make a couple changes:

## Important

### 1. Add a benchmark to demonstrate the performance improvement

This PR's entire motivation is performance, but there's no benchmark data to verify the claim. Go makes this straightforward — add a benchmark alongside the existing tests:

```go
func BenchmarkDistance_ShortRange(b *testing.B) {
    // Typical transit query: ~860m apart (fast-path)
    for i := 0; i < b.N; i++ {
        Distance(47.6062, -122.3321, 47.6101, -122.3421)
    }
}

func BenchmarkDistance_LongRange(b *testing.B) {
    // Cross-country: NYC to LA (Vincenty fallback)
    for i := 0; i < b.N; i++ {
        Distance(40.7128, -74.0060, 34.0522, -118.2437)
    }
}
```

Then run `go test -bench=. -benchmem ./internal/utils/` on the current `main` branch and on your branch, and include both results in the PR description. This makes the optimization self-documenting and protects against future regressions.

## Fit and Finish

### 2. Verify whether the tolerance increase is actually needed (`geometry_test.go:148`)

The "Very close points" test tolerance was increased from 1.0m to 2.0m. For points ~13.5 meters apart, both the Vincenty formula and the equirectangular approximation should give nearly identical results (I get ~13.95m for both in back-of-envelope calculations). Please verify the actual computed value — if the equirectangular result is within the original 1.0m tolerance, keep it at 1.0m. If it genuinely needs 2.0m, add a brief comment explaining why (e.g., `// Wider tolerance due to equirectangular approximation`).

## Strengths

- **Right optimization, right place**: `Distance` is called in hot loops (shape projection in `trips_helper.go`, proximity sorting in `stops_for_location`). Cutting trig operations here has real impact.
- **Clean fallback**: The threshold check is cheap (two `math.Abs` + comparisons) and the Vincenty path remains untouched for long distances.
- **Good test**: `TestDistance_EquirectangularFastPath` verifies both accuracy and symmetry of the new path.

ARCoder181105:
Hey @aaronbrethorst 
Thanks for the feedback. I've pushed the requested changes:

**1. Test Tolerance Reverted**
I reverted the "Very close points" test tolerance back to `1.0m`. As you calculated, the equirectangular approximation yields ~13.95m. Since the test expects 13.5m, the 0.45m difference safely passes within the original 1.0m tolerance! No bump needed.

**2. Benchmarks Added**
I added the benchmarks to `geometry_test.go` and ran them locally. The results show exactly why this optimization is needed—the fast-path is **over 12x faster** (a ~92% reduction in CPU time per operation).

**Benchmark on `main` (Exact formula for everything):**

```text
BenchmarkDistance_ShortRange-16          8405379               140.6 ns/op        0 B/op           0 allocs/op
BenchmarkDistance_LongRange-16           8232061               144.7 ns/op        0 B/op           0 allocs/op

```

**Benchmark on this PR (`perf/optimize-distance-calc`):**

```text
BenchmarkDistance_ShortRange-16        100000000               11.68 ns/op        0 B/op           0 allocs/op
BenchmarkDistance_LongRange-16           8198871               144.8 ns/op        0 B/op           0 allocs/op

```
The test are passing perfectly

image

aaronbrethorst (changes requested):
Hey Aditya, nice update — both items from my previous review are resolved. The benchmarks confirm the optimization delivers exactly what you claimed (I ran them locally and got similar numbers: ~4ns vs ~56ns on short-range, a 14x improvement), and the tolerance is correctly at 1.0m. Just one small thing left:

## Blocking

### 1. `gofmt` trailing whitespace in `geometry_test.go`

Line 440 has a trailing tab on a blank line. Please run:

```bash
go fmt ./internal/utils/geometry_test.go
```

Then amend the commit.

---

Once that's in, this is ready to merge. The equirectangular fast-path is a clean, well-tested optimization that targets exactly the right hot path for transit queries.

ARCoder181105:
Hi @aaronbrethorst 
Done , implemented the command now 
This is ready for review

PR #430, opened 2026-02-19
chore: add doc comment and DB error logging for alerts optimization

3rabiii:
### Description
This is a follow-up to the recently merged alerts optimization #395 PR to implement the non-blocking 

### Changes Made
* **Doc Comment:** Added standard Go documentation for the exported `GetAlertsByIDs` method in `realtime.go`.
* **Graceful Error Logging:** Added `slog.WarnContext` in `GetSituationIDsForTrip` (`trips_helper.go`) to log non-`sql.ErrNoRows` database errors. This improves observability for operators during database degradation without breaking the existing fallback behavior (trip-only alert matching).

image

@aaronbrethorst 
Ready for review!

### DIFF
--- a/internal/gtfs/realtime.go
+++ b/internal/gtfs/realtime.go
@@ -2,6 +2,8 @@ package gtfs
 
 import (
 	"context"
+	"database/sql"
+	"errors"
 	"fmt"
 	"io"
 	"log/slog"
@@ -111,6 +113,7 @@ func (manager *Manager) GetAlertsForRoute(routeID string) []gtfs.Alert {
 	return alerts
 }
 
+// GetAlertsByIDs returns alerts matching the given trip, route, or agency IDs.
 func (manager *Manager) GetAlertsByIDs(tripID, routeID, agencyID string) []gtfs.Alert {
 	manager.realTimeMutex.RLock()
 	defer manager.realTimeMutex.RUnlock()
@@ -122,7 +125,7 @@ func (manager *Manager) GetAlertsByIDs(tripID, routeID, agencyID string) []gtfs.
 			continue
 		}
 		for _, entity := range alert.InformedEntities {
-			if entity.TripID != nil && entity.TripID.ID == tripID {
+			if entity.TripID != nil && tripID != "" && entity.TripID.ID == tripID {
 				alerts = append(alerts, alert)
 				break
 			}
@@ -152,7 +155,18 @@ func (manager *Manager) GetAlertsForTrip(ctx context.Context, tripID string) []g
 			route, err := manager.GtfsDB.Queries.GetRoute(ctx, routeID)
 			if err == nil {
 				agencyID = route.AgencyID
+			} else if !errors.Is(err, sql.ErrNoRows) {
+				slog.WarnContext(ctx, "Failed to fetch route for alerts; degrading to trip+route matching only",
+					slog.String("trip_id", tripID),
+					slog.String("route_id", routeID),
+					slog.Any("error", err),
+				)
 			}
+		} else if !errors.Is(err, sql.ErrNoRows) {
+			slog.WarnContext(ctx, "Failed to fetch trip for alerts",
+				slog.String("trip_id", tripID),
+				slog.Any("error", err),
+			)
 		}
 	}
 
--- a/internal/restapi/trips_helper.go
+++ b/internal/restapi/trips_helper.go
@@ -2,7 +2,10 @@ package restapi
 
 import (
 	"context"
+	"database/sql"
+	"errors"
 	"fmt"
+	"log/slog"
 	"math"
 	"sort"
 	"time"
@@ -845,7 +848,18 @@ func (api *RestAPI) GetSituationIDsForTrip(ctx context.Context, tripID string) [
 			route, err := api.GtfsManager.GtfsDB.Queries.GetRoute(ctx, routeID)
 			if err == nil {
 				agencyID = route.AgencyID
+			} else if !errors.Is(err, sql.ErrNoRows) {
+				api.Logger.Warn("Failed to fetch route for alerts; degrading to trip+route matching only",
+					slog.String("trip_id", tripID),
+					slog.String("route_id", routeID),
+					slog.Any("error", err),
+				)
 			}
+		} else if !errors.Is(err, sql.ErrNoRows) {
+			api.Logger.Warn("Failed to fetch trip for alerts; degrading to trip matching only",
+				slog.String("trip_id", tripID),
+				slog.Any("error", err),
+			)
 		}
 	}
 


aaronbrethorst (changes requested):
Hey Adel, thanks for taking the time to improve observability around the alerts optimization. Surfacing previously-silent database errors is exactly the kind of follow-up work that makes a codebase more operable. Before we can merge this, I will need you to make a couple changes:

## Critical

1. **Use `errors.Is` instead of `!=` for sentinel error comparison** (`internal/restapi/trips_helper.go:819`): The PR uses `err != sql.ErrNoRows`, but idiomatic Go uses `errors.Is` for sentinel comparisons because it handles wrapped errors correctly. If `sql.ErrNoRows` is ever wrapped (e.g., by middleware or a future sqlc change), the `!=` check would fail and start producing false-positive warnings for every "trip not found" result. The codebase already uses the correct pattern in `trip_for_vehicle_handler.go:178`. Change to:

   ```go
   } else if !errors.Is(err, sql.ErrNoRows) {
   ```

   (You'll need to add `"errors"` to the import block.)

2. **The `GetRoute` error path is completely unlogged** (`internal/restapi/trips_helper.go:815-818`): The PR adds logging when `GetTrip` fails, but the immediately following `GetRoute` call still silently swallows its errors. If the trip exists but the route lookup fails due to a real DB error, `agencyID` will silently remain empty and agency-level alerts will be missed with no indication of why. Please add similar logging:

   ```go
   route, err := api.GtfsManager.GtfsDB.Queries.GetRoute(ctx, routeID)
   if err == nil {
       agencyID = route.AgencyID
   } else if !errors.Is(err, sql.ErrNoRows) {
       slog.WarnContext(ctx, "Failed to fetch route for alerts; degrading to trip+route matching only",
           slog.String("trip_id", tripID),
           slog.String("route_id", routeID),
           slog.Any("error", err),
       )
   }
   ```

3. **`GetAlertsForTrip` in `realtime.go:145-161` has the identical silent failure pattern**: This function performs the same trip-lookup-then-route-lookup sequence but has zero error logging. Since it's called from `arrival_and_departure_for_stop_handler.go` and `trip_for_vehicle_handler.go`, operators would only get alerts for failures on *some* endpoints but silently miss identical failures on others. Please apply the same error logging treatment there as well.

## Important

4. **Use `api.Logger` instead of the `slog` package-level functions** (`internal/restapi/trips_helper.go:820`): Every other log call in `internal/restapi/` uses `api.Logger.Warn(...)` or `api.Logger.Error(...)`. The `api.Logger` is the application's configured structured logger and may have different handlers or formatters than the default `slog` package logger. Using `slog.WarnContext` directly could send these logs to a different destination than all other REST API logs. Since `GetSituationIDsForTrip` is a method on `*RestAPI`, you have access to `api.Logger`.

## Fit and Finish

5. **The `GetAlertsByIDs` doc comment is slightly misleading about empty-string handling** (`internal/gtfs/realtime.go:114`): The comment says "returns alerts matching the given trip, route, or agency IDs," which implies symmetric treatment. However, the code treats `tripID` differently: `routeID` and `agencyID` have explicit `!= ""` guards (lines 130, 134), but `tripID` does not (line 126). This means passing `""` for `tripID` could match entities with an empty trip ID, while `""` for the others is safely skipped. This is a pre-existing issue, but since you're adding the doc comment, it would be good to either add a `tripID != ""` guard for consistency, or note the asymmetry in the comment.

3rabiii:
@aaronbrethorst 
Ready for review!

aaronbrethorst (changes requested):
Hey Adel, you're making great progress on the alerts observability improvements. All five items from my last review are properly addressed — `errors.Is`, `api.Logger`, the `tripID` empty-string guard, and both trip/route error logging paths in both files all look correct. However, the latest commit introduced a new issue that needs to be fixed before we can merge:

## Critical

1. **Double `realTimeMutex.RLock()` in `GetAlertsForTrip` will deadlock** (`internal/gtfs/realtime.go:173-176`): The new code acquires `realTimeMutex.RLock()` at line 173, then calls `GetAlertsByIDs` which acquires `realTimeMutex.RLock()` *again* at line 118. Go's `sync.RWMutex` is **not reentrant** — if a writer is pending (which happens every time real-time data updates at `realtime.go:252`), the second `RLock` will block, causing a deadlock:

   1. Request goroutine: acquires `RLock` (line 173)
   2. Background goroutine: calls `Lock()` to update real-time data (line 252), waits for all readers to release
   3. Request goroutine: tries `RLock` again inside `GetAlertsByIDs` (line 118) — **blocked** because Go's RWMutex blocks new readers when a writer is pending
   4. **Deadlock**: writer waits for step 1's RLock, request waits for the writer

   The Go docs explicitly warn about this: *"Note that this precludes recursive locking."*

   **Fix**: Remove lines 173-174. `GetAlertsByIDs` already handles its own locking, so the extra lock in `GetAlertsForTrip` is both unnecessary and dangerous:

   ```go
   // Remove these two lines:
   manager.realTimeMutex.RLock()
   defer manager.realTimeMutex.RUnlock()
   ```

3rabiii:
@aaronbrethorst 
Ready for review!

aaronbrethorst (changes requested):
Hey Adel, the error logging and `tripID` guard from round 1 all look correct -- `errors.Is`, `api.Logger` in `trips_helper.go`, and symmetric empty-string checks in `GetAlertsByIDs` are exactly right. However, the critical deadlock from round 2 is still present:

## Critical

1. **Double `realTimeMutex.RLock()` in `GetAlertsForTrip` still present -- will deadlock** (`internal/gtfs/realtime.go:173-176`): Lines 173-174 acquire `realTimeMutex.RLock()`, then line 176 calls `GetAlertsByIDs`, which acquires `realTimeMutex.RLock()` again at line 118. Go's `sync.RWMutex` is not reentrant. When a real-time data update calls `realTimeMutex.Lock()` at line 252 between these two reads, the second `RLock` blocks and deadlocks:

   1. Request goroutine: `RLock` at line 173
   2. Background goroutine: `Lock()` at line 252 to update real-time data -- waits for readers
   3. Request goroutine: `RLock` at line 118 inside `GetAlertsByIDs` -- **blocked** by pending writer
   4. **Deadlock**

   Tests pass because there's no concurrent writer during testing, but this will deadlock in production under real-time feed update load.

   **Fix**: Remove lines 173-174. `GetAlertsByIDs` already handles its own locking:
   ```go
   // Remove these two lines from GetAlertsForTrip:
   manager.realTimeMutex.RLock()
   defer manager.realTimeMutex.RUnlock()
   ```

   So lines 172-176 become simply:
   ```go
   return manager.GetAlertsByIDs(tripID, routeID, agencyID)
   ```

3rabiii:
"Hi Aaron, I'm a bit confused because I believe I already removed those two RLock/RUnlock lines in my previous commit. I've double-checked both my local environment and the 'Files changed' tab here on GitHub, and I can't seem to find them in GetAlertsForTrip anymore. Could you please double-check if you might be looking at an older cached version of the PR? Or if they are still showing up for you, could you point me to the exact line number on the current HEAD?
@aaronbrethorst

Issue #432, opened 2026-02-20
Critical: Race conditions and nested lock dependencies in GTFS manager


3rabiii:
### Overview
During a recent code review, two critical concurrency issues were identified in the GTFS `Manager` that could lead to server panics or deadlocks under concurrent `ForceUpdate` operations.

### Race Condition (TOCTOU) in `GetAlertsForTrip`
- **Location:** `internal/gtfs/realtime.go`
- **Problem:** The method checks `manager.GtfsDB != nil` but does not hold a static lock before dereferencing it. Between the check and the use, a concurrent `ForceUpdate()` can close the DB or set `GtfsDB` to nil, causing a nil pointer dereference (panic). Unlike other methods, this lacks both an internal lock and a caller contract comment.

### Race Condition & Nested Locks in `GetVehicleForTrip`
- **Location:** `internal/gtfs/gtfs_manager.go`
- **Problem:** The method accesses `GtfsDB` relying solely on a doc-comment contract (`// IMPORTANT: Caller must hold manager.RLock()`). If a caller forgets, a data race occurs. Furthermore, the method internally acquires `realTimeMutex.RLock()`. If a caller *does* hold the static lock, it creates a dangerous nested lock dependency that can lead to deadlocks.

### Expected Behavior
- Database access should be intrinsically thread-safe without relying on doc-comment contracts.
- Nested lock dependencies must be eliminated to prevent potential deadlocks.

PR #433, opened 2026-02-20
fix: resolve critical TOCTOU race conditions and nested lock dependencies in GTFS manager

3rabiii:
### Description
This PR addresses the Critical Concurrency / Race Condition issues flagged in the recent review. It enforces thread-safe database access and removes dangerous doc-comment contracts.

### 1. Fixed : TOCTOU Race Condition in `GetAlertsForTrip`
- **Issue:** The method checked `manager.GtfsDB != nil` but didn't acquire a static lock, risking a panic if `ForceUpdate` fired concurrently.
- **Fix:** Removed the ineffective `// IMPORTANT: Caller must hold...` comment and safely encapsulated `manager.realTimeMutex.RLock()` and `defer manager.realTimeMutex.RUnlock()` *inside* the function. Callers (like handlers) no longer need to worry about the internal GTFS DB locking logic.

### 2. Fixed : Race Condition & Nested Locks in `GetVehicleForTrip`
- **Issue:** The method relied on callers holding `manager.RLock()` to safely query `GtfsDB`, and then acquired `realTimeMutex.RLock()` internally, creating a dangerous nested lock dependency that could lead to deadlocks.
- **Fix:** - Dropped the doc-comment contract.
  - Scoped `manager.RLock()` strictly around the database queries (`GetTrip` and `GetTripsByBlockID`).
  - Explicitly released `manager.RUnlock()` immediately after the DB queries complete and *before* acquiring `realTimeMutex.RLock()`.
  - Added safe unlocks on all early error returns.
  - This guarantees thread-safe DB access while completely eliminating the nested lock risk.
 
image

@aaronbrethorst 
fixes : #432

### DIFF
--- a/internal/gtfs/gtfs_manager.go
+++ b/internal/gtfs/gtfs_manager.go
@@ -368,24 +368,21 @@ func (manager *Manager) VehiclesForAgencyID(agencyID string) []gtfs.Vehicle {
 // GetVehicleForTrip retrieves a vehicle for a specific trip ID or finds the first vehicle that is part of the block
 // for that trip. Note we depend on getting the vehicle that may not match the trip ID exactly,
 // but is part of the same block.
+// IMPORTANT: Caller must hold manager.RLock() before calling this method.
 func (manager *Manager) GetVehicleForTrip(tripID string) *gtfs.Vehicle {
 	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
 	defer cancel()
 
 	logger := slog.Default().With(slog.String("component", "gtfs_manager"))
 
-	manager.RLock()
-
 	requestedTrip, err := manager.GtfsDB.Queries.GetTrip(ctx, tripID)
 	if err != nil {
-		manager.RUnlock()
 		logging.LogError(logger, "could not get trip", err,
 			slog.String("trip_id", tripID))
 		return nil
 	}
 
 	if !requestedTrip.BlockID.Valid {
-		manager.RUnlock()
 		logger.Debug("trip has no block ID, cannot find vehicle by block",
 			slog.String("trip_id", tripID))
 		return nil
@@ -394,8 +391,6 @@ func (manager *Manager) GetVehicleForTrip(tripID string) *gtfs.Vehicle {
 	requestedBlockID := requestedTrip.BlockID.String
 
 	blockTrips, err := manager.GtfsDB.Queries.GetTripsByBlockID(ctx, requestedTrip.BlockID)
-
-	manager.RUnlock()
 
 	if err != nil {
 		logging.LogError(logger, "could not get trips for block", err,
--- a/internal/gtfs/realtime.go
+++ b/internal/gtfs/realtime.go
@@ -140,10 +140,8 @@ func (manager *Manager) GetAlertsByIDs(tripID, routeID, agencyID string) []gtfs.
 }
 
 // GetAlertsForTrip returns alerts matching the trip, its route, or agency.
+// IMPORTANT: Caller must hold manager.RLock() before calling this method.
 func (manager *Manager) GetAlertsForTrip(ctx context.Context, tripID string) []gtfs.Alert {
-	manager.realTimeMutex.RLock()
-	defer manager.realTimeMutex.RUnlock()
-
 	var routeID string
 	var agencyID string
 


aaronbrethorst (changes requested):
Hey Adel, thanks for diving into the concurrency model of the GTFS manager — race conditions are notoriously tricky to reason about, and it's clear you've thought carefully about the lock ordering. The `GetAlertsForTrip` fix is a genuine improvement that closes a real data race. However, the `GetVehicleForTrip` change introduces a new deadlock risk that needs to be addressed before we can merge.

## Critical

### 1. Recursive `RLock` in `GetVehicleForTrip` creates a deadlock under writer contention

**File:** `internal/gtfs/gtfs_manager.go:377`

Every caller of `GetVehicleForTrip` already holds `manager.RLock()` (which is `staticMutex.RLock()`):

- `tripDetailsHandler` — line 106: `api.GtfsManager.RLock()` → `BuildTripStatus` → `GetVehicleForTrip`
- `tripForVehicleHandler` — line 107: `api.GtfsManager.RLock()` → `BuildTripStatus` → `GetVehicleForTrip`
- `arrivalsAndDeparturesForStop` — line 106: `api.GtfsManager.RLock()` → `GetVehicleForTrip`
- `arrivalAndDepartureForStopHandler` — line 124: `api.GtfsManager.RLock()` → `GetVehicleForTrip`

With this PR, `GetVehicleForTrip` now also calls `manager.RLock()` internally. This creates a recursive read-lock on the same goroutine. Go's `sync.RWMutex` does **not** support reentrant locking, and the [Go documentation explicitly warns against this](https://pkg.go.dev/sync#RWMutex.RLock):

> *"If a goroutine holds a RWMutex for reading and another goroutine might call Lock, no goroutine should expect to be able to acquire a read lock until the initial read lock is released."*

Here's the concrete deadlock scenario:

1. **Handler goroutine**: acquires `staticMutex.RLock()` (reader count = 1)
2. **ForceUpdate goroutine**: calls `staticMutex.Lock()` → blocks, and **prevents new readers**
3. **Handler goroutine** (in `GetVehicleForTrip`): calls `staticMutex.RLock()` → **blocked** (pending writer)
4. **Deadlock**: the handler can't release its first `RLock` because it's blocked in step 3, and `ForceUpdate` can't proceed because the handler still holds the `RLock` from step 1.

`ForceUpdate` fires periodically, so this deadlock **will** occur in production — it's not just theoretical.

Additionally, the manual `manager.RUnlock()` calls at lines 381, 388, and 398 don't achieve their stated goal of "releasing the static lock before acquiring the realtime lock" — they only decrement the reader count from 2 to 1. The handler's `RLock` still keeps `staticMutex` held, so nested locking still occurs at lines 411-412.

**Suggested fix:** Revert `GetVehicleForTrip` to its previous approach: document the caller contract and let callers manage the static lock. The "doc-comment contract" pattern is well-established across this codebase (used by 15+ other methods) and is the idiomatic approach for this lock architecture.

If you want to move toward self-contained locking, that's a larger refactor — you'd need to audit every caller to remove their `RLock`/`RUnlock` pairs and have each method acquire its own lock. That's a worthy goal but beyond the scope of this PR.

## Important

### 2. mann-patwa's review comment deserves a response

mann-patwa correctly identified the core concern: *"If we lock at the DB access level and the caller already holds a lock, we could still run into deadlocks."* This is the exact issue described above. Please engage with their feedback — it's spot-on.

## Fit and Finish

### 3. `GetAlertsForTrip` change is correct — consider adding a brief comment

The `realTimeMutex.RLock()` addition to `GetAlertsForTrip` is a genuine fix. Previously, `GetAlertsByIDs` (called at line 161) accessed `manager.realTimeAlerts` without holding `realTimeMutex`, which was a data race. The new internal lock correctly protects this access.

One thing to note: the `GtfsDB` access at lines 150-158 (`GetTrip`, `GetRoute`) is still only protected because all current callers happen to hold `manager.RLock()`. Since you removed the doc-comment contract, consider adding a brief comment noting that DB access safety depends on the caller context, or add `manager.RLock()`/`manager.RUnlock()` scoped around just those DB calls (being careful to release it before `GetAlertsByIDs` to avoid holding two locks longer than necessary).

Thanks again, and I look forward to merging this once the `GetVehicleForTrip` deadlock is resolved.

PR #437, opened 2026-02-20
perf: reduce global lock contention in RateLimitMiddleware

### LINKED ISSUE #436: Performance: Reduce global lock contention in RateLimitMiddleware
**Description:**
While reviewing `internal/restapi/rate_limit_middleware.go`, I noticed that `getLimiter()` acquires a full write lock (`rl.mu.Lock()`) for every single incoming HTTP request.

Since the vast majority of traffic comes from returning API keys, the global write lock is currently only being used to update the `lastSeen` timestamp. Under high traffic, this forces concurrent requests to wait in line sequentially, creating an unnecessary bottleneck.

**Proposed Solution:**
We can easily optimize this by switching to a Read-Write lock pattern:

1. Change `lastSeen` in the `rateLimitClient` struct from `time.Time` to `atomic.Int64` (to store Unix nanoseconds).
2. Use an `RLock()` (fast path) to update the timestamp and return the limiter for existing clients.
3. Fall back to a full `Lock()` (slow path) only when creating a rate limiter for a *new* API key.

**Example Implementation:**

```go
func (rl *RateLimitMiddleware) getLimiter(apiKey string) *rate.Limiter {
	// Update timestamp using only a Read Lock
	rl.mu.RLock()
	if client, exists := rl.limiters[apiKey]; exists {
		client.lastSeen.Store(rl.clock.Now().UnixNano())
		rl.mu.RUnlock()
		return client.limiter
	}
	rl.mu.RUnlock()

	// Full write lock (with double-checking) to create a new client
	rl.mu.Lock()
	defer rl.mu.Unlock()
	// ... create and store new client limiter ...
}

```

**Impact:**
This allows requests from existing clients to pass through the rate limiter entirely in parallel, significantly improving the server's vertical scaling and maximum throughput.

I already have the implementation and the updated tests working locally. I'd be happy to submit a PR if this approach looks good to you!

ARCoder181105:
### Description

This PR addresses the concurrency bottleneck identified in the rate-limiting middleware where a global write lock (`rl.mu.Lock()`) was being acquired for every incoming HTTP request. By optimizing the lock usage, we can significantly increase the server's maximum request throughput under heavy load.

**Closes #436**

### Changes Made

* **Switched to Atomic Timestamps:** Changed the `lastSeen` field in `rateLimitClient` from `time.Time` to `atomic.Int64` (storing Unix nanoseconds) to allow thread-safe updates without requiring a full write lock.
* **Implemented Fast-Path / Slow-Path:** * **Fast Path:** Existing clients now only acquire an `RLock()` to update their timestamp and retrieve their limiter.
* **Slow Path:** New clients fall back to a full `Lock()` with double-checked locking to safely initialize a new limiter.

* **Updated Cleanup Ticker:** Modified the background garbage collection loop to read from the atomic `lastSeen` value and convert it back to `time.Time` for eviction calculations.
* **Updated Tests:** Refactored `rate_limit_cleanup_test.go` to safely interact with the new `atomic.Int64` type when fast-forwarding time and asserting eviction logic.

### Impact

* **Higher Concurrency:** Requests from returning API keys can now pass through the middleware in parallel rather than waiting sequentially for a write lock.
* **Lower Tail Latency:** Reduces CPU wait times during high-traffic spikes.

### DIFF
--- a/internal/restapi/rate_limit_cleanup_test.go
+++ b/internal/restapi/rate_limit_cleanup_test.go
@@ -4,11 +4,13 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httptest"
+	"sync"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/time/rate"
 	"maglev.onebusaway.org/internal/clock"
 )
 
@@ -49,17 +51,7 @@ func TestRateLimitMiddleware_CleanupKeepsActiveClients(t *testing.T) {
 	mockClock.Advance(6 * time.Minute)
 
 	// Trigger cleanup manually
-	middleware.mu.Lock()
-	now := mockClock.Now()
-	threshold := 10 * time.Minute
-	for key, client := range middleware.limiters {
-		if !middleware.exemptKeys[key] {
-			if now.Sub(client.lastSeen) > threshold {
-				delete(middleware.limiters, key)
-			}
-		}
-	}
-	middleware.mu.Unlock()
+	middleware.cleanupOnce()
 
 	// Verify limiter still exists (last seen was only 6 minutes ago)
 	middleware.mu.RLock()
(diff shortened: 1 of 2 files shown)


aaronbrethorst (approved):
Hey Aditya, you've nailed every item from my review. The `lastSeen` field comment, the double-checked locking test with 50 goroutines, the capitalized comment, the zero-value guard, and the `cleanupOnce` extraction that eliminates the duplicated cleanup logic in tests — all addressed cleanly in a single commit. I particularly appreciate that you tackled the `cleanupOnce` refactor (which I'd offered as a follow-up) right here in this PR; it makes the test suite significantly more maintainable.

There's nothing left to change — this PR is ready to merge. The RLock fast-path will be a meaningful improvement for high-traffic deployments where returning API keys were previously serialized through a write lock on every request.

PR #439, opened 2026-02-20
fix: allow colon in ID pattern validation to support GTFS IDs

### LINKED ISSUE #438: bug: valid GTFS IDs containing colons are rejected by ID validation
### Context
Some GTFS agencies use colons (`:`) in their IDs (e.g., 29_101:179). The current validIDPattern rejects these as invalid since it only allows [a-zA-Z0-9_.-].

This is a false positive validation issue, IDs in this service are treated as opaque strings and are NOT interpreted as URLs, SQL fragments, or executable content. Allowing (`:`) in this context does not introduce injection risk.

Also the [GTFS specification](https://gtfs.org/schedule/reference/) does not restrict which characters can appear in IDs.

These IDs can be found in `shape_id` field of some agencies e.g., Puget Sound Transit agencies 29, 40) include colon-containing shape_id values.

### Current behavior:
Requests with colon-containing IDs return 400 Bad Request.
```
// currently:
validIDPattern = regexp.MustCompile(`^[a-zA-Z0-9_.-]+$`)
```

### Expected behavior:
Colon-containing IDs should be accepted as valid.

### Recommendation
`util/validation.go` file:
```
// change to allow colon
validIDPattern = regexp.MustCompile(`^[a-zA-Z0-9_.:-]+$`) 
```

uloamaka:
### Description
This PR addresses the input ID validation regex logic which incorrectly rejects IDs containing colons (:).

### Changes
- Allow colon in `validIDPattern` to support GTFS agencies that use colons in their IDs (e.g., agency_N23:S01)
- The colon poses no SQL injection or XSS risk

Fix: #438

### DIFF
--- a/internal/utils/validation.go
+++ b/internal/utils/validation.go
@@ -9,8 +9,8 @@ import (
 
 // Compiled regular expressions for validation
 var (
-	// Allow alphanumeric, underscore, hyphen, dot - common in transit IDs
-	validIDPattern = regexp.MustCompile(`^[a-zA-Z0-9_.-]+$`)
+	// Allow alphanumeric, underscore, hyphen, dot, colon - common in transit IDs
+	validIDPattern = regexp.MustCompile(`^[a-zA-Z0-9_.:-]+$`)
 
 	// Detect potentially dangerous characters - more focused on injection patterns
 	dangerousPattern = regexp.MustCompile(`[<>]|--|\/\*|\*\/|;.*--`)
--- a/internal/utils/validation_test.go
+++ b/internal/utils/validation_test.go
@@ -64,6 +64,11 @@ func TestValidateID(t *testing.T) {
 			id:      "agency.123_stop.456",
 			wantErr: false,
 		},
+		{
+			name:    "valid ID with colon",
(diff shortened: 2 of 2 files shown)


aaronbrethorst (approved):
Good catch, Godsgift — real agencies do use colons in their GTFS IDs, and rejecting those is a legitimate bug.

This is a clean, well-motivated fix. The regex change is correct (`-` stays literal at the end of the character class), the test case covers the new behavior, and colons in opaque ID strings pose no injection risk since all database access goes through parameterized sqlc queries.

## What I Checked

- The regex `^[a-zA-Z0-9_.:-]+$` is syntactically correct — no accidental range created; `-` remains at the end of the character class so all four special characters (`_`, `.`, `:`, `-`) are literals
- The new test case uses a realistic GTFS-shaped ID (`agency_shp_654:456`)
- All existing tests still pass — no regressions
- Lint is clean
- IDs are used as parameterized query inputs via sqlc, so no SQL injection concern
- The linked issue (#438) provides solid justification with real-world examples from Puget Sound agencies

No issues found. Merging.

PR #444, opened 2026-02-21
feat: Enhance region bounds calculation to use stops as a fallback an…

mann-patwa:
# Description

This PR introduces several improvements and bug fixes to the routes for location handler and the underlying spatial bounding box calculations. 

Previously, when querying routes for a specific location, the handler had a logic flaw where it could drop valid routes if multiple transit agencies happened to use the same internal route ID. Furthermore, the region bounding box check (used to fail-fast on out-of-bounds searches) would silently disable itself if the GTFS feed omitted the optional `shapes.txt` file, causing unnecessary database load for out-of-bounds queries.

### Changes Made

- **Fix `routeIDs` collision bug:** Updated [routes_for_location_handler.go](file:///Users/mannpatwa/Desktop/maglev/internal/restapi/routes_for_location_handler.go#L105-L113) to use a globally unique identifier `utils.FormCombinedID(routeRow.AgencyID, routeRow.ID)` as the map key. This ensures routes from different agencies with identical string IDs (e.g., both having route "1") are no longer incorrectly flagged as duplicates.
- **Implement fallback for missing shape data:** Modified [ComputeRegionBounds](cci:1://file:///Users/mannpatwa/Desktop/maglev/internal/gtfs/shapes.go:4:0-75:1) in [shapes.go](file:///Users/mannpatwa/Desktop/maglev/internal/gtfs/shapes.go#L6-L9) to accept both `shapes` and [stops](cci:2://file:///Users/mannpatwa/Desktop/maglev/internal/restapi/stops_for_route_handler.go:17:0-20:1). If a GTFS feed does not provide geographic shapes, the system now safely [falls back to computing the region's total bounding box](file:///Users/mannpatwa/Desktop/maglev/internal/gtfs/shapes.go#L42-L68) using the minimum and maximum coordinates of all [stops](cci:2://file:///Users/mannpatwa/Desktop/maglev/internal/restapi/stops_for_route_handler.go:17:0-20:1).
- **Add fallback for empty Sub-route short names:** Added a [null-safety check](file:///Users/mannpatwa/Desktop/maglev/internal/restapi/routes_for_location_handler.go#L110-L113) before route initialization. If `ShortName` is empty, it now falls back to using the `LongName` to populate the `nullSafeShortName` parameter in `models.NewRoute`.
- **Optimize map assignment:** Moved the `agencyIDs` tracking logic [inside the duplicate-check block](file:///Users/mannpatwa/Desktop/maglev/internal/restapi/routes_for_location_handler.go#L108) so we only track agencies for routes we actually append to the results.
- **Update Tests:** Updated [shapes_test.go](file:///Users/mannpatwa/Desktop/maglev/internal/gtfs/shapes_test.go#L109) test cases to accommodate the new function signature.

### Testing
- `make test` runs successfully with no regressions.
- Verified bounding box fallback logic on a real GTFS dataset ([testdata/raba.zip](cci:7://file:///Users/mannpatwa/Desktop/maglev/testdata/raba.zip:0:0-0:0)). An integration test was run on a duplicated version of the zip archive where `shapes.txt` was manually deleted.
    - **Original Output**: `Bounds WITH shapes: lat=40.328705, lon=-122.101745, latSpan=1.218870, lonSpan=0.991490`
    - **Fallback Output**: `Bounds WITHOUT shapes: lat=40.330402, lon=-122.098197, latSpan=1.213889, lonSpan=0.984394`
  
  This confirmed that the missing shape coordinates correctly triggered the fallback logic to find the minimum/maximum bounds using the 375 bus stops, generating a nearly identical (but slightly tighter around the stop-posts) fallback bounding box.

### DIFF
--- a/internal/gtfs/shapes.go
+++ b/internal/gtfs/shapes.go
@@ -12,30 +12,33 @@ func ComputeRegionBounds(shapes []gtfs.Shape, stops []gtfs.Stop) *RegionBounds {
 	var minLat, maxLat, minLon, maxLon float64
 	first := true
 
+	updateBounds := func(lat, lon float64) {
+		if first {
+			minLat = lat
+			maxLat = lat
+			minLon = lon
+			maxLon = lon
+			first = false
+			return
+		}
+		if lat < minLat {
+			minLat = lat
+		}
+		if lat > maxLat {
+			maxLat = lat
+		}
+		if lon < minLon {
+			minLon = lon
+		}
+		if lon > maxLon {
+			maxLon = lon
+		}
+	}
+
 	if len(shapes) > 0 {
 		for _, shape := range shapes {
 			for _, point := range shape.Points {
-				if first {
-					minLat = point.Latitude
-					maxLat = point.Latitude
-					minLon = point.Longitude
-					maxLon = point.Longitude
-					first = false
-					continue
-				}
-
-				if point.Latitude < minLat {
-					minLat = point.Latitude
-				}
-				if point.Latitude > maxLat {
-					maxLat = point.Latitude
-				}
-				if point.Longitude < minLon {
-					minLon = point.Longitude
-				}
-				if point.Longitude > maxLon {
-					maxLon = point.Longitude
-				}
+				updateBounds(point.Latitude, point.Longitude)
 			}
 		}
 	} else {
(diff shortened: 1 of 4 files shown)


mann-patwa:
Hi @aaronbrethorst , can you review this !

mann-patwa:
This PR is ready for review!
I have created an issue for #451 to track the fallback logic for nullSafeShortName.

PR #446, opened 2026-02-21
perf: implement ETag conditional caching for static endpoints

### LINKED ISSUE #445: Optimization: Implement ETag / 304 Not Modified for static endpoints & fix middleware bug
### Summary

Static GTFS endpoints (agencies, routes, stops, shapes) use `Cache-Control: public, max-age=300` but have no conditional request support. Once `max-age` expires, clients are forced to re-download the full JSON payload even if the underlying GTFS dataset has not changed in days. This causes unnecessary bandwidth consumption and redundant JSON serialization on every cache expiry cycle.

While reviewing this, I also found a bug in `caching_middleware.go` that would silently break any ETag implementation added in the future, so both need to be addressed together.

### 1. ETag Support

A SHA-256 `FileHash` is already computed and stored on every GTFS import in `gtfsdb/helpers.go` — this can be used directly as the `ETag` value with no extra runtime hashing cost at all.

* **Proposed change:** Expose `FileHash` on the `Manager` struct after a successful load in `internal/gtfs/static.go`, then add a new middleware for static endpoints that reads the incoming `If-None-Match` header.
* If it matches the current hash, return `304 Not Modified` with an empty body immediately, skipping all database queries and JSON serialization. If it does not match, proceed normally and write `ETag: "<hash>"` on the response.
* **Scope:** This must be scoped to static tier endpoints only. Applying it to real-time endpoints like `arrivals-and-departures-for-stop` or `vehicles-for-agency` would incorrectly serve stale vehicle positions as unchanged, which is a correctness bug.

### 2. Middleware Bug

`cacheControlWriter.WriteHeader` in `caching_middleware.go` currently treats `304` as a non-2xx response and overwrites its `Cache-Control` header with `no-cache, no-store, must-revalidate`. This causes clients to discard their cached copy on every `304` response, completely defeating the purpose of conditional requests. The existing tests in `caching_middleware_test.go` only cover 200 and 404 responses so this has gone undetected.

```go
// current — 304 incorrectly hits the else branch
if code >= 200 && code < 300 {
    w.Header().Set("Cache-Control", w.headerValue)
} else {
    w.Header().Set("Cache-Control", "no-cache, no-store, must-revalidate")
}

// fix
if (code >= 200 && code < 300) || code == http.StatusNotModified {
    w.Header().Set("Cache-Control", w.headerValue)
}

```

ARCoder181105:
### Description

This PR implements **Conditional HTTP Requests (`ETag` / `304 Not Modified`)** for static GTFS endpoints. It allows returning clients to skip downloading multi-megabyte JSON payloads when the underlying GTFS dataset hasn't changed.

While building this, I also fixed a latent bug in the caching middleware that was incorrectly overwriting cache headers on `304` responses.

closes #445 

### Changes Made

* **Zero-cost Hashing:** Added `SystemETag` to the `gtfs.Manager`. It automatically populates using the existing SHA-256 `FileHash` generated during database import, avoiding any runtime hashing overhead.
* **Middleware Bug Fix:** Updated `cacheControlWriter` in `caching_middleware.go` to preserve `Cache-Control` headers for `304 Not Modified` responses.
* **New Middleware:** Added `ETagMiddleware` to intercept `If-None-Match` headers. If the hashes match, it short-circuits the request and returns a `304` with an empty body.
* **Targeted Application:** Created the `etagStatic` helper in `routes.go` and applied it **strictly to static endpoints** (agencies, routes, stops, shapes, etc.). Real-time endpoints were deliberately excluded so vehicle data is never served stale.

### Impact

* **Bandwidth:** Drops payload size from several Megabytes to ~200 bytes for returning clients.
* **CPU:** Bypasses routing, database queries, and JSON serialization entirely on cache hits.
* **Stability:** Ensures clients actually cache `304` responses by fixing the middleware conflict.

### DIFF
--- a/internal/gtfs/gtfs_manager.go
+++ b/internal/gtfs/gtfs_manager.go
@@ -55,6 +55,7 @@ type Manager struct {
 	blockLayoverIndices            map[string][]*BlockLayoverIndex
 	regionBounds                   *RegionBounds
 	isHealthy                      bool
+	systemETag                     string      // systemETag stores the SHA-256 hash of the currently loaded GTFS static dataset.
 	isReady                        atomic.Bool // Tracks whether initial data loading is complete
 
 	feedTrips    map[string][]gtfs.Trip
@@ -103,6 +104,12 @@ func InitGTFSManager(config Config) (*Manager, error) {
 		return nil, fmt.Errorf("error building GTFS database: %w", err)
 	}
 	manager.GtfsDB = gtfsDB
+
+	// Populate systemETag from import metadata
+	metadata, err := gtfsDB.Queries.GetImportMetadata(context.Background())
+	if err == nil && metadata.FileHash != "" {
+		manager.systemETag = fmt.Sprintf(`"%s"`, metadata.FileHash)
+	}
 
 	// Build spatial index for fast stop location queries
 	ctx := context.Background()
@@ -567,6 +574,14 @@ func (manager *Manager) IsServiceActiveOnDate(ctx context.Context, serviceID str
 	default:
 		return 0, nil
 	}
+}
+
+// GetSystemETag retrieves the SystemETag in a thread-safe manner.
+// It acquires the static data read lock to prevent data races during GTFS reloads.
+func (manager *Manager) GetSystemETag() string {
+	manager.staticMutex.RLock()
+	defer manager.staticMutex.RUnlock()
+	return manager.systemETag
 }
 
 // IsHealthy returns true if the GTFS data is loaded and valid.
--- a/internal/gtfs/static.go
+++ b/internal/gtfs/static.go
@@ -314,6 +314,18 @@ func (manager *Manager) ForceUpdate(ctx context.Context) error {
 
 	manager.lastUpdated = time.Now()
 
+	metadata, err := manager.GtfsDB.Queries.GetImportMetadata(ctx)
+	if err != nil {
+		logging.LogError(logger, "Failed to fetch import metadata for ETag generation during hot-swap", err)
+		manager.systemETag = ""
+	} else if metadata.FileHash != "" {
+		manager.systemETag = fmt.Sprintf(`"%s"`, metadata.FileHash)
+		logging.LogOperation(logger, "system_etag_updated_successfully", slog.String("etag", manager.systemETag))
+	} else {
+		logging.LogOperation(logger, "import_metadata_empty_filehash_clearing_etag")
+		manager.systemETag = ""
+	}
+
 	manager.isHealthy = true
 
 	logging.LogOperation(logger, "gtfs_static_data_updated_hot_swap",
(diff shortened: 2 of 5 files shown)


aaronbrethorst (approved):
Aditya -- clean execution on every front. The middleware ordering fix is exactly right, the RFC 7232 compliance is there, the thread-safe `GetSystemETag()` method is a good pattern, and the wildcard handling was a nice touch. Two small optimization suggestions below, but nothing blocking.

---

## Previous Feedback Status

All 4 items from round 1 have been addressed:

| # | Item | Status |
|---|------|--------|
| 1 | ETag middleware bypasses API key validation and rate limiting | **Fixed** -- `etagStatic` now wraps the handler at the innermost level, passed into `rateLimitAndValidateAPIKey` so auth always runs first |
| 2 | 304 response must include ETag header (RFC 7232) | **Fixed** -- `w.Header().Set("ETag", etag)` before `w.WriteHeader(http.StatusNotModified)` |
| 3 | `SystemETag` read without holding read lock | **Fixed** -- `GetSystemETag()` method acquires `staticMutex.RLock()`, called from `etagStatic` |
| 4 | Consider handling `If-None-Match: *` | **Fixed** -- `inm == etag || inm == "*"` with wildcard test case |

---

## Fit and Finish

### 1. `etagStatic` rebuilds the middleware chain on every request (`routes.go:55`)

```go
return func(w http.ResponseWriter, r *http.Request) {
    ETagMiddleware(getETagFunc)(http.HandlerFunc(handler)).ServeHTTP(w, r)
}
```

`ETagMiddleware(getETagFunc)` and the subsequent `(http.HandlerFunc(handler))` call create two new closures per request. Since neither the ETag function nor the handler change between requests, the chain can be built once at registration time:

```go
func etagStatic(api *RestAPI, handler func(http.ResponseWriter, *http.Request)) func(http.ResponseWriter, *http.Request) {
    getETagFunc := func() string {
        if api.GtfsManager != nil {
            return api.GtfsManager.GetSystemETag()
        }
        return ""
    }

    wrapped := ETagMiddleware(getETagFunc)(http.HandlerFunc(handler))
    return wrapped.ServeHTTP
}
```

Simpler, and no per-request allocations. A small win in a PR that's all about performance.

### 2. `SystemETag` is exported but only accessed via `GetSystemETag()` from outside the package

The field is written internally in `ForceUpdate` and `setStaticGTFS`, and read externally via the lock-protected method. Making it lowercase (`systemETag`) would enforce that external callers always go through the thread-safe method, preventing future code from accidentally reading the field directly.

---

## Verification

- `make test` -- all packages pass
- `make lint` -- 0 issues
- Middleware ordering: ETag runs inside auth/rate-limit chain (confirmed via `routes.go` line 88)
- RFC 7232: ETag header set on 304 responses (confirmed via `caching_middleware.go` line 63)
- Thread safety: `GetSystemETag()` uses `staticMutex.RLock()`, writes use `staticMutex.Lock()` (confirmed via `static.go` lines 256, 337)
- Static vs real-time: endpoints correctly categorized -- schedule/stop/route/shape/agency get ETag; vehicles/arrivals/trip-details/current-time do not

## Strengths

- All 4 round 1 items addressed with no regressions.
- Smart reuse of the existing SHA-256 `FileHash` -- zero runtime hashing overhead.
- The 304 cache-control bug fix (`cacheControlWriter` preserving headers for 304) is a genuine catch that was hiding a real problem.
- Test coverage is thorough: match, mismatch, wildcard, empty ETag, and the 304 cache-control preservation.
- Route categorization is thoughtful -- real-time endpoints correctly excluded from ETag.
- `GetSystemETag()` method is the right abstraction for thread-safe access.

LGTM -- merge.

aaronbrethorst:
@ARCoder181105 oops, I just noticed there's a merge conflict. please fix the issue with routes.go and then we'll get this merged.

PR #452, opened 2026-02-22
Fix nullsafe shortname fallback

### LINKED ISSUE #451: nullSafeShortName Fallback Logic
# Inconsistent nullSafeShortName Fallback Logic Across Endpoints
## Description
Currently, there is an inconsistency in how the `nullSafeShortName` (or `shortName`) of a route is populated across different API endpoints when the `short_name` field is missing from the underlying GTFS data. 
Previously, [routes_for_location_handler.go](cci:7://file:///Users/mannpatwa/Desktop/maglev/internal/restapi/routes_for_location_handler.go:0:0-0:0) contained custom fallback logic that would use the route's `long_name` if the `short_name` was empty:
```go
shortName := routeRow.ShortName.String
if shortName == "" {
    shortName = routeRow.LongName.String
}
```
However, every other call site in the codebase — such as trip_handler.go, stop_handler.go, arrivals_and_departure_for_stop_handler.go, trips_for_route_handler.go, and filters.go — simply passes route.ShortName.String directly without any fallback.

This results in the same route returning different shortName values depending on which endpoint is queried. The custom fallback has been temporarily removed from `routes_for_location_handler.go` to enforce immediate consistency, but a broader architectural fix is needed to address missing short names properly.

mann-patwa:
fixes #451 

Updated the signature of `models.NewRoute()` to completely drop the nullSafeShortName parameter (now it correctly expects 9 arguments instead of 10).
Centralized the fallback logic exclusively within the `models.NewRoute` constructor:
```go
nullSafeShortName := shortName
if nullSafeShortName == "" {
    nullSafeShortName = longName
}
```

### DIFF
--- a/internal/gtfs/shapes_test.go
+++ b/internal/gtfs/shapes_test.go
@@ -80,7 +80,7 @@ func TestGetRegionBounds(t *testing.T) {
 			expectedLonSpan: 0.0,
 		},
 		{
-			name: "Fallback to Stops",
+			name:   "Fallback to Stops",
 			shapes: []gtfs.Shape{},
 			stops: []gtfs.Stop{
 				{Latitude: fptr(47.0), Longitude: fptr(-122.0)},
--- a/internal/models/routes.go
+++ b/internal/models/routes.go
@@ -15,7 +15,12 @@ type Route struct {
 	URL               string    `json:"url"`
 }
 
-func NewRoute(id, agencyID, shortName, longName, description string, routeType RouteType, url, color, textColor, nullSafeShortName string) Route {
+func NewRoute(id, agencyID, shortName, longName, description string, routeType RouteType, url, color, textColor string) Route {
+	nullSafeShortName := shortName
+	if nullSafeShortName == "" {
+		nullSafeShortName = longName
+	}
+
 	return Route{
 		AgencyID:          agencyID,
 		Color:             color,
--- a/internal/models/routes_test.go
+++ b/internal/models/routes_test.go
@@ -2,8 +2,9 @@ package models
 
 import (
 	"encoding/json"
-	"github.com/stretchr/testify/assert"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
 )
 
 func TestRouteCreation(t *testing.T) {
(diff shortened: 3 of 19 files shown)


mann-patwa:
This is ready for review @aaronbrethorst !

aaronbrethorst (changes requested):
Hey Mann, this is a well-motivated refactoring — centralizing the `nullSafeShortName` fallback into `NewRoute()` eliminates redundant logic across 17 call sites and makes the constructor the single source of truth. The mechanical changes are all correct (I verified that every caller was passing the same value for both `shortName` and `nullSafeShortName`). Tests pass, lint is clean, gofmt is clean. Before we can merge this, I will need you to make a couple changes:

## Important

### 1. Add a test for the fallback logic (`routes_test.go`)

The entire purpose of this PR is centralizing the `shortName → longName` fallback, but no test actually exercises that path. Every existing test either passes a non-empty `shortName` (so the fallback never triggers) or passes both as empty strings. Add a test case that verifies the behavior:

```go
func TestRouteNullSafeShortNameFallback(t *testing.T) {
    // When shortName is empty, NullSafeShortName should fall back to longName
    route := NewRoute("1", "agency-1", "", "Downtown Express", "", 3, "", "", "")
    assert.Equal(t, "Downtown Express", route.NullSafeShortName)
    assert.Equal(t, "", route.ShortName) // ShortName itself stays empty

    // When shortName is present, NullSafeShortName should use it
    route2 := NewRoute("2", "agency-1", "DX", "Downtown Express", "", 3, "", "", "")
    assert.Equal(t, "DX", route2.NullSafeShortName)
}
```

### 2. Squash into a single commit

The branch has 2 commits:

```
ef49557 cleanup
fcdbdde Refactor: centralize routes short_name fallback logic
```

Please squash these into a single commit.

## Fit and Finish

### 3. Remove unused variable (`routes_test.go:19`)

`nullSafeShortName := "A"` is still declared in `TestRouteCreation` but is no longer passed to `NewRoute()`. It's only used in the assertion on line 34. Since `NewRoute` now derives `NullSafeShortName` internally, you can replace both lines with:

```go
assert.Equal(t, "A", route.NullSafeShortName)
```

### 4. Unrelated whitespace changes in `stops_for_location_handler.go`

The two changes in this file (trailing whitespace on `return` statements, lines 186 and 222) are unrelated to the nullSafeShortName refactoring. They're harmless, but for future PRs, keeping unrelated fixes separate makes review easier.

## Strengths

- **Correct mechanical change**: I verified all 17 call sites were passing the same value for both `shortName` and `nullSafeShortName`, so the consolidation is safe.
- **Net deletion**: 40 additions vs 56 deletions — the codebase gets simpler.
- **Single responsibility**: The constructor is now the only place that knows about the fallback, which is exactly where it belongs.

PR #457, opened 2026-02-22
Fix data race in routesForAgencyHandler by adding RLock

### LINKED ISSUE #455: bug: routesForAgencyHandler missing RLock before accessing GTFS manager
## Description

`routesForAgencyHandler` in `internal/restapi/routes_for_agency_handler.go` calls `FindAgency()` and `RoutesForAgencyID()` without first acquiring `manager.RLock()`. Both methods document the contract: "IMPORTANT: Caller must hold manager.RLock() before calling this method."

This is a data race during `ForceUpdate` (which acquires a write lock and replaces the underlying maps). Other handlers like `vehiclesForAgencyHandler` correctly acquire the read lock.

## Steps to Reproduce

Run with `-race` flag under load while a `ForceUpdate` is in progress.

## Fix

Add `RLock()/defer RUnlock()` to the handler, matching the pattern in `vehiclesForAgencyHandler`:

```go
func (api *RestAPI) routesForAgencyHandler(w http.ResponseWriter, r *http.Request) {
    id := utils.ExtractIDFromParams(r)

    if err := utils.ValidateID(id); err != nil {
        // ...validation error handling...
        return
    }

    api.GtfsManager.RLock()
    defer api.GtfsManager.RUnlock()

    agency := api.GtfsManager.FindAgency(id)
    // ... rest of handler
```

Found during review of #372.

tejasva-vardhan:
Fixes #455

Fix data race in routesForAgencyHandler by adding proper RLock usage.

The handler was calling FindAgency() and RoutesForAgencyID() without acquiring GtfsManager.RLock(), violating the documented contract.

Changes:
- Added utils.ValidateID(id) for input validation
- Added RLock()/RUnlock() before accessing GtfsManager
- Followed the locking pattern used in vehiclesForAgencyHandler

This ensures thread safety during ForceUpdate without altering existing business logic.

### DIFF
--- a/internal/restapi/routes_for_agency_handler.go
+++ b/internal/restapi/routes_for_agency_handler.go
@@ -10,6 +10,9 @@ import (
 func (api *RestAPI) routesForAgencyHandler(w http.ResponseWriter, r *http.Request) {
 	id, _ := utils.GetIDFromContext(r.Context())
 
+	api.GtfsManager.RLock()
+	defer api.GtfsManager.RUnlock()
+
 	agency := api.GtfsManager.FindAgency(id)
 
 	if agency == nil {


aaronbrethorst (changes requested):
Hey Tejasva, good catch on the missing read lock -- `FindAgency()` and `RoutesForAgencyID()` both document "IMPORTANT: Caller must hold manager.RLock() before calling this method", and this handler wasn't complying. The fix follows the exact pattern from `vehiclesForAgencyHandler` and is correct. One small thing before we merge:

## Fit and Finish

### 1. Remove the redundant `ValidateID` call (`routes_for_agency_handler.go:12-15`)

This handler is registered via `withID()` in `routes.go:81`, which wraps it with `ValidateIDMiddleware`. That middleware already calls `utils.ValidateID(id)` and returns a 400 if validation fails -- the handler only runs with a pre-validated ID from `GetIDFromContext`. The `ValidateID` call inside the handler is harmless but redundant, and no other handler in the codebase does this (including `vehiclesForAgencyHandler`, which is the pattern you referenced):

```go
func (api *RestAPI) routesForAgencyHandler(w http.ResponseWriter, r *http.Request) {
	id, _ := utils.GetIDFromContext(r.Context())

	api.GtfsManager.RLock()
	defer api.GtfsManager.RUnlock()

	agency := api.GtfsManager.FindAgency(id)
```

Just remove the `ValidateID` block (lines 12-15 in the diff) to keep the handler consistent with the rest of the codebase.

## Strengths

- **Correct diagnosis**: The data race documented in #455 is real, and the fix is exactly right -- acquire `RLock()` before accessing in-memory GTFS data, release via `defer`.
- **Clean single commit**: One logical change, one commit, clear commit message.
- **CI green**: Lint and tests pass on both platforms.

## Recommended Action

Remove the redundant `ValidateID` block, amend the commit, and this is ready to merge.

tejasva-vardhan:
Thanks for the review! Removed the redundant ValidateID call since it's already handled by the middleware. Updated the PR.

PR #459, opened 2026-02-23
Fix N+1 query issue in vehiclesForAgencyHandler

### LINKED ISSUE #450: Optimize vehicles-for-agency handler by batching route lookups
Problem:
The vehicles-for-agency handler performs a route lookup (GetRoute) inside a loop for each vehicle, which can lead to N+1 query issues and degraded performance with larger datasets.

Location:
internal/restapi/vehicles_for_agency_handler.go

Suggested Fix:
- Collect unique route IDs from vehicles
- Fetch routes in a single call using GetRoutesByIDs
- Build a map of routeID → route
- Use this map inside the loop instead of repeated DB calls

This pattern is already used in other handlers like trips_for_route_handler.go.

tejasva-vardhan:
Fixes #450

Fix N+1 query issue in vehiclesForAgencyHandler.

The handler previously called GetRoute inside a loop for each vehicle, resulting in multiple database queries.

Changes:
 1.Collected unique route IDs from vehicles
 2.Fetched routes in a single call using GetRoutesByIDs
 3.Built a map of routeID → route
 4.Replaced per-vehicle GetRoute calls with map lookups

This follows the same pattern used in trips_for_route_handler.go and trips_for_location_handler.go.

This reduces database calls from O(N) to O(1), improving performance for larger datasets.

Note: Tests may fail locally if CGO is disabled (sqlite3 requires CGO). This is an environment issue, not related to this change.

### DIFF
--- a/internal/restapi/vehicles_for_agency_handler.go
+++ b/internal/restapi/vehicles_for_agency_handler.go
@@ -4,6 +4,7 @@ import (
 	"net/http"
 	"time"
 
+	"maglev.onebusaway.org/gtfsdb"
 	"maglev.onebusaway.org/internal/models"
 	"maglev.onebusaway.org/internal/utils"
 )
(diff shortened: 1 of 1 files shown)


aaronbrethorst (approved):
Tejasva, solid optimization — replacing per-vehicle `GetRoute` calls with a single `GetRoutesByIDs` batch fetch is the right fix for the N+1 query issue. The pattern matches what's used in `trips_for_route_handler.go` and `trips_for_location_handler.go`, and the empty-slice case is handled safely by sqlc's generated code. Tests pass, lint is clean. Merging.

---

## Strengths

- **Correct N+1 elimination**: Collects unique route IDs, batch-fetches once, then uses map lookups in the loop — textbook fix.
- **Safe error handling**: The new `GetRoutesByIDs` error path returns a proper 500 instead of silently skipping all routes, which is actually better than the original behavior (where a DB connection failure in the loop would silently produce responses with missing route references).
- **Follows established patterns**: Same deduplication-then-batch approach used elsewhere in the codebase.

PR #462, opened 2026-02-23
Verify trip status

Ahmedhossamdev:
Fix: #108
Replaces: #408

### DIFF
--- a/internal/restapi/trips_helper_test.go
+++ b/internal/restapi/trips_helper_test.go
@@ -15,35 +15,6 @@ import (
 	"maglev.onebusaway.org/internal/utils"
 )
 
-// TestFindNextStop tests the findNextStop helper for correct next-stop lookup and edge cases.
-func TestFindNextStop(t *testing.T) {
-	seq2 := uint32(2)
-	seq3 := uint32(3)
-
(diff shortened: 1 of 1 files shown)


Ahmedhossamdev:
> BuildTripStatus uses mismatched trip IDs for stop times vs shape queries. GetStopTimesForTrip on line 63 uses activeTripRawID (which may differ from the requested trip if the vehicle was reassigned within a block), but GetShapePointsByTripID on line 121 uses the original tripID. If the vehicle is on a different trip within the block, the shape data and stop times will be for different trips, producing incorrect position/distance calculations. Please use a consistent trip ID for both, and add a comment explaining the choice.


For this I think for this we should use the activeTripId because it's changing based of the GTFS-RT data.

Ahmedhossamdev:
@aaronbrethorst Done

aaronbrethorst:
@Ahmedhossamdev oops, I just noticed that there is a merge conflict! Get it fixed and we'll get this merged :)

Ahmedhossamdev:
> @Ahmedhossamdev oops, I just noticed that there is a merge conflict! Get it fixed and we'll get this merged :)

Cool

Issue #448, opened 2026-02-22
Support multiple GTFS-RT feeds from different agencies


aaronbrethorst:
## Summary

Maglev currently supports a single GTFS-RT feed configuration (one set of trip updates, vehicle positions, and alerts URLs). The existing Java OBA server supports multiple GTFS-RT feeds from different agencies against a single static GTFS bundle. Maglev should support this too.

## Problem

Many transit regions have a single consolidated static GTFS bundle but multiple agencies each publishing their own GTFS-RT feeds. For example, a regional deployment might have:

- Agency A publishing trip updates + vehicle positions at `https://a.example.com/*.pb`
- Agency B publishing trip updates + vehicle positions + alerts at `https://b.example.com/*.pb`
- Agency C publishing only vehicle positions at `https://c.example.com/*.pb`

Today, Maglev can only poll one set of GTFS-RT URLs, so it can't serve real-time data for multi-agency regions.

## Background: How the Java version works

The Java implementation (`GtfsRealtimeSource.java`, ~1,375 lines) creates one poller instance per feed. Each instance:

1. Polls its three feed URLs (trip updates, vehicle positions, alerts) on a configurable interval (default 30s)
2. Parses the protobuf responses
3. Groups trip updates + vehicle positions by trip
4. Pushes `VehicleLocationRecord`s to a downstream listener
5. Manages alerts (create/update/delete)
6. Expires stale vehicles not seen for 15 minutes

Multiple feeds are supported by instantiating multiple poller beans via Spring XML config — each with its own URLs, agency IDs, refresh interval, and HTTP headers.

### Per-feed configuration options in Java

| Property | Description |
|---|---|
| `tripUpdatesUrl` | HTTP endpoint for trip updates |
| `vehiclePositionsUrl` | HTTP endpoint for vehicle positions |
| `alertsUrl` | HTTP endpoint for service alerts |
| `agencyId` / `agencyIds` | Which transit agency this feed belongs to |
| `refreshInterval` | Polling period in seconds (default: 30) |
| `headersMap` | Custom HTTP headers (e.g. API keys) |
| `enabled` | On/off switch |

### Config changes

Extend `gtfs-rt-feeds` in the JSON config to support per-feed configuration. Each feed entry gets its own agency ID(s), refresh interval, HTTP headers, and enabled flag:

```json
{
  "gtfs-rt-feeds": [
    {
      "id": "puget-sound",
      "agency-ids": ["1"],
[...]

PR #465, opened 2026-02-24
Adds support for multiple concurrent GTFS-RT feeds from different agencies against a single static GTFS bundle matching the capability of the Java OBA server.

Ahmedhossamdev:
Fix: #448 

### Key Changes

**Multi-Feed Architecture**
- Each GTFS-RT feed now runs its own independent polling goroutine
- Real-time data (vehicles, trips, alerts) stored in per-feed sub-maps to prevent data clobbering
- Feed refresh only updates its own partition other feeds' data remains intact
- Errors are isolated: one feed failing doesn't affect others

**Stale Vehicle Expiry**
- Implements per-feed stale vehicle tracking (15 min window)
- Vehicles are only expired when not seen by their specific feed
- Fixes a bug where refreshing one feed could prematurely expire vehicles from another

**Config & Documentation**
- Consolidates all HTTP headers under the `headers` field (removes legacy `realtime-auth-header-*`)
- Updates config schema with new per-feed options (`id`, `agency-ids`, `headers`, `refresh-interval`, `enabled`)
- Updates README and CLAUDE.md with new config examples

### Testing
- Adds tests for multi-feed merging and data isolation
- Adds tests for per-feed stale vehicle expiry
- All existing tests pass

### DIFF
--- a/internal/gtfs/parallel_realtime_test.go
+++ b/internal/gtfs/parallel_realtime_test.go
@@ -191,13 +191,9 @@ func TestRealTimeDataConsistency(t *testing.T) {
 
 	wg.Wait()
 
-	// Should not panic or cause data corruption
-	trips := manager.GetRealTimeTrips()
-	vehicles := manager.GetRealTimeVehicles()
-
-	// Slices may be nil or empty due to invalid URLs, but function should not panic
-	assert.True(t, trips != nil || len(trips) == 0, "Trips should be accessible without panic")
-	assert.True(t, vehicles != nil || len(vehicles) == 0, "Vehicles should be accessible without panic")
+	// Should not panic or cause data corruption; invalid URLs produce no data.
+	assert.Empty(t, manager.GetRealTimeTrips(), "no trips should be stored when all URLs are unreachable")
+	assert.Empty(t, manager.GetRealTimeVehicles(), "no vehicles should be stored when all URLs are unreachable")
 }
 
 // newTestManager creates a minimal Manager for tests that only exercise realtime code
--- a/internal/gtfs/realtime_test.go
+++ b/internal/gtfs/realtime_test.go
@@ -210,3 +210,79 @@ func TestLoadRealtimeData_Non200StatusCode(t *testing.T) {
 		})
 	}
 }
+
+func TestEnabledFeeds(t *testing.T) {
+	tests := []struct {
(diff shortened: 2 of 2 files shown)


internal/gtfs/realtime.go
```diff
@@ -191,102 +155,259 @@ func (manager *Manager) GetAlertsForStop(stopID string) []gtfs.Alert {
...
-	// Wait for both to complete
 	wg.Wait()
 
 	// Check for context cancellation
 	if ctx.Err() != nil {
 		return
 	}
 
-	// Update data if at least one fetch succeeded
 	manager.realTimeMutex.Lock()
```
mann-patwa: We can acquire the lock later, just before calling `rebuildMergedRealtimeLocked`. At this point, we’re only writing to the individual feeds, so there’s no need to lock the aggregated feeds yet.

aaronbrethorst (approved):
Hey Ahmed, you've done a thorough job addressing all fifteen items from the first review. I verified each one against the current diff:

- Dead standalone functions removed, tests updated to use `rebuildMergedRealtimeLocked` (item 1)
- Duplicate feed ID validation returns an error from `ToGtfsConfigData()` (item 2)
- `agency-ids` documented as reserved for future use (item 3)
- Logging now distinguishes success, all-fail-with-stale-data (Warn), and all-fail-no-data (Error) (item 4)
- Stale comment updated to reference `pollFeed` (item 5)
- `SetRealTimeTripsForTest` stores under `"_test"` feed ID and rebuilds merged view (item 6)
- Empty-string guard added to trip lookup in `rebuildMergedRealtimeLocked` (item 7)
- Feed IDs sorted before iteration for deterministic merge order (item 8)
- Warning logged for incomplete legacy auth header pair (item 9)
- Stale vehicle test exercises the real `updateFeedRealtime` code path (item 10)
- Dead test code removed, tautological assertions replaced, distinct test data used across feeds, concurrent update test added, `enabledFeeds()` test added (items 11-15)

The per-feed lock concern raised by @mann-patwa is handled by the existing `realTimeMutex` — all feed map writes and merged-view rebuilds happen under the same lock, so concurrent feed goroutines serialize correctly.

There's nothing left to change — this PR is ready to merge. Well-architected multi-feed support with clean data isolation and solid test coverage.

PR #486, opened 2026-02-27
fix(gtfs-rt): prevent stale vehicle feeds from overwriting newer data

### LINKED ISSUE #476: GTFS-Realtime: stale vehicle positions can overwrite newer data (no freshness guard)
### Summary
Maglev currently allows older (stale) GTFS-Realtime vehicle position data
to overwrite newer vehicle data. There is no freshness or timestamp guard
when updating `realTimeVehicles`.

This behavior is reproducible and documented by a unit test.

### Why this matters
In real-world GTFS-RT deployments, feeds can arrive out of order due to:
- Network retries or delays
- CDN caching
- Feed restarts
- Multiple upstream data sources

Allowing stale vehicle data to overwrite newer data can cause vehicles to
jump backwards, disappear, or show incorrect positions in client apps.

### Current behavior
`updateGTFSRealtime()` replaces `manager.realTimeVehicles` whenever a fetch
succeeds, without checking timestamps or data freshness.

### Expected behavior (needs maintainer guidance)
Possible options include:
- Ignore vehicle updates older than the latest seen timestamp
- Track per-vehicle freshness and reject stale updates
- Log stale updates without applying them
- Make freshness behavior configurable

### Reproduction
A unit test demonstrates the current behavior:
1. Newer realtime vehicle data is applied
2. Older (stale) vehicle data is applied afterward
3. The older data overwrites the newer data

### Environment
- OS: Windows 10
- Go: latest
- Maglev: main branch
- Tests run with `-tags sqlite_fts5`

I’m happy to submit a PR once desired behavior is confirmed.

AtharvaKathe18:
### Summary
Prevents stale GTFS-Realtime vehicle position feeds from overwriting newer data by introducing a per-feed freshness guard based on the GTFS-RT FeedMessage timestamp.

This fixes an issue where out-of-order, retried, or delayed realtime feeds could cause vehicles to jump backwards, disappear, or show incorrect positions.

### What changed
- Track last-applied GTFS-RT vehicle feed timestamp per feed
- Reject vehicle updates where feed timestamp ≤ last applied timestamp
- Still run vehicle expiry cleanup even when stale updates are rejected
- Ensure expired vehicles are removed from both lastSeen tracking and feedVehicles
- No behavior change for trip updates or service alerts

### Why this matters
GTFS-Realtime feeds can arrive out of order due to retries, caching, or upstream restarts. The GTFS-RT spec expects consumers to handle this safely. This change aligns Maglev with that expectation.

### Tests
- All existing tests pass
- Added/updated tests covering stale vehicle overwrite behavior and expiry
- Verified with:  
  `go test -tags sqlite_fts5 ./internal/gtfs`

### Related issue
Fixes #476

### DIFF
--- a/internal/gtfs/realtime_test.go
+++ b/internal/gtfs/realtime_test.go
@@ -295,9 +295,12 @@ func TestEnabledFeeds(t *testing.T) {
 // are rejected and vehicles from the newer feed are preserved. This tests
 // the feed-level freshness guard that prevents out-of-order feed updates.
 func TestStaleFeedRejected(t *testing.T) {
+	// Read the test data before creating the server to ensure proper error handling
+	data, err := os.ReadFile(filepath.Join("../../testdata", "raba-vehicle-positions.pb"))
+	require.NoError(t, err, "failed to read RABA vehicle positions test data")
+
 	// Create a test server that serves the same RABA vehicle data
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		data, _ := os.ReadFile(filepath.Join("../../testdata", "raba-vehicle-positions.pb"))
 		w.Header().Set("Content-Type", "application/x-protobuf")
 		_, _ = w.Write(data)
 	}))
@@ -318,6 +321,14 @@ func TestStaleFeedRejected(t *testing.T) {
 	firstPoll := manager.GetRealTimeVehicles()
 	require.NotEmpty(t, firstPoll, "first poll should load vehicles")
 	firstCount := len(firstPoll)
+
+	// Verify the feed has a FeedHeader timestamp — this test
+	// exercises the freshness guard, which only applies to feeds
+	// with a non-zero CreatedAt.
+	manager.realTimeMutex.RLock()
+	require.NotZero(t, manager.feedVehicleTimestamp[feed.ID],
+		"RABA feed must have FeedHeader timestamp for this test to be meaningful")
+	manager.realTimeMutex.RUnlock()
 
 	// Simulate a stale feed by manually setting the stored timestamp to a very
 	// large value (future timestamp), so the next update will be rejected.


aaronbrethorst (changes requested):
Hey Atharva -- Solid work addressing all five items from the first round. The critical `uint64` overflow is properly guarded with `IsZero()`, the `cleanupExpiredVehicles` extraction is clean, `isVehicleStale` has thorough unit tests, and `TestStaleVehicleExpiry` no longer depends on `time.Sleep`. Just a couple of items to tighten up the test coverage before this is ready to merge.

---

## Prior Feedback Status

All 5 items from the first review have been addressed:

| # | Item | Status |
|---|------|--------|
| 1 | `uint64` overflow permanently blocks vehicle updates for feeds without FeedHeader timestamp | **Fixed** -- `IsZero()` guard on `vehicleData.CreatedAt` bypasses freshness comparison for feeds without a FeedHeader timestamp |
| 2 | `TestRealtimeAcceptsStaleVehicleData` doesn't test new behavior | **Fixed** -- replaced with `TestStaleFeedRejected` that exercises `updateFeedRealtime` with a manually-set future timestamp |
| 3 | Add unit tests for `isVehicleStale` | **Fixed** -- `TestIsVehicleStale` covers all 6 cases (both present: older/newer/equal, either nil, both nil) |
| 4 | `time.Sleep(100ms)` fragile in `TestStaleVehicleExpiry` | **Fixed** -- test rewritten to call `cleanupExpiredVehicles` directly, no sleep needed |
| 5 | Extract duplicate cleanup logic | **Fixed** -- `cleanupExpiredVehicles` extracted and used in the stale-feed branch |

---

## Important

### 1. `TestStaleFeedRejected` may pass vacuously if RABA data lacks a FeedHeader timestamp (`realtime_test.go:316-330`)

The test relies on the freshness guard rejecting a stale feed. But if the RABA protobuf data has no FeedHeader timestamp (i.e., `CreatedAt.IsZero()` is true), the guard is bypassed and **every update is always applied** — the test would still pass because the same data produces the same vehicle count, but it wouldn't actually exercise the rejection path.

The RABA data almost certainly has a FeedHeader timestamp since it's real production data, but the test should verify this prerequisite. Add an assertion after the first poll:

```go
// Verify the feed has a FeedHeader timestamp — this test
// exercises the freshness guard, which only applies to feeds
// with a non-zero CreatedAt.
manager.realTimeMutex.RLock()
require.NotZero(t, manager.feedVehicleTimestamp[feed.ID],
    "RABA feed must have FeedHeader timestamp for this test to be meaningful")
manager.realTimeMutex.RUnlock()
```

This way, if the test data is ever replaced with data that lacks a timestamp, the test will fail loudly rather than passing silently.

### 2. `os.ReadFile` error silently discarded in `TestStaleFeedRejected` (`realtime_test.go:300`)

```go
data, _ := os.ReadFile(filepath.Join("../../testdata", "raba-vehicle-positions.pb"))
```

Compare with `TestStaleVehicleExpiry` in `multi_feed_test.go:88-89`:

```go
data, err := os.ReadFile(filepath.Join("../../testdata", "raba-vehicle-positions.pb"))
require.NoError(t, err)
```

If the file is missing, the handler silently serves an empty body, and the test would fail at `require.NotEmpty(t, firstPoll)` with a confusing error message. Use `require.NoError` for consistency and clearer failure messages. Note: the `t` variable from the test won't be available inside the HTTP handler closure, so you'll need to capture it or use a different approach (e.g., read the file before creating the server).

---

## Fit and Finish

### 3. Consider `time.Time` instead of `uint64` for timestamp storage

The `feedVehicleTimestamp` map still uses `uint64` (`gtfs_manager.go:65`). The `IsZero()` guard prevents the overflow bug, but `time.Time` would be more idiomatic and eliminate the class of problems entirely:

```go
feedVehicleTimestamp map[string]time.Time
```

Then the comparison becomes:
```go
if !vehicleData.CreatedAt.IsZero() && !vehicleData.CreatedAt.After(manager.feedVehicleTimestamp[feedID]) {
    // stale
}
```

Not blocking — the current implementation is correct. Just something to consider for clarity.

---

## Verification

- Tests: all pass
- Lint: 0 issues
- Single commit: clean

## Strengths

- **Clean extraction of `cleanupExpiredVehicles`**: The helper is well-scoped, properly documented, and handles the nil-map edge case.
- **Thorough `isVehicleStale` tests**: All 6 cases covered, including the nil-timestamp edge cases.
- **`TestStaleVehicleExpiry` rewrite**: Deterministic and directly tests the cleanup function without flaky timing.
- **`IsZero()` guard is exactly right**: The simplest correct fix for the overflow bug — no over-engineering.

---

The two important items are straightforward test hardening fixes. Once those are addressed, this is good to merge.

Issue #491, opened 2026-02-28
Fix: Rate Limiter Retry-After Time Always Rounds Down to 0


Jason10293:
### Description:
In `rate_limit_middleware`, the `retryAfter `duration for finite rate limits is computed via `time.Duration(rl.rateLimit`). Converting `rl.rateLimit` (a float64 req/sec number) to `time.Duration` truncates it to an integer nanosecond count, and the subsequent division would then uses integer division. For any rate above 1 req/s, this produces 0, so the `Retry-After` header is always zero seconds, making it useless.

## Current Behaviour:
When extracting the value from the header after hitting an error `429: too many requests`, we can get that the `Retry-After` value is 0 for values >0. 
``` Go
var retryAfter time.Duration
switch rl.rateLimit {
// ... rate limit = inf and 0 case ...
default:
	retryAfter = time.Duration(1) / time.Duration(rl.rateLimit) // Results in 0!
}

// ... Setting headers ...
w.Header().Set("Retry-After", strconv.Itoa(int(retryAfter.Seconds())))
// ... Setting headers ...
```
<img width="448" height="93" alt="Image" src="https://github.com/user-attachments/assets/578c61f8-86fc-461a-9d27-edafb2b3f7c5" />

## Expected Behavior:
Instead of returning 0, the api header should return an integer value of 1 sec (or any other value that makes sense).

## Proposed Fix:
Update the calculation to be `retryAfter = time.Duration(float64(time.Second) / float64(rl.rateLimit))` This will perform a floating point calculation on the retry after time instead. From there set the header to be the `math.Ceil` of the retry after time (In other words, 1 for rate limits $`\geq1`$), as this value must be an integer.

``` Go
var retryAfter time.Duration
switch rl.rateLimit {
// ... #rate limit = inf and 0 case ...
default:
	retryAfter = time.Duration(float64(time.Second) / float64(rl.rateLimit)) // Use floating point division instead
}

// ... Setting headers ...
w.Header().Set("Retry-After", strconv.Itoa(int(math.Ceil(retryAfter.Seconds())))) // Take the math.Ceil of the retry-after value 
// ... Setting headers ...
```

## Note:
The "default" retry after values doesn't necessarily have to be 1 sec, could be any value honestly. I just chose 1 sec for the sake of example.

PR #492, opened 2026-02-28
fix: correct Retry-After header calculation in rate limiter

Jason10293:
This change aims to fix the bug detailed in #491

## Summary:

- Bugfix: `sendRateLimitExceeded `now computes `retryAfter `using floating-point math instead of integer division, so sub-second rate limits no longer collapse to zero.
- RFC compliance: `Retry-After` headers now round up to the nearest whole second via `math.Ceil(retryAfter.Seconds())`, ensuring clients always receive a meaningful integer delay.
- Tests: `TestRateLimitMiddleware_HitRateLimit` updated to assert the corrected `Retry-After` values for both low and high finite rate limits.

### DIFF
--- a/internal/restapi/rate_limit_middleware.go
+++ b/internal/restapi/rate_limit_middleware.go
@@ -3,6 +3,7 @@ package restapi
 import (
 	"encoding/json"
 	"log/slog"
+	"math"
 	"net/http"
 	"strconv"
 	"strings"
@@ -152,12 +153,12 @@ func (rl *RateLimitMiddleware) sendRateLimitExceeded(w http.ResponseWriter, r *h
 	case rate.Inf:
 		retryAfter = time.Second // Should not happen, but fallback
 	default:
-		retryAfter = time.Duration(1) / time.Duration(rl.rateLimit)
+		retryAfter = time.Duration(float64(time.Second) / float64(rl.rateLimit))
 	}
 
 	// Set headers
 	w.Header().Set("Content-Type", "application/json")
-	w.Header().Set("Retry-After", strconv.Itoa(int(retryAfter.Seconds())))
+	w.Header().Set("Retry-After", strconv.Itoa(int(math.Ceil(retryAfter.Seconds()))))
 	w.Header().Set("X-RateLimit-Limit", strconv.Itoa(rl.burstSize))
 	w.Header().Set("X-RateLimit-Remaining", "0")
 	w.WriteHeader(http.StatusTooManyRequests)
--- a/internal/restapi/rate_limit_middleware_test.go
+++ b/internal/restapi/rate_limit_middleware_test.go
@@ -3,8 +3,10 @@ package restapi
 import (
 	"encoding/json"
 	"fmt"
+	"math"
 	"net/http"
 	"net/http/httptest"
+	"strconv"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


aaronbrethorst (changes requested):
Hey Jason -- nice catch on the integer division bug. The analysis in #491 was thorough and the fix is correct. A few things to address before we merge.

## Critical

No critical issues.

## Important

### 1. Duplicate test case names

`rate_limit_middleware_test.go:447-448` — two entries both named `"low rate limit"`:

```go
{name: "low rate limit", rateLimit: 2},
{name: "low rate limit", rateLimit: 20},
```

Go appends `#01` to the duplicate so it doesn't break, but when a test fails you can't tell which case it was without checking suffixes. Please give each a unique name, e.g. `"rate limit 2"`, `"rate limit 20"`, `"rate limit 100"`, `"rate limit 200"`.

### 2. All test cases assert the same expected value (`1`)

Because `math.Ceil(1.0 / N)` = `1` for all N >= 1, every test case expects `Retry-After: 1`. This means the test would also pass if the implementation hardcoded `Retry-After: 1`. It doesn't prove the calculation actually works.

Please add at least one case where the expected value is *not* 1. For example, a sub-second rate with a longer interval:

```go
// In a separate test or by extending the table to accept an interval parameter:
middleware := initRateLimitMiddleware(1, 2*time.Second)
// rate.Every(2s / 1) = 0.5 events/sec
// Retry-After = ceil(1/0.5) = ceil(2.0) = 2
```

This proves the math actually flows through correctly rather than always landing on the same answer.

### 3. Add `rateLimit: 1` to the test table

This is the boundary where someone might assume the old code worked. It didn't — `time.Duration(1) / time.Duration(1.0)` = 1 nanosecond, which rounds to 0 seconds. Including it explicitly documents that the fix covers the boundary.

## Fit and Finish

### 4. Formatting: blank line and comment spacing

`rate_limit_middleware_test.go:440-442`:

```go
	})

}
func TestRateLimitMiddleware_CorrectRetryAfterTime(t *testing.T) {
	//Testing finite rate limits
```

Two small style issues:
- There's an extra blank line inside the closing of `TestRateLimitMiddleware_EdgeCases`, and then no blank line between the `}` and the new function. The rest of the file consistently uses a single blank line between top-level functions.
- `//Testing` needs a space after `//` per Go convention: `// Testing finite rate limits`.

Should be:

```go
	})
}

func TestRateLimitMiddleware_CorrectRetryAfterTime(t *testing.T) {
	// Testing finite rate limits
```

### 5. Use `strconv.Atoi` instead of `strconv.ParseFloat`

`rate_limit_middleware_test.go:477`:

```go
retryAfter, err := strconv.ParseFloat(retryAfterStr, 64)
```

The header is set via `strconv.Itoa()`, so it's always an integer string. Using `strconv.Atoi()` is more precise and would catch a regression where the header accidentally includes a decimal.

### 6. Use a concrete API key in the test

`rate_limit_middleware_test.go:466`:

```go
req := httptest.NewRequest(http.MethodGet, "/test?key=", nil)
```

Empty `key=` routes through the `__no_key__` sentinel path, which isn't what this test is about. Use `?key=test-key` to keep the test focused on the Retry-After calculation.

## Verdict

the fix itself is correct and well-motivated. Items 1-3 above need to be addressed (unique test names, at least one test case that expects a value other than 1, and the boundary case). Items 4-6 are quick cleanup.

PR #507, opened 2026-02-28
fix: add panic recovery middleware for HTTP handlers

### LINKED ISSUE #494: Add panic recovery middleware for HTTP handlers
## Problem

There is exactly one `recover()` in production code — in the metrics collector goroutine (`internal/metrics/metrics.go:137`). HTTP request handler goroutines have **no panic recovery**. An unhandled panic in any handler will crash the entire server process, not just fail the individual request.

There are also three `panic()` calls in `internal/gtfs/advanced_direction_calculator.go` (lines 46, 60, 74) for configuration errors that could be triggered by a misconfigured caller at runtime.

## Impact

A single bad request or edge case that triggers a panic kills the process for all users.

## Fix

Add a recovery middleware as the outermost layer of the middleware chain that:
1. Catches panics
2. Logs the stack trace
3. Returns a 500 response to the client
4. Keeps the server running

## Location

- Middleware chain is configured in `cmd/api/app.go` (`CreateServer`)
- Existing middleware registered in `internal/restapi/routes.go`

## Priority

**Ship blocker** — small effort, critical protection.

tejasva-vardhan:
This PR addresses the issue of missing panic recovery in HTTP handlers.

As discussed in the issue, a panic in any handler currently crashes the entire server instead of failing just that request. This introduces a recovery middleware to handle such cases safely.

The middleware:
 catches panics from handlers
 logs the panic along with the stack trace
 returns a 500 response (only if a response has not already been written)
 ensures the server continues running

It is added as the outermost middleware so that all requests are protected.

This change does not affect normal request handling, but improves reliability under unexpected failures.

Fixes #494

### DIFF
--- a/internal/restapi/recovery_middleware.go
+++ b/internal/restapi/recovery_middleware.go
@@ -23,6 +23,13 @@ func (rw *recoveryResponseWriter) WriteHeader(code int) {
 	rw.ResponseWriter.WriteHeader(code)
 }
 
+func (rw *recoveryResponseWriter) Write(b []byte) (int, error) {
+	if !rw.wroteHeader {
+		rw.wroteHeader = true
+	}
+	return rw.ResponseWriter.Write(b)
+}
+
 // NewRecoveryMiddleware returns middleware that recovers from panics in handlers,
 // logs the panic with stack trace, and returns HTTP 500 (JSON) if no response was sent.
 func NewRecoveryMiddleware(logger *slog.Logger, c clock.Clock) func(http.Handler) http.Handler {
@@ -32,13 +39,18 @@ func NewRecoveryMiddleware(logger *slog.Logger, c clock.Clock) func(http.Handler
 			defer func() {
 				if rec := recover(); rec != nil {
 					stack := debug.Stack()
+					reqID, _ := r.Context().Value(RequestIDKey).(string)
 					var err error
 					if e, ok := rec.(error); ok {
 						err = e
 					} else {
 						err = fmt.Errorf("%v", rec)
 					}
-					logging.LogError(logger, "handler panic recovered", err, slog.String("path", r.URL.Path), slog.String("stack", string(stack)))
+					logging.LogError(logger, "handler panic recovered", err,
+						slog.String("path", r.URL.Path),
+						slog.String("method", r.Method),
+						slog.String("request_id", reqID),
+						slog.String("stack", string(stack)))
 					if !rw.wroteHeader {
 						w.Header().Set("Content-Type", "application/json")
 						w.WriteHeader(http.StatusInternalServerError)
@@ -53,7 +65,12 @@ func NewRecoveryMiddleware(logger *slog.Logger, c clock.Clock) func(http.Handler
 							Text:        "internal server error",
 							Version:     1,
 						}
-						_ = json.NewEncoder(w).Encode(response)
+						if err := json.NewEncoder(w).Encode(response); err != nil {
+							logging.LogError(logger, "failed to encode panic recovery response", err,
+								slog.String("path", r.URL.Path),
+								slog.String("method", r.Method),
+								slog.String("request_id", reqID))
+						}
 					}
 				}
 			}()
--- a/internal/restapi/recovery_middleware_test.go
+++ b/internal/restapi/recovery_middleware_test.go
@@ -2,6 +2,7 @@ package restapi
 
 import (
 	"encoding/json"
+	"fmt"
 	"io"
 	"log/slog"
 	"net/http"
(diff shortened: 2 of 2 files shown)


ARCoder181105:
@tejasva-vardhan 
Add test for the middleware

aaronbrethorst (changes requested):
Hey Tejasva, nice work adding panic recovery -- this is genuinely important for production reliability, and the overall design is solid: per-request wrapper, goroutine-scoped recovery, and a response format that matches the existing `serverErrorResponse` in `errors.go` exactly. There are a couple of issues to address before we can merge, though.

## Critical

### 1. `recoveryResponseWriter` must also override `Write()` to track response state

`recovery_middleware.go:16-24`

The wrapper only intercepts `WriteHeader()`, but in Go's `net/http`, calling `Write()` without an explicit `WriteHeader()` triggers an implicit `WriteHeader(200)` on the **original** writer -- bypassing the wrapper entirely. This means `wroteHeader` stays `false` even after headers and body bytes have been sent to the client.

If a handler calls `w.Write(data)` and then panics, the recovery code sees `wroteHeader == false`, concludes no response has been sent, and writes a 500 JSON response on top of the already-committed 200. The client receives a corrupted response: HTTP 200 status with the handler's partial body concatenated with the error JSON.

Your own codebase has the correct pattern in `caching_middleware.go:46-51`, which overrides both `WriteHeader` and `Write`:

```go
func (rw *recoveryResponseWriter) Write(b []byte) (int, error) {
    if !rw.wroteHeader {
        rw.wroteHeader = true
    }
    return rw.ResponseWriter.Write(b)
}
```

Please add this method and a test for the "panic after `Write()` without `WriteHeader()`" scenario to confirm recovery does NOT attempt to overwrite an already-started response.

## Important

### 2. Log the JSON encoding error instead of discarding it

`recovery_middleware.go:56`

```go
_ = json.NewEncoder(w).Encode(response)
```

Every other error response in `errors.go` logs encoding failures (lines 28-31, 51-53, 87-89). The recovery middleware has access to `logger` -- please follow the same pattern:

```go
if err := json.NewEncoder(w).Encode(response); err != nil {
    logging.LogError(logger, "failed to encode panic recovery response", err,
        slog.String("path", r.URL.Path))
}
```

This matters most precisely when things are already broken (the panic recovery path). Losing observability during error recovery is the worst time to lose it.

### 3. Add a test for `panic(error)` to cover the type-switch branch

`recovery_middleware_test.go`

The existing test panics with a string (`panic("test panic")`), which only exercises the `fmt.Errorf("%v", rec)` branch. The `if e, ok := rec.(error); ok` branch (line 36-38) is never tested. Please add a test that panics with an `error` value (e.g., `panic(fmt.Errorf("typed error"))`) and verifies the same 500 JSON response is returned.

## Fit and Finish

### 4. Consider enriching the panic log with request context

`recovery_middleware.go:41`

The panic log includes `path` and `stack`, but since recovery is outermost, panicked requests bypass the request logging middleware entirely -- meaning no request log entry is emitted for them. Consider adding `method` and `request_id` (available from context since `RequestIDMiddleware` runs inside recovery) to make the panic log more useful for debugging:

```go
reqID, _ := r.Context().Value(RequestIDKey).(string)
logging.LogError(logger, "handler panic recovered", err,
    slog.String("method", r.Method),
    slog.String("path", r.URL.Path),
    slog.String("request_id", reqID),
    slog.String("stack", string(stack)),
)
```

This is optional but would significantly improve the debugging experience.

---

**What's good about this PR:**

- The response format (`code`, `currentTime`, `text`, `version`) matches `serverErrorResponse` in `errors.go` exactly, including `Version: 1`
- Using `models.ResponseCurrentTime(c)` with an injected clock is the right approach -- consistent and testable
- The `wroteHeader` guard is the right design (just needs `Write` coverage too)
- Tests use `MockClock` with deterministic time assertions -- good test discipline
- Middleware placement as outermost is the correct choice for a recovery handler

tejasva-vardhan:
Thanks for the detailed review!

- Fixed response state tracking by overriding Write() in addition to WriteHeader()
- Added a test case for panic after Write() to ensure the response is not overwritten
- Logged JSON encoding errors using logging.LogError to match existing patterns
- Added a test for panic with an error type to cover the type assertion branch
- Enriched panic logs with method and request_id for better debugging context

All changes are minimal and follow the existing code style. Tests pass locally.

Please review again.

aaronbrethorst (approved):
Hey Tejasva, great work on the follow-up -- you've addressed every item from the first review thoroughly and cleanly. The `Write()` override, JSON encoding error logging, `panic(error)` test, and enriched log context with `method` and `request_id` are all exactly right.

I ran through the code with a fine-tooth comb this time and I'm satisfied this is ready to merge. A couple of optional observations below for your consideration, but none are blockers.

## Fit and Finish

### 1. Recovery path writes to `w` instead of `rw`

`recovery_middleware.go:55-68`

The recovery block writes the error response through the original `w` rather than the wrapped `rw`:

```go
if !rw.wroteHeader {
    w.Header().Set("Content-Type", "application/json")  // w, not rw
    w.WriteHeader(http.StatusInternalServerError)         // w, not rw
    // ...
    json.NewEncoder(w).Encode(response)                   // w, not rw
}
```

This works correctly today because `rw.ResponseWriter` IS `w` and recovery is outermost, so nothing wraps `w` above it. But writing through `rw` would be more self-consistent -- it makes the code easier to reason about and would be the right call if middleware ordering ever changes. Totally optional.

### 2. Consider a test for panic after explicit `WriteHeader()` (without `Write()`)

`recovery_middleware_test.go`

The existing "panic after write" test covers `Write()` setting `wroteHeader = true`, but there's no test where a handler calls `WriteHeader(206)` and then panics. This would exercise the `WriteHeader` interception path specifically and verify the 206 status is preserved. Quick addition if you want it:

```go
func TestRecoveryMiddleware_PanicAfterWriteHeaderOnly(t *testing.T) {
    logger := slog.New(slog.NewTextHandler(io.Discard, nil))
    mockClock := clock.NewMockClock(time.Now())

    handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
        w.WriteHeader(http.StatusPartialContent)
        panic("panic after WriteHeader")
    })

    wrapped := NewRecoveryMiddleware(logger, mockClock)(handler)
    req := httptest.NewRequest(http.MethodGet, "/test", nil)
    rr := httptest.NewRecorder()
    wrapped.ServeHTTP(rr, req)

    if rr.Code != http.StatusPartialContent {
        t.Errorf("expected status 206, got %d", rr.Code)
    }
}
```

---

**What's good about this PR:**

- All four items from the first review are addressed correctly
- `Write()` override correctly tracks implicit `WriteHeader(200)` behavior, matching the pattern in `caching_middleware.go`
- JSON encoding error logging follows the same `logging.LogError` pattern used throughout `errors.go`
- The `error` type-switch test (`TestRecoveryMiddleware_PanicWithErrorType`) fully exercises the branch at line 44
- Enriched panic logs (`method`, `request_id`, `stack`, `path`) give good debugging context
- The test suite is well-structured: 4 tests covering the core behavioral contract, deterministic time assertions via `MockClock`, and full JSON response validation
- 94.7% line coverage on 80 lines of implementation

Approved -- nice, clean implementation.

PR #509, opened 2026-02-28
fix: enable SQLite WAL mode to prevent locks (#493)

### LINKED ISSUE #493: SQLite WAL mode not enabled — will cause database lock errors under load
## Problem

The comment at `gtfsdb/helpers.go:971` mentions WAL mode, but `PRAGMA journal_mode=WAL` is never actually set anywhere in the codebase. The database is configured with 25 max open connections (`MaxOpenConns=25`), but without WAL mode, SQLite uses the default rollback journal where **readers block writers and writers block readers**.

Under any real concurrent traffic, this will produce `database is locked` errors.

## Location

- `gtfsdb/helpers.go` — `configureSQLitePerformance` function
- The comment on line 971 references WAL mode but no pragma is applied

## Fix

Add `PRAGMA journal_mode=WAL` to `configureSQLitePerformance` alongside the existing `cache_size` and `temp_store` pragmas.

## Priority

**Ship blocker** — trivial fix, high impact. This is the single most likely cause of production incidents on day one.

ARCoder181105:
## Description
This PR addresses issue #493 by explicitly enabling Write-Ahead Logging (WAL) for SQLite databases. 

Previously, `WAL` mode was referenced in the connection pool comments but was never actually applied in the database configuration. Without WAL mode, SQLite falls back to the default rollback journal where readers block writers and writers block readers. With `MaxOpenConns` set to 25 for file-based databases, this would inevitably lead to `database is locked` errors under concurrent production load.

## Changes Made
* **`gtfsdb/helpers.go`**: Added `PRAGMA journal_mode=WAL` to the `configureSQLitePerformance` function to allow concurrent readers and a single writer.
* **`gtfsdb/sqlite_performance_test.go`**: 
  * Updated `TestFileDatabaseConnectionPool` to verify that `wal` mode is properly enabled for file-based databases.
  * Updated `TestSQLitePerformancePragmasApplied` to verify that `:memory:` databases safely fall back to `memory` journal mode (as SQLite does not support WAL in-memory).

## Related Issues
Closes #493 

## Checklist
- [x] I have performed a self-review of my own code
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

### DIFF
--- a/gtfsdb/helpers.go
+++ b/gtfsdb/helpers.go
@@ -938,6 +938,8 @@ func configureSQLitePerformance(ctx context.Context, db *sql.DB) error {
 		{"PRAGMA cache_size=-64000", "Set cache size to 64MB"},
 		// Store temp tables and indices in memory for faster operations
 		{"PRAGMA temp_store=MEMORY", "Store temporary data in memory"},
+		// Enable Write-Ahead Logging to allow concurrent readers and a single writer
+		{"PRAGMA journal_mode=WAL", "Enable WAL mode"},
 	}
 
 	logger := slog.Default().With(slog.String("component", "sqlite_performance"))
--- a/gtfsdb/sqlite_performance_test.go
+++ b/gtfsdb/sqlite_performance_test.go
@@ -38,6 +38,13 @@ func TestSQLitePerformancePragmasApplied(t *testing.T) {
 	require.NoError(t, err)
 	// 2 = MEMORY
 	assert.Equal(t, 2, tempStore, "Temp store should be set to MEMORY (2)")
+
+	// Verify journal_mode PRAGMA
+	var journalMode string
+	err = client.DB.QueryRowContext(ctx, "PRAGMA journal_mode").Scan(&journalMode)
+	require.NoError(t, err)
+	// In-memory databases don't support WAL mode, so SQLite automatically uses 'memory'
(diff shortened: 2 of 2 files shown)


aaronbrethorst (approved):
Hey Aditya -- Clean and surgical fix. The `configureConnectionPool` comments have been referencing WAL mode since the connection pool was set up, but the PRAGMA was never actually applied. This PR closes that gap with minimal, correct code and appropriate test coverage.

---

## No Issues Found

The change is correct:
- `PRAGMA journal_mode=WAL` is a database-level persistent setting for file-based databases, so it correctly applies across all pooled connections.
- For `:memory:` databases, SQLite silently falls back to `memory` journal mode (no error), which is the right behavior since the connection pool is already limited to 1 for in-memory databases.
- The test assertions verify both cases: `"wal"` for file databases, `"memory"` for in-memory databases.

## Verification

- Tests: all pass
- Lint: 0 issues
- Single commit: clean

## Strengths

- **Minimal and focused**: 2 lines of production code, exactly what was needed.
- **Both cases tested**: File-based databases get WAL mode verified, in-memory databases get the fallback behavior verified.
- **Accurate comments**: The inline comment correctly describes what WAL mode provides.

---

Looks good -- merge it!

PR #513, opened 2026-02-28
fix: add exponential backoff retry for initial GTFS data load (#496)

### LINKED ISSUE #496: GTFS static data load has no retry on startup
## Problem

In `internal/gtfs/gtfs_manager.go` (`InitGTFSManager`), the initial GTFS static data download is attempted exactly once. If the feed URL is unreachable at startup — due to a transient network issue, DNS hiccup, or upstream outage — the entire application exits immediately with no retry.

The periodic 24-hour update cycle (`internal/gtfs/static.go`) handles failures gracefully by logging and continuing to the next cycle, but the initial load does not.

## Impact

In a containerized environment, the orchestrator may restart the process, but this creates unnecessary churn and delays. A brief upstream blip at the wrong moment causes a full outage.

## Fix

Add retry with exponential backoff (e.g., 3-5 attempts with 5s/15s/30s/60s delays) before giving up on initial load. The existing health check already reports unhealthy when GTFS data isn't loaded, so readiness probes will correctly gate traffic.

## Location

- `internal/gtfs/gtfs_manager.go` — `InitGTFSManager`
- `internal/gtfs/static.go` — `downloadAndExtractGTFS`

## Priority

Medium-High

ARCoder181105:
### **Description:**

**Problem:**
Previously, during application startup in `InitGTFSManager`, the initial GTFS static data download was attempted exactly once. If the feed URL was unreachable due to a transient network issue, DNS hiccup, or upstream outage, the application would exit immediately. This caused unnecessary container churn and delays in orchestrated environments.

**Solution:**
This PR introduces a retry mechanism with exponential backoff for the initial GTFS static data load and database build processes.

### **Changes Made:**

* Updated `InitGTFSManager` in `internal/gtfs/gtfs_manager.go` to wrap `loadGTFSData` and `buildGtfsDB` in a retry loop.
* Implemented a maximum of 5 attempts (1 initial + 4 retries) with an exponential backoff schedule of `[5s, 15s, 30s, 60s]`.
* Added structured logging to capture and report each failed attempt before stalling.
* If all 5 attempts fail, the application gracefully surfaces the accumulated error instead of entering an indeterminate state.

### **Testing:**

* Ran `make test` and verified all tests pass (`ok maglev.onebusaway.org/internal/gtfs 15.567s`).
* **Note on tests:** Existing unit tests in `gtfs_manager_test.go` continue to pass without modification because they rely on local mock files (`raba.zip`). Since local file loading succeeds on the first attempt, the retry loop immediately breaks and avoids the `time.Sleep()` backoff delays, ensuring test execution remains fast.

Closes #496

### DIFF
--- a/cmd/api/app.go
+++ b/cmd/api/app.go
@@ -67,10 +67,10 @@ func ParseAPIKeys(apiKeysFlag string) []string {
 // BuildApplication creates and initializes the Application with all dependencies.
 // This includes creating the logger, initializing the GTFS manager, and creating the direction calculator.
 // Returns an error if GTFS manager initialization fails.
-func BuildApplication(cfg appconf.Config, gtfsCfg gtfs.Config) (*app.Application, error) {
+func BuildApplication(ctx context.Context, cfg appconf.Config, gtfsCfg gtfs.Config) (*app.Application, error) {
 	logger := slog.New(slog.NewTextHandler(os.Stdout, nil))
 
-	gtfsManager, err := gtfs.InitGTFSManager(gtfsCfg)
+	gtfsManager, err := gtfs.InitGTFSManager(ctx, gtfsCfg)
 	if err != nil {
 		return nil, fmt.Errorf("failed to initialize GTFS manager: %w", err)
 	}
--- a/cmd/api/app_test.go
+++ b/cmd/api/app_test.go
@@ -67,6 +67,8 @@ func TestParseAPIKeys(t *testing.T) {
 }
 
 func TestBuildApplicationWithMemoryDB(t *testing.T) {
+	ctx := context.Background()
+
 	// Get path to test data
 	testDataPath := filepath.Join("..", "..", "testdata", "raba.zip")
 
@@ -89,7 +91,7 @@ func TestBuildApplicationWithMemoryDB(t *testing.T) {
 		Verbose:      false,
 	}
 
-	coreApp, err := BuildApplication(cfg, gtfsCfg)
+	coreApp, err := BuildApplication(ctx, cfg, gtfsCfg)
 
 	require.NoError(t, err, "BuildApplication should not return an error")
 	assert.NotNil(t, coreApp, "Application should not be nil")
@@ -99,6 +101,8 @@ func TestBuildApplicationWithMemoryDB(t *testing.T) {
 }
 
 func TestBuildApplicationWithTestData(t *testing.T) {
+	ctx := context.Background()
+
 	// Get path to test data
 	testDataPath := filepath.Join("..", "..", "testdata", "raba.zip")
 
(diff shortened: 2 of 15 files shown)


aaronbrethorst (changes requested):
Aditya, excellent work addressing all the feedback from round 1. Every item has been properly resolved — the partial SQLite cleanup, the connection leak fix, the local file skip, structured logging, success-after-retry logging, and the full context.Context plumbing with signal-based cancellation. The `signal.NotifyContext` in `main.go` is a particularly nice touch that goes beyond what was suggested. Just one small item to fix before merging.

## Prior Feedback Status

All 6 items from the first review have been addressed:

| # | Item | Status |
|---|------|--------|
| 1 | `buildGtfsDB` leaves partial SQLite file on disk | **Fixed** — `os.Remove` before retry at `gtfs_manager.go:608` |
| 2 | `buildGtfsDB` leaks open SQLite connection on failure | **Fixed** — `client.Close()` on error path in `static.go:103` |
| 3 | Retrying local file loads is unnecessary | **Fixed** — `maxAttempts = 1` when `isLocalFile` at `gtfs_manager.go:573` |
| 4 | Use structured logging attributes | **Fixed** — `slog.Int`/`slog.Duration` fields at `gtfs_manager.go:587-591` |
| 5 | Log success after retry recovery | **Fixed** — `gtfs_manager.go:639-641` |
| 6 | Accept `context.Context` for cancellable startup | **Fixed** — full context plumbing through `InitGTFSManager`, cancellable `select`/`time.After` sleep, and `signal.NotifyContext` in `main.go` |

## Important Issues (1 found)

### 1. `go fmt` not run on two files

Two test files have formatting issues that `go fmt` would fix:

- **`internal/gtfs/shutdown_test.go`** (lines 27, 73, 109): Missing space after comma:
  ```go
  // Current:
  manager, err := InitGTFSManager(context.Background(),config)
  // Should be:
  manager, err := InitGTFSManager(context.Background(), config)
  ```

- **`internal/gtfs/hot_swap_test.go`** (line 260): Missing spaces around `:=`:
  ```go
  // Current:
  ctx:=context.Background()
  // Should be:
  ctx := context.Background()
  ```

**Fix:** Run `go fmt ./...` and commit the changes. This is a project requirement per CLAUDE.md.

## Strengths

- Clean, thorough implementation of all round 1 feedback
- The `signal.NotifyContext` in `main.go` elegantly connects Ctrl+C handling to the retry loop's context cancellation
- Independent tracking of `staticData` vs `gtfsDB` avoids redundant work
- The backoff schedule [5s, 15s, 30s, 60s] is sensible
- Context is properly threaded through to `GetImportMetadata`, `buildStopSpatialIndex`, and the initial real-time feed fetch

## Checklist

- [x] Tests pass
- [x] Lint clean
- [ ] Run `go fmt ./...` (Important #1)
- [x] All round 1 feedback addressed

## Recommended Action

1. Run `go fmt ./...` and commit
2. Ready to merge after that

PR #517, opened 2026-02-28
gtfsdb: replace GetBulkInsertBatchSize with SafeBatchSize(fieldsPerRow)

### LINKED ISSUE #483: GetBulkInsertBatchSize: compute batch size from fields-per-row instead of returning a hardcoded constant
## Problem

`GetBulkInsertBatchSize()` in `gtfsdb/config.go` returns a single hardcoded value
(`DefaultBulkInsertBatchSize = 3000`) regardless of how many fields each INSERT
statement binds per row.

SQLite enforces a hard upper bound of **32,766 bound parameters per statement**
(`SQLITE_MAX_VARIABLE_NUMBER`). The safe maximum batch size therefore depends
entirely on the number of fields per row:

    max_batch_size = 32766 / fields_per_row

Currently, both batched bulk-insert functions use the same batch size of 3000:

| Function              | Fields/row | Variables used (batch=3000) | Limit  | Utilisation |
|-----------------------|------------|----------------------------|--------|-------------|
| `bulkInsertStopTimes` | 10         | 30,000                     | 32,766 |**91%**       |
| `bulkInsertShapes`    | 5          | 15,000                     | 32,766 | **46%**   |

`bulkInsertShapes` can safely use a batch size of **6,553** — more than double the
current value — halving the number of `tx.ExecContext` calls, SQL string
allocations, and sort operations for large shape datasets.

## Proposed Fix

Replace the no-argument `GetBulkInsertBatchSize()` method with a `SafeBatchSize(fieldsPerRow int) int` helper:

```go
const sqliteMaxVariables = 32766 // SQLITE_MAX_VARIABLE_NUMBER hard limit

func SafeBatchSize(fieldsPerRow int) int {
    if fieldsPerRow <= 0 {
        return DefaultBulkInsertBatchSize
    }
    return sqliteMaxVariables / fieldsPerRow
}
```

**I already have a PR prepared implementing this change and can submit it upon approval.**
@aaronbrethorst

Adityatorgal17:
## Summary
SQLite enforces a hard limit of 32766 bound parameters per statement
(`SQLITE_MAX_VARIABLE_NUMBER`). The safe max batch size therefore depends
on the number of fields bound per row:

    maxBatchSize = 32766 / fieldsPerRow

## Changes

- Add `sqliteMaxVariables = 32766` constant
- Replace `GetBulkInsertBatchSize()` with `SafeBatchSize(fieldsPerRow int)`
  which computes the optimal batch size per caller
- `bulkInsertStopTimes`: `SafeBatchSize(10)` → 3276 rows/batch (was 3000)
- `bulkInsertShapes`:    `SafeBatchSize(5)`  → 6553 rows/batch (was 3000)
  - Effectively halves the number of INSERT calls for large shape datasets
- Add `TestSafeBatchSize` covering:
  - edge cases
  - both callers

Fixes #483

### DIFF
--- a/gtfsdb/config.go
+++ b/gtfsdb/config.go
@@ -3,11 +3,13 @@ package gtfsdb
 import "maglev.onebusaway.org/internal/appconf"
 
 const (
-	// DefaultBulkInsertBatchSize is the default batch size for multi-row INSERTs.
-	// SQLite's default SQLITE_MAX_VARIABLE_NUMBER is 32766 (with SQLITE_ENABLE_COLUMN_METADATA).
-	// We use 3000 records with 10 fields per record = 30,000 variables per batch,
-	// which is well under the limit and provides good performance.
+	// DefaultBulkInsertBatchSize is the fallback batch size used when fieldsPerRow is
+	// unknown or <= 0.
 	DefaultBulkInsertBatchSize = 3000
+
+	// sqliteMaxVariables is the hard upper bound on bound parameters per SQLite
+	// statement (SQLITE_MAX_VARIABLE_NUMBER).
+	sqliteMaxVariables = 32766
 )
 
 // Config holds configuration options for the Client
@@ -18,10 +20,9 @@ type Config struct {
 	verbose bool                // Enable verbose logging
 
 	// Performance tuning
-	// BulkInsertBatchSize controls how many records are inserted per multi-row INSERT statement.
-	// Default is 1000. Larger values can improve performance but may hit SQLite's
-	// SQLITE_MAX_VARIABLE_NUMBER limit (default 999).
-	// Set to 0 to use the default value.
+	// BulkInsertBatchSize is reserved for future use. Batch sizes are now computed
+	// automatically via SafeBatchSize based on SQLITE_MAX_VARIABLE_NUMBER (32766)
+	// and the number of fields bound per row.
 	BulkInsertBatchSize int
 }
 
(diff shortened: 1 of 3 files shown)


aaronbrethorst (approved):
Hey Aditya — this is a clean, well-reasoned refactor. The math is solid (`32766 / fieldsPerRow` with integer truncation guarantees you never exceed `SQLITE_MAX_VARIABLE_NUMBER`), the field counts are correct (I verified both INSERT statements against the schema), and the table-driven test with the invariant assertion is a nice touch. The shapes batch size effectively doubles from 3000 to 6553, which is a real performance win for large datasets.

## Fit and Finish

### 1. `BulkInsertBatchSize` field is now dead code

The `BulkInsertBatchSize` field on `Config` is still present, initialized to 3000 in `NewConfig()`, and documented as "reserved for future use" — but `SafeBatchSize()` completely ignores it. No code path reads this field anymore.

**Suggestion**: Either remove it entirely, or if you intend to keep it as an optional cap/override, have `SafeBatchSize` incorporate it (e.g., `min(sqliteMaxVariables/fieldsPerRow, c.BulkInsertBatchSize)` when the field is set). The current state where the field exists and is initialized but never consulted is a bit misleading.

**File**: `gtfsdb/config.go:22-26, 34`

### 2. Consider naming the magic numbers at call sites

The field counts `10` and `5` are hardcoded arguments that must stay in sync with the INSERT column list, the placeholder tuple, and the `args` append — four places total. The inline comments help, but a named constant near each function would make the coupling explicit:

```go
const stopTimeFieldsPerRow = 10
batchSize := c.config.SafeBatchSize(stopTimeFieldsPerRow)
```

This also lets you reuse the constant in the `make([]interface{}, 0, len(batch)*stopTimeFieldsPerRow)` allocation.

**File**: `gtfsdb/helpers.go:586, 756`

### 3. Stale comment in shapes bulk insert

The comment on line 849 says "150k records with batch size 3000" — but with `SafeBatchSize(5)`, the batch size is now 6553, making 50 batches approximately 327k records. Minor, but worth a quick update.

**File**: `gtfsdb/helpers.go:849`

## Strengths

- The mathematical basis is correct and clearly documented — `32766 / fieldsPerRow` is the right formula
- Field counts verified against schema: stop_times has exactly 10 user-supplied columns, shapes has exactly 5
- Table-driven test with both edge cases (zero, negative) and real-world values (10, 5)
- The `got*tt.fieldsPerRow <= sqliteMaxVariables` invariant assertion is a smart addition
- Clean migration — no remaining callers of the old `GetBulkInsertBatchSize` method
- Shapes batch size doubles from 3000 to 6553, halving INSERT calls for large shape datasets

I'd love to see a followup PR with the requests I outlined above, but I don't think they're worth blocking merging. Nice work!

PR #518, opened 2026-02-28
chore: remove redundant error log on batch route fetch failure

AhmedAlian7:
# Remove redundant error log in route fetching

### Summary
This is a quick follow-up PR to clean up a minor logging issue from the previous merge. 
- **main PR #488**

### Changes
- **Removed** `api.Logger.Error("Failed to batch fetch routes for schedule stop", "error", err)` from the GTFS batch fetching logic.


aaronbrethorst (approved):
Nice cleanup, Ahmed -- good eye catching this leftover from #488.

`serverErrorResponse` (`errors.go:35`) already logs the error with the request path, so the explicit `Logger.Error` call on line 146 was producing a duplicate log entry for every failure. Removing it is the right call.

**Merge** -- no changes needed.

PR #519, opened 2026-02-28
feat: add large GTFS dataset support for performance benchmarks

### LINKED ISSUE #502: Acquire large-agency GTFS test data for realistic performance testing
## Context

This is part of the performance testing initiative. The current test data (RABA — Roseville Area) is a small agency. Performance problems that only manifest with large datasets won't be caught.

## What to do

Add one or two large-agency GTFS static datasets as test fixtures so benchmarks and load tests can run against realistic data volumes.

### Suggested agencies (all public domain)

| Agency | Approx. stops | Approx. routes | Approx. trips | GTFS feed |
|--------|--------------|----------------|---------------|-----------|
| TriMet (Portland) | ~6,500 | ~80 | ~12,000 | https://developer.trimet.org/GTFS.shtml |
| MBTA (Boston) | ~9,500 | ~200 | ~60,000 | https://www.mbta.com/developers/gtfs |
| King County Metro (Seattle) | ~8,500 | ~200 | ~30,000 | https://kingcounty.gov/en/dept/metro/rider-tools/mobile-and-web-tools/for-developers |

Pick **one** to start — TriMet is a good balance of size and complexity.

### What to deliver

1. **A script** in `testdata/` or `scripts/` that downloads and prepares the large GTFS dataset
2. **Do NOT check the zip into git** — it's too large. Instead:
   - Add a `testdata/README.md` documenting how to download it
   - Add a `scripts/download-perf-data.sh` that fetches it
   - Use a Go build tag (e.g., `//go:build perftest`) so large-data benchmarks are opt-in
3. **A pre-built SQLite database** for the large dataset (or instructions to build one with the existing import tooling)

### Usage pattern

```go
//go:build perftest

func BenchmarkLargeAgencyStopsForLocation(b *testing.B) {
    api := createLargeAgencyApi(b) // loads TriMet or similar
    // ...
}
```

```bash
# Regular tests (fast, RABA only)
make test

# Performance tests with large data
go test -tags perftest -bench=. ./internal/restapi/
```

### Acceptance criteria

- [ ] Download script for at least one large agency's GTFS data
- [ ] `createLargeAgencyApi` test helper behind a build tag
- [ ] Documentation in `testdata/README.md`
- [ ] `.gitignore` entry for the large dataset files
- [ ] Verify the data loads successfully with the existing GTFS import pipeline

## Good first issue

This is approachable — it's mostly scripting and test infrastructure, no deep domain knowledge required.

tejasva-vardhan:
This PR adds support for running benchmarks on a larger, more realistic GTFS dataset as part of the performance testing initiative.

Right now, most tests use the RABA dataset, which is quite small. That makes it hard to catch performance issues that only show up with real-world data sizes. This change introduces a simple way to work with a larger dataset without affecting normal test runs.

This fixes #502

## What’s included

- Added a download script for a large public GTFS dataset (TriMet)
  - `scripts/download-perf-data.sh`
- Dataset is **not committed** to the repo
  - `testdata/perf/` is added to `.gitignore`
- Added documentation in `testdata/README.md`
  - explains how to download and use the dataset
- Introduced a `perftest` build tag so large-data benchmarks are opt-in
- Added `createLargeAgencyApi` helper to load the dataset using the existing GTFS pipeline
- Added a sample benchmark using the large dataset

The SQLite database is not pre-generated or committed. It is built automatically using the existing import pipeline when benchmarks are run.

TriMet was chosen as a starting point since it provides a good balance of size and complexity while still being publicly available.

## How to run

```bash
# Download the dataset
./scripts/download-perf-data.sh

# Run performance benchmarks with large dataset
go test -tags=perftest -bench=. -benchmem ./internal/restapi/

### CODE REVIEW

**Review by aaronbrethorst — CHANGES_REQUESTED** [review]
Tejasva, this is a well-thought-out approach to large-dataset benchmarking. The `perftest` build tag keeps CI fast while making realistic benchmarks available on demand, and the download script plus documentation make it easy for anyone to get started. Just one item to fix before merging.

## Important Issues (1 found)

### 1. `download-perf-data.sh` is not executable

The script is missing the execute bit, so `./scripts/download-perf-data.sh` will fail with "Permission denied":

```
$ ls -la scripts/download-perf-data.sh
-rw-r--r--  ...  download-perf-data.sh
```

**Fix:**
```bash
chmod +x scripts/download-perf-data.sh
```

Then commit the permission change. Git tracks the executable bit, so this will persist for everyone who clones the repo.

## Fit-and-Finish (2 found)

### 1. Consider renaming `perf_test_helpers.go` to `perf_test_helpers_test.go`

The file only contains test helper code (imports `testing`, uses `testing.TB`, called exclusively from `_test.go` files). Go convention is to use the `_test.go` suffix for such files — this ensures the code is only compiled during `go test` and can't accidentally leak into a production binary.

I realize the project has a pre-existing instance of this pattern in `models/test_helpers.go`, so this isn't blocking, but it's worth fixing while you're here.

### 2. Benchmark doesn't verify successful responses

The benchmark fires requests but never checks the HTTP status. If the endpoint returns errors (e.g., due to missing data or misconfiguration), the benchmark would silently measure error-path performance. Consider adding a one-time sanity check before the timer loop:

```go
// Verify the endpoint works before benchmarking
rec := httptest.NewRecorder()
mux.ServeHTTP(rec, req)
if rec.Code != http.StatusOK {
    b.Fatalf("expected 200, got %d: %s", rec.Code, rec.Body.String())
}
b.ResetTimer()
for i := 0; i < b.N; i++ {
    // ...
}
```

## Strengths

- The `perftest` build tag is the right approach — keeps CI fast, makes large-data benchmarks opt-in
- Data is not committed to the repo; `.gitignore` entry and download script are clean
- `testdata/README.md` documentation is clear and actionable
- Proper use of `testing.TB` and `tb.Helper()` in the helper function
- Good error message when the dataset is missing, pointing the user to the download script
- Reuses the existing `InitGTFSManager` pipeline rather than custom loading code

## Checklist

- [x] Tests pass
- [x] Lint clean (including with `-tags=perftest`)
- [ ] Make download script executable (Important #1)
- [ ] Consider `_test.go` suffix for helper file (Fit-and-Finish #1)
- [ ] Consider sanity check in benchmark (Fit-and-Finish #2)

## Recommended Action

1. Fix the script permission (`chmod +x`)
2. Consider the two fit-and-finish suggestions

PR #522, opened 2026-02-28
Fix/n plus one routes for stops

### LINKED ISSUE #304: Fix N+1 query in buildStopList — replace per-stop GetRouteIDsForStop with batch GetRoutesForStops
**Labels:** `performance`, `bug`

---

### Description

`buildStopList` currently fires a separate `GetRouteIDsForStop` query for every stop in the loop. If there are 40 stops, that's 40 individual database queries executing back to back.

### Current Behavior

```go
func (rb *referenceBuilder) buildStopList(stops []gtfsdb.Stop) {
	rb.stopList = make([]models.Stop, 0, len(stops))

	for _, stop := range stops {
		routeIds, err := rb.api.GtfsManager.GtfsDB.Queries.GetRouteIDsForStop(rb.ctx, stop.ID) // 🔴 fires once per stop
		if err != nil {
			continue
		}

		routeIdsString := rb.processRouteIds(routeIds)
		rb.stopList = append(rb.stopList, rb.createStop(stop, routeIdsString))
	}
}
```

**Query breakdown for 40 stops:**
- `GetStopsByIDs` → 1 query (upstream, batch fetched correctly)
- `GetRouteIDsForStop` → 40 queries (one per stop inside the loop)
- **Total: 41 queries**

### Expected Behavior

All route IDs for all stops should be fetched in a single batch query using the existing `GetRoutesForStops` method, which already accepts a slice of stop IDs. A map is then built from the result so that each stop's routes are looked up in memory — no database calls inside the loop.

**Query breakdown for 40 stops:**
- `GetStopsByIDs` → 1 query
- `GetRoutesForStops` → 1 query
- **Total: 2 queries**

### Proposed Fix

```go
func (rb *referenceBuilder) buildStopList(stops []gtfsdb.Stop) {
	rb.stopList = make([]models.Stop, 0, len(stops))

	stopIDs := make([]string, 0, len(stops))
	for _, stop := range stops {
		stopIDs = append(stopIDs, stop.ID)
	}

	routesForStops, err := rb.api.GtfsManager.GtfsDB.Queries.GetRoutesForStops(rb.ctx, stopIDs)
	if err != nil {
		return
	}

	stopRouteMap := make(map[string][]string)
	for _, r := range routesForStops {
		stopRouteMap[r.StopID] = append(stopRouteMap[r.StopID], r.ID)
	}

	for _, stop := range stops {
		routeIdsString := stopRouteMap[stop.ID]
		if routeIdsString == nil {
			routeIdsString = []string{}
		}
		rb.stopList = append(rb.stopList, rb.createStop(stop, routeIdsString))
	}
}
```

### Notes

- `GetRoutesForStops` is already used in `arrivalAndDepartureForStopHandler` — no new queries need to be written.
- Verify the field names on the `GetRoutesForStops` return struct (`StopID`, `ID`) match before merging.
- This is a classic N+1 query pattern. Any other reference builder loops that call per-item queries should be audited for the same issue.

Adityatorgal17:
## Problem

Every call to `trips-for-location` triggered N+1 database queries inside `buildStopList` — one `GetRouteIDsForStop` query per stop in the bounding box.

With the stop limit capped at 100, this resulted in up to **101 queries** just to build the stop list for a single API request.

---

## Root Cause

`buildStopList` performed a per-stop database query instead of batching route lookups.

## Fix

Replace the per-stop query loop with a single batch query:

```
GetRouteIDsForStops(stopIDs)
```

Then:

- Build `stopID → []combinedRouteID` map in memory
- Extract raw route IDs using `utils.ExtractCodeID`
- Store raw IDs in `presentRoutes`

This eliminates N+1 queries and restores correct route references.

---

## Changes

| File | Change |
|------|--------|
| `trips_for_location_handler.go` | Replaced `buildStopList` with batch implementation; removed dead `processRouteIds` |
| `trips_for_location_handler_test.go` | Added `TestTripsForLocationHandler_ReferencesContainStopsAndRoutes` |

---
## Testing

Ran all required checks:

- `make fmt`
- `make lint`
- `make test`
---

## Closes

Fixes #304  
Supersedes #305

### DIFF
--- a/internal/restapi/trips_for_location_handler.go
+++ b/internal/restapi/trips_for_location_handler.go
@@ -504,29 +504,47 @@ func (rb *referenceBuilder) collectTripIDs(trips []models.TripsForLocationListEn
 
 func (rb *referenceBuilder) buildStopList(stops []gtfsdb.Stop) {
 	rb.stopList = make([]models.Stop, 0, len(stops))
+	if len(stops) == 0 {
+		return
+	}
+
+	stopIDs := make([]string, 0, len(stops))
+	for _, stop := range stops {
+		stopIDs = append(stopIDs, stop.ID)
+	}
+
+	routesForStops, err := rb.api.GtfsManager.GtfsDB.Queries.GetRouteIDsForStops(rb.ctx, stopIDs)
+	if err != nil {
+		logging.LogError(rb.api.Logger, "failed to batch fetch routes for stops", err)
+		return
+	}
+
+	// Build in-memory map: stopID → []combinedRouteID (e.g. "40_100479").
+	// Also register the raw route ID (e.g. "100479") in presentRoutes so that
+	// collectAgenciesAndRoutes can fetch full route details via GetRoutesByIDs,
+	// which queries WHERE routes.id IN (?), using raw IDs — not combined ones.
+	stopRouteMap := make(map[string][]string)
+	for _, r := range routesForStops {
+		combinedID, ok := r.RouteID.(string)
+		if !ok {
+			continue
+		}
+		stopRouteMap[r.StopID] = append(stopRouteMap[r.StopID], combinedID)
+		if rawID, err := utils.ExtractCodeID(combinedID); err == nil {
+			rb.presentRoutes[rawID] = models.Route{}
+		}
+	}
+
 	for _, stop := range stops {
 		if rb.ctx.Err() != nil {
 			return
 		}
-
-		routeIds, err := rb.api.GtfsManager.GtfsDB.Queries.GetRouteIDsForStop(rb.ctx, stop.ID)
-		if err != nil {
-			continue
+		combinedRouteIDs := stopRouteMap[stop.ID]
+		if combinedRouteIDs == nil {
+			combinedRouteIDs = []string{}
 		}
-
-		routeIdsString := rb.processRouteIds(routeIds)
-		rb.stopList = append(rb.stopList, rb.createStop(stop, routeIdsString))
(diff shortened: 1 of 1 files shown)


aaronbrethorst (approved):
Aditya, this is a really clean fix. You've not only eliminated the N+1 query problem in `buildStopList`, but you've also fixed a subtle pre-existing bug in the process — the old `processRouteIds` was storing combined IDs (e.g. "40_100479") in `presentRoutes`, but `collectAgenciesAndRoutes` passes those to `GetRoutesByIDs`, which queries `WHERE routes.id IN (?)` against raw IDs (e.g. "100479"). The combined IDs would never match, so route references were silently empty. Your use of `utils.ExtractCodeID` to store the raw ID fixes that. Nicely done.

## Critical Issues (0 found)

None.

## Important Issues (0 found)

None.

## Strengths

- Eliminates up to 100 per-stop queries with a single batch `GetRouteIDsForStops` call
- Fixes the pre-existing bug where `presentRoutes` held combined IDs that couldn't match `GetRoutesByIDs`'s raw ID lookup
- The checked type assertion (`r.RouteID.(string)` with `, ok`) is safer than the old unchecked panic-on-failure version
- Good test that validates end-to-end data integrity — stops reference routes that actually appear in `references.routes`
- Clear comment at `trips_for_location_handler.go:522-525` explaining the raw vs combined ID distinction

## Checklist

- [x] Tests pass
- [x] Lint clean
- [x] N+1 eliminated
- [x] Route references bug fixed
- [x] New test validates the fix

---

nice work, merging.

PR #524, opened 2026-02-28
feat: Add missing indexes to `trips` and `stop_times` tables to improve GTFS read performance.

### LINKED ISSUE #523: Missing Database Indexes for Query Optimization
### Summary
Critical queries in the system currently lack proper indexes, which leads to full table scans when handling large datasets. Missing composite indexes on the `trips` and `stop_times` tables significantly degrade performance for high-frequency operations, especially in production environments with 1000+ stops and 10000+ trips. Adding these indexes will enable range scans and constant-time lookups, eliminating O(N) filtering.

### Proposed Changes
Add three missing composite indexes to `gtfsdb/schema.sql` to speed up specific GTFS reads:
1. **`idx_trips_route_service` on `(route_id, service_id)`**
   - Optimizes `GetTripsForRouteInActiveServiceIDs()` by filtering trips by both route AND service directly via the index.
2. **`idx_stop_times_stop_arrival` on `(stop_id, arrival_time, departure_time)`**
   - Optimizes `GetStopTimesForStopInWindow()` to enable rapid range scans instead of filtering all stop times linked to a stop.
3. **`idx_trips_route_headsign` on `(route_id, trip_headsign, shape_id)`**
   - Optimizes `GetShapesGroupedByTripHeadSign()` by rapidly filtering trips by route, headsign, and shape for faster shape direction calculations.

*Note: Ensure that any `LIMIT 1` queries relying on these index groupings are made deterministic using an `ORDER BY` clause to avoid arbitrary result sorting variations across different SQLite deployments.*

### Testing
- [ ] Ensure database schema migrations execute cleanly on existing SQLite deployments.
- [ ] Run automated sqlc test suites (`make models`, `make test`) to ensure generated query plans work as expected.

### Impact
- **Performance**: Introduces a potential 10-50x speedup for the affected API endpoints dependent on these queries (e.g., calculating schedule stops, route paths).
- **Scalability**: Safe-guards against long query durations and database locking issues during high-volume GTFS real-time synchronization.

AhmedAlian7:
## Summary
This PR adds three critical composite database indexes to optimize GTFS spatial and temporal queries, resolving full table scan bottlenecks on large datasets. It also applies a deterministic sort patch to sqlc query definitions to maintain backwards compatibility with tests under the new index structures.

- Closes #523 

## Changes
- **`gtfsdb/schema.sql`**: 
  - Added `DROP INDEX IF EXISTS idx_trips_route_id` to remove the overlapping single-column index.
  - Added `idx_trips_route_service` on `trips(route_id, service_id)` to optimize `GetTripsForRouteInActiveServiceIDs`.
  - Added `DROP INDEX IF EXISTS idx_stop_times_stop_arrival` in order to safely upgrade the previous index migration.
  - Added `idx_stop_times_stop_arrival` on `stop_times(stop_id, arrival_time, departure_time)` to optimize window scanning in `GetStopTimesForStopInWindow`.
  - Added `idx_trips_route_headsign` on `trips(route_id, trip_headsign, shape_id)` to optimize grouping in `GetShapesGroupedByTripHeadSign`.
- **`gtfsdb/query.sql`**: Added `ORDER BY id LIMIT 1` to `GetShapesGroupedByTripHeadSign` to guarantee deterministic selection after the insertion of the new composite index.
- **`internal/restapi/trips_helper_test.go`**: Fixed a previously uncaught nil-pointer dereference in `TestGetFirstStopOfNextTripInBlock_WithBlockContinuation` which occurred when using unbounded `time.Now()` for mocked dataset calculations.

## Testing
- [x] Tested locally with standard test suite (`make test`, `make lint`).
- [x] Verified `make models` safely generates the deterministic sqlc queries.
- [x] Passed successfully over standard endpoint tests including previously failing `TestStopsForRouteHandlerEndToEnd`.

## Impact
- **Response Speeds**: Expect a 10-50x speed improvement for shape queries and stop-time window lookups on larger deployed instances configurations (10,000+ trips).
- **Stability**: Prevents sporadic application lock-ups stemming from DB full table scans during concurrent real-time syncs and API fetches.

### DIFF
--- a/gtfsdb/schema.sql
+++ b/gtfsdb/schema.sql
@@ -421,7 +421,13 @@ CREATE INDEX IF NOT EXISTS idx_problem_reports_stop_code
 -- Missing indexes to prevent full table scans on gtfs reads
 
 -- migrate
+DROP INDEX IF EXISTS idx_trips_route_id;
+
+-- migrate
 CREATE INDEX IF NOT EXISTS idx_trips_route_service ON trips (route_id, service_id);
+
+-- migrate
+DROP INDEX IF EXISTS idx_stop_times_stop_arrival;
 
 -- migrate
 CREATE INDEX IF NOT EXISTS idx_stop_times_stop_arrival ON stop_times (stop_id, arrival_time, departure_time);


aaronbrethorst (changes requested):
Hey Ahmed — nice work identifying these index gaps. Composite indexes on `(route_id, service_id)` and `(route_id, trip_headsign, shape_id)` are well-targeted for the queries they serve, and the deterministic `ORDER BY id` addition to `GetShapesGroupedByTripHeadSign` is the right fix for non-deterministic row selection. There's one critical migration issue that needs to be fixed before this can ship.

## Critical Issues

### 1. `idx_stop_times_stop_arrival` upgrade is silently broken by `IF NOT EXISTS`

Line 336 of `schema.sql` already creates:
```sql
CREATE INDEX IF NOT EXISTS idx_stop_times_stop_arrival ON stop_times (stop_id, arrival_time);
```

Line 427 (new in this PR) attempts to create the same-named index with an additional column:
```sql
CREATE INDEX IF NOT EXISTS idx_stop_times_stop_arrival ON stop_times (stop_id, arrival_time, departure_time);
```

Since the index name `idx_stop_times_stop_arrival` already exists from the earlier migration, the `IF NOT EXISTS` clause causes the 3-column version to silently no-op on any database that has already run the original migration. **The new index definition will never be applied to existing databases.**

Tests pass because they use fresh in-memory databases where all migrations run sequentially, but the earlier migration creates the 2-column index first and the new one is skipped.

The PR description mentions "Added `DROP INDEX IF EXISTS idx_stop_times_stop_arrival` in order to safely upgrade the previous index migration" and "Added `DROP INDEX IF EXISTS idx_trips_route_id` to remove the overlapping single-column index" — but neither DROP statement appears in the actual diff.

**Fix**: Add a `DROP INDEX IF EXISTS` migration step before the new `CREATE INDEX` statement, similar to the existing pattern at lines 332-336:

```sql
-- migrate
DROP INDEX IF EXISTS idx_stop_times_stop_arrival;

-- migrate
CREATE INDEX IF NOT EXISTS idx_stop_times_stop_arrival ON stop_times (stop_id, arrival_time, departure_time);
```

**File**: `gtfsdb/schema.sql:426-427`

## Important Issues

### 2. `idx_trips_route_id` is now redundant but not dropped

The existing `idx_trips_route_id ON trips (route_id)` at line 324 is now redundant — the new `idx_trips_route_service ON trips (route_id, service_id)` covers all queries that filter on `route_id` alone (SQLite can use a composite index for leading-column lookups). Keeping both wastes disk space and slows down inserts for every trip row.

The PR description mentions dropping this index, but no DROP statement is in the diff.

**Fix**: Add a `DROP INDEX IF EXISTS idx_trips_route_id` migration step before or after the new composite index creation.

**File**: `gtfsdb/schema.sql`

## Suggestions

### 3. Test fix is good but consider using a date that matches the RABA test data

The change from `time.Now()` to `time.Date(2024, 1, 1, ...)` is a solid fix for the non-deterministic test behavior. If the RABA test dataset has a specific service date range, using a date within that range would make the test more meaningful. Minor point — the current fix is functional.

**File**: `internal/restapi/trips_helper_test.go:1723`

## Strengths

- Well-targeted composite indexes that match actual query patterns
- `idx_trips_route_service` directly serves `GetTripsForRouteInActiveServiceIDs` (WHERE route_id AND service_id IN)
- `idx_trips_route_headsign` is a covering index for the `GetShapesGroupedByTripHeadSign` subquery
- `ORDER BY id LIMIT 1` ensures deterministic shape selection after index changes
- Test fix correctly replaces non-deterministic `time.Now()` with a fixed date and upgrades `assert.NotNil` to `require.NotNil` to prevent nil-pointer dereference

## Recommended Action

1. **Add the missing DROP INDEX statements** — without these, the `stop_times` index upgrade is silently broken on existing databases
2. **Drop the now-redundant `idx_trips_route_id`** as mentioned in the PR description
3. Re-run `make test && make lint` and resubmit

PR #521, opened 2026-02-28
feat: Implement protected API key authentication for problem report endpoints.

AhmedAlian7:
# Description

This PR implements a security layer for the problem report retrieval endpoints (`/api/where/problem-reports-for-trip` and `/api/where/problem-reports-for-stop`). Following maintainer feedback, these endpoints are now restricted to authorized transit agency staff and internal tools to protect sensitive data. Problem report submissions remain open to standard API keys to encourage user feedback.

- Closes #461

## Changes

1. **Configuration**: 
   - Added a new `protected-api-keys` property to `config.schema.json` and internal configuration structs.
   - Authorized keys can be configured via `config.json` or the `GTFS_PROTECTED_API_KEYS` environment variable.

2. **Middleware Implementation**:
   - Created `validateProtectedAPIKey` middleware to extract and validate API keys from the `X-API-Key` header (with a fallback to the `key` query parameter).
   - Utilizes `crypto/subtle.ConstantTimeCompare` for secure key comparison to mitigate timing attacks.
   - Unauthorized requests immediately receive a `401 Unauthorized` response.

3. **Route Protection**:
   - Applied a new `withProtectedCombinedID` helper to the retrieval endpoints, which validates the protected API key before proceeding with standard ID validation, rate limiting, and compression.
   - Write-endpoints (submissions) continue to use standard validation (`withCombinedID`).

4. **Testing**:
   - Added unit tests for the `validateProtectedAPIKey` middleware verifying correct behavior with valid, invalid, missing, and standard unauthorized keys.
   - Updated integration tests for problem report retrieval handlers to enforce `401 Unauthorized` responses when standard keys are used, ensuring tests pass successfully with the new protection.
   - Resolved `appconf` test regressions caused by new configuration requirements.

PR #531, opened 2026-03-01
chore: fix missing trailing newlines in JSON configs

AhmedAlian7:
**This PR restores the missing newlines to the JSON configuration files. This addresses maintainer feedback from PR #521 to prevent unnecessary diff noise in future commits.**


aaronbrethorst (approved):
Ahmed — clean and to the point. Thanks for circling back to fix these from the #521 feedback.

## Verdict

Approve. No issues.

PR #536, opened 2026-03-02
fix: propagate context into static GTFS HTTP download

suhr25:
**This PR makes **GTFS downloads respect context cancellation**. Previously, `ForceUpdate(ctx)` accepted a context but the HTTP request ignored it. Changes are mainly in `internal/gtfs/static.go`, with small updates in `gtfs_manager.go`.**

## FIX 

* **Issue:** HTTP requests were created without context, so cancellation/timeouts from `ForceUpdate` had no effect.
* **Fix:** Pass `ctx` through `loadGTFSData` → `rawGtfsData` and use `http.NewRequestWithContext`.

```go
// BEFORE
req, err := http.NewRequest("GET", source, nil)

// AFTER
req, err := http.NewRequestWithContext(ctx, "GET", source, nil)
```

## VERIFICATION

* Run: `go test ./...`
* Test with a slow GTFS endpoint and a short timeout:

  * Call `ForceUpdate` with ~500ms context
  * **Before:** waits ~5 minutes
  * **After:** exits quickly with `context deadline exceeded`

### DIFF
--- a/cmd/api/app.go
+++ b/cmd/api/app.go
@@ -67,10 +67,10 @@ func ParseAPIKeys(apiKeysFlag string) []string {
 // BuildApplication creates and initializes the Application with all dependencies.
 // This includes creating the logger, initializing the GTFS manager, and creating the direction calculator.
 // Returns an error if GTFS manager initialization fails.
-func BuildApplication(cfg appconf.Config, gtfsCfg gtfs.Config) (*app.Application, error) {
+func BuildApplication(ctx context.Context, cfg appconf.Config, gtfsCfg gtfs.Config) (*app.Application, error) {
 	logger := slog.New(slog.NewTextHandler(os.Stdout, nil))
 
-	gtfsManager, err := gtfs.InitGTFSManager(context.Background(), gtfsCfg)
+	gtfsManager, err := gtfs.InitGTFSManager(ctx, gtfsCfg)
 	if err != nil {
 		return nil, fmt.Errorf("failed to initialize GTFS manager: %w", err)
 	}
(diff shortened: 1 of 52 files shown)


aaronbrethorst (approved):
Hey Suhrid — clean fix. The `ForceUpdate` method accepted a context but the HTTP request underneath was ignoring it entirely. Threading `ctx` through `loadGTFSData` → `rawGtfsData` → `http.NewRequestWithContext` is exactly the right fix, and keeping `context.Background()` for the startup path in `InitGTFSManager` is correct since there's no meaningful caller context at that point.

## Critical

No critical issues.

## Important

No important issues.

## Fit and Finish

No fit-and-finish issues. Single commit, minimal diff, correct change.

## Note

PR #513 is also modifying `InitGTFSManager` and `loadGTFSData` to accept `context.Context`. You'll need to fix this merge conflict before I can merge your PR.

aaronbrethorst (changes requested):
Hey Suhrid, thanks for resolving the merge conflict. The context propagation fix itself is still clean, and the conflict resolution correctly preserved the `ctx` threading through the startup retry loop in `InitGTFSManager`.

One small thing to fix before we can merge:

## Fit and Finish

1. **Run `go fmt`** — The merge conflict resolution introduced inconsistent struct field alignment in `internal/gtfs/config.go`. Lines 28–31 (`Env`, `Verbose`, `EnableGTFSTidy`, `StartupRetries`) use narrower alignment than the fields above them. Running `go fmt ./...` will fix this automatically.

fletcherw:
Closing stale PR, rawGTFSData already uses NewRequestWithContext now.

PR #539, opened 2026-03-02
perf: replace N+1 queries with batch fetches in arrivals-and-departur…

### LINKED ISSUE #538: perf: eliminate N+1 queries in arrivals-and-departures-for-stop
## Problem

The `arrivals-and-departures-for-stop` handler performs multiple
per-iteration database queries that scale linearly with:

- Number of active arrivals
- Number of stops referenced in the response

Specifically:

1. `GetStopTimesForTrip(tripID)` is called once per arrival to compute
   `totalStopsInTrip`.
2. `GetStop` and `GetRoutesForStops` are called once per stop in
   `stopIDSet` when building references.

This results in O(A + 2S) queries per request, where:

- A = active arrivals
- S = stops in stopIDSet

Under large datasets and concurrent load, this can significantly impact
latency and database throughput.

## Proposed Fix

- Replace per-arrival stop time queries with a single
  `GetStopTimesForTripIDs` batch call.
- Replace per-stop reference queries with
  `GetStopsByIDs` and `GetRoutesForStops` batch calls.
- Perform in-memory lookups inside loops.

## Expected Outcome

Reduce database queries to a fixed number per request, improving
scalability and load performance without changing API behavior.

Adityatorgal17:
## Description

This PR removes two sequential query bottlenecks in:

    GET /api/where/arrivals-and-departures-for-stop/{id}

(internal/restapi/arrivals_and_departure_for_stop.go)

Per-iteration DB calls have been replaced with batch queries that already
exist in the `gtfsdb` layer.

Closes #538 

---

### 1. totalStopsInTrip — N+1 inside main arrival loop (HIGH impact)

**Before**

`GetStopTimesForTrip(tripID)` was called once per active arrival inside
the main loop solely to compute `len(result)` for `totalStopsInTrip`.

**After**

A single:

    GetStopTimesForTripIDs(uniqueTripIDs)

call is made after the existing batch trip/route fetch.

Results are aggregated into:

    tripStopCountMap[tripID]

The loop body now performs:

    totalStopsInTrip := tripStopCountMap[st.TripID]

which is O(1).

---

### 2. Stop reference building — 2N queries in stopIDSet loop (MEDIUM impact)

**Before**

For each stop in `stopIDSet`, two queries were executed:

- `GetStop(ctx, stopID)`
- `GetRoutesForStops(ctx, []string{stopID})`

**After**

All stop IDs are collected into a slice and fetched in two batch calls:

    GetStopsByIDs(ctx, stopIDsSlice)
    GetRoutesForStops(ctx, stopIDsSlice)

Results are indexed into:

- `stopsMap`
- `routesByStop`

The loop now performs only in-memory lookups.

---

## Impact

**Performance**

Reduces per-request DB queries from:

    O(A + 2S)

to:

    O(1) fixed batch queries

Where:
- A = number of active arrivals
- S = number of stops in stopIDSet

**API Contract**

- No response structure changes
- No ordering changes
- No behavioral changes

**SQL**

- No new queries introduced
- Only existing batch query functions used

---

## Testing

- Ran `make fmt`
- Ran `make lint`
- Ran `make test`

### DIFF
--- a/internal/restapi/arrivals_and_departure_for_stop.go
+++ b/internal/restapi/arrivals_and_departure_for_stop.go
@@ -247,6 +247,19 @@ func (api *RestAPI) arrivalsAndDeparturesForStopHandler(w http.ResponseWriter, r
 		tripsLookup[trip.ID] = trip
 	}
 
+	// Batch-fetch stop counts per trip to avoid per-arrival N+1 queries for totalStopsInTrip.
+	tripStopCountMap := make(map[string]int, len(uniqueTripIDs))
+	if len(uniqueTripIDs) > 0 {
+		allStopTimesForTrips, err := api.GtfsManager.GtfsDB.Queries.GetStopTimesForTripIDs(ctx, uniqueTripIDs)
+		if err != nil {
+			api.Logger.Warn("failed to batch fetch stop times for trips", slog.Any("error", err))
+		} else {
+			for _, st := range allStopTimesForTrips {
+				tripStopCountMap[st.TripID]++
+			}
+		}
+	}
+
 	for _, ast := range allActiveStopTimes {
 		st := ast.GetStopTimesForStopInWindowRow
 
@@ -419,16 +432,7 @@ func (api *RestAPI) arrivalsAndDeparturesForStopHandler(w http.ResponseWriter, r
 			predictedDepartureTime = 0
 		}
 
-		tripStopTimes, err := api.GtfsManager.GtfsDB.Queries.GetStopTimesForTrip(ctx, st.TripID)
-		var totalStopsInTrip int
-		if err != nil {
-			api.Logger.Debug("failed to get stop times for trip",
-				slog.String("tripID", st.TripID),
-				slog.Any("error", err))
-			totalStopsInTrip = 0
-		} else {
-			totalStopsInTrip = len(tripStopTimes)
-		}
+		totalStopsInTrip := tripStopCountMap[st.TripID]
 
 		blockTripSequence := api.calculateBlockTripSequence(ctx, st.TripID, serviceMidnight)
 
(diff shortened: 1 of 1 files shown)


aaronbrethorst (approved):
Hey Aditya — solid optimization work. Replacing per-iteration DB queries with batch fetches is the right pattern, and using existing `GetStopTimesForTripIDs`, `GetStopsByIDs`, and `GetRoutesForStops` queries keeps the SQL layer unchanged. The PR description is also well-organized with clear before/after explanations. A couple of small things to be aware of.

## Critical

No critical issues.

## Important

No important issues.

## Fit and Finish

### 1. `GetStopTimesForTripIDs` fetches all columns just to count rows

`arrivals_and_departure_for_stop.go:252-260`:

```go
allStopTimesForTrips, err := api.GtfsManager.GtfsDB.Queries.GetStopTimesForTripIDs(ctx, uniqueTripIDs)
// ...
for _, st := range allStopTimesForTrips {
    tripStopCountMap[st.TripID]++
}
```

The query is `SELECT * FROM stop_times WHERE trip_id IN (...)`, which fetches every column (arrival_time, departure_time, stop_id, headsign, etc.) when only `trip_id` is needed for counting. For a request with 30 active arrivals averaging 40 stops per trip, that's ~1200 full rows transferred just to increment a counter.

A purpose-built `SELECT trip_id, COUNT(*) FROM stop_times WHERE trip_id IN (...) GROUP BY trip_id` would avoid transferring unused columns. That said, I understand the design choice to use existing queries — this can be optimized later if profiling shows it matters.

### 2. Subtle behavior change on batch route query failure

In the old code, if `GetRoutesForStops` failed for a single stop, that stop was skipped entirely (`continue`). In the new code, if the batch `GetRoutesForStops` fails, all stops are included in references but with empty `RouteIDs`. This is arguably better behavior (a stop existing with no routes is more useful than a missing stop reference), but worth noting as a behavioral difference.

## Strengths

- Uses existing batch queries — no SQL changes needed
- Single commit, focused change
- Clear PR description with before/after and complexity analysis
- Log levels are appropriate: `Warn` for batch failures, `Debug` for individual missing stops
- The `if len(uniqueTripIDs) > 0` guard avoids unnecessary empty-slice queries

## Verdict

Merge. Tests pass, lint clean, single commit. The optimization is correct and uses established patterns from the codebase. The fit-and-finish items are informational — no changes required.

Issue #540, opened 2026-03-02
Concurrency: Fix Inconsistent Lock Ordering Causing Potential Deadlock in VehiclesForAgencyID()


AhmedAlian7:
### Problem

`VehiclesForAgencyID()` introduces a potential deadlock due to inconsistent lock acquisition ordering. The function requires the caller to hold `manager.RLock()` (static mutex). However, it internally calls `GetRealTimeVehicles()`, which acquires a separate mutex (`realTimeMutex`). If another goroutine acquires `realTimeMutex` first and then attempts to acquire `manager.RLock()`, a circular wait condition may occur, resulting in a deadlock.

### File References

- `internal/gtfs/gtfs_manager.go:324-341` (comment states caller must hold `manager.RLock()` before calling the method)
- Line 331 calls `manager.GetRealTimeVehicles()` which acquires `realTimeMutex`

### Code Reference

    // IMPORTANT: Caller must hold manager.RLock() before calling this method.
    func (manager *Manager) VehiclesForAgencyID(agencyID string) []gtfs.Vehicle {
        routes := manager.RoutesForAgencyID(agencyID)

        for _, v := range manager.GetRealTimeVehicles() { // Acquires realTimeMutex!

### Risk

Potential circular wait between `staticMutex` (via `RLock`) and `realTimeMutex`. High risk under concurrent workloads. Could lead to service stalls, blocked goroutines, and degraded system stability.

### Proposed Solution

1. Audit all lock acquisition orders across the codebase.
2. Establish and document a formal lock ordering policy.
3. Refactor to avoid holding multiple locks simultaneously where possible.
4. Alternatively, call `GetRealTimeVehicles()` before acquiring `manager.RLock()` after verifying correctness.
5. Add explicit lock order documentation in `gtfs_manager.go`.

### Testing

Run full test suite with race detector:

    go test -race ./...

Validate no deadlock scenarios under concurrent load testing.

PR #541, opened 2026-03-02
fix: eliminate potential deadlock in VehiclesForAgencyID and enforce lock ordering policy

AhmedAlian7:
## Description

- closes: #540

This PR resolves a concurrency issue in `VehiclesForAgencyID()` by eliminating inconsistent lock ordering and formally documenting a lock acquisition policy within `gtfs_manager.go`.

Previously, the method required callers to hold `staticMutex` via `RLock()` and then internally called `GetRealTimeVehicles()`, which acquires `realTimeMutex`. This created a nested locking scenario that could lead to circular wait conditions if another code path acquired the locks in reverse order.

---

### 1. Lock Ordering Policy — `gtfs_manager.go`

Added formal documentation to the `Manager` struct establishing that:

> `staticMutex` must always be acquired before `realTimeMutex`.

This defines a clear global lock acquisition order and prevents circular-wait deadlocks system-wide.

---

### 2. Refactored `VehiclesForAgencyID` — `gtfs_manager.go`

**Before**
- Required caller to hold `staticMutex` (`RLock()`).
- Internally called `GetRealTimeVehicles()` which acquires `realTimeMutex`.
- Nested locking created deadlock risk.

**After**
- Method manages its own locking internally.
- Callers must NOT hold any `Manager` locks.
- Uses a two-phase locking approach:

Phase 1:
- Acquire `staticMutex.RLock()`
- Collect route IDs
- Release `staticMutex`

Phase 2:
- Call `GetRealTimeVehicles()` (independently acquires `realTimeMutex`)
- Filter vehicles using precomputed route ID map

Additional improvements:
- Preallocated `routeIDs` map capacity using `len(routes)`
- Avoided repeated `GetRealTimeVehicles()` calls by storing result in `rtVehicles`
- Simplified conditional filtering logic

---

### 3. Reduced Handler Lock Scope — `vehicles_for_agency_handler.go`

**Before**
- `RLock()` held for the entire handler execution, including:
  - DB queries
  - Response model construction

**After**
- `RLock()` held only for the `FindAgency` call
- Lock released immediately afterward
- Significantly reduced lock contention window

---

## Impact

- Eliminates potential circular wait deadlock between `staticMutex` and `realTimeMutex`
- Removes unsafe external locking requirement from `VehiclesForAgencyID`
- Establishes explicit lock ordering policy for future development
- Reduces lock scope in REST handler
- Improves concurrency safety under high load

### DIFF
--- a/internal/restapi/vehicles_for_agency_handler.go
+++ b/internal/restapi/vehicles_for_agency_handler.go
@@ -41,8 +41,8 @@ func (api *RestAPI) vehiclesForAgencyHandler(w http.ResponseWriter, r *http.Requ
 		}
 	}
 	routeIDs := make([]string, 0, len(routeIDSet))
-	for id := range routeIDSet {
-		routeIDs = append(routeIDs, id)
+	for routeID := range routeIDSet {
+		routeIDs = append(routeIDs, routeID)
 	}
 	routes, err := api.GtfsManager.GtfsDB.Queries.GetRoutesByIDs(ctx, routeIDs)
 	if err != nil {
@@ -50,8 +50,8 @@ func (api *RestAPI) vehiclesForAgencyHandler(w http.ResponseWriter, r *http.Requ
 		return
 	}
 	routeByID := make(map[string]gtfsdb.Route, len(routes))
-	for _, r := range routes {
-		routeByID[r.ID] = r
+	for _, route := range routes {
+		routeByID[route.ID] = route
 	}
 
 	// Maps to build references


aaronbrethorst (approved):
Ahmed, this is a really clean concurrency fix. The two-phase locking approach in `VehiclesForAgencyID` is the right design -- acquiring and releasing `staticMutex` before independently acquiring `realTimeMutex` eliminates the nested lock pattern entirely, which is even better than just enforcing lock ordering. The lock ordering policy documentation on the `Manager` struct is a welcome addition that will help future contributors avoid similar issues.

I verified the design from several angles and everything checks out:

- **TOCTOU gap between phases**: Static data could theoretically change between collecting route IDs and filtering vehicles, but this is benign for a read-only API serving best-effort snapshots. Missed or stale vehicles resolve on the next request.
- **Handler safety after early `RUnlock()`**: The `agency` pointer remains valid because Go's GC keeps the old `gtfs.Static` data alive as long as the handler holds a reference. The old struct is never mutated -- it's replaced entirely via a new map.
- **All callers verified**: The handler, handler test, and concurrency test all call `VehiclesForAgencyID` without holding external locks, consistent with the new contract.
- **Variable renames**: `id` → `routeID` and `r` → `route` are pure renames that eliminate variable shadowing -- nice touch.

No issues found. Approved.

PR #543, opened 2026-03-02
perf(docker): optimize build times with BuildKit and GHA caching

### LINKED ISSUE #542: Optimize Docker builds with BuildKit and GitHub Actions caching
### The Problem

Currently, our Docker builds (both local and in CI) are taking longer than necessary. Because we aren't using BuildKit cache mounts, whenever a small code change is made, Docker invalidates the layer and redownloads all the Go modules from scratch. Furthermore, the Go compiler doesn't have a persistent build cache, so it has to recompile the entire project every time instead of just the files that changed.

### The Solution

We can drastically speed up build times by making two small changes:

1. **Dockerfiles:** Add `--mount=type=cache` to the `go mod download` and `go build` steps in `Dockerfile` and `Dockerfile.dev`. This tells Docker to save the downloaded dependencies and compiled object files locally between builds.
2. **GitHub Actions:** Add `cache-from: type=gha` and `cache-to: type=gha,mode=max` to our `docker-publish.yml` workflow. This ensures that the Docker cache is saved to GitHub's native cache storage, so CI runs don't start from a cold cache every time.

### Benefits

* **Much faster local builds:** Changing a single `.go` file will only recompile that specific file.
* **Faster CI/CD pipelines:** Reduces GitHub Actions build minutes.
* **Bandwidth savings:** No more redownloading the entire dependency graph on every build.

I already have these changes ready on a local branch. If this sounds good, I can open a Pull Request right away!

ARCoder181105:
## Description
This PR significantly reduces Docker build times for both local development and CI/CD pipelines by implementing BuildKit cache mounts and GitHub Actions cache backend support.

Closes #542

## Changes Made
* **`Dockerfile` & `Dockerfile.dev`**: Added `--mount=type=cache` to the `go mod download` and `go build` steps. 
  * Caching `/go/pkg/mod` prevents redundant dependency downloads.
  * Caching `/root/.cache/go-build` allows the Go compiler to reuse compiled object files, recompiling only the code that has changed.
* **`.github/workflows/docker-publish.yml`**: Configured `cache-from: type=gha` and `cache-to: type=gha,mode=max` in the `buildx` and `buildx-release` jobs to persist the Docker build cache across GitHub Actions workflow runs.

## Impact
* **Faster CI/CD**: Workflow execution time will drop significantly after the initial cache population.
* **Faster Local Builds**: Developers making small code changes will experience near-instant rebuilds.
* **Network Efficiency**: Saves bandwidth by not redownloading the entire `go.mod` dependency tree on every run.

## Testing
* [x] CI pipeline will validate the caching mechanism upon creation of this PR (watch for `importing cache manifest from type=gha` in the build logs).

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -62,6 +62,8 @@ jobs:
           tags: |
             ghcr.io/onebusaway/maglev:latest
             ghcr.io/onebusaway/maglev:${{ github.sha }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
 
   # Push to Docker Hub on published releases (non-prerelease only)
   buildx-release:
@@ -104,3 +106,5 @@ jobs:
           tags: |
             opentransitsoftwarefoundation/maglev:${{ env.IMAGE_TAG }}
             opentransitsoftwarefoundation/maglev:latest
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,8 @@ WORKDIR /build
 
 # Copy dependency files first for better layer caching
 COPY go.mod go.sum ./
-RUN go mod download
+RUN --mount=type=cache,target=/go/pkg/mod \
+    go mod download
 
 # Copy source code
 COPY . .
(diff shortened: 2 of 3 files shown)


aaronbrethorst (changes requested):
Hey Aditya — nice initiative on the Docker build optimization. BuildKit cache mounts are the right tool for avoiding redundant `go mod download` and incremental recompilation, and the GHA caching setup for CI is the standard pattern. One issue to fix before merge.

## Critical

No critical issues.

## Important

### 1. Cache mount permission mismatch in `Dockerfile.dev` will break dev builds

`Dockerfile.dev:35-36`:

```dockerfile
RUN --mount=type=cache,target=/go/pkg/mod \
    go mod download
```

The `USER maglev` directive (line 29) means this `RUN` executes as uid 1000. However, BuildKit cache mounts default to uid/gid 0 with mode 0755 — the `maglev` user can read but not write. The `go mod download` will fail with a permission error.

Note that the existing comment on line 21 explicitly flags this concern:
```dockerfile
# Critical Fix: chown -R /go ensures the maglev user can access the Go module cache volume
```

The `chown` on line 26 fixed it for the regular filesystem, but the cache mount overlays that directory and reintroduces the problem.

**Fix**: Add `uid` and `gid` to the cache mount:
```dockerfile
RUN --mount=type=cache,target=/go/pkg/mod,uid=1000,gid=1000 \
    go mod download
```

The production `Dockerfile` is fine — it runs the build stage as root, so the default cache mount permissions work.

## Fit and Finish

No fit-and-finish issues.

## Strengths

- Clean, focused change — 11 lines added, 3 removed across 3 files
- Single commit, clear PR description with impact analysis
- Both `go mod download` and `go build` get cache mounts in the production Dockerfile (module cache + compiler cache)
- GHA cache configuration (`type=gha`, `mode=max`) is the recommended pattern for GitHub Actions
- Correctly applied to both the `buildx` and `buildx-release` jobs

## Verdict

Request changes. The `Dockerfile.dev` cache mount needs `uid`/`gid` parameters to match the non-root user. One-line fix — should be quick to address.

Issue #527, opened 2026-03-01
GTFS-Realtime: Missing per-vehicle timestamp validation causes incorrect vehicle jumps


AtharvaKathe18:
### Summary
Maglev currently applies GTFS-Realtime vehicle position updates without validating
the per-vehicle `timestamp` field. Even when a feed is fresh at the feed level,
individual vehicle entities may be stale or out of order.

This can cause vehicles to jump backwards, regress in time, or briefly show
incorrect positions — especially in unreliable network conditions.

### Why this matters
In real-world deployments (especially low-infrastructure environments):

- Vehicles may report intermittently
- Individual vehicles can lag while the feed continues updating
- GTFS-RT feeds may batch old vehicle updates together with newer ones

Without per-vehicle timestamp validation, Maglev may accept older vehicle
positions that overwrite newer data.

This is particularly problematic for the Vehicle Tracker project, where
vehicle data is crowd-sourced and network reliability cannot be assumed.

### Current behavior
- Feed-level freshness is handled
- Vehicle entities are accepted as long as the feed update is applied
- No comparison is made between:
  - Existing vehicle timestamp
  - Incoming vehicle timestamp

### Expected behavior
Possible approaches (open to maintainer guidance):

- Reject per-vehicle updates if incoming `vehicle.Timestamp` is older
- Keep the most recent vehicle position per vehicle ID
- Log but ignore stale per-vehicle updates
- Make behavior configurable (strict vs permissive)

### Proposed work
- Add per-vehicle timestamp comparison during vehicle merge
- Ensure newer vehicle positions are never overwritten by older ones
- Add unit tests covering:
  - Newer → older vehicle updates
  - Missing timestamps
  - Mixed fresh/stale vehicles in same feed

### Scope
- Affects `internal/gtfs/realtime.go`
- Testable via unit tests (no API changes required)
- Backwards compatible

### Related
- #476 (feed-level stale vehicle handling)

I’m happy to take this on and submit a PR if the direction is acceptable.

PR #550, opened 2026-03-02
realtime: ignore stale per-vehicle updates and log them

AtharvaKathe18:
###  Description

This PR adds per-vehicle timestamp validation when applying GTFS-Realtime vehicle position updates.
While Maglev already handles feed-level freshness, individual vehicle entities within a fresh feed may still be stale or out of order. This can cause newer vehicle positions to be overwritten by older ones, leading to vehicles jumping backwards or briefly showing incorrect locations.

### **Problem**

In real-world deployments, especially in low-infrastructure or unreliable network environments:
Vehicles may report intermittently
Individual vehicles may lag while the feed continues updating
GTFS-RT feeds may batch old and new vehicle updates together
Currently, vehicle entities are applied without validating their individual timestamps, allowing stale vehicle updates to overwrite newer data.

### **Solution**

Added per-vehicle timestamp comparison during vehicle merge
Incoming vehicle updates are ignored if their timestamp is older than the currently stored value for the same vehicle ID
Ensures the most recent vehicle position is always preserved
Stale per-vehicle updates do not affect other vehicles in the same feed

### **Tests**

Added unit tests covering:
Newer → older per-vehicle updates
Vehicle updates with missing timestamps
Mixed fresh and stale vehicle entities within the same feed

### **Scope**

Affects: internal/gtfs/realtime.go
No API changes
Fully backwards compatible

###  Related  Issues

#527  – GTFS-Realtime feed-level stale vehicle handling

### DIFF
--- a/internal/gtfs/realtime.go
+++ b/internal/gtfs/realtime.go
@@ -345,7 +345,13 @@ func (manager *Manager) updateFeedRealtime(ctx context.Context, feedCfg RTFeedCo
 
 				if prev, exists := prevByID[v.ID.ID]; exists {
 					if isVehicleStale(prev, v) {
-						// Keep newer existing vehicle, drop stale update
+						// Log and keep the newer existing vehicle, dropping the stale update
+						logging.LogOperation(logger, "skipping_stale_vehicle_entity",
+							slog.String("feed", feedID),
+							slog.String("vehicle_id", v.ID.ID),
+							slog.Time("existing_timestamp", *prev.Timestamp),
+							slog.Time("incoming_timestamp", *v.Timestamp),
+						)
 						validVehicles = append(validVehicles, prev)
 						continue
 					}
--- a/internal/gtfs/realtime_test.go
+++ b/internal/gtfs/realtime_test.go
@@ -1,8 +1,10 @@
 package gtfs
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"log/slog"
 	"net/http"
 	"net/http/httptest"
 	"os"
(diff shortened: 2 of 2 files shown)


aaronbrethorst (approved):
Atharva, nice work turning an invisible operational event into a visible one. Adding logging to the stale-vehicle-skip path is the right call -- operators need to know when individual vehicle updates are being rejected, especially in the unreliable-network scenarios your PR description identifies.

I verified the safety and correctness from several angles:

- **Nil-pointer safety**: The `isVehicleStale` guard (line 55) returns `false` when either timestamp is nil, so the `*prev.Timestamp` and `*v.Timestamp` dereferences at lines 352-353 are guaranteed safe. The `v.ID` nil check at line 342 guards `v.ID.ID`. All good.
- **Logger safety**: `LogOperation` has an explicit nil-logger guard, and `FromContext` falls back to `slog.Default()`. Safe.
- **Test coverage**: Three well-structured integration tests covering stale rejection, mixed fresh/stale feeds, and nil timestamps. The tests exercise the full pipeline through `updateFeedRealtime` rather than just unit-testing `isVehicleStale`, which gives confidence that the logging works in context.
- **`encodeVehicleFeed` helper**: Clean design, auto-generates entity IDs, uses `panic` on marshal failure (appropriate for test helpers).

## Fit and Finish

### 1. Redundant import alias

`internal/gtfs/realtime_test.go:21` -- The import alias is unnecessary since the package is already named `logging`:

```go
// Current:
logging "maglev.onebusaway.org/internal/logging"

// Should be:
"maglev.onebusaway.org/internal/logging"
```

Every other file in `internal/gtfs/` imports this package without an alias.

### 2. Consider log volume in production (optional)

The per-vehicle log fires at `Info` level via `LogOperation`. Unlike the feed-level stale skip at line 316 (which fires at most once per feed per poll), this fires once per stale vehicle per poll. In the exact "low-infrastructure or unreliable network environments" you're targeting, a feed with 500 vehicles could easily have 50+ stale entities per poll cycle, producing ~100 Info lines per minute per feed.

One option to consider: use `Debug` level for the per-vehicle detail and emit a single aggregated `Info`-level line after the loop with the count of skipped vehicles. But the current approach is consistent with the existing pattern at line 316, so this is entirely optional.

No blocking issues. Approved.

PR #556, opened 2026-03-02
Fix: Unify predicted arrival/departure times across singular and plural endpoints

### LINKED ISSUE #555: Inconsistent predicted times between singular and plural arrivals endpoints
## Bug

The `arrival-and-departure-for-stop` (singular) and `arrivals-and-departures-for-stop` (plural) endpoints return **different predicted arrival/departure times** for the same stop, trip, and real-time data.

## Root Cause

Prediction logic is duplicated across two independent code paths that have diverged:

- **Plural handler** (`arrivals_and_departure_for_stop.go`) — inline logic with 3-tier fallback
- **Singular handler** (`arrival_and_departure_for_stop_handler.go`) — `getPredictedTimes()` with only 2-tier fallback

The singular endpoint is **missing the trip-level `TripUpdate.Delay` fallback**, so when an agency provides only trip-level delay (no per-stop `StopTimeUpdates`), the singular endpoint returns `predicted=false` while the plural endpoint correctly returns the delayed time.

## Impact

Riders see "Bus arriving in 7 min" in the arrivals list, but tapping the detail view shows "No real-time data." Common with smaller agencies that provide trip-level delays only.

## Affected Files

- `internal/restapi/arrival_and_departure_for_stop_handler.go` — `getPredictedTimes()` missing Tier 3 fallback
- `internal/restapi/arrivals_and_departure_for_stop.go` — duplicated inline prediction logic

## Fix

1. Add trip-level `Delay` fallback to `getPredictedTimes()`
2. Refactor plural handler to call shared `getPredictedTimes()` instead of inline logic
3. Add tests for trip-level delay fallback and cross-endpoint consistency

ARCoder181105:
## Summary

Fixes inconsistent predicted arrival/departure times between the singular (`arrival-and-departure-for-stop`) and plural (`arrivals-and-departures-for-stop`) endpoints by consolidating duplicated prediction logic into a shared `getPredictedTimes()` function.

## Problem

The two endpoints used separate, diverged code paths for computing real-time predictions from GTFS-RT data. The singular endpoint was missing the trip-level `TripUpdate.Delay` fallback, causing it to return `predicted=false` when agencies provide only trip-level delay without per-stop `StopTimeUpdates`.

### `arrival_and_departure_for_stop_handler.go`
- Add trip-level `Delay` fallback (Tier 3) to `getPredictedTimes()`
- Update return signature to `(int64, int64, bool)`
- Fix early exit when `StopTimeUpdates` is empty but trip-level `Delay` exists
- Fix `propagatedDelay` not resetting when closest prior stop lacks `Delay`
- Add operator precedence parentheses for clarity

### `arrivals_and_departure_for_stop.go`
- Replace inline prediction logic with call to shared `getPredictedTimes()`
- Remove `vehicle != nil` gate from prediction path

### `arrival_and_departure_for_stop_handler_test.go`
- Update existing tests for new 3-return-value signature
- Add `TestGetPredictedTimes_TripLevelDelayFallback`
- Add `TestGetPredictedTimes_DelayPropagationLogic`

### `arrivals_and_departures_for_stop_handler_test.go`
- Add `TestPluralArrivals_TripLevelDelayWithoutVehicle`

## Testing

All three prediction tiers verified for both endpoints:
1. ✅ Exact stop match
2. ✅ Propagated delay from closest prior stop
3. ✅ Trip-level `TripUpdate.Delay` fallback (previously broken on singular)

***Closes #555***

### DIFF
--- a/internal/restapi/arrival_and_departure_for_stop_handler.go
+++ b/internal/restapi/arrival_and_departure_for_stop_handler.go
@@ -280,9 +280,11 @@ func (api *RestAPI) arrivalAndDepartureForStopHandler(w http.ResponseWriter, r *
 		predictedArrivalTime = scheduledArrivalTimeMs
 		predictedDepartureTime = scheduledDepartureTimeMs
 
-		predictedArrival, predictedDeparture := api.getPredictedTimes(tripID, stopCode, targetStopTime.StopSequence, scheduledArrivalTime, scheduledDepartureTime)
+		// getPredictedTimes now returns 3 values (arr, dep, isPredicted)
+		// and includes trip-level Delay fallback for consistency with the plural handler
+		predictedArrival, predictedDeparture, isPredicted := api.getPredictedTimes(tripID, stopCode, targetStopTime.StopSequence, scheduledArrivalTime, scheduledDepartureTime)
 
-		if predictedArrival != 0 && predictedDeparture != 0 {
+		if isPredicted {
 			predictedArrivalTime = predictedArrival
 			predictedDepartureTime = predictedDeparture
 			predicted = true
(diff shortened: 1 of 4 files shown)


aaronbrethorst (approved):
Excellent unification work, Aditya. Consolidating the duplicated prediction logic into the shared `getPredictedTimes()` function eliminates a real inconsistency between the singular and plural endpoints — the singular endpoint was silently dropping predictions when agencies only provide trip-level `Delay` without per-stop `StopTimeUpdates`. The `propagatedDelay = 0` reset fix and the removal of the `vehicle != nil` gate are both correct and well-tested improvements.

No blocking issues found. This is ready to merge.

## Verification

- All tests pass
- Lint clean (0 issues)
- Unit consistency verified: `getPredictedTimes()` works internally in nanoseconds (`time.Duration` values), converts to milliseconds only at the final `.UnixMilli()` call — no precision loss
- The `vehicle != nil` removal is safe: `getPredictedTimes()` only reads `TripUpdate` data, which is independent of vehicle presence. The new `TestPluralArrivals_TripLevelDelayWithoutVehicle` test directly validates this.
- The 3-tier fallback ordering (exact match > propagated delay > trip-level delay) is structurally enforced by the `if/else if/else` chain at lines 591–605

## Strengths

- **Fixes a real production bug**: The singular endpoint was missing trip-level `Delay` fallback (Tier 3), causing it to return `predicted=false` for agencies that provide only trip-level delay without per-stop data. Now both endpoints behave identically.
- **`propagatedDelay = 0` reset** (line 579): Fixes a subtle bug where a closer prior stop with no delay data would inherit the delay from a farther prior stop. The reset ensures only the closest stop's actual delay is propagated.
- **Removing the `vehicle != nil` gate**: A valid GTFS-RT scenario is a TripUpdate with delay data but no vehicle position — this now correctly produces predictions. Good pairing of positive test (`TripLevelDelayWithoutVehicle`) and negative test (`VehiclePositionAloneDoesNotPredict`).
- **Strong test coverage**: New tests directly target the exact bug scenarios (empty StopTimeUpdates with trip-level delay, predictions without vehicle).

## Suggestion

**Clarify operator precedence on line 539** (`arrival_and_departure_for_stop_handler.go`):

The current parenthesization is misleading:
```go
if realTimeTrip == nil || (len(realTimeTrip.StopTimeUpdates) == 0) && realTimeTrip.Delay == nil {
```

The parens only wrap `len(...) == 0` but not the whole `&&` clause, making it look like the `||` groups with the `len` check. Go's operator precedence (`&&` binds tighter than `||`) makes this work correctly, but a future reader will likely misread it. Consider:

```go
if realTimeTrip == nil || (len(realTimeTrip.StopTimeUpdates) == 0 && realTimeTrip.Delay == nil) {
```

This makes the intent — "bail out if trip is nil, OR if there's neither stop-time data nor trip-level delay" — unambiguous.

PR #562, opened 2026-03-03
fix(restapi): replace incorrect ctx.Err() handling with clientCanceledResponse

### LINKED ISSUE #561: Context cancellation produces wrong HTTP status in API handlers
## Problem

When a client disconnects mid-request or a server-side deadline fires, two incorrect behaviours occur across the REST API:

### **Silent 200 OK**

Handlers that check `if ctx.Err() != nil { return }` return with no body and the default **200 OK** status from `httptest.ResponseRecorder`. The client (if still connected) receives an empty success response.

### **Wrong 500 Internal Server Error**

Handlers that call `serverErrorResponse(w, r, ctx.Err())` log:

```
ERROR msg="internal server error" error="context canceled"
```

and return **500 Internal Server Error**. A cancelled request is not a server fault.

---

Both cases are also triggered indirectly: any `if err != nil { serverErrorResponse(...) }` guard after a DB query fires when the DB returns `context.Canceled` as its query error (Go's `database/sql` propagates the context error), producing the same false **500**.

---

## Expected behaviour

- `context.Canceled` (client disconnected): log at **Info** level, write **no response** — the connection is gone, any write is futile.

- `context.DeadlineExceeded` (server timeout): log at **Warn** level, return **504 Gateway Timeout** — semantically correct, not a server fault.

Adityatorgal17:
Fixes #561

### Summary

This PR corrects improper handling of `ctx.Err()` across REST API handlers, ensuring that request cancellations and timeouts produce semantically correct HTTP responses and log levels.

---

### Problem

When a request context was cancelled or exceeded its deadline, handlers produced incorrect responses:

| Pattern | Actual Response | Expected Response |
|----------|----------------|------------------|
| `if ctx.Err() != nil { return }` | `200 OK` (empty body) | No response / `504` |
| `if ctx.Err() != nil { serverErrorResponse(...) }` | `500 Internal Server Error` | No response / `504` |
| `if err != nil { serverErrorResponse(...) }` (DB propagated ctx error) | `500 Internal Server Error` | No response / `504` |

This resulted in:
- False `200 OK` responses
- Incorrect `500 Internal Server Error` on client disconnect
- Misleading error-level logs for non-server faults

---

### Changes

- Introduced `clientCanceledResponse(w, r, err)` helper in `errors.go`:
  - `context.Canceled` → `Logger.Info("request canceled by client")` + return without writing a response
  - `context.DeadlineExceeded` → `Logger.Warn("request deadline exceeded")` + `504 Gateway Timeout`

- Replaced direct `ctx.Err()` handling in all 15 affected handlers.

- Added context-aware guarding inside `serverErrorResponse` so DB-propagated context errors are handled automatically without modifying every call site.

---

### Testing

- Updated/verified:
  - `context_cancellation_test.go`
  - `block_handler_test.go`

- Tests accept:
  - `200` (Canceled, no write)
  - `504` (DeadlineExceeded)

- `make test` passes across all packages.

### DIFF
--- a/internal/restapi/block_handler_test.go
+++ b/internal/restapi/block_handler_test.go
@@ -551,13 +551,8 @@ func TestBlockHandlerContextCancellation(t *testing.T) {
 		api.SetRoutes(mux)
 		mux.ServeHTTP(w, req)
 
-		// context.Canceled → no response written (client gone), code stays 200 from httptest.NewRecorder default
-		// context.DeadlineExceeded → 504 Gateway Timeout (server took too long, not a server fault)
-		// Either outcome is correct; the old expectation of 500 was semantically wrong.
-		assert.True(t,
-			w.Code == http.StatusGatewayTimeout ||
-				w.Code == http.StatusOK ||
-				w.Code == http.StatusInternalServerError,
-			"Expected 504 (deadline exceeded), 200 (canceled, no body written), or 500 (legacy), got: %d", w.Code)
+		// With a 1ns timeout and 1µs sleep the context is always DeadlineExceeded
+		assert.Equal(t, http.StatusGatewayTimeout, w.Code,
+			"Expected 504 Gateway Timeout for DeadlineExceeded context, got: %d", w.Code)
 	})
 }
--- a/internal/restapi/errors.go
+++ b/internal/restapi/errors.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"log/slog"
 	"net/http"
 
 	"maglev.onebusaway.org/internal/logging"
@@ -46,7 +47,7 @@ func (api *RestAPI) serverErrorResponse(w http.ResponseWriter, r *http.Request, 
 		api.clientCanceledResponse(w, r, err)
 		return
 	}
-	api.Logger.Error("internal server error", "error", err, "path", r.URL.Path)
+	logging.LogError(api.Logger, "internal server error", err, slog.String("path", r.URL.Path))
 	// Send a 500 Internal Server Error response
 	response := struct {
 		Code        int    `json:"code"`
@@ -100,7 +101,7 @@ func (api *RestAPI) validationErrorResponse(w http.ResponseWriter, r *http.Reque
 	w.WriteHeader(http.StatusBadRequest)
 	err := json.NewEncoder(w).Encode(response)
 	if err != nil {
-		api.Logger.Error("failed to encode validation error response", "error", err)
+		logging.LogError(api.Logger, "failed to encode validation error response", err)
 	}
 }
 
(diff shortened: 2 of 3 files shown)


aaronbrethorst (approved):
Aditya, clean work on the follow-up commit. All four feedback items from the first review have been addressed properly, and the resulting code is consistent and well-structured.

## Verification of Previous Review Items

### Critical #1: `http.Error()` with JSON body -- RESOLVED

`errors.go:130-146` — The `DeadlineExceeded` case now uses `w.Header().Set("Content-Type", "application/json")`, `w.WriteHeader(http.StatusGatewayTimeout)`, and `json.NewEncoder(w).Encode(response)`, matching the pattern used by `invalidAPIKeyResponse`, `serverErrorResponse`, and `validationErrorResponse`. The response struct also includes `currentTime` and `version` fields for consistency with the other error responses.

### Suggestion #2: Default case in switch -- RESOLVED

`errors.go:149-155` — A `default` case now logs a warning with the unexpected error type, path, and method. This makes the function self-documenting and prevents silent failures if called with an unexpected error.

### Suggestion #3: Redundant ctx.Err() checks in stops_for_location_handler.go -- RESOLVED

The belt-and-suspenders pattern (checking `ctx.Err()` before `serverErrorResponse`) has been removed. The remaining `ctx.Err()` checks at lines 97, 204, and 238 are standalone loop/pre-work guards, which is the correct pattern.

### Suggestion #4: Overly permissive test assertion -- RESOLVED

`block_handler_test.go:554-556` — The test now uses `assert.Equal(t, http.StatusGatewayTimeout, w.Code)` for a deterministic assertion, since the 1ns timeout + 1µs sleep guarantees `DeadlineExceeded`.

## Observations

These are non-blocking notes for awareness, not required changes:

### 1. Consider capturing `ctx.Err()` in a variable at call sites

Several call sites use the pattern:
```go
if ctx.Err() != nil {
    api.clientCanceledResponse(w, r, ctx.Err())
    return
}
```

While Go's standard library guarantees `ctx.Err()` is monotonic (once non-nil, it stays non-nil), capturing it in a variable is slightly cleaner:
```go
if err := ctx.Err(); err != nil {
    api.clientCanceledResponse(w, r, err)
    return
}
```

This is a style preference, not a correctness issue.

### 2. `default` case in `clientCanceledResponse` doesn't write a response

The `default` case logs a warning but doesn't write an HTTP response, which would result in an implicit 200 with empty body. In practice this is fine -- the only callers are explicit `ctx.Err()` checks (which always produce `Canceled` or `DeadlineExceeded`) and the `serverErrorResponse` guard (which uses `errors.Is` for the same two errors). A future-proofing option would be to write a 500 in the default case, but you'd need to inline it to avoid recursion with `serverErrorResponse`. Not blocking.

## Checklist

- [x] Tests pass
- [x] Lint clean (0 issues)
- [x] Critical #1 fixed: proper JSON Content-Type for 504 response
- [x] Suggestion #2 addressed: default case added
- [x] Suggestion #3 addressed: redundant checks removed
- [x] Suggestion #4 addressed: deterministic test assertion
- [x] All 15+ handlers consistently use `clientCanceledResponse`
- [x] `serverErrorResponse` guard intercepts DB-propagated context errors

## Strengths

- The `serverErrorResponse` guard at lines 38-49 is a force-multiplier — every handler that calls `serverErrorResponse` for DB errors automatically gets correct context cancellation handling without needing to be modified
- The 504 response struct now includes `currentTime` and `version` fields, matching the format of all other error responses in the file
- Log levels are semantically correct: Info for client cancellation, Warn for deadline exceeded
- Test is now deterministic and specifically validates the expected behavior

Issue #566, opened 2026-03-04
fix: vehicle lookup silently fails when optional static GTFS fields are null


Ahmedhossamdev:
## The issue

- While testing the server with a different GTFS feed, I noticed that some trips do not have a `block_id` in the static GTFS data. This caused
`GetVehicleForTrip` to silently return `nil` even when the vehicle was
present in the RT feed because the only lookup path was block-based.

- Trips without a `block_id` would hit the early return and never check`realTimeVehicleLookupByTrip`, so endpoints like `trip-details` and
`trip-for-vehicle` would show no vehicle data despite the vehicle being actively reported in the feed.

- This is likely a broader pattern any lookup that depends on optional static GTFS fields without a fallback will silently degrade when those
fields are absent, depending on the feed.

**Orientation (bearing)**
Some feeds do not include a `bearing` value in vehicle positions. When absent, orientation was left at `0.0` (Go zero value), which is misleading

PR #567, opened 2026-03-04
fix: graceful fallbacks for absent optional GTFS fields in vehicle lookup and orientation

Ahmedhossamdev:
Fix: #566 

## Summary

Both issues follow the same root cause: optional GTFS/GTFS-RT fields being absent cause silent degradation instead of a graceful fallback. Discovered while testing against a sound-transit feed where `block_id` and `bearing` are not always present.

### Vehicle lookup was silently returning nil for trips without `block_id`

`GetVehicleForTrip` relied entirely on block-based matching. If a trip had no `block_id` in static GTFS, it hit an early return and never checked `realTimeVehicleLookupByTrip` even when the vehicle was actively reporting that exact trip ID in the RT feed. Endpoints like `trip-details` and `trip-for-vehicle` would show no vehicle data as a result.

Fixed by adding a direct `realTimeVehicleLookupByTrip[tripID]` fast path at the top of `GetVehicleForTrip`, before the static DB query and block-based scan. If the RT feed reports the exact trip ID, the vehicle is returned immediately regardless of whether `block_id` is set in static GTFS.

### Orientation was always `0` when bearing is absent from the feed

When a feed does not include a `bearing` value in vehicle positions, orientation was left at `0.0` (Go zero value). Clients cannot distinguish "facing East" from "no data".

Fixed by inferring orientation from the heading of the closest shape segment at the vehicle's GPS position when `bearing` is nil

### DIFF
--- a/internal/restapi/trips_helper.go
+++ b/internal/restapi/trips_helper.go
@@ -1098,7 +1098,7 @@ func interpolateDistance(cumulativeDistances []float64, segmentLength float64, i
 }
 
 // inferOrientationFromShape computes the OBA orientation (degrees, 0=East, 90=North)
-// for a vehicle at (lat, lon) by finding the closest shape segment and returning its
+// for a vehicle at (lat, lon) by finding the closest shape segment and returning its heading.
 // Returns -1 if the shape has fewer than 2 points.
 func inferOrientationFromShape(lat, lon float64, shape []gtfs.ShapePoint) float64 {
 	if len(shape) < 2 {


aaronbrethorst (approved):
Ahmed, all three review items are resolved and the tests are well-crafted. Let's get this merged.

## Previous Review Items -- All Resolved

- [x] Unit tests for `inferOrientationFromShape` (5 cases: fewer than 2 points, east, north, south, closest segment selection)
- [x] Test for vehicle lookup fast path (`TestGetVehicleForTrip_DirectTripIDLookup` — creates a Manager with RT data and no block_id, verifies direct trip-ID lookup works)
- [x] Truncated doc comment fixed ("returning its heading.")

## Current Review

No issues found.

## Strengths

- The orientation test cases handle wrap-through-zero correctly (`wantMin: 355, wantMax: 5`) — that's a subtle edge case and the assertion logic is clean
- `TestGetVehicleForTrip_DirectTripIDLookup` properly uses `rebuildMergedRealtimeLocked()` to set up the lookup maps, testing the real code path rather than mocking
- The fast path in `GetVehicleForTrip` has clean lock handling — `realTimeMutex.RUnlock()` on both the hit and miss paths before returning or continuing
- The `vehicle.Position.Bearing == nil` guard correctly avoids overwriting orientation when bearing data is present
- The closest-segment selection in the "east segment over distant north segment" test case proves the geometry math works with multi-segment shapes

PR #574, opened 2026-03-04
Fix/advanced direction calc panics

### LINKED ISSUE #573: Panic Used for Concurrency Control in AdvancedDirectionCalculator
**Problem:** `AdvancedDirectionCalculator` uses `panic()` to enforce single-threaded initialization. Production code should never panic for control flow, as it forces an application crash.

**File References:**
- `internal/gtfs/advanced_direction_calculator.go` (multiple panic locations)

**Current Code:**
```go
func (adc *AdvancedDirectionCalculator) SetVarianceThreshold(threshold float64) {
    if adc.initialized.Load() {
        panic("SetVarianceThreshold called after concurrent operations have started")
    }
}
```

**Problem:** If called from the wrong place in startup, this crashes the entire application.

**Proposed Solution:**
1. Return `error` instead of panicking: `func (...) SetStandardDeviationThreshold(...) error`.
2. Update callers to handle initialization errors gracefully.
3. Update all panic sites.
4. Add initialization order documentation.

AhmedAlian7:
## Description

- closes #573 

This pull request addresses an issue where `AdvancedDirectionCalculator` was using `panic()` as a form of concurrency control during initialization. According to best practices, production code should not panic for control flow.

### Changes Made
- Modified `AdvancedDirectionCalculator` initialization methods (`SetStandardDeviationThreshold`, `SetShapeCache`, `SetContextCache`) to return `error` instead of `panic` when called after initialization.
- Added input validation to `SetStandardDeviationThreshold` to require values strictly greater than zero.
- Updated caller `DirectionPrecomputer.SetStandardDeviationThreshold` to return an `error`.
- Updated caller `DirectionPrecomputer.PrecomputeAllDirections` to handle the `SetShapeCache` error gracefully by propagating it up the stack.
- Added explanatory documentation in `DirectionPrecomputer` clarifying the asymmetric fatality between `loadShapeCache` and `SetShapeCache`.
- Updated caller `InitializeGlobalCache` to handle errors correctly and fail the global cache warmup safely instead of crashing the server.
- Updated unit tests (`advanced_direction_calculator_test.go` and `shape_cache_test.go`) to test against the new `error` returns rather than expecting `recover()`.
- Renamed `TestSetContextCache_PanicAfterInit` to `TestSetContextCache_ReturnsErrorAfterInit`, tested validation bounds on `SetStandardDeviationThreshold`, and properly propagated asynchronous errors in `TestSetContextCache_ConcurrentAccess`.

## Impact
- **Stability**: Calling the aforementioned functions out-of-order no longer crashes the GTFS application but instead results in a properly logged and handled error.
- **Code Quality**: Enforces Go idiomatic error returns over exception-style panics.

## Testing
- Unit tests (`make test`) succeed and all modified assertion logic tests out correctly.

### DIFF
--- a/internal/gtfs/advanced_direction_calculator.go
+++ b/internal/gtfs/advanced_direction_calculator.go
@@ -46,6 +46,9 @@ func (adc *AdvancedDirectionCalculator) SetStandardDeviationThreshold(threshold 
 	if adc.initialized.Load() {
 		return errors.New("SetStandardDeviationThreshold called after concurrent operations have started")
 	}
+	if threshold <= 0 {
+		return errors.New("standard deviation threshold must be greater than zero")
+	}
 	adc.standardDeviationThreshold = threshold
 	return nil
 }
@@ -67,8 +70,7 @@ func (adc *AdvancedDirectionCalculator) SetShapeCache(cache map[string][]gtfsdb.
 
 // SetContextCache injects the bulk-loaded context data.
 // IMPORTANT: This must be called before any concurrent calculation operations begin.
-// Returns an error if called after internal state has been initialized (i.e., after the first
-// fallback to shape-based calculation).
+// Returns an error if called after CalculateStopDirection has been invoked.
 func (adc *AdvancedDirectionCalculator) SetContextCache(cache map[string][]gtfsdb.GetStopsWithShapeContextRow) error {
 	adc.cacheMutex.Lock()
 	defer adc.cacheMutex.Unlock()
--- a/internal/gtfs/advanced_direction_calculator_test.go
+++ b/internal/gtfs/advanced_direction_calculator_test.go
@@ -205,6 +205,16 @@ func TestStandardDeviationThreshold(t *testing.T) {
 	err := calc.SetStandardDeviationThreshold(1.0)
 	assert.NoError(t, err)
 	assert.Equal(t, 1.0, calc.standardDeviationThreshold)
+
+	// Test invalid zero threshold
+	err = calc.SetStandardDeviationThreshold(0.0)
+	assert.Error(t, err)
+	assert.Equal(t, 1.0, calc.standardDeviationThreshold) // Should remain unchanged
+
+	// Test invalid negative threshold
+	err = calc.SetStandardDeviationThreshold(-1.0)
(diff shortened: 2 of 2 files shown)


aaronbrethorst (approved):
Hey Ahmed, nice work converting those panic calls to proper error returns. This is exactly the kind of improvement that makes a Go codebase more production-ready. The change is clean, well-scoped, and the test updates are thorough.

## Verdict: Merge

This PR is ready to merge. All tests pass, lint is clean, and the changes are correct. All callers of the three modified methods have been updated — nothing was missed.

## Suggestions

These are minor observations, not blockers:

1. **Missing test for `SetShapeCache` after-init error path**: There's a test for `SetContextCache` returning an error after init (`TestSetContextCache_ReturnsErrorAfterInit`), but no analogous test for `SetShapeCache`. A quick test that sets `initialized.Store(true)` and asserts `SetShapeCache` returns the expected error would round out the coverage. Similarly, `SetStandardDeviationThreshold`'s after-init guard is tested only indirectly. Not a blocker — the code paths are straightforward and the pattern is well-covered by the `SetContextCache` test.

2. **`TestSetShapeCacheThreadSafety` comment is slightly stale** (`shape_cache_test.go:231`): The comment says "This test documents that SetShapeCache is NOT thread-safe" — but `SetShapeCache` actually *is* thread-safe (it acquires `cacheMutex`). Pre-existing, not your responsibility.

3. **Threshold validation error messages not asserted** (`advanced_direction_calculator_test.go:210-217`): The zero/negative threshold tests assert `assert.Error(t, err)` but don't check the error message. The `SetContextCache` after-init test does assert the message. Minor inconsistency — fine either way.

## Strengths

- Clean, idiomatic Go error handling replacing panics
- All production callers properly check and propagate errors with `fmt.Errorf` wrapping
- Good addition of input validation for the threshold (> 0 check) — prevents silent misconfiguration where a zero threshold would effectively disable variance filtering
- The concurrent access test using an error channel instead of `recover()` is much cleaner and won't accidentally swallow unrelated panics
- Error propagation chain is complete: `Set*Cache` -> `InitializeGlobalCache` -> `BuildApplication` -> application fails to start
- PR description is well-written with clear rationale

PR #549, opened 2026-03-02
feat(gtfsdb): implement database foundation for frequencies.txt (Phase 1)

3rabiii:
### Overview
This PR implements **Phase 1 (Database Foundation)** of the GTFS `frequencies.txt` integration plan. It establishes the database schema, models, and data ingestion pipeline required to support frequency-based trips, without modifying any public API handlers yet.

### Changes Included
- **Schema & Models:** Added `frequencies` table to `schema.sql` and regenerated `sqlc` models.
- **Database Queries:** Added queries for `GetFrequenciesForTrip`, `GetFrequenciesForTrips` (batch), `CreateFrequency`, and `ClearFrequencies` in `query.sql`.
- **Ingestion Logic:** Implemented `bulkInsertFrequencies` and integrated frequency parsing from `go-gtfs` into `processAndStoreGTFSDataWithSource()`.
- **Data Management:** Updated `clearAllGTFSData()` to clear the frequencies table and added frequency counts to `staticDataCounts()`.

### Testing
Added robust database-level testing to ensure ingestion reliability:
- `bulk_insert_frequency_test.go`: Verifies correct storage in SQLite.
- `synthetic_gtfs_frequency_test.go`: Tests ingestion using synthetic GTFS data.
- `conditional_import_test.go`: Ensures the system does not crash and degrades gracefully when a GTFS feed lacks a `frequencies.txt` file.

*(Note: Phase 2 and Phase 3 covering API handler integration and schedule endpoints will follow in subsequent PRs after merging this.)*

image

@aaronbrethorst
closes : #548

PR #582, opened 2026-03-04
feat: GTFS frequencies Phase 2 - Core Models & Business Logic

### LINKED ISSUE #581: Feature: GTFS frequencies support - Phase 2 (Models & Business Logic)
Following the successful merge of Phase 1 #549 , this issue tracks the implementation of **Phase 2** for GTFS `frequencies.txt` support.
this phase focuses strictly on the data structures, core business logic, and performance infrastructure. The actual population of the frequency data into the various API response handlers (e.g., `schedule_for_stop`, `trips_for_location`) is intentionally left for **Phase 3**.

### Scope of Work for Phase 2:

* **Model Enrichment:** Add `Frequency` model and update `ScheduleFrequencies` and `Schedule.Frequency` to use concrete typed structs instead of `[]interface{}` and `int64`.
* **Core Business Logic:** Implement `frequency_helper.go` with functions like `GetActiveHeadway` and `ExpandFrequencyTrips` (handling `exact_times=1` expansions in memory, mimicking OBA Java's `FrequencyBlockTripIndex` behavior).
* **Performance Optimization:** Implement an O(1) in-memory cache (`frequencyTripIDs` map) inside `gtfs_manager.go` during static data load. This completely prevents N+1 database query issues in handlers.
* **Timezone Accuracy:** Fix nanosecond-to-millisecond time conversions to properly calculate the start of the day based on the agency's specific local timezone rather than UTC truncation.
* **JSON Serialization Parity:** Ensure strict backward compatibility with the OBA API (e.g., empty frequency arrays serialize to `[]`, missing single frequencies serialize to `null`, and internal logic fields are omitted).
* **Testing:** Add comprehensive unit tests and concurrency tests (`TestConcurrentFrequencyAccess`) to ensure the new frequency cache is completely thread-safe during hot-swaps of static data.

3rabiii:
Resolves #581 

Following the merged Phase 1 #549 , this PR implements **Phase 2** of the GTFS frequencies support. this phase focuses strictly on the data structures, core business logic, and performance infrastructure. 

**Note :** The actual population of the frequency data into the various API response handlers (e.g., `schedule_for_stop`, `trips_for_location`) is intentionally left for **Phase 3** (the immediate next PR). The handlers updated in this PR have been safely modified to use the new types while passing `nil`/empty slices.

### Key Changes & Optimizations:
* **Model Enrichment:** Added `Frequency` model and updated `ScheduleFrequencies` and `Schedule.Frequency` to use concrete types instead of `[]interface{}` and `int64`.
* **O(1) In-Memory Cache (Performance):** Implemented an in-memory `frequencyTripIDs` map inside `gtfs_manager.go` during static data load. This ensures `IsFrequencyBasedTrip` is an O(1) lookup, completely preventing N+1 database query issues in handlers.
* **Timezone Accuracy:** Fixed the nanosecond-to-millisecond time conversion in `NewFrequencyFromDB` to properly calculate the start of the day based on the agency's specific local timezone rather than UTC truncation.
* **JSON Serialization Parity:** Ensured strict backward compatibility with the OBA API (e.g., empty frequency arrays serialize to `[]`, missing single frequencies serialize to `null`, and internal logic fields like `exactTimes` are properly omitted via `json:"-"`).
* **Java Parity Logic:** Implemented `ExpandFrequencyTrips` to handle `exact_times=1` (schedule-based) expansions in memory, mimicking OBA Java's `FrequencyBlockTripIndex` behavior.
* **Thread Safety:** Added concurrent access testing (`TestConcurrentFrequencyAccess`) to ensure the new frequency cache is completely thread-safe during hot-swaps of static data.

image

@aaronbrethorst 
closes: #581

### DIFF
--- a/internal/gtfs/concurrency_test.go
+++ b/internal/gtfs/concurrency_test.go
@@ -292,3 +292,67 @@ func (manager *Manager) safeGetAgencies() []gtfs.Agency {
 	}
 	return manager.gtfsData.Agencies
 }
+
+func TestConcurrentFrequencyAccess(t *testing.T) {
+	// Test that our new O(1) in-memory frequency map is thread-safe
+	// when being read by API handlers while the static data is reloading
+
+	manager := &Manager{
+		frequencyTripIDs: make(map[string]struct{}),
+	}
+	manager.frequencyTripIDs["freq_trip_1"] = struct{}{}
+
+	var wg sync.WaitGroup
+	done := make(chan struct{})
+
+	// Start Readers (simulating API handlers calling IsFrequencyBasedTrip)
+	for i := 0; i < 20; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for {
+				select {
+				case <-done:
+					return
+				default:
+					// Read from the cache. This uses manager.staticMutex.RLock() internally
+					_ = manager.IsFrequencyBasedTrip("freq_trip_1")
+					_ = manager.IsFrequencyBasedTrip("normal_trip")
+					time.Sleep(time.Microsecond)
+				}
+			}
+		}()
+	}
+
+	// Start Writer (simulating static data reload updating the cache)
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		for i := 0; i < 10; i++ {
+			select {
+			case <-done:
+				return
+			default:
+				// Simulate the hot-swap of the frequency cache
+				newFreqMap := make(map[string]struct{})
+				newFreqMap["freq_trip_2"] = struct{}{}
+				newFreqMap["freq_trip_3"] = struct{}{}
+
+				manager.staticMutex.Lock()
+				manager.frequencyTripIDs = newFreqMap
+				manager.staticMutex.Unlock()
+
+				time.Sleep(5 * time.Millisecond)
+			}
+		}
+	}()
+
+	// Let the race happen for a short time
+	time.Sleep(50 * time.Millisecond)
+	close(done)
+	wg.Wait()
+
+	// Verify the cache was actually updated by the writer
+	isFreq := manager.IsFrequencyBasedTrip("freq_trip_2")
+	assert.True(t, isFreq, "Expected freq_trip_2 to be in the frequency cache after writer finishes")
+}
--- a/internal/gtfs/frequency_helper.go
+++ b/internal/gtfs/frequency_helper.go
@@ -1,0 +1,102 @@
+package gtfs
+
+import (
+	"context"
+	"time"
+
+	"maglev.onebusaway.org/gtfsdb"
+)
+
+// IsFrequencyBasedTrip returns true if the given trip has any frequency entries in the database.
+func (manager *Manager) IsFrequencyBasedTrip(tripID string) bool {
+	manager.staticMutex.RLock()
+	defer manager.staticMutex.RUnlock()
+
+	_, isFreq := manager.frequencyTripIDs[tripID]
+	return isFreq
+}
+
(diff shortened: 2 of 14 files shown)


3rabiii:
@aaronbrethorst 
Ready for review!

aaronbrethorst (changes requested):
Adel, impressive follow-up — you've addressed every item from round 1 cleanly. The cache rebuild is in both `ForceUpdate()` and `setStaticGTFS()`, the slice pre-allocation is correct, the `*Frequency` pointer approach for null serialization is exactly right, and the new tests (post-midnight, concurrent access, field preservation) are well-crafted. One item to fix before merge.

## Important Issues

### 1. SQLite WAL/SHM files committed to the repository

**Files**: `testdata/raba-test.db-shm` (32KB), `testdata/raba-test.db-wal` (4MB)

These are ephemeral SQLite WAL-mode runtime files — they're generated when `raba-test.db` is opened and contain transient journal state. They should not be in version control: they bloat the repo, can cause stale journal issues for other developers, and may behave differently across platforms.

**Fix**:
```bash
git rm --cached testdata/raba-test.db-shm testdata/raba-test.db-wal
```

Note: PR #580 already added `*.db-wal` and `*.db-shm` to `.gitignore`, so after removing these tracked files they won't come back.

## Suggestions

### 2. Frequency cache silently emptied on DB query failure during hot-swap

In `ForceUpdate()` (`static.go:321-329`), if `GetFrequencyTripIDs` fails, the error is logged but `manager.frequencyTripIDs` is set to the new empty map. This means a transient DB error during a GTFS reload silently makes all frequency-based trips appear as normal trips. Consider preserving the previous cache on failure:

```go
if ids, freqErr := client.Queries.GetFrequencyTripIDs(ctx); freqErr == nil {
    freqTripIDs := make(map[string]struct{})
    for _, id := range ids {
        freqTripIDs[id] = struct{}{}
    }
    manager.frequencyTripIDs = freqTripIDs
} else {
    logging.LogError(logger, "failed to reload frequency trip IDs during hot-swap; retaining previous cache", freqErr)
}
```

The same pattern applies in `InitGTFSManager` (`gtfs_manager.go:213-222`) and `setStaticGTFS` (`static.go:382-391`), though for the initial load there's no previous cache to preserve.

### 3. `setStaticGTFS` frequency rebuild is dead code on initial startup

`InitGTFSManager` calls `setStaticGTFS(staticData)` at line 196 *before* `manager.GtfsDB = gtfsDB` at line 197. The nil guard at `static.go:373` (`if manager.GtfsDB != nil`) skips the entire frequency block on the first call. The cache is correctly populated later at lines 213-222, so there's no bug — but the dead code path is misleading. A brief comment in `setStaticGTFS` noting "GtfsDB may be nil during initial construction; frequency cache is populated by InitGTFSManager directly" would prevent future confusion.

## Strengths

- Every round 1 item addressed thoroughly and correctly
- The `frequencyTripIDs` cache is properly protected by `staticMutex` in all read/write paths
- `ExpandFrequencyTrips` is clean and well-tested — the base offset shift logic correctly handles templates whose first stop doesn't start at the window start
- `NewFrequencyFromDB` correctly uses the agency's local timezone for start-of-day calculation
- Test coverage is excellent: boundary conditions, over-midnight GTFS times, field preservation, JSON serialization parity, concurrent access
- The `*Frequency` pointer approach for null serialization is the right design — matches OBA production API behavior

## Verdict

**Request changes** — remove the two SQLite WAL/SHM files from tracking. The suggestions above are non-blocking improvements. Once the binary files are removed, this is ready to merge.

Issue #585, opened 2026-03-04
Bug: Trip directionId serialized as number instead of string


3rabiii:
### Description
The Go `maglev` server serializes the `Trip.DirectionID` field as a JSON number (`int64`), which breaks strict parity with the legacy Java server.

### Context & Impact
In the legacy Java implementation (`TripV2Bean.java`), `directionId` is strongly typed as a `String` (e.g., `"0"` or `"1"`). 
Mobile clients (Android/iOS) using strict JSON parsers (like Codable or Retrofit/Moshi) expect this field to be a string. Receiving an integer causes type mismatch exceptions, leading to immediate app crashes on the client side.

### Expected Behavior (Java Parity)
```json
{
  "directionId": "0"
}
```
### Actual Behavior (Maglev/Go)
```
{
  "directionId": 0
}
```
### Proposed Solution
* Change DirectionID type from int64 to string in models.Trip and models.NewTripReference.
* Update all API handlers (e.g., arrivals_and_departure_for_stop, trips_for_route_handler, block_handler) to convert the database int64 value to a string using strconv.FormatInt(..., 10).
* Update all associated unit tests to assert against string values instead of floats/ints.

PR #586, opened 2026-03-04
fix(api): change trip directionId to string for java parity

3rabiii:
### Description
This PR fixes a critical type mismatch bug where `Trip.DirectionID` was being serialized as an `int64` instead of a `string`. This breaks strict parity with the legacy OBA Java server (`TripV2Bean`) and causes immediate crashes on mobile clients (Android/iOS) that use strict JSON parsing.

### What Changed
- **Models:** Changed `DirectionID` from `int64` to `string` in `models.Trip` and `models.NewTripReference()`.
- **Handlers:** Updated all relevant API handlers (e.g., `trips_for_route`, `arrivals_and_departure`, `block_handler`, etc.) to convert the database `int64` integer to a string using `strconv.FormatInt(trip.DirectionID.Int64, 10)`.
- **Tests:** Updated all corresponding unit tests to assert against string values (e.g., `"1"`) instead of numeric values (e.g., `float64(1)`).

### Testing
- All unit tests have been updated and are passing successfully (`make test`).
- The endpoint now correctly produces the standard `"directionId": "1"` format, exactly matching the Java server.
image

@aaronbrethorst 
fixes: #585

### DIFF
--- a/internal/models/trip.go
+++ b/internal/models/trip.go
@@ -2,7 +2,7 @@ package models
 
 type Trip struct {
 	BlockID        string `json:"blockId"`
-	DirectionID    int64  `json:"directionId"`
+	DirectionID    string `json:"directionId"`
 	ID             string `json:"id"`
 	RouteID        string `json:"routeId"`
 	ServiceID      string `json:"serviceId"`
@@ -24,7 +24,7 @@ func NewTripResponse(trip *Trip, timeZone string, peakOffPeak int) *TripResponse
 	}
 }
 
-func NewTripReference(id, routeID, serviceID, headSign, shortName string, directionID int64, blockID, shapeID string) *Trip {
+func NewTripReference(id, routeID, serviceID, headSign, shortName string, directionID string, blockID, shapeID string) *Trip {
 	return &Trip{
 		BlockID:        blockID,
 		DirectionID:    directionID,
--- a/internal/models/trip_details_test.go
+++ b/internal/models/trip_details_test.go
@@ -13,7 +13,7 @@ func TestNewTripDetails(t *testing.T) {
 		RouteID:        "route_456",
 		ServiceID:      "service_789",
 		TripHeadsign:   "Downtown Terminal",
-		DirectionID:    1,
+		DirectionID:    "1",
 		BlockID:        "block_1",
 		ShapeID:        "shape_1",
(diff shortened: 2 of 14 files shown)


aaronbrethorst (approved):
Hey Adel, great catch on the type mismatch — this is exactly the kind of subtle bug that can cause real headaches for mobile clients with strict JSON parsing. The fix is thorough and well-researched.

## Critical Issues

None found.

## Important Issues

None found.

## Suggestions

### 1. Minor style inconsistency in test

**File:** `internal/restapi/trip_handler_test.go:46`

The test uses `fmt.Sprintf("%d", trips[0].DirectionId)` while all handler code uses `strconv.FormatInt(..., 10)`. Functionally identical, but using `strconv.FormatInt` would be more consistent with the pattern established throughout the PR. Truly a nitpick — not blocking.

### 2. Empty-string test case for `NewTripReferenceWithEmptyValues`

**File:** `internal/models/trip_test.go:36`

The test passes `"0"` as the directionID for the "empty values" case. Since `DirectionID` is now a string, the natural "empty" value would be `""` rather than `"0"`. Consider whether this test should assert against `""` to test what happens when no direction is known. Not blocking since the handler code always produces a numeric string via `strconv.FormatInt`.

## Verification

- Confirmed the [OpenAPI spec](https://github.com/OneBusAway/sdk-config/blob/main/openapi.yml) defines `directionId` as `type: string`
- Exhaustive search of all `DirectionID` references across the codebase — every conversion site is covered
- `stops_for_route_handler.go:249` has a local `DirectionID int64` in a grouping struct — this is correct as-is (used for map keys, never assigned to `models.Trip`)
- `trips_for_route_handler.go:464` copies `models.Trip.DirectionID` (string-to-string) — correct
- `reference_utils.go:113` uses `entity.DirectionID.String()` on a GTFS-RT alert entity — unrelated
- NULL `direction_id` in DB produces `"0"` (same as pre-existing behavior) — acceptable
- All tests pass, linter clean

## Strengths

- Thorough — every handler that constructs a `models.Trip` is updated consistently
- Tests updated to match the new string type
- Clear PR description with screenshot evidence of the fix
- Correctly identified the root cause and linked to the issue

**Verdict: Merge.** This is a correct, well-scoped bug fix with full coverage across the codebase. Ship it!

PR #588, opened 2026-03-04
fix(gtfs,restapi): reject invalid agency timezones and prevent trips-…

### LINKED ISSUE #587: trips-for-location can panic when GTFS agency timezone is invalid (Time.In(nil)), causing recovery-middleware 500
## Description

While testing `/api/where/trips-for-location.json`, we discovered a potential panic caused by invalid GTFS agency timezone values.

In `parseAndValidateRequest` (`trips_for_location_handler.go`), the code previously did:

- `time.LoadLocation(currentAgency.Timezone)` with ignored error  
- then `api.Clock.Now().In(currentLocation)`

If `currentAgency.Timezone` is invalid, `LoadLocation` fails and `currentLocation` becomes `nil`.  
Calling `Time.In(nil)` panics. Recovery middleware catches it and returns a generic **500** (`version:1`), but this is still a runtime panic in normal request flow.

---

## Why this matters

- Runtime panic in request path (even if recovered) is fragile behavior.
- One bad agency timezone in GTFS data can break endpoint requests.
- The API returns generic internal errors instead of deterministic handling.

---

## Steps to reproduce

Build GTFS zip with invalid timezone in `agency.txt`:

```bash
rm -rf /tmp/badgtfs && mkdir -p /tmp/badgtfs
unzip -q testdata/raba.zip -d /tmp/badgtfs
sed -i 's#America/Los_Angeles#Invalid/Zone#g' /tmp/badgtfs/agency.txt
(cd /tmp/badgtfs && zip -qr /tmp/raba-bad-timezone.zip .)
```
## Configure GTFS feed

Point config to the modified GTFS zip:

```json
"gtfs-static-feed": {
  "url": "/tmp/raba-bad-timezone.zip",
  "enable-gtfs-tidy": false
}
```
## Run the server and reproduce

Start the server and call the endpoint:

```bash
curl "http://localhost:4000/api/where/trips-for-location.json?key=test&lat=40.5865&lon=-122.3917&latSpan=0.1&lonSpan=0.1"
```
## Observed

- **Response:** `500 internal server error`
- **Logs:** panic recovered (`handler panic recovered` in recovery middleware)

## Expected

- No panic in request handling.
- Invalid timezone should be handled deterministically:
  - ideally rejected during **GTFS load/startup** (fail fast / unhealthy state),
  - and request path should return a **controlled error** if invalid data somehow slips through.

## Suggested Direction

- Validate all `agency.timezone` values during **static GTFS load and hot-swap**.
- In request handlers, avoid panic-prone usage by checking `LoadLocation` errors and returning controlled errors  
  (**no fallback to UTC**).

Adityatorgal17:
Fixes #587

## Summary

Fix panic path in `trips-for-location` caused by invalid agency timezone and enforce timezone validity at GTFS load time.

---

### 1. Added strict GTFS agency timezone validation during static data load

**File:** `internal/gtfs/static.go`

- Added helper: `validateStaticAgencyTimezones(...)`
- `loadGTFSData(...)` now fails if any agency has an invalid timezone.

---

### 2. Added defensive request-time timezone resolution in trips-for-location

**File:** `internal/restapi/trips_for_location_handler.go`

- Replaced ignored `time.LoadLocation` error with `resolveAgencyLocation(...)`
- Explicit error propagation introduced (no UTC fallback).

---

### 3. Added tests

- `internal/gtfs/static_timezone_validation_test.go`
- `internal/restapi/trips_for_location_handler_test.go` (`TestResolveAgencyLocation`)

---

## Behavior Before

- Invalid agency timezone could lead to `Time.In(nil)` panic in `trips-for-location`.
- Panic was recovered by middleware and surfaced as generic **HTTP 500**.

---

## Behavior After

- Invalid timezone is rejected during **GTFS load/startup/hot-swap** (fail fast).
- `trips-for-location` no longer has a panic path for timezone resolution.
- Handler returns controlled error flow if invalid timezone somehow appears.
- **No UTC fallback introduced.**

---

## Validation

- `make test` passed.
- Targeted new tests passed.

### DIFF
--- a/internal/gtfs/static.go
+++ b/internal/gtfs/static.go
@@ -140,7 +140,7 @@ func loadGTFSData(source string, isLocalFile bool, config Config) (*gtfs.Static,
 }
 
 func validateStaticAgencyTimezones(staticData *gtfs.Static) error {
-	for _, agency := range staticData.Agencies {
+	for i, agency := range staticData.Agencies {
 		tz := strings.TrimSpace(agency.Timezone)
 		// Go treats LoadLocation("") as UTC, so we consider this an error for GTFS validation purposes
 		if tz == "" {
@@ -149,6 +149,8 @@ func validateStaticAgencyTimezones(staticData *gtfs.Static) error {
 		if _, err := time.LoadLocation(tz); err != nil {
 			return fmt.Errorf("agency %q has invalid timezone %q: %w", agency.Id, tz, err)
 		}
+		// Write the trimmed value back so downstream LoadLocation calls use the clean string
+		staticData.Agencies[i].Timezone = tz
 	}
 	return nil
 }
--- a/internal/gtfs/static_timezone_validation_test.go
+++ b/internal/gtfs/static_timezone_validation_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	"github.com/OneBusAway/go-gtfs"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
@@ -24,7 +25,8 @@ func TestValidateStaticAgencyTimezones(t *testing.T) {
 				{Id: "a1", Timezone: ""},
 			},
 		}
-		require.Error(t, validateStaticAgencyTimezones(staticData))
+		err := validateStaticAgencyTimezones(staticData)
+		require.Contains(t, err.Error(), "empty timezone")
 	})
 
 	t.Run("invalid timezone", func(t *testing.T) {
@@ -34,5 +36,43 @@ func TestValidateStaticAgencyTimezones(t *testing.T) {
 			},
 		}
 		require.Error(t, validateStaticAgencyTimezones(staticData))
+	})
+
+	t.Run("whitespace-padded timezone is normalized and valid", func(t *testing.T) {
+		staticData := &gtfs.Static{
+			Agencies: []gtfs.Agency{
+				{Id: "a1", Timezone: "  America/Los_Angeles  "},
+			},
+		}
+		require.NoError(t, validateStaticAgencyTimezones(staticData))
+		assert.Equal(t, "America/Los_Angeles", staticData.Agencies[0].Timezone)
+	})
+
+	t.Run("whitespace-only timezone is rejected", func(t *testing.T) {
+		staticData := &gtfs.Static{
+			Agencies: []gtfs.Agency{
+				{Id: "a1", Timezone: "   "},
+			},
+		}
+		err := validateStaticAgencyTimezones(staticData)
+		require.Contains(t, err.Error(), "empty timezone")
+	})
+
+	t.Run("multiple agencies second has invalid timezone", func(t *testing.T) {
+		staticData := &gtfs.Static{
+			Agencies: []gtfs.Agency{
+				{Id: "a1", Timezone: "America/Los_Angeles"},
+				{Id: "a2", Timezone: "Invalid/Zone"},
+			},
+		}
+		err := validateStaticAgencyTimezones(staticData)
+		require.Contains(t, err.Error(), "a2")
+	})
+
+	t.Run("empty agency list passes validation", func(t *testing.T) {
+		staticData := &gtfs.Static{
+			Agencies: []gtfs.Agency{},
+		}
+		require.NoError(t, validateStaticAgencyTimezones(staticData))
 	})
 }


aaronbrethorst (changes requested):
Aditya, great job addressing the feedback from the first round — the `resolveAgencyLocation` wrapper is gone and you're now checking the `time.LoadLocation` error inline, exactly matching the pattern in `trips_for_route_handler.go`. The empty timezone validation with the explicit `tz == ""` check is a nice touch too, since `LoadLocation("")` silently returns UTC. Almost there — one thing to clean up.

## Prior Feedback Verification

- [x] `resolveAgencyLocation` wrapper removed — error checked inline at `trips_for_location_handler.go:143-146`
- [x] Empty timezone edge case handled — `validateStaticAgencyTimezones` explicitly rejects `tz == ""` with a clear error message
- [x] Empty timezone test added — `static_timezone_validation_test.go:21-28`

## Important Issues (1 found)

### 1. `TestResolveAgencyLocation` tests the Go standard library, not project code

`trips_for_location_handler_test.go:14-26` — This test was written for the `resolveAgencyLocation` helper that was removed in this revision. What remains is two calls to `time.LoadLocation` with no project code involved:

```go
func TestResolveAgencyLocation(t *testing.T) {
    t.Run("valid timezone", func(t *testing.T) {
        loc, err := time.LoadLocation("America/Los_Angeles") // tests Go stdlib
```

Delete this test entirely. The timezone validation is already well-tested by `TestValidateStaticAgencyTimezones` in `static_timezone_validation_test.go`, and the inline error check in the handler is exercised by the existing integration tests.

## Strengths

- Load-time validation is the right architectural choice — fail at GTFS load rather than at request time
- The `strings.TrimSpace` on the timezone before validation catches whitespace-only values
- Clean error messages that include the agency ID for debugging
- The inline error handling in the handler matches the established pattern in `trips_for_route_handler.go`

## Checklist

- [x] Tests pass
- [x] Lint clean
- [x] Previous feedback addressed (wrapper removed, empty tz handled)
- [ ] Remove orphaned `TestResolveAgencyLocation` test

## Recommended Action

1. Delete `TestResolveAgencyLocation` (lines 14-26 of `trips_for_location_handler_test.go`)

aaronbrethorst (changes requested):
Hey Aditya, nice work getting all the prior feedback addressed — the inline error check is clean and matches the established pattern in `trips_for_route_handler.go`. One more thing to fix before this is ready.

## Verdict: Request Changes

## Prior Feedback Verification

- [x] `resolveAgencyLocation` wrapper removed — error checked inline at `trips_for_location_handler.go:149-152`
- [x] Empty timezone edge case handled — `validateStaticAgencyTimezones` explicitly rejects `tz == ""`
- [x] `TestResolveAgencyLocation` deleted — no orphaned test code remains

## Important Issues (1 found)

### 1. `validateStaticAgencyTimezones` trims whitespace but doesn't store the trimmed value back

`static.go:142-154` — The function calls `strings.TrimSpace(agency.Timezone)` and validates the trimmed result, but the range loop operates on a copy of the agency struct, so the original `agency.Timezone` in the slice is never updated. A timezone like `" America/Los_Angeles "` would pass validation (the trimmed value is valid), but every downstream `time.LoadLocation(" America/Los_Angeles ")` call would fail.

Fix: iterate by index and write the normalized value back:

```go
func validateStaticAgencyTimezones(staticData *gtfs.Static) error {
    for i, agency := range staticData.Agencies {
        tz := strings.TrimSpace(agency.Timezone)
        if tz == "" {
            return fmt.Errorf("agency %q has empty timezone", agency.Id)
        }
        if _, err := time.LoadLocation(tz); err != nil {
            return fmt.Errorf("agency %q has invalid timezone %q: %w", agency.Id, tz, err)
        }
        staticData.Agencies[i].Timezone = tz
    }
    return nil
}
```

Alternatively, if you prefer not to mutate the input, validate the raw `agency.Timezone` directly (without trimming) — whitespace in a GTFS timezone field is always a data error.

## Suggestions

These are not blockers:

1. **Add a test for whitespace-padded timezone** (`static_timezone_validation_test.go`): This would catch the bug above. Also consider adding cases for: whitespace-only strings (`"   "`), multiple agencies where the second has an invalid timezone, and an empty agency list.

2. **Assert error message content in the "empty timezone" test** (`static_timezone_validation_test.go:27`): Since `time.LoadLocation("")` returns `(UTC, nil)` — no error — the explicit `tz == ""` check is the *only* defense against empty timezones silently becoming UTC. Adding `require.Contains(t, err.Error(), "empty timezone")` would ensure the correct branch is exercised.

3. **Inconsistent timezone error handling across handlers** (out of scope): The codebase now has three strategies for `time.LoadLocation` failures — return server error (`trips_for_route_handler.go`, `trips_for_location_handler.go`), fallback to UTC via `LoadLocationWithUTCFallBack` (5 handlers), and inline UTC fallback (`stops_for_route_handler.go`). Now that load-time validation exists, a follow-up PR could standardize all handlers on the error-returning pattern.

## Strengths

- Load-time validation is the right architectural choice — fail fast at GTFS load rather than at request time
- The `strings.TrimSpace` call is a good defensive measure against whitespace-only values
- Error messages include the agency ID and timezone value for debugging
- Clean defense-in-depth: validate at load time *and* handle errors at the handler level
- All previous review feedback fully addressed

## Checklist

- [x] Tests pass
- [x] Previous feedback addressed (wrapper removed, empty tz handled, orphaned test deleted)
- [ ] Fix trim-but-don't-store bug in `validateStaticAgencyTimezones`

## Recommended Action

1. Fix the trim-but-don't-store issue (important — creates a validation/runtime mismatch)
2. Consider the test suggestions

Issue #601, opened 2026-03-05
Configurable log level and output format


aaronbrethorst:
## Summary

Log level is hardcoded to `slog.LevelInfo` and the output format (text vs JSON) cannot be changed without modifying code. Operators need to adjust log verbosity for debugging and switch to JSON output for log aggregation pipelines in production.

## Current Behavior

In `cmd/api/app.go`, the application logger uses `slog.NewTextHandler` with a hardcoded level:

- The bootstrap logger in `main.go` uses `slog.NewTextHandler` with no level configuration.
- The request logger (line ~153 in `app.go`) is hardcoded to `slog.LevelInfo` with JSON output.
- The `internal/logging/` package has `NewStructuredLogger()` which creates JSON loggers, but this isn't used as the main application logger.

There is no config option, CLI flag, or environment variable to control log level or format.

## Expected Behavior

Add two new configuration options:

- **`log-level`** — one of `debug`, `info`, `warn`, `error` (default: `info`)
- **`log-format`** — one of `text`, `json` (default: `text`)

These should be settable via the JSON config file and/or environment variables, consistent with how other config options work.

## Context for Contributors

- Configuration is managed in `internal/appconf/json_config.go` — add the new fields to `Config` struct and validation in `validateConfig()`.
- The config schema is in `config.schema.json` — update it to include the new fields with enum validation.
- Logger creation is in `cmd/api/app.go` — modify to use the config values when creating `slog.Handler`.
- The `internal/logging/structured_logging.go` package already has JSON handler support via `NewStructuredLogger()`.
- `slog.LevelVar` can be used for a dynamically-settable log level.
- Follow the existing pattern for config defaults (see `applyDefaults()` in `json_config.go`).

Jason10293:
working on this now!

PR #622, opened 2026-03-06
feat: add configurable logging options for log level and format

Jason10293:
This PR aims to address the enhancement in #601 .

## Key Changes
- Added configurable `log-level` (`debug|info|warn|error`, default `info`) and `log-format` (`text|json`, default `text`) to app config.
- Added defaults and validation for both fields in config loading, plus schema support in `config.schema.json`.
- Updated logger construction to use configured level/format for both app logs and request logs (removed hardcoded behavior).
- Added env var overrides: `MAGLEV_LOG_LEVEL` and `MAGLEV_LOG_FORMAT`.
- Updated `--dump-config`, `config.example.json`, and `README.markdown` to include the new options.
- Updated tests in `internal/appconf/json_config_test.go` for defaults, validation, and env override behavior.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -89,6 +89,8 @@ Example `config.json`:
   "env": "production",
   "api-keys": ["key1", "key2", "key3"],
   "rate-limit": 50,
+  "log-level": "info",
+  "log-format": "json",
   "gtfs-static-feed": {
     "url": "https://example.com/gtfs.zip",
     "auth-header-name": "Authorization",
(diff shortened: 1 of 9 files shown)


aaronbrethorst (approved):
Jason, all four items from my previous review have been addressed — nice turnaround. The duplicate logger is gone, validation tests are in place, `config_full.json` covers the new fields, and the blank line formatting is consistent.

Tests pass, lint is clean, and the code reads well. Ship it.

PR #627, opened 2026-03-06
fix: consolidate TripStatus types (issue #619)

### LINKED ISSUE #619: Consolidate duplicate TripStatus types into single OpenAPI-spec-compliant struct
## Problem

The codebase has two divergent TripStatus types for the same OpenAPI schema:

- `TripStatus` in `internal/models/vehicle.go` — 14 fields (stripped-down)
- `TripStatusForTripDetails` in `internal/models/trip_details.go` — 27 fields (comprehensive)

The OpenAPI spec defines **one** `TripStatus` schema used everywhere. Having two types causes:
- `TripStatus` in vehicle responses is missing 13 fields vs the spec
- Changes to the spec require updating two structs
- Different response shapes for the same conceptual object confuse API consumers
- Field drift over time as one type gets updated but not the other

These should be unified into a single struct matching the full spec schema, with optional fields using pointer types/omitempty for cases where data isn't available.

## Files
- `internal/models/vehicle.go` (`TripStatus`)
- `internal/models/trip_details.go` (`TripStatusForTripDetails`)
- All handlers that construct either type

## Severity
Medium — architectural issue causing API parity drift.

Adityatorgal17:
## Description

The codebase previously contained **two divergent `TripStatus` types** representing the same OpenAPI schema:

- `TripStatus` in `internal/models/vehicle.go` — a stripped-down version with ~14 fields.
- `TripStatusForTripDetails` in `internal/models/trip_details.go` — a comprehensive version with ~27 fields.

However, the OpenAPI specification defines **a single `TripStatus` schema used across all endpoints**. Maintaining two separate structs introduced several problems:

- `TripStatus` used in vehicle responses was **missing multiple fields defined in the spec**.
- API responses for the same conceptual object had **different shapes across endpoints**.
- Future updates to the OpenAPI spec would require **duplicated changes across two structs**.
- Risk of **schema drift** as one struct could evolve independently of the other.

This PR consolidates both types into a **single unified `TripStatus` struct** that matches the full OpenAPI schema.

Optional fields remain safely usable by endpoints that do not populate them through pointer types and/or `omitempty`.

---

## Changes

- Removed the stripped-down `TripStatus` struct from:
  - `internal/models/vehicle.go`
- Renamed `TripStatusForTripDetails` → `TripStatus`.
- Updated all usages across the codebase to reference the unified type.
- Ensured all endpoints now return a consistent `TripStatus` response structure matching the OpenAPI schema.

---

## Affected Areas

The unified `TripStatus` type is now used across multiple endpoints and helpers, including:

- `vehicles-for-agency`
- `trips-for-route`
- `trip-details`
- `trip-for-vehicle`
- `arrivals-and-departures-for-stop`

---

## Files Modified

Key updates include:

- `internal/models/vehicle.go` — removed the smaller `TripStatus` struct
- `internal/models/trip_details.go` — renamed `TripStatusForTripDetails` → `TripStatus`
- `internal/models/trips_for_route.go`
- `internal/models/arrival_and_departure.go`
- `internal/restapi/trips_helper.go`
- `internal/restapi/vehicles_helper.go`
- `internal/restapi/trip_details_handler.go`
- `internal/restapi/trip_for_vehicle_handler.go`
- `internal/restapi/trips_for_route_handler.go`
- `internal/restapi/arrival_and_departure_for_stop_handler.go`
- `internal/restapi/arrivals_and_departure_for_stop.go`

Tests updated accordingly:

- `internal/models/arrival_and_departure_test.go`
- `internal/models/trip_details_test.go`
- `internal/restapi/vehicles_helper_test.go`
- `internal/restapi/trips_helper_test.go`

---

## Impact

- Ensures **API responses match the OpenAPI schema consistently across endpoints**.
- Eliminates duplication and reduces maintenance overhead.
- Prevents **schema drift** between different `TripStatus` representations.
- Improves clarity for API consumers by providing **a single canonical `TripStatus` object**.

---

## Testing

All checks passed successfully:

```bash
make fmt
make test
```

## Fixes #619

### DIFF
--- a/internal/models/arrival_and_departure_test.go
+++ b/internal/models/arrival_and_departure_test.go
@@ -33,10 +33,9 @@ func TestNewArrivalAndDeparture(t *testing.T) {
 	occupancyStatus := "MANY_SEATS_AVAILABLE"
 	predictedOccupancy := "FEW_SEATS_AVAILABLE"
 	historicalOccupancy := "STANDING_ROOM_ONLY"
-	tripStatus := &TripStatus{
-		VehicleID: vehicleID,
-		Status:    "in_progress",
-	}
+	tripStatus := NewTripStatus()
+	tripStatus.VehicleID = vehicleID
+	tripStatus.Status = "in_progress"
 	situationIDs := []string{"situation_1", "situation_2"}
 
 	arrival := NewArrivalAndDeparture(
@@ -88,10 +87,9 @@ func TestNewArrivalAndDeparture(t *testing.T) {
 }
 
 func TestArrivalAndDepartureJSON(t *testing.T) {
-	tripStatus := &TripStatus{
-		VehicleID: "vehicle_789",
-		Status:    "in_progress",
-	}
+	tripStatus := NewTripStatus()
+	tripStatus.VehicleID = "vehicle_789"
+	tripStatus.Status = "in_progress"
 
 	lastUpdateTime := int64(1609462700000)
 	arrival := ArrivalAndDeparture{
--- a/internal/models/trip_details.go
+++ b/internal/models/trip_details.go
@@ -35,7 +35,7 @@ func NewEmptyTripDetails() *TripDetails {
 type TripStatus struct {
 	ActiveTripID               string     `json:"activeTripId"`
 	BlockTripSequence          int        `json:"blockTripSequence"`
-	ClosestStop                string     `json:"closestStop,omitempty"`
+	ClosestStop                string     `json:"closestStop"`
 	ClosestStopTimeOffset      *int       `json:"closestStopTimeOffset,omitempty"`
 	DistanceAlongTrip          *float64   `json:"distanceAlongTrip,omitempty"`
 	Frequency                  *Frequency `json:"frequency,omitempty"`
@@ -48,7 +48,7 @@ type TripStatus struct {
 	NextStopTimeOffset         *int       `json:"nextStopTimeOffset,omitempty"`
 	OccupancyCapacity          *int       `json:"occupancyCapacity,omitempty"`
 	OccupancyCount             *int       `json:"occupancyCount,omitempty"`
-	OccupancyStatus            string     `json:"occupancyStatus,omitempty"`
+	OccupancyStatus            string     `json:"occupancyStatus"`
 	Orientation                *float64   `json:"orientation,omitempty"`
 	Phase                      string     `json:"phase"`
 	Position                   Location   `json:"position"`
@@ -64,9 +64,8 @@ type TripStatus struct {
 	Scheduled                  bool       `json:"scheduled"` // (Scheduled = !Predicted) ,this field is not part of the OpenAPI TripStatus schema but is retained for compatibility with existing API consumers. Tracked as a known spec deviation.
 }
 
-// NewTripStatus returns a TripStatus with safe zero-value defaults.
-// Always use this instead of &TripStatus{} to ensure SituationIDs is
-// initialized to []string{} (never nil), avoiding null vs [] inconsistency in JSON.
+// SituationIDs initialized to []string{} for Go-side convenience.
+// JSON output is unaffected because omitempty omits both nil and empty slices.
 func NewTripStatus() *TripStatus {
 	return &TripStatus{
 		SituationIDs: []string{},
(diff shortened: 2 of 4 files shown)


aaronbrethorst (changes requested):
Aditya, nice work addressing the feedback from the first round — the `omitempty` tags, `*Location` pointer migration, and `NewTripStatus()` constructor are all handled correctly. The core consolidation is solid and the nil guards around `LastKnownLocation` are careful.

A few things to tighten up before merging:

## Critical Issues (1 found)

### 1. `NewTripStatus()` comment is misleading — `omitempty` on slices omits both `nil` and `[]string{}`

`internal/models/trip_details.go:67-74`

The constructor comment says:

```go
// Always use this instead of &TripStatus{} to ensure SituationIDs is
// initialized to []string{} (never nil), avoiding null vs [] inconsistency in JSON.
```

This is incorrect. The `SituationIDs` field now has `json:"situationIds,omitempty"` (line 59). Go's `encoding/json` omits slices tagged `omitempty` for **both** `nil` and `[]string{}` — so the initialization has zero effect on JSON output. The field is omitted either way.

**Fix**: Either (a) remove `omitempty` from `SituationIDs` if you actually need `[]` vs `null` control in the JSON, or (b) update the comment to accurately reflect that the initialization is just a Go-side convenience (preventing nil-related surprises in Go code, though `range`/`append` already handle nil slices fine). Given that `situationIds` isn't in the OpenAPI spec's `required` list for TripStatus, omitting it when empty is fine — but the comment must not claim something that isn't true.

## Important Issues (2 found)

### 1. `closestStop` has `omitempty` but is required in the OpenAPI spec

`internal/models/trip_details.go:38`

```go
ClosestStop string `json:"closestStop,omitempty"`
```

The OpenAPI spec lists `closestStop` in the `required` array for TripStatus. With `omitempty`, the field is omitted when the closest stop hasn't been determined (empty string) rather than appearing as `""`. Per CLAUDE.md: "All API endpoints MUST behave identically to what is defined in this OpenAPI spec."

I understand this matches the old vehicle.go behavior and was specifically requested in the first review round. If this is an intentional deviation, add a brief comment (like the `Scheduled` field has) so future readers know it's deliberate, not accidental.

### 2. `occupancyStatus` has `omitempty` but is required in the OpenAPI spec

`internal/models/trip_details.go:51`

Same situation — `occupancyStatus` is in the spec's `required` list. Same fix: document the deviation or remove `omitempty`.

## Fit and Finish (2 found)

### 1. The `Scheduled`/`Predicted` redundancy is a future defect magnet

`internal/models/trip_details.go:64`

These two fields are always logically inverse (`Scheduled = !Predicted`). Having both as independent mutable fields means any future mutation site that changes one without updating the other introduces an inconsistency. The `vehiclesForAgencyHandler` (line 95) sets `tripStatus.Scheduled = true` without setting `Predicted = false`, relying on the zero value of `bool` — works today, fragile tomorrow.

Not a blocker for this PR, but worth tracking as a follow-up: consider computing `Scheduled` during JSON marshaling via a custom marshaler, or adding a `SetPredicted(bool)` method that sets both.

### 2. Test code bypasses the constructor

Six test sites in `trip_details_test.go` and `arrival_and_departure_test.go` construct `TripStatus` with struct literals (`&TripStatus{...}` or `TripStatus{...}`) rather than `NewTripStatus()`. While tests may not care about serialization defaults, this normalizes the anti-pattern. Consider updating them to use the constructor for consistency.

## Strengths

- Correct structural decision: one canonical type eliminates divergence risk
- All handler references updated consistently — zero remaining `TripStatusForTripDetails` references
- `LastKnownLocation` pointer migration handled safely with proper nil guards in `trips_helper.go` (lines 139, 150)
- Tests updated with `require.NotNil` assertions before pointer dereference
- `NewTripStatus()` used consistently in all production code paths
- The `Position` (value type) vs `LastKnownLocation` (pointer type) asymmetry correctly models required vs optional semantics from the OpenAPI spec
- Net code reduction with improved consistency

## Checklist

- [x] Tests pass
- [x] All references updated (zero `TripStatusForTripDetails` remaining)
- [x] `LastKnownLocation` nil guards correct
- [x] Constructor used in production code
- [ ] Fix misleading `NewTripStatus()` comment re: `omitempty` behavior (Critical #1)
- [ ] Document `closestStop`/`occupancyStatus` `omitempty` as intentional spec deviations (Important #1, #2)

## Verdict: Request Changes

The consolidation is the right move and the implementation is clean. Fix the misleading constructor comment and document the `omitempty` spec deviations, then this is good to merge.

PR #632, opened 2026-03-06
Migrate ReferencesModel fields to strict types (#616)

### LINKED ISSUE #616: ReferencesModel uses `[]interface{}` instead of typed slices for Routes, Situations, StopTimes, Trips
## Problem

At `internal/models/references.go:6-11`, four fields use `[]interface{}` instead of typed slices:

```go
type ReferencesModel struct {
    Agencies   []AgencyReference `json:"agencies"`
    Routes     []interface{}     `json:"routes"`      // should be []Route
    Situations []interface{}     `json:"situations"`   // should be []Situation
    StopTimes  []interface{}     `json:"stopTimes"`    // should be []StopTime
    Stops      []Stop            `json:"stops"`
    Trips      []interface{}     `json:"trips"`        // should be []Trip
}
```

This causes several problems:
1. **No compile-time guarantee** that referenced objects have the correct structure
2. **Situations can't properly serialize** consequence objects with their required `condition` and `conditionDetails` sub-fields
3. **Any handler that appends the wrong type** will produce silently malformed JSON
4. **No IDE autocompletion** or static analysis on these fields

Should be changed to `[]Route`, `[]Situation`, `[]StopTime`, `[]Trip`.

## Files
- `internal/models/references.go`
- All handlers that populate references will need updating

## Severity
High — weakens type safety, risks malformed JSON responses.

Vedd-Patel:
## Description

Fixes #616 

This PR resolves #616 by refactoring the `ReferencesModel` to use strictly typed slices instead of generic `[]interface{}` arrays. This improves type safety, makes the API response structures predictable, and prevents runtime type assertion errors.

### Changes Made:
* **Models:** Updated `ReferencesModel` in `internal/models` to use strict types (e.g., `[]models.Route`, `[]models.Trip`, `[]models.StopTime`, `[]models.Situation`).
* **Handlers:** Updated all affected REST API handlers (including `trips_for_location_handler.go`, `schedule_for_route_handler.go`, `vehicles_for_agency_handler.go`, and `stops_for_route_handler.go`) to build and return strictly typed slices.
* **Helpers:** Updated mapping logic and ID formatting (`utils.FormCombinedID`) to ensure data perfectly matches the new struct definitions.
* **Tests:** Refactored the test suite (`schedule_for_route_handler_test.go`, `filters_test.go`, etc.) to use direct struct field access instead of legacy `map[string]interface{}` assertions.

## Testing
* [x] `make test` runs successfully across all packages, specifically passing all `internal/restapi` and `internal/utils` tests.
* [x] Verified that Combined IDs (e.g., `AgencyID_StopID`) are correctly preserved in the new struct outputs.

### DIFF
--- a/internal/restapi/route_search_handler.go
+++ b/internal/restapi/route_search_handler.go
@@ -126,7 +126,7 @@ func (api *RestAPI) routeSearchHandler(w http.ResponseWriter, r *http.Request) {
 		Agencies:   agencies,
 		Routes:     []models.Route{},
 		Situations: situations,
-		StopTimes:  []models.StopTime{},
+		StopTimes:  []models.RouteStopTime{},
 		Stops:      []models.Stop{},
 		Trips:      []models.Trip{},
 	}
--- a/internal/restapi/routes_for_location_handler.go
+++ b/internal/restapi/routes_for_location_handler.go
@@ -147,7 +147,7 @@ func (api *RestAPI) routesForLocationHandler(w http.ResponseWriter, r *http.Requ
 		Agencies:   agencies,
 		Routes:     []models.Route{},
 		Situations: situations,
-		StopTimes:  []models.StopTime{},
+		StopTimes:  []models.RouteStopTime{},
 		Stops:      []models.Stop{},
 		Trips:      []models.Trip{},
 	}
--- a/internal/restapi/stops_for_location_handler.go
+++ b/internal/restapi/stops_for_location_handler.go
@@ -254,7 +254,7 @@ func (api *RestAPI) stopsForLocationHandler(w http.ResponseWriter, r *http.Reque
 		Agencies:   agencies,
 		Routes:     routes,
 		Situations: situations,
-		StopTimes:  []models.StopTime{},
+		StopTimes:  []models.RouteStopTime{},
 		Stops:      []models.Stop{},
 		Trips:      []models.Trip{},
 	}


aaronbrethorst (approved):
Hey Ved, great work addressing all the feedback from round 1 -- the type migration is thorough and well-executed across all 27 files. One small thing to clean up after merging:

## Status: Approve (with nit)

Tests pass, `go fmt` is clean, and the only lint error is the pre-existing `errcheck` in `gtfs_manager_test.go` (unrelated to this PR).

## Round 2 Follow-up

### Fit and Finish

1. **Dead code: `BuildRouteReferencesAsInterface` has no remaining callers.**

   **`internal/restapi/reference_utils.go:68-80`**

   Both callers (`stops_for_agency_handler.go` and `stops_for_route_handler.go`) were updated to call `BuildRouteReferences` directly, but the wrapper function that converts `[]models.Route` back to `[]interface{}` was left behind. Delete it -- it no longer serves a purpose.

## Round 1 Item Status

| # | Item | Status |
|---|------|--------|
| 1 | Lint errors (S1011 loop -> append) | Fixed |
| 2 | StopTime references lose fields | Fixed -- `StopTimes` is now `[]RouteStopTime` |
| 3 | Incomplete vehicles_for_agency migration | Fixed -- `tripRefs` is `map[string]models.Trip` with `FormCombinedID` |
| 4 | Mixed indentation (spaces vs tabs) | Fixed -- `go fmt` passes clean |
| 5 | `stopTimesRefs` still `[]interface{}` | Fixed -- now `[][]models.RouteStopTime` |
| 6 | Unused `uniqueStopIDs` | N/A -- was actually used |
| 7 | Redundant `route := r` in filters_test | Fixed |
| 8 | Test no longer validates `tripId` | Fixed -- test now asserts `tripId` is present |

## Suggestions (carried forward from round 2)

### 1. `NewTripReference()` returns `*Trip` -- inconsistent with other constructors

**`internal/models/trip.go:27`**

Every other model constructor returns by value. `NewTripReference()` returns `*Trip`, forcing every handler to dereference with `*tripRef`. Changing it to return `Trip` would eliminate the dereference pattern across 6 files. Not blocking -- could be a quick follow-up.

### 2. Consider using `NewEmptyReferences()` to reduce boilerplate

Several handlers construct `ReferencesModel` with explicit empty slices for every field. `NewEmptyReferences()` exists but isn't widely used -- adopting it would reduce the repetitive `[]models.Route{}`, `[]models.Situation{}`, etc.

PR #636, opened 2026-03-08
chore: address PR #607 review feedback

AhmedAlian7:
## Summary
Follow-up for #607 PR review to address non-blocking suggestions.

## Changes
- **Explicit MaxHeaderBytes default:** Added a comment in `app.go` to clarify that `1 << 20` explicitly matches Go's default `http.DefaultMaxHeaderBytes`.
- **Improved Size Limit Tests:** Updated `TestSizeLimitMiddleware_ExceedsLimit` in `size_limit_middleware_test.go` to simulate returning a `413 Request Entity Too Large` error, ensuring the test verifies that clients receive the right HTTP status when the body exceeds the limit check.

### DIFF
--- a/cmd/api/app.go
+++ b/cmd/api/app.go
@@ -161,7 +161,7 @@ func CreateServer(coreApp *app.Application, cfg appconf.Config) (*http.Server, *
 		IdleTimeout:    time.Minute,
 		ReadTimeout:    5 * time.Second,
 		WriteTimeout:   10 * time.Second,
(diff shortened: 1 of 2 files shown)


aaronbrethorst (approved):
Hey Ahmed, this is a tidy follow-up — the clarified `MaxHeaderBytes` comment and the improved test that now verifies the 413 status code both look good. The test structure is better too, with the explicit `if err != nil` guard and early return making the intent clearer.

There's nothing left to change — this PR is ready to merge.

Issue #620, opened 2026-03-06
Add automated OpenAPI conformance testing for API responses


aaronbrethorst:
## Problem

The project has 261 handler tests with good coverage, but none validate that response JSON structures match the OpenAPI spec. As the spec evolves, or as new fields are added/removed, there's no automated check that responses remain conformant.

This gap analysis found multiple field-level mismatches (missing fields, JSON tag case errors, type mismatches) that existing tests didn't catch because they test behavior, not schema conformance.

## Proposed Solution

Add a conformance testing layer that validates API response bodies against the OpenAPI spec. Options include:

1. **Spec-generated Go types**: Use an OpenAPI code generator to create reference types, then compare serialized responses against them
2. **JSON Schema validation**: Extract JSON schemas from the OpenAPI spec and validate response bodies in tests
3. **Golden file testing**: Capture expected response shapes and diff against actual outputs

This should run as part of `make test` to catch regressions automatically.

## Severity
Medium — prevents the class of silent spec drift bugs found in this analysis.

PR #638, opened 2026-03-08
feat: add OpenAPI conformance testing layer and resolve spec drift bugs

3rabiii:
### Objective
This PR introduces an automated conformance testing layer to validate API response bodies against the OpenAPI specification, addressing the gap where responses could silently drift from the expected schema. 

By implementing this, we ensure strict SDK compatibility and automatically catch any future structural regressions during `make test`.

### Implementation Details
- **Validation Engine:** Integrated `github.com/getkin/kin-openapi` (v0.133.0) for robust OpenAPI 3.0 schema validation.
- **Test Coverage:** Added `openapi_conformance_test.go` encompassing **27 conformance tests** across 5 distinct groups:
  - Static (20 endpoints)
  - Location (3 endpoints)
  - Search (2 endpoints)
  - RealTime (1 endpoint)
  - Errors (2 endpoints)
- **Spec Reference:** Added a local `openapi.yml` (with a minor patch to resolve a data wrapper inconsistency in `stops-for-agency`).

### Bugs Caught & Fixed (Immediate ROI)
The new conformance layer immediately identified and helped resolve **8 spec drift bugs** across the codebase:
* **Missing `limitExceeded`:** Added `"limitExceeded": false` to all `NewEntryResponse` formations.
* **`TripStatusForTripDetails` required fields:** Transitioned 8 fields from pointer/`omitempty` to value types to ensure they are always present.
* **Null array in `schedule-for-stop`:** Added a nil guard to prevent null `stopRouteSchedules`.
* **Null array in `block` handler:** Initialized `RouteIDs`/`StaticRouteIDs` as empty slices to prevent null references.
* **Null arrays in `stops-for-location`:** Fixed nil slices for `list`, `agencies`, and `routes`.
* **Null array in `trips-for-location`:** Added a nil guard for the `trips` reference.
* **Missing `outOfRange` in `search/route`:** Switched to using `NewListResponseWithRange`.
* **Critical missing fields in `VehicleStatus` / `TripStatus`:** Updated the `vehicle.go` model to include strictly required fields (`tripId`, `occupancyCapacity`, `occupancyCount`, `occupancyStatus`, `predicted`, etc.) and updated the respective handlers, **simultaneously resolving the critical SDK-blocker issue.**
### All Tests Passed
image

@aaronbrethorst 
fixes: #620

### DIFF
--- a/cmd/api/app.go
+++ b/cmd/api/app.go
@@ -162,8 +162,11 @@ func CreateServer(coreApp *app.Application, cfg appconf.Config) (*http.Server, *
 	// Apply global compression around the entire mux
 	compressedMux := restapi.CompressionMiddleware(mux)
 
+	// Add freshness middleware
+	freshnessHandler := api.FreshnessMiddleware(compressedMux)
+
 	// Wrap with security middleware
-	secureHandler := api.WithSecurityHeaders(compressedMux)
+	secureHandler := api.WithSecurityHeaders(freshnessHandler)
 
 	// Add metrics middleware
 	metricsHandler := restapi.MetricsHandler(coreApp.Metrics)(secureHandler)
@@ -184,7 +187,7 @@ func CreateServer(coreApp *app.Application, cfg appconf.Config) (*http.Server, *
 		IdleTimeout:    time.Minute,
 		ReadTimeout:    5 * time.Second,
 		WriteTimeout:   10 * time.Second,
-		MaxHeaderBytes: 1 << 20, // 1 MB
+		MaxHeaderBytes: 1 << 20, // 1 MB (explicit; matches Go's default)
 		ErrorLog:       slog.NewLogLogger(coreApp.Logger.Handler(), slog.LevelError),
 	}
 
--- a/internal/gtfs/gtfs_manager.go
+++ b/internal/gtfs/gtfs_manager.go
@@ -44,6 +44,7 @@ type Manager struct {
 	GtfsDB                         *gtfsdb.Client
 	routesByAgencyID               map[string][]*gtfs.Route
 	lastUpdated                    time.Time
+	lastUpdatedUnixNanos           atomic.Int64 // Lock-free freshness tracking
 	isLocalFile                    bool
 	realTimeTrips                  []gtfs.Trip
 	realTimeVehicles               []gtfs.Vehicle
@@ -85,6 +86,9 @@ type Manager struct {
 
 	// Exported metrics client dependency
 	Metrics *metrics.Metrics
+
+	// Tracks the last successful update time per feed
+	feedLastUpdate map[string]time.Time
 }
 
 // clearFeedData removes stale data for a specific feed when the staleness threshold is crossed
@@ -98,6 +102,8 @@ func (manager *Manager) clearFeedData(feedID string) {
 
 	delete(manager.feedVehicleTimestamp, feedID)
 	delete(manager.feedVehicleLastSeen, feedID)
+
+	delete(manager.feedLastUpdate, feedID)
 
 	manager.rebuildMergedRealtimeLocked()
 }
@@ -240,6 +246,7 @@ func InitGTFSManager(ctx context.Context, config Config) (*Manager, error) {
 		feedTrips:                      make(map[string][]gtfs.Trip),
 		feedVehicles:                   make(map[string][]gtfs.Vehicle),
 		feedAlerts:                     make(map[string][]gtfs.Alert),
+		feedLastUpdate:                 make(map[string]time.Time),
 		feedAgencyFilter:               make(map[string]map[string]bool),
 		feedVehicleLastSeen:            make(map[string]map[string]time.Time),
 		feedVehicleTimestamp:           make(map[string]uint64),
@@ -853,6 +860,50 @@ func (manager *Manager) SetRealTimeTripsForTest(trips []gtfs.Trip) {
 
 	manager.feedTrips["_test"] = trips
 	manager.rebuildMergedRealtimeLocked()
+}
+
+// GetStaticLastUpdated returns the timestamp when static GTFS data was last loaded lock-free.
+func (manager *Manager) GetStaticLastUpdated() time.Time {
+	nanos := manager.lastUpdatedUnixNanos.Load()
+	if nanos == 0 {
+		return time.Time{}
+	}
+	// Append .UTC() here to ensure the RFC3339 string always ends in 'Z'
+	return time.Unix(0, nanos).UTC()
+}
+
+// GetFeedUpdateTimes returns a copy of the last update times for all realtime feeds.
+func (manager *Manager) GetFeedUpdateTimes() map[string]time.Time {
+	manager.realTimeMutex.RLock()
+	defer manager.realTimeMutex.RUnlock()
+
+	// Return a copy to prevent concurrent map read/write outside the lock
+	result := make(map[string]time.Time, len(manager.feedLastUpdate))
+	for k, v := range manager.feedLastUpdate {
+		result[k] = v
+	}
+	return result
+}
+
+// SetFeedUpdateTime safely records the time a feed was successfully updated.
+func (manager *Manager) SetFeedUpdateTime(feedID string, t time.Time) {
+	manager.realTimeMutex.Lock()
+	defer manager.realTimeMutex.Unlock()
+
+	// Ensure map is initialized (helpful for tests)
+	if manager.feedLastUpdate == nil {
+		manager.feedLastUpdate = make(map[string]time.Time)
+	}
+
+	manager.feedLastUpdate[feedID] = t
+}
+
+// SetStaticLastUpdatedForTest manually sets the static data timestamp for testing purposes.
+func (manager *Manager) SetStaticLastUpdatedForTest(t time.Time) {
+	manager.staticMutex.Lock()
+	defer manager.staticMutex.Unlock()
+	manager.lastUpdated = t
+	manager.lastUpdatedUnixNanos.Store(t.UnixNano())
 }
 
 // AddTestAlert is a helper method used ONLY for testing to inject mock alerts safely.
--- a/internal/gtfs/gtfs_manager_mock.go
+++ b/internal/gtfs/gtfs_manager_mock.go
@@ -63,6 +63,8 @@ type MockVehicleOptions struct {
 	CurrentStopSequence *uint32
 	StopID              *string
 	CurrentStatus       *gtfs.CurrentStatus
+	OccupancyStatus     *gtfs.OccupancyStatus
+	NoTrip              bool // NoTrip creates a vehicle with Trip == nil, simulating a GTFS-RT vehicle.with no current trip assignment, which VehiclesForAgencyID filters out.
 }
 
 func (m *Manager) MockAddVehicleWithOptions(vehicleID, tripID, routeID string, opts MockVehicleOptions) {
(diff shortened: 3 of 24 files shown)


aaronbrethorst (changes requested):
Hey Adel, thank you for the thorough follow-up commit — you've addressed the three critical issues from my previous review correctly. The `*TripStatus`, `*Location`, and `LastKnownLocation` changes are all exactly right, and the else branch in `vehicles_for_agency_handler.go` for default trip status is well-implemented. The conformance test infrastructure itself is really clean and will be a valuable guardrail going forward.

There are a few remaining items to address before we can merge.

## Critical

1. **`VehicleStatus.LastUpdateTime` and `LastLocationUpdateTime` still violate the OpenAPI spec** (`internal/models/vehicle.go:5-6`)

   These fields remain as `*int64` with `omitempty`:

   ```go
   LastLocationUpdateTime *int64 `json:"lastLocationUpdateTime,omitempty"`
   LastUpdateTime         *int64 `json:"lastUpdateTime,omitempty"`
   ```

   The OpenAPI spec lists both as **required** on `VehicleStatus`. When a vehicle has no timestamp (the `if vehicle.Timestamp != nil` guard at `vehicles_for_agency_handler.go:76`), these fields are omitted entirely from the JSON response — violating the spec. The conformance tests pass only because all RABA test vehicles happen to have timestamps.

   This is the same class of bug you already correctly fixed in `TripStatus`, where you changed these identical fields from `*int64` to `int64`. Apply the same fix here for consistency:

   ```go
   LastLocationUpdateTime int64 `json:"lastLocationUpdateTime"`
   LastUpdateTime         int64 `json:"lastUpdateTime"`
   ```

   Then update `vehicles_for_agency_handler.go` to assign the value directly instead of via pointer.

2. **`TripStatus` timestamp zero values produce January 1, 1970** (`internal/restapi/vehicles_helper.go:116-118`)

   With `LastUpdateTime` and `LastLocationUpdateTime` now as `int64` value types, when `BuildVehicleStatus` encounters a stale or nil vehicle (line 116-118), it sets `Status` and `Phase` but does not set timestamps. They remain at `0`, which serializes as the Unix epoch — January 1, 1970.

   Every scheduled (non-realtime) trip will have `"lastUpdateTime": 0` in its `tripStatus`. API clients receiving this will interpret it as a real timestamp from 1970.

   **Fix**: Set a meaningful fallback in `BuildVehicleStatus` when no vehicle timestamp exists — use the current time or service date as a default. Similarly, ensure the `else` branch in `vehicles_for_agency_handler.go:174-178` (the new default trip status for vehicles without trips) sets these timestamps.

## Important

3. **Vendored OpenAPI spec has no provenance comment** (`testdata/openapi.yml`)

   This was item #5 from my previous review and hasn't been addressed yet. The file begins directly with `openapi: 3.0.0` with no indication of where it came from, what patches were applied, or how to update it.

   **Fix**: Add a YAML comment header:

   ```yaml
   # Source: https://github.com/OneBusAway/sdk-config/blob/main/openapi.yml
   # Fetched: 2026-03-08
   # Local modifications: [describe any patches applied]
   openapi: 3.0.0
   ```

4. **`TripStatus.NextStop` has `omitempty` removed, but the OpenAPI spec does not list it as required** (`internal/models/trip_details.go:47`)

   The PR changes `NextStop` from `json:"nextStop,omitempty"` to `json:"nextStop"`. However, `nextStop` is NOT in the OpenAPI spec's `required` array for TripStatus. When a vehicle is at the last stop on a route, this field will emit an empty string `""` instead of being omitted.

   **Fix**: Revert to `json:"nextStop,omitempty"`, or if you believe the field should always be present, update the vendored spec's required list to match and file an issue upstream.

5. **`OccupancyCapacity`/`OccupancyCount` code comment contradicts the type change** (`internal/restapi/trips_helper.go:32-33`)

   The comment says: *"OccupancyCapacity and OccupancyCount are left nil (null in JSON) when no data."* But these fields are now `int` value types, making nil impossible. They will always serialize as `0`.

   This is a tension between the OpenAPI spec (which lists them as required) and the original code intent (which wanted to distinguish "no data" from "zero occupancy"). Please make a deliberate choice: if `0` is acceptable for "no data", remove the now-misleading comment. If the distinction matters for downstream consumers, keep them as `*int` and document the spec deviation.

6. **Missing conformance tests for key endpoints**

   The conformance test suite covers 19+ endpoints, which is great, but several important ones are missing:

   - **`arrivals-and-departures-for-stop`** — the most important transit endpoint, powers real-time arrival boards. Complex response structure with nested entry.
   - **`arrival-and-departure-for-stop`** — single-arrival variant, also modified in this PR.
   - **`trip-for-vehicle`** — requires real-time data; should be tested alongside `vehicles-for-agency` in the real-time test group.

   These don't need to block the initial merge, but consider filing a follow-up issue to track adding them.

7. **Real-time conformance test doesn't verify vehicles were loaded** (`internal/restapi/openapi_conformance_test.go:402`)

   After creating the GTFS manager in `TestOpenAPIConformance_RealTimeEndpoints`, there's no assertion that real-time vehicles were actually loaded. If the protobuf files fail to parse, the test would validate an empty list response — which passes conformance but provides no meaningful coverage of the vehicle serialization logic you modified.

   **Fix**: After creating the API, add:

   ```go
   vehicles := api.GtfsManager.VehiclesForAgencyID(agencyID)
   require.NotEmpty(t, vehicles, "Real-time vehicles must be loaded for conformance testing")
   ```

## Fit and Finish

8. **`VehicleStatus.OccupancyCapacity`/`OccupancyCount` use `int` with `omitempty` — zero is silently dropped** (`internal/models/vehicle.go:10-11`)

   With `int` and `omitempty`, a value of `0` causes these fields to be omitted. A bus with zero passengers is a valid state, but it becomes indistinguishable from "no data". The same fields on `TripStatus` use `int` *without* `omitempty`. Consider aligning the approach.

9. **`TripStatus.Position` is a value type `Location` but is not nullable in the spec** (`internal/models/trip_details.go:54`)

   When no vehicle position exists, `Position` serializes as `{"lat":0,"lon":0}` (null island). The spec marks `position` as a required object but not nullable. This is a pre-existing issue and not introduced by this PR, but worth noting since the conformance framework could be extended to catch it.

10. **Per-request server creation in conformance tests is wasteful** (`internal/restapi/openapi_conformance_test.go:80`)

    `serveAndCaptureRawJSON` creates a new `httptest.NewServer` for every request. Within `TestOpenAPIConformance_StaticEndpoints`, this creates and tears down 19+ servers. A single persistent server per test function would be more efficient. Not blocking, but a nice cleanup.

Ahmedhossamdev:
Hey @3rabiii, great work!
I have a few questions:
1. Will the openapi.yml be updated only if it changes, or will it be fetched with every test from https://github.com/OneBusAway/sdk-config?
2. If you've made any changes to openapi.yml, please open an issue here: https://github.com/OneBusAway/sdk-config

aaronbrethorst (changes requested):
Hey Adel, excellent work on this third iteration. You've systematically addressed the critical and important issues from both previous reviews — the `*TripStatus`/`*Location` pointer fixes, the timestamp fallback defaults, the provenance header on `openapi.yml`, and the vehicle-loaded assertion in the real-time test. The conformance test infrastructure is clean, well-structured, and has already proven its value by catching 8 real spec drift bugs. This is ready to merge after one small fix.

## Important

1. **Stale comments reference nil behavior on `int` fields** (`internal/restapi/trips_helper.go:32-33, 47`)

   Two comments still describe nil/null behavior that's impossible with the current types:

   - Line 32-33: *"OccupancyCapacity and OccupancyCount are left nil (null in JSON) when no data. Java OBA uses nullable Integer types that serialize conditionally."*
   - Line 47: *"We intentionally leave OccupancyCapacity as nil here."*

   Both `OccupancyCapacity` and `OccupancyCount` are now `int` (not `*int`) in `TripStatus`, so they can never be nil — they default to `0`. A developer reading these comments would incorrectly believe the fields could be null in JSON output.

   **Fix**: Update line 32-33 to: *"OccupancyCapacity and OccupancyCount default to 0 when no data is available."* Update line 47 to: *"We intentionally leave OccupancyCapacity at its zero value (0) here, as GTFS-RT OccupancyPercentage has no direct mapping to OBA's capacity-based model."*

## Fit and Finish

2. **`SituationIDs` comment claims `omitempty` omits empty slices — it doesn't** (`internal/models/trip_details.go:67-68`)

   The comment says: *"JSON output is unaffected because omitempty omits both nil and empty slices."* This is incorrect — Go's `encoding/json` `omitempty` omits nil slices but NOT empty (`[]string{}`) slices. Since `NewTripStatus()` initializes `SituationIDs` to `[]string{}`, the field will always appear as `[]` in JSON regardless of `omitempty`. The comment masks a minor logic conflict. Consider removing the comment or clarifying the actual behavior.

3. **Missing conformance tests for three modified endpoints**

   The conformance suite covers 19+ endpoints impressively, but three endpoints that were modified in this PR lack conformance tests:

   - `arrivals-and-departures-for-stop` — the most important transit endpoint, powers real-time arrival boards
   - `arrival-and-departure-for-stop` — single-arrival variant, also modified
   - `trip-for-vehicle` — requires real-time data, should be in the real-time test group

   Not blocking — these are harder to test with static data. Consider filing a follow-up issue to track adding them.

4. **`Scheduled` field comment references tracking without a link** (`internal/models/trip_details.go:64`)

   The comment says *"Tracked as a known spec deviation"* but doesn't reference where it's tracked. Consider changing to *"This is a known deviation from the OpenAPI spec"* or adding a GitHub issue link.

## Previous Review Items — Verified Addressed

- [x] `VehicleStatus.LastUpdateTime`/`LastLocationUpdateTime` changed to `int64` value types with `currentTime` fallback
- [x] `TripStatus.LastKnownLocation` uses `*Location` with no `omitempty`
- [x] Default `TripStatus` with `status: "default"` and `phase: "scheduled"` for vehicles without trips
- [x] OpenAPI spec provenance header added at top of `testdata/openapi.yml`
- [x] Real-time test asserts `vehicles` are loaded before conformance check
- [x] `VehicleStatus.TripStatus` is `*TripStatus` without `omitempty` — always serialized
- [x] Upstream spec issue filed at OneBusAway/sdk-config#1

## Verdict

**Merge** once the merge conflict is resolved. Please fix the stale comments in `trips_helper.go` (item #1) while you're rebasing.

aaronbrethorst (approved):
Hey Adel, you've done a thorough job addressing the feedback from three rounds of review. The stale comments in `trips_helper.go` are now accurate, the incorrect `omitempty` comment on `SituationIDs` is removed, the `Scheduled` field comment no longer references phantom tracking, and the merge conflict resolution is clean. All tests pass, and lint is clean. The conformance testing infrastructure you've built is genuinely valuable — catching 8 spec drift bugs on its first run speaks for itself.

There's one more stale comment that slipped through the latest update, and a couple of pre-existing items worth tracking as follow-ups. None of these block the merge.

## Fit and Finish

1. **One more stale comment from the pointer-to-value migration** (`internal/restapi/trips_helper.go:304`)

   ```go
   // Dereference schedule deviation safely, default to 0 if not set
   schedDev := int64(status.ScheduleDeviation)
   ```

   `ScheduleDeviation` is now `int` (not `*int`), so there's nothing to dereference and no nil-default behavior. The code is just a type conversion. Either remove the comment or update it to something like `// Convert to int64 for arithmetic below`.

2. **Minor typo in `Scheduled` field comment** (`internal/models/trip_details.go:64`)

   There's a stray comma-space after the closing parenthesis: `!Predicted) ,this` — should be `!Predicted). This`.

3. **Pre-existing: `NewArrivalsAndDepartureResponse` missing `limitExceeded`** (`internal/models/response.go:49-61`)

   This PR correctly added `limitExceeded` to `NewEntryResponse`, `NewListResponse`, and `NewListResponseWithRange`, but `NewArrivalsAndDepartureResponse` was not modified (it wasn't in scope). Its `data` map lacks `limitExceeded`. Consider addressing this alongside the follow-up conformance tests in issue #665, since this endpoint is also missing from the conformance suite.

PR #639, opened 2026-03-08
fix(models): align VehicleStatus with OpenAPI spec (#613)

### LINKED ISSUE #613: VehicleStatus and TripStatus (vehicle.go) missing 16+ fields required by OpenAPI spec
## Problem

`VehicleStatus` and `TripStatus` in `internal/models/vehicle.go` are missing numerous fields that the OpenAPI spec requires. Vehicle tracking is a core consumer use case, and any client using the OneBusAway SDKs will see null/missing values for fields it expects.

### `VehicleStatus` missing fields (4):
- `tripId`
- `occupancyCapacity`
- `occupancyCount`
- `occupancyStatus`

### `TripStatus` (vehicle.go) missing fields (12):
- `frequency`
- `lastKnownDistanceAlongTrip`
- `lastKnownLocation`
- `lastKnownOrientation`
- `lastLocationUpdateTime`
- `lastUpdateTime`
- `occupancyCapacity`
- `occupancyCount`
- `occupancyStatus`
- `scheduledDistanceAlongTrip`
- `situationIds`
- `vehicleId`

The OpenAPI spec uses a single `TripStatus` schema everywhere, but `vehicle.go:18-34` has a stripped-down version with only 14 of the 27 required fields.

## Files
- `internal/models/vehicle.go`

## Severity
Critical — launch blocker for SDK-compatible clients.

Adityatorgal17:
### Background
PR #627 resolved the 12 missing fields in `TripStatus`.  
This PR completes the alignment by fixing the remaining missing fields in `VehicleStatus`.

---

#### `internal/models/vehicle.go`

| Field | Change | Reason |
|------|--------|--------|
| `tripId` | Added (`string`, no `omitempty`) | Required by OpenAPI spec |
| `lastUpdateTime` | Removed `omitempty` | Required by OpenAPI spec |
| `lastLocationUpdateTime` | Removed `omitempty` | Required by OpenAPI spec |
| `location` | Removed `omitempty` | Required by OpenAPI spec |
| `tripStatus` | Removed `omitempty` | Required by OpenAPI spec |
| `occupancyStatus` | Added (`string`, `omitempty`) | Optional per spec |
| `occupancyCapacity` | Added (`*int`, `omitempty`) | Optional per spec |
| `occupancyCount` | Added (`*int`, `omitempty`) | Optional per spec |

---

#### `vehicles_for_agency_handler.go`

- Populate `tripId` using `FormCombinedID(agencyID, vehicle.Trip.ID.ID)`
- Propagate `occupancyStatus` from GTFS-RT to both `VehicleStatus` and `TripStatus`
- `occupancyCapacity` and `occupancyCount` intentionally left `nil`  
  - GTFS-RT only provides an enum + percentage
  - Maglev has no agency fleet capacity database

---
#### `vehicles_for_agency_handler_test.go`(Updated)
---

### Tests Passed

```bash
make test
```

### DIFF
--- a/internal/gtfs/gtfs_manager_mock.go
+++ b/internal/gtfs/gtfs_manager_mock.go
@@ -63,6 +63,8 @@ type MockVehicleOptions struct {
 	CurrentStopSequence *uint32
 	StopID              *string
 	CurrentStatus       *gtfs.CurrentStatus
+	OccupancyStatus     *gtfs.OccupancyStatus
+	NoTrip              bool // NoTrip creates a vehicle with Trip == nil, simulating a GTFS-RT vehicle.with no current trip assignment, which VehiclesForAgencyID filters out.
 }
 
 func (m *Manager) MockAddVehicleWithOptions(vehicleID, tripID, routeID string, opts MockVehicleOptions) {
@@ -75,19 +77,26 @@ func (m *Manager) MockAddVehicleWithOptions(vehicleID, tripID, routeID string, o
 		}
 	}
 	now := time.Now()
-	v := gtfs.Vehicle{
-		ID:        &gtfs.VehicleID{ID: vehicleID},
-		Timestamp: &now,
-		Trip: &gtfs.Trip{
+
+	var trip *gtfs.Trip
+	if !opts.NoTrip {
+		trip = &gtfs.Trip{
 			ID: gtfs.TripID{
 				ID:      tripID,
 				RouteID: routeID,
 			},
-		},
+		}
+	}
+
+	v := gtfs.Vehicle{
+		ID:                  &gtfs.VehicleID{ID: vehicleID},
+		Timestamp:           &now,
+		Trip:                trip,
 		Position:            opts.Position,
 		CurrentStopSequence: opts.CurrentStopSequence,
 		StopID:              opts.StopID,
 		CurrentStatus:       opts.CurrentStatus,
+		OccupancyStatus:     opts.OccupancyStatus,
 	}
 	m.realTimeVehicles = append(m.realTimeVehicles, v)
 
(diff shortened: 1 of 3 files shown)


aaronbrethorst (approved):
Nice follow-up, Aditya — you've addressed every item from the prior review cleanly and thoroughly. The occupancy propagation test is exactly what was needed, the nil-trip regression guard is clever (sharing `rawRouteID` so the vehicle *would* leak through if the filter were removed), and the null-value assertions in the model test make the serialization contract precise. Well done.

## Critical

_None._

## Important

_None._

## Fit and Finish

1. **`tripId` format not explicitly asserted.** `TestVehiclesForAgencyHandler_OccupancyPropagation` implicitly covers the new `tripId` assignment (`vehicles_for_agency_handler.go:95`), but never asserts its value. Adding `assert.Equal(t, agencyID+"_"+tripID, vehicle["tripId"])` would catch a regression if `FormCombinedID` were accidentally removed or its arguments swapped — the current tests would still pass if `tripId` were set to just the raw trip ID without the agency prefix. Minor, but easy to add.

2. **Typo in mock comment.** `gtfs_manager_mock.go:67`: "simulating a GTFS-RT vehicle.with no current trip" → "vehicle with" (period instead of space).

## Summary

All four items from the prior review have been addressed:

| Prior Feedback | Status |
|---|---|
| Test for occupancy propagation path | Addressed — `TestVehiclesForAgencyHandler_OccupancyPropagation` exercises both `vehicleStatus` and `tripStatus` propagation |
| Test for vehicle-without-trip case | Addressed — `TestVehiclesForAgencyHandler_VehicleWithoutTrip` with `NoTrip` mock option |
| Model test should verify null values | Addressed — explicit `assert.Nil` checks on pointer fields + empty-string check on `tripId` |
| Conditional assertions could silently pass | Addressed — `t.Log` calls added, and the new mock-based tests use deterministic data with `require` guards |

Tests pass, lint is clean (pre-existing `errcheck` in `gtfs_manager_test.go` is unrelated to this PR). The model changes correctly align with the OpenAPI spec. Ready to merge.

**Verdict: Merge** ✅

PR #641, opened 2026-03-09
chore: clean up dead sequence guard and standardize error logging

AhmedAlian7:
## Description
- This PR addresses follow-up suggestions PR #590

### Changes
* **Removed dead guard clause (`arrival_and_departure_for_stop_handler.go`):** Dropped the redundant `StopSequence` check. `GetTargetStopTimeWithTotalStopsBySequence` already filters by `stop_sequence` in the `WHERE` clause, making the subsequent sequence match check unnecessary.
* **Standardized error logging (`trips_helper.go`):** Updated `getFirstStopOfNextTripInBlock` to use `slog.String("error", err.Error())` instead of `slog.Any`, matching the logging style used throughout the rest of the file to make log queries easier to write.

### DIFF
--- a/internal/restapi/arrival_and_departure_for_stop_handler.go
+++ b/internal/restapi/arrival_and_departure_for_stop_handler.go
@@ -230,11 +230,6 @@ func (api *RestAPI) arrivalAndDepartureForStopHandler(w http.ResponseWriter, r *
 		}
 	}
 
-	if params.StopSequence != nil && int64(*params.StopSequence) != targetRow.StopSequence {
-		api.sendNotFound(w, r)
-		return
-	}
-
 	targetStopTime := struct {
 		ArrivalTime   int64
 		DepartureTime int64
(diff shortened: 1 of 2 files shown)


aaronbrethorst (approved):
Nice cleanup, Ahmed! Both changes are well-reasoned and correct.

## Review Summary

### Changes Reviewed
1. **Removed dead guard clause** (`arrival_and_departure_for_stop_handler.go`): The `StopSequence` check after the `GetTargetStopTimeWithTotalStopsBySequence` query was indeed redundant — the SQL query already filters by `stop_sequence` in its `WHERE` clause (line 1162 of `query.sql`), so if the sequence didn't match, the query would have already returned `sql.ErrNoRows`.

2. **Standardized error logging** (`trips_helper.go`): Changed `slog.Any("error", err)` to `slog.String("error", err.Error())` to match the dominant pattern in the file. Consistent logging makes log queries easier to write and reason about.

## Feedback

### Fit and Finish

1. **Remaining `slog.Any("error", err)` calls** (`trips_helper.go:701`, `trips_helper.go:707`): There are two more `slog.Any("error", err)` calls in `getSituationIDsForTrip` that use the old pattern. Since this PR is about standardizing the logging style, it might be nice to update those too for full consistency. Not a blocker — could be a follow-up.

## Verdict: **Merge** :white_check_mark:

Both changes are correct and well-scoped. Tests pass, no regressions.

PR #667, opened 2026-03-11
fix(restapi): log DB errors in trips-for-route block loop (#659)

### LINKED ISSUE #659: trips-for-route handler silently ignores DB errors in block loop
While looking through `internal/restapi/trips_for_route_handler.go`, I noticed that database errors inside the block-processing loop are currently ignored.

In particular, errors from `GetTripsInBlock` and `GetActiveTripInBlockAtTime` lead to:

```go
if err != nil {
    continue
}
```

So if one of these queries fails for a block, the handler just skips that block and continues processing.

This could potentially lead to incomplete results being returned without any indication that something went wrong. It might also make production debugging harder since the errors are not logged.

Would it make sense to add logging for these cases (for example including the block ID), and possibly a metric to track how often this happens?

Happy to open a PR to add logging if that sounds reasonable.

tejasva-vardhan:
## Fix: log DB errors in `trips-for-route` block loop

Fixes #659

While looking through `internal/restapi/trips_for_route_handler.go`, I noticed that errors from the database queries `GetTripsInBlock` and `GetActiveTripInBlockAtTime` were being ignored inside the block-processing loop.

Previously the code handled errors like this:

```go
if err != nil {
    continue
}
```

This meant that if a database query failed for a block, the handler would silently skip that block and continue processing. In practice this could lead to incomplete results being returned without any indication that something went wrong, and it also makes production debugging harder since the failure isn’t visible in logs.

### What this change does

This change adds structured logging when those queries fail. The handler now logs the error along with the `block_id` before continuing to the next block.

Example:

```go
logging.LogError(api.Logger, "failed to fetch trips in block", err, slog.String("block_id", blockID))
```

### Behavior

* No change to API responses or control flow
* The handler still skips the block on error
* Errors are now visible in logs for easier debugging

Overall this is a small change that improves observability without affecting existing behavior.

### DIFF
--- a/internal/restapi/trips_for_route_handler.go
+++ b/internal/restapi/trips_for_route_handler.go
@@ -3,6 +3,7 @@ package restapi
 import (
 	"context"
 	"database/sql"
+	"errors"
 	"log/slog"
 	"net/http"
 	"strconv"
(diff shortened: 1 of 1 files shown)


aaronbrethorst (approved):
Hey Tejasva, thanks for improving observability in the block-processing loop. Surfacing these previously-swallowed errors will make production debugging significantly easier, and your `sql.ErrNoRows` special-casing for `GetActiveTripInBlockAtTime` is exactly right — no active trip at a given time is expected, not exceptional.

There's nothing left to change — this PR is ready to merge. I've opened #693 to track upgrading some related `Debug`-level logging in `buildTripReferences` in the same file, but that's separate from this work. Nice job.

PR #676, opened 2026-03-11
fix(metrics): keep DB pool metrics accurate across GTFS hot-swap

### LINKED ISSUE #675: DB connection pool Prometheus metrics become stale after static GTFS hot-swap
### Description

`maglev_db_connections_*` and `maglev_db_wait_seconds_total` are collected from the initial `*sql.DB` pointer and are not rebound after static GTFS hot-swap replaces the DB client. After a successful hot-swap, API traffic uses the new DB client, but metrics can report stale/zero values from the old (closed) pool.

### Why this matters

This breaks observability and can mislead alerts/dashboards that rely on DB pool health metrics, especially in systems with periodic feed updates.

### Code pointers

- **Collector starts once with initial DB:** `cmd/api/app.go` — `StartDBStatsCollector(gtfsManager.GtfsDB.DB, 15*time.Second)`
- **Collector reads captured DB pointer:** `internal/metrics/metrics.go` — `stats := db.Stats()`
- **Hot-swap closes/replaces DB:** `internal/gtfs/static.go` — closes old DB and assigns `manager.GtfsDB = client`

### Reproduction

1. Start server and scrape `/metrics`.
2. Trigger static GTFS hot-swap (via internal `ForceUpdate` path).
3. Scrape `/metrics` again after collector interval.
4. Observe:
   - **Before swap:** non-zero `maglev_db_connections_open` / `idle`
   - **After swap:** DB metrics drop to zero/stale
   - API endpoints still return `200` and serve data

### Expected

DB pool metrics should reflect the currently active DB client after hot-swap.

### Actual

DB pool metrics continue reflecting the old DB handle captured when collector started.

### Suggested fix

- Make DB stats collection reference the current DB dynamically (or restart/rebind collector on each successful DB swap).
- Ensure thread-safe pointer updates and add regression test for *"metrics remain valid across `ForceUpdate`"*.

Adityatorgal17:
### Summary

Fixes #675 

Fix DB pool Prometheus metrics becoming stale after static GTFS hot-swap.

Previously, `StartDBStatsCollector` captured the initial `*sql.DB` at startup, so after `ForceUpdate` replaced `GtfsDB`, collector stats could reflect the old/closed pool.

### What changed

1. **Updated metrics collector API** to accept a DB provider callback:
   - `StartDBStatsCollector(dbProvider func() *sql.DB, interval time.Duration)`

2. **Updated collector loop** to fetch current DB on each tick:
   - If provider returns `nil`, set DB gauges (`open`, `in_use`, `idle`) to `0`.
   - If DB pointer changes, reset wait-duration baseline for the new pool.

3. **Updated app wiring:**
   - In `cmd/api/app.go`, pass a lock-safe provider closure (`GtfsManager.RLock`/`RUnlock`) that returns the current `GtfsDB.DB`.

4. **Added/updated tests** in `internal/metrics/metrics_test.go`:
   - `nil` provider
   - Provider returning `nil` DB
   - Idempotent start behavior
   - DB swap regression test (`TestStartDBStatsCollector_FollowsDBSwap`)

### Why

`sql.DBStats.WaitDuration` is cumulative per pool, and GTFS hot-swap replaces the pool pointer. Reading a fixed pointer after swap gives stale metrics and can incorrectly show zero/old values while traffic uses the new DB.

### Validation
```bash
make fmt
make test
```

### DIFF
--- a/cmd/api/app.go
+++ b/cmd/api/app.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"database/sql"
 	"encoding/json"
 	"fmt"
 	"log/slog"
@@ -120,9 +121,16 @@ func BuildApplication(ctx context.Context, cfg appconf.Config, gtfsCfg gtfs.Conf
 		Metrics:             appMetrics,
 	}
 
-	// Start DB stats collector if database is available
-	if gtfsManager != nil && gtfsManager.GtfsDB != nil && gtfsManager.GtfsDB.DB != nil {
-		appMetrics.StartDBStatsCollector(gtfsManager.GtfsDB.DB, 15*time.Second)
+	// Start DB stats collector using a provider so metrics follow DB hot-swap.
+	if gtfsManager != nil {
+		appMetrics.StartDBStatsCollector(func() *sql.DB {
+			gtfsManager.RLock()
+			defer gtfsManager.RUnlock()
+			if gtfsManager.GtfsDB == nil {
+				return nil
+			}
+			return gtfsManager.GtfsDB.DB
+		}, 15*time.Second)
 	}
 
 	return coreApp, nil
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -164,8 +164,8 @@ func NewWithLogger(logger *slog.Logger) *Metrics {
 // The interval specifies how often to collect stats.
 // This method is idempotent - calling it multiple times has no effect after the first call.
 // Call Shutdown() to stop the collector.
-func (m *Metrics) StartDBStatsCollector(db *sql.DB, interval time.Duration) {
-	if db == nil {
+func (m *Metrics) StartDBStatsCollector(dbProvider func() *sql.DB, interval time.Duration) {
+	if dbProvider == nil {
 		return
 	}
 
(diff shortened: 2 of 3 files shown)


aaronbrethorst (approved):
Hey Aditya, nice catch on the stale metrics after hot-swap — this is one of those bugs that's easy to overlook because everything *appears* to work until you cross-reference Prometheus with actual traffic patterns. The provider callback pattern is the right solution here: it keeps the collector decoupled from the DB lifecycle while ensuring it always reads from the live pool.

A few things I particularly liked:

- The `WaitDuration` baseline reset on pointer change is exactly correct — cumulative counters from a fresh `*sql.DB` can't be diffed against the old pool's baseline, and you handled that cleanly with the `continue` after setting gauges.
- The RLock-scoped closure in `app.go` is minimal and safe — it can't deadlock with `ForceUpdate`'s write lock since it never holds any other lock.
- The `TestStartDBStatsCollector_FollowsDBSwap` test with `atomic.Pointer` and a held transaction is a clever way to prove the swap is actually detected.
- Zeroing the gauges when the provider returns `nil` is the right call — stale values from a closed pool would be worse than zeros.

There's nothing left to change — this PR is ready to merge.

PR #678, opened 2026-03-11
fix(metrics): guard non-positive DB stats collector interval

### LINKED ISSUE #677: DB stats collector panics on non-positive interval and can become non-restartable
### Description

`StartDBStatsCollector` calls `time.NewTicker(interval)`. If `interval <= 0`, `time.NewTicker` panics.

Even though the panic is recovered, the collector can still become effectively non-restartable because `collectorStarted` is already set before ticker creation:
```go
if !m.collectorStarted.CompareAndSwap(false, true) { return }
```

So after panic/recover, later calls to `StartDBStatsCollector` can return immediately (CAS fails), leaving DB metrics collection unavailable unless that flag is reset.

### Impact

- DB metrics (`maglev_db_connections_*`, `maglev_db_wait_seconds_total`) may stop updating.
- Recovery path depends on panic behavior instead of explicit input validation.
- Operational observability can silently degrade.
-  Keep the recover block for unknown panics, but treat non-positive interval as normal validation logic.

### Expected behavior

Collector should validate interval before startup state is mutated:
- If `interval <= 0`, log and return without setting started state.
- A later valid call should start the collector normally.

### Proposed fix

- Add explicit guard for `interval <= 0` before `collectorStarted.CompareAndSwap(...)`.
- Add regression test to verify:
  1. Zero/negative interval does not start collector.
  2. Subsequent positive interval starts collector successfully.

Adityatorgal17:
Fixes #677 

### Summary

Adds explicit validation in DB stats collector for non-positive intervals and a regression test.

### What changed

1. **`internal/metrics/metrics.go`:**
   - Added early return for `interval <= 0` in `StartDBStatsCollector`.
   - Logs an error (if logger is configured).
   - Validation happens before `collectorStarted.CompareAndSwap(...)`.

2. **`internal/metrics/metrics_test.go`:**
   - Added `TestStartDBStatsCollector_NonPositiveInterval`.
   - Confirms:
     - `interval == 0` does not start collector.
     - `interval < 0` does not start collector.
     - Subsequent valid interval starts collector successfully.

### Why

Avoids relying on panic/recover for input validation and keeps `StartDBStatsCollector`  restartable.

### Validation
```bash
make test
```

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -168,6 +168,12 @@ func (m *Metrics) StartDBStatsCollector(dbProvider func() *sql.DB, interval time
 	if dbProvider == nil {
 		return
 	}
+	if interval <= 0 {
+		if m.logger != nil {
+			m.logger.Error("invalid DB stats collector interval", "interval", interval)
+		}
+		return
+	}
 
 	// Prevent spawning multiple collectors
 	if !m.collectorStarted.CompareAndSwap(false, true) {
--- a/internal/metrics/metrics_test.go
+++ b/internal/metrics/metrics_test.go
@@ -59,6 +59,26 @@ func TestStartDBStatsCollector_ProviderCanReturnNilDB(t *testing.T) {
 	}, time.Second, 20*time.Millisecond)
 }
 
+func TestStartDBStatsCollector_NonPositiveInterval(t *testing.T) {
+	db, err := sql.Open(gtfsdb.DriverName, ":memory:")
+	require.NoError(t, err)
+	defer func() { _ = db.Close() }()
+
+	m := New()
+
+	m.StartDBStatsCollector(func() *sql.DB { return db }, 0)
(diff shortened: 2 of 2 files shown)


aaronbrethorst (approved):
Aditya, nice work on this — the fix is clean and correctly targeted. Placing the `interval <= 0` guard before `CompareAndSwap` is exactly right: it keeps invalid input from permanently locking out the collector, which was the core bug described in #677.

The merge conflicts from the previous review have been resolved, and the PR is now mergeable. Tests all pass.

## Feedback

### Fit and Finish

1. **Nit: log level could be `Warn` instead of `Error`** — `internal/metrics/metrics.go:173`. Passing a non-positive interval is a programmer mistake / misconfiguration, not a runtime failure. `Warn` is arguably more appropriate here since nothing is broken — the collector simply doesn't start. This is not a blocker; `Error` is defensible too.

## Verdict

**Merge.** This is a well-scoped fix with good test coverage. The guard, placement, and test cases all make sense.

PR #680, opened 2026-03-12
feat: GTFS Frequencies Phase 3 (API Integration & Handlers)

### LINKED ISSUE #666: Feature: GTFS Frequencies Phase 3 (API Integration & Handlers)
This issue tracks **Phase 3** (the final phase) of the GTFS Frequencies implementation.

With Phase 1 (Storage/DB) and Phase 2 #582 (Core Logic & Helpers) already merged, this phase focuses entirely on wiring the frequency data into the API Handlers and Response Models.

### Implementation

**Part 1: Models & Core Handlers**
* `stop_time_schedule.go`: Implement `ScheduleFrequency` struct and update slices.
* `trip_details_handler.go`: Populate `Frequency`, `TripStatusForTripDetails.Frequency`, and `Schedule.Frequency`.
* `trips_helper.go` (`BuildTripSchedule`): Populate `Schedule.Frequency`.
* `arrival_and_departure_for_stop_handler.go`: Populate `Frequency` and adjust predicted times based on headways for `exact_times=0`.
* `trip_for_vehicle_handler.go`: Populate frequency for the vehicle's active trip.

**Part 2: Batch Queries & Schedule Handlers**
* `schedule_for_stop_handler.go`: Build `ScheduleFrequencies` using the new struct. **Dynamically expand in-memory stop times for `exact_times=1`.**
* ~`schedule_for_route_handler.go`: Include frequency data per-trip-direction group.~ *(Dropped for legacy parity)*
* `trips_for_route_handler.go`: Populate frequency fields using highly optimized batch queries to avoid N+1.
* `trips_for_location_handler.go`: Populate using batch queries utilizing robust negative caching and fallback querying.

### Testing Requirements
* Create `createTestApiWithFrequencyData(t)` fixture helper with mixed trip types (`exact_times=0`, `exact_times=1`, normal).
* Add subtests for `trip_details_handler_test.go`.
* Add subtests for `trips_for_route_handler_test.go` & `trips_for_location_handler_test.go`.
* Add subtests for `schedule_for_stop_handler_test.go` (including expansion math).
* Add subtests for `arrival_and_departure_for_stop_handler_test.go`.
* Verify standard regression suite passes perfectly with normal (non-frequency) data.

---

> [!NOTE]
> **Final Scope Adjustments for Legacy Parity**
> To guarantee exact behavioral parity with the legacy Java system, the following scope decisions were applied during implementation:
> * `schedule_for_route_handler` frequency inclusion was **dropped** (the legacy system did not include it).
> * Time-window filtering changes in `trips_for_location_handler` were **dropped** to strictly maintain the legacy GTFS-RT vehicle requirement.
> * `ScheduleFrequency` struct and `trip_details` refactoring were already handled in [models.Frequency conflates two distinct legacy objects — split into Frequency and ScheduleFrequency #1147].

3rabiii:
### Description
This PR implements **Phase 3** (the final phase) of the GTFS Frequencies feature. It wires the database storage and core logic (built in Phases 1 & 2) #582  directly into the API handlers and response models.

### Key Changes
This PR is cleanly separated into 3 logical commits to make reviewing easier:

**1. Core Models & Single-Trip Handlers:**
* Added the new `ScheduleFrequency` struct to `stop_time_schedule.go` for the schedule endpoints.
* Integrated `NewFrequencyFromDB` for proper Nanosecond to Epoch MS time conversion.
* Updated `trip_details_handler.go`, `trip_for_vehicle_handler.go`, and `arrivals_and_departure_for_stop.go` to look up and populate the `Frequency` object for frequency-based trips (handling both `exact_times=0` active window matching and `exact_times=1`).

**2. Batch Queries & Schedule Handlers:**
* Updated `schedule_for_stop` and `schedule_for_route` handlers to build `ScheduleFrequencies` arrays and expand virtual stop times for `exact_times=1` trips.
* Updated `trips_for_route` and `trips_for_location` to use batch queries (`GetFrequenciesForTrips`) to populate the headway `*int64` field efficiently without N+1 query issues.

**3. Testing:**
* Consolidated all frequency testing into a new comprehensive suite: `frequency_integration_test.go`.
* Includes subtests for all affected handlers, covering `exact_times=0`, `exact_times=1`, and `null`/empty fallbacks.
* Added a regression suite to ensure endpoints still behave perfectly when no frequency data is present.

image

@aaronbrethorst 
closes : #666


aaronbrethorst (changes requested):
Adel, this is a clean Phase 3 — wiring the frequency data through every handler that needs it, with batch queries to avoid N+1 problems and a comprehensive test suite covering both `exact_times` modes. The `ScheduleFrequency` type for the schedule endpoints is the right call since those responses need `serviceId` and `tripId` context that the trip-details `Frequency` struct doesn't carry. The test structure with `insertFrequencyData` + deferred cleanup is solid.

Two items need fixing before merge, plus a suggestion for a follow-up cleanup.

## Important

1. **`time.Now()` used instead of `api.Clock.Now()` in two places** — breaks mock clock testing

   - `internal/restapi/trips_helper.go:223`:
     ```go
     active := gtfsInternal.GetActiveHeadwayForTime(frequencies, serviceDate, time.Now().In(serviceDate.Location()))
     ```
     Should be:
     ```go
     active := gtfsInternal.GetActiveHeadwayForTime(frequencies, serviceDate, api.Clock.Now().In(serviceDate.Location()))
     ```

   - `internal/restapi/trips_for_location_handler.go:821`:
     ```go
     serviceDate := time.Date(time.Now().Year(), time.Now().Month(), time.Now().Day(), 0, 0, 0, 0, currentLocation)
     ```
     Should use `api.Clock.Now()`. The function already receives a `currentLocation` — it just needs the clock-aware time.

   The rest of the codebase consistently uses `api.Clock.Now()` (e.g., `trips_for_location_handler.go:40`). Using `time.Now()` means these code paths will produce wrong results in tests and make the frequency window lookup non-deterministic relative to the mock clock.

2. **Tests silently pass when the endpoint doesn't return 200** (`frequency_integration_test.go`, multiple locations)

   Many test cases guard their assertions with `if model.Code == 200 { ... }`, which means the test passes without verifying anything if the endpoint returns an error. Examples:

   - `TestTripForVehicleFrequencyIntegration` lines 510, 546
   - `TestTripsForLocationFrequencyIntegration` lines 442, 460
   - `TestSingularArrivalAndDepartureFrequencyIntegration` lines 862, 879

   These should use `require.Equal(t, 200, model.Code)` before the assertions. If the endpoint can legitimately return non-200 (e.g., trip not serving this stop), use `t.Skip()` with a clear reason rather than silently passing.

## Suggestions

3. **Extract the repeated frequency lookup into a helper** (`trips_helper.go`)

   The same ~15-line block appears in four places:
   - `trip_details_handler.go:175-207`
   - `trip_for_vehicle_handler.go:113-143`
   - `trips_helper.go:214-232`
   - `arrival_and_departure_for_stop_handler.go:389-404`

   Something like:
   ```go
   func (api *RestAPI) lookupTripFrequency(ctx context.Context, tripID string, serviceDate time.Time, now time.Time) *models.Frequency {
       frequencies, err := api.GtfsManager.GetFrequenciesForTrip(ctx, tripID)
       if err != nil || len(frequencies) == 0 {
           return nil
       }
       active := gtfsInternal.GetActiveHeadwayForTime(frequencies, serviceDate, now)
       if active == nil {
           active = &frequencies[0]
       }
       freq := models.NewFrequencyFromDB(*active, serviceDate)
       return &freq
   }
   ```

   This would reduce each call site to a single line and make the fallback-to-first-entry behavior consistent everywhere.

aaronbrethorst (changes requested):
Adel, this is a solid final phase -- wiring frequency data through every handler that needs it, with batch queries in the right places and a comprehensive test suite. The `lookupTripFrequency` extraction and the `ScheduleFrequency` type for schedule endpoints are both good design calls. The three fix-up commits addressing the prior review are clean.

There are a few items that need attention before merge.

## Critical Issues (3 found)

1. **N+1 query regression in `buildScheduleFromMemory` and `buildScheduleForTrip`** [code-reviewer, code-simplifier]

   Both functions make per-trip `GetFrequenciesForTrip` DB calls, but they're called inside loops that already have batch-fetched frequency data available nearby:

   - `buildScheduleFromMemory` (`trips_for_location_handler.go:829`) is called per-trip from the `buildTripsForLocationEntries` loop (line 369), which *also* does a batch fetch at lines 401-419. Each trip generates two frequency lookups.
   - `buildScheduleForTrip` (`trips_for_location_handler.go:452-458`) has the same issue, and is also called per-trip from `trips_for_route_handler.go:256`.

   For a response with N trips, this adds N unnecessary DB queries. Pass the already-fetched `freqsByTrip` map into these functions instead of re-querying.

2. **`context.Background()` used instead of request context** [code-reviewer, silent-failure-hunter]

   `trips_for_location_handler.go:829`:
   ```go
   freqs, freqErr := api.GtfsManager.GetFrequenciesForTrip(context.Background(), trip.ID)
   ```

   `buildScheduleFromMemory` doesn't accept a `ctx` parameter. If a client disconnects, this query keeps running. Add `ctx context.Context` as a parameter and pass it through from the caller.

3. **Missing `headwaySecs <= 0` guard risks infinite loop** [silent-failure-hunter]

   `schedule_for_route_handler.go:246-250`:
   ```go
   headwaySec := freq.HeadwaySecs
   startSec := freq.StartTime / int64(time.Second)
   endSec := freq.EndTime / int64(time.Second)
   for depTime := startSec; depTime < endSec; depTime += headwaySec {
   ```

   If a corrupted frequency row has `headway_secs=0`, this is an infinite loop. The same expansion in `schedule_for_stop_handler.go:290-291` correctly guards with `if headwayNanos <= 0 { continue }`. Add the same guard here.

## Important Issues (5 found)

4. **Inconsistent frequency selection in `buildScheduleForTrip`/`buildScheduleFromMemory`** [code-reviewer, code-simplifier]

   The `lookupTripFrequency` helper correctly calls `GetActiveHeadwayForTime` to find the frequency window active at the current time, falling back to `freqs[0]`. But `buildScheduleForTrip` (line 455) and `buildScheduleFromMemory` (line 830) always take `freqs[0]` unconditionally -- they never attempt to find the active window. This means the schedule's frequency field may differ from the trip-level frequency field for the same trip. Once you resolve the N+1 issue above, consider reusing `lookupTripFrequency` or at minimum using `GetActiveHeadwayForTime`.

5. **`lookupTripFrequency` silently discards database errors** [silent-failure-hunter]

   `trips_helper.go:1153-1155`:
   ```go
   frequencies, err := api.GtfsManager.GetFrequenciesForTrip(ctx, tripID)
   if err != nil || len(frequencies) == 0 {
       return nil
   }
   ```

   This is called from 4 handlers. A DB error (SQLite busy, corruption, etc.) looks identical to "no frequency data" with zero logging. Separate the error path and add a `Warn` log for the error case.

6. **Inconsistent DB access patterns** [code-simplifier]

   Two handlers bypass the Manager and call `GtfsDB.Queries.GetFrequenciesForTrips()` directly:
   - `arrivals_and_departure_for_stop.go:262`
   - `schedule_for_route_handler.go:114`

   While three others go through `api.GtfsManager.GetFrequenciesForTrips()`. All five should use the Manager method consistently -- if caching or validation is added to the Manager method later, the direct-access call sites will silently skip it.

7. **Missing `require.Equal(t, 200, model.Code)` in `TestTripDetailsFrequencyIntegration`** [code-reviewer, pr-test-analyzer]

   All three subtests in `TestTripDetailsFrequencyIntegration` (lines 275-335) are missing the 200 status check that the prior review requested. The other test functions have it.

8. **Tests with lenient assertions that silently pass** [pr-test-analyzer, silent-failure-hunter]

   Several tests use `t.Log()` and continue/return instead of failing when the core assertion can't be verified:
   - `TestTripsForRouteFrequencyIntegration` lines 394-396: `foundFreq` is logged but never asserted
   - `TestTripsForLocationFrequencyIntegration` lines 443-464: loop exits without asserting trip was found
   - `TestTripForVehicleFrequencyIntegration` lines 524-527: `if ok && freq != nil` guard means nil frequency passes silently
   - `TestArrivalsAndDeparturesFrequencyIntegration` line 792: `t.Log` instead of `t.Error`

   These tests can pass while the feature is completely broken. If the test data doesn't guarantee the condition will be met, use `t.Skip` (not `t.Log`) so it's visible in test output that the assertion was skipped.

## Suggestions (3 found)

9. **Extract duplicated batch-fetch-group-populate pattern** [code-simplifier]

   The ~20-line block at `trips_for_route_handler.go:292-312` and `trips_for_location_handler.go:400-420` is nearly identical. A shared helper like `populateHeadwayFrequencies(ctx, tripIDs, results)` would eliminate the duplication.

10. **Extract `resolveActiveFrequency` pure helper** [code-simplifier]

    The "find active frequency, fall back to first" logic appears in `lookupTripFrequency` and inline in the arrivals handler. Extracting the core resolution into a pure function would let `lookupTripFrequency` and the batch arrivals path share the logic without sacrificing the N+1 optimization.

11. **`time.Sleep(500ms)` in `createTestApiWithFrequencyRealTime`** [pr-test-analyzer]

    Line 239 waits a fixed duration for real-time data to load. This creates flaky tests on slow CI. Consider polling for the data to be available with a short timeout.

## Strengths

- The `ScheduleFrequency` type correctly separates schedule-endpoint needs (serviceId, tripId context) from the trip-details `Frequency` struct
- Batch frequency fetches in the arrivals, schedule, and route handlers properly avoid N+1 queries for the entry-level frequency field
- The `lookupTripFrequency` helper extraction (from prior review feedback) is clean and reused across 4 handlers
- Comprehensive regression test (`TestExistingEndpointsNoFrequencyRegression`) ensures existing endpoints still work without frequency data
- Both `exact_times` modes are tested across the handlers where the distinction matters

## Recommended Action

1. Fix critical issues 1-3 (N+1 regression, context.Background, infinite loop guard)
2. Address important issues 4-8 (consistency, error logging, test assertions)
3. Consider suggestions 9-11 for cleanup

burma-shave:
Closing as stale — this PR predates #1144 (merged 2026-07-08), which changed `trips-for-route`/`trips-for-location`'s `Frequency` field from `*int64` to `*models.Frequency`; this PR still assumes the old `*int64` shape and won't build against current `main`.

The `ScheduleFrequency` split introduced here for `schedule-for-stop` was the right instinct — #1147 arrives at the same conclusion independently, with a fuller shape (`stopHeadsign`/`arrivalEnabled`/`departureEnabled` included) and a shared `FrequencyWindow` base. Future Phase 3 work should build on #1147's corrected types rather than resurrecting this branch.

PR #713, opened 2026-03-14
test(gtfsdb): tighten latency/perf error handling and expand slow-que…

### LINKED ISSUE #705: Tighten error handling in query latency test suite
Follow-up from PR #648 review. The production `slowQueryDB` wrapper is solid, but there are several error handling gaps in the test/benchmark code that could mask failures or produce misleading profiling results.

### 1. Silent error swallowing in `TestConnectionPoolTuning` goroutines
**`gtfsdb/query_latency_test.go:547-549`**

Query errors inside goroutines are silently discarded. If all queries fail for a given pool size, that row is silently omitted from the results table. Should use an error channel matching the pattern from `TestQueryLatencyUnderConcurrentLoad`.

### 2. `loadPerfFixture` catches all import errors as non-fatal
**`gtfsdb/query_latency_perftest_test.go:51-54`**

The catch block swallows all import errors (corrupted zip, disk full, schema mismatch), not just the expected duplicate-import case. If the import fails and the database is empty, benchmarks run against an empty database with no test failure. Fix by checking `latencyIsEmpty` after a failed import and calling `b.Fatalf` if the database has no data.

### 3. Missing `rows.Err()` check in `TestExplainQueryPlans`
**`gtfsdb/query_latency_test.go:429-452`**

After the `for rows.Next()` loop, `rows.Err()` is never checked. Per the `database/sql` contract, `rows.Next()` can return `false` due to an iteration error.

### 4. `latencyFetchRouteIDsForStop` / `latencyFetchActiveServiceIDs` conflate errors with empty results
**`gtfsdb/query_latency_test.go:126-150`**

Both helpers log a `WARNING` and return `nil` on error, making database errors indistinguishable from "no data." Consider failing hard with `tb.Fatalf` since these are test setup helpers.

### 5. `ExecContext` and `QueryRowContext` have no dedicated slow-query tests
**`gtfsdb/helpers_test.go:133-175`**

`TestSlowQueryDB_LogsSlowQueries` only exercises `QueryContext`. The other two DBTX methods should also be tested to verify `maybeLog` is called with the correct `op` attribute.

### 6. `maybeLog` error-inclusion branch is untested
**`gtfsdb/helpers.go:104`**

No test triggers a slow query that also returns an error. The `if err != nil` branch that appends an error attribute to the log record has zero test coverage.

Adityatorgal17:
follow-up PR for #648 and fixes #705 
 
This PR addresses review feedback around silent failures and missing test coverage in GTFS DB latency/perf and slow-query wrapper tests.

### Changes made:

- `TestConnectionPoolTuning` now captures goroutine query errors via `errCh` and reports them, instead of silently returning.
- `loadPerfFixture` now treats import errors as fatal when DB remains empty (`latencyIsEmpty` + `b.Fatalf`).
- `TestExplainQueryPlans` now checks `rows.Err()` after iteration.
- `latencyFetchRouteIDsForStop` and `latencyFetchActiveServiceIDs` now fail hard (`tb.Fatalf`) on query errors.
- Added dedicated slow-query tests for:
  - `ExecContext` (`op=ExecContext`)
  - `QueryRowContext` (`op=QueryRowContext`)
- Added coverage for `maybeLog` error path by forcing a slow failing query and asserting error attribute is logged.

### Validation:

- Passed: `make test`

### DIFF
--- a/gtfsdb/helpers_test.go
+++ b/gtfsdb/helpers_test.go
@@ -108,14 +108,14 @@ func TestProcessAndStoreGTFSData_ValidationFailurePreservesData(t *testing.T) {
 
 	for name, content := range files {
 		f, err := w.Create(name)
-		assert.NoError(t, err)
+		require.NoError(t, err)
 
 		_, err = f.Write([]byte(content))
-		assert.NoError(t, err)
+		require.NoError(t, err)
 	}
 
 	err = w.Close()
-	assert.NoError(t, err)
+	require.NoError(t, err)
 
 	invalidBytes := buf.Bytes()
 
@@ -141,21 +141,7 @@ func TestSlowQueryDB_LogsSlowQueries(t *testing.T) {
 	require.NoError(t, err)
 
 	// Capture slog output via a custom handler.
-	type logRecord struct {
-		msg   string
-		level slog.Level
-		attrs map[string]any
-	}
-	var captured []logRecord
-	handler := &captureHandler{fn: func(r slog.Record) {
-		attrs := make(map[string]any)
-		r.Attrs(func(a slog.Attr) bool {
-			attrs[a.Key] = a.Value.Any()
-			return true
-		})
-		captured = append(captured, logRecord{msg: r.Message, level: r.Level, attrs: attrs})
-	}}
-	logger := slog.New(handler)
+	logger, captured := newCaptureLogger(t)
 
 	ctx := context.Background()
 
@@ -165,7 +151,7 @@ func TestSlowQueryDB_LogsSlowQueries(t *testing.T) {
 	rows, err := wrapper.QueryContext(ctx, "SELECT 1")
 	require.NoError(t, err)
 	require.NoError(t, rows.Close())
-	assert.Empty(t, captured, "threshold=0 must not emit any log records")
+	assert.Empty(t, *captured, "threshold=0 must not emit any log records")
 
 	// Use a fake clock advancing 10 ms per call to ensure the query exceeds
 	// the threshold and avoid Windows timer resolution issues.
@@ -179,10 +165,13 @@ func TestSlowQueryDB_LogsSlowQueries(t *testing.T) {
 	rows, err = wrapper.QueryContext(ctx, "SELECT 1")
 	require.NoError(t, err)
 	require.NoError(t, rows.Close())
-	require.NotEmpty(t, captured, "threshold=1ns must emit a slow_query record")
-	assert.Equal(t, "slow_query", captured[0].msg)
-	assert.Equal(t, slog.LevelWarn, captured[0].level)
-	assert.Equal(t, "QueryContext", captured[0].attrs["op"])
+	require.NotEmpty(t, *captured, "threshold=1ns must emit a slow_query record")
+	rec := (*captured)[0]
+	assert.Equal(t, "slow_query", rec.msg)
+	assert.Equal(t, slog.LevelWarn, rec.level)
+	assert.Equal(t, "QueryContext", rec.attrs["op"])
+	assert.Contains(t, rec.attrs, "duration")
+	assert.Contains(t, rec.attrs, "query")
 }
 
 func TestSlowQueryDB_LogsSlowExecContext(t *testing.T) {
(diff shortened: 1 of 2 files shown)


aaronbrethorst (changes requested):
Hey Aditya, nice work tightening up the error handling in these test files. The `tb.Logf` to `tb.Fatalf` upgrades and the `errCh` pattern for surfacing goroutine errors are exactly the kind of improvements that prevent tests from passing while exercising none of the real logic. Good stuff.

## Critical Issues (0 found)

No critical issues.

## Important Issues (2 found)

1. **[silent-failure-hunter]**: The index listing block in `TestExplainQueryPlans` has the same `rows.Err()` omission that this PR fixes 20 lines above it. It also silently swallows query errors (`if err == nil { ... }` with no else branch) and scan errors (`if scanErr == nil` just skips the log line). This is the exact class of bug the PR is meant to address. [`gtfsdb/query_latency_test.go:461-476`]

   ```go
   // Current: silently skips on query error, no rows.Err() check
   if err == nil {
       defer func() { _ = idxRows.Close() }()
       for idxRows.Next() {
           ...
           if scanErr := idxRows.Scan(...); scanErr == nil {
               t.Logf(...)
           }
       }
   }
   ```

   Suggestion: add `t.Errorf` for the query error else branch, `t.Errorf` for scan errors, and `idxRows.Err()` after the loop -- matching the pattern used in the EXPLAIN block above.

2. **[silent-failure-hunter]**: `latencyIsWALEnabled` discards the PRAGMA query error at line 153, so a failed check silently reports `WAL=false`. This could mislead someone tuning connection pool settings. [`gtfsdb/query_latency_test.go:153`]

   ```go
   // Current
   _ = db.QueryRowContext(ctx, "PRAGMA journal_mode").Scan(&mode)

   // Suggested
   if err := db.QueryRowContext(ctx, "PRAGMA journal_mode").Scan(&mode); err != nil {
       t.Errorf("failed to check WAL mode: %v", err)
       return false
   }
   ```

## Suggestions (3 found)

1. **[pr-test-analyzer]**: The `logRecord` struct + `captureHandler` setup (~20 lines) is duplicated identically across four test functions. Consider extracting a shared helper like `newCaptureLogger(t) (*slog.Logger, *[]logRecord)` to reduce boilerplate. This is a style preference -- Go tests commonly repeat setup for clarity, so feel free to leave as-is if you prefer. [`gtfsdb/helpers_test.go:144-157, 196-209, 233-246, 272-285`]

2. **[pr-test-analyzer]**: The new tests assert `op` and `msg` attributes but don't verify that `"duration"` and `"query"` attributes are present. These are the most operationally useful fields for debugging slow queries. A quick `assert.Contains(t, captured[0].attrs, "duration")` would lock in the contract. [`gtfsdb/helpers_test.go`]

3. **[silent-failure-hunter]**: `rows.Columns()` error is discarded at line 426 in `TestExplainQueryPlans`. If it fails, `cols` is nil and the subsequent scan will silently produce wrong results. [`gtfsdb/query_latency_test.go:426`]

## Strengths

- The `tb.Logf` to `tb.Fatalf` upgrade in fixture helpers is the right call -- these are setup functions where failure means the test has no valid data.
- The `errCh` pattern in `TestConnectionPoolTuning` correctly surfaces goroutine errors that were previously swallowed silently.
- The `rows.Err()` check in `TestExplainQueryPlans` follows proper Go idiom for detecting mid-iteration database errors.
- The error-attribute test (`TestSlowQueryDB_LogsSlowErrorsWithErrorAttribute`) properly validates both the presence and content of the error attribute with a type assertion before checking the value.
- The fake clock pattern (`wrapper.now`) avoids flaky tests from OS timer resolution, and all new tests correctly use it.

## Recommended Action

1. Fix the index listing block in `TestExplainQueryPlans` (important issue #1) -- it's inconsistent with the fix you made in the same function.
2. Consider fixing the WAL check (important issue #2) since it's a one-line change.
3. Consider the suggestions at your discretion.

aaronbrethorst (approved):
Aditya, thorough follow-up — you addressed every item from the last review, including the optional suggestions. The `newCaptureLogger` extraction in particular cleans things up nicely.

## Previous Feedback Status

- [x] Index listing block in `TestExplainQueryPlans` now reports query/scan errors and checks `idxRows.Err()` (important #1)
- [x] `latencyIsWALEnabled` checks PRAGMA query error instead of discarding it (important #2)
- [x] `newCaptureLogger` helper extracted to deduplicate `logRecord`/`captureHandler` boilerplate (suggestion #1)
- [x] New tests assert `duration` and `query` attributes alongside `op` and `msg` (suggestion #2)
- [x] `rows.Columns()` error checked in `TestExplainQueryPlans` (suggestion #3)

## Checklist

- [x] Tests pass
- [x] All prior feedback addressed
- [x] Error handling consistent throughout: `tb.Fatalf` for setup failures, `t.Errorf` for non-fatal assertions
- [x] `errCh` pattern correctly surfaces goroutine errors in `TestConnectionPoolTuning`

## Verdict

**Merge** — clean and complete.

PR #720, opened 2026-03-16
perf: resolve N+1 query patterns in schedule and route handlers

### LINKED ISSUE #719: Refactor Schedule and Route Handlers to Resolve N+1 Query Patterns
## Description
A performance bottleneck has been identified in several REST API handlers due to the presence of N+1 database query patterns. Specifically, related data (routes, trips, blocks, shapes) is being fetched within loops rather than utilizing batch queries. 

The primary affected file is `internal/restapi/schedule_for_stop_handler.go` (lines 140-150), along with helpers utilized by `arrivals_and_departure_for_stop.go`.

## Impact
* **Latency Overhead:** Request times are currently 100-200ms. The target latency is 5-10ms.
* **Database Load:** Represents an approximate 100x increase in query volume for the affected endpoints.

### 1. `gtfsdb` Layer Updates
Add necessary batch definitions to support bulk-fetching:
* Update `gtfsdb/query.sql` with new batch query methods:
    * `GetBlockIDsByTripIDs`: Maps trip IDs to their respective block IDs.
    * `GetShapePointsByTripIDs`: Retrieves shape points for multiple trips in a single query.
* Execute `make models`.

### 2. Handler Refactoring (`internal/restapi/`)

**Arrivals and Departures (`arrivals_and_departure_for_stop.go`)**
* **Current State:** Loops over `allActiveStopTimes`, calling `getBlockDistanceToStop`, `GetTrip`, and `GetRoute` conditionally per item.
* **Required Changes:** * Implement a two-pass approach.
    * Pass 1 (Pre-fetch): Extract all `TripIDs` and `ActiveTripIDs`. Batch fetch trips, routes, block IDs, stop times, and shape points concurrently into a local cache/map.
    * Pass 2 (Build): Construct the models using the pre-fetched maps, removing all `api.GtfsDB.Queries` calls from the iteration.

**Block and Shape Distances (`block_distance_helper.go` & `shape_distance_helpers.go`)**
* Refactor `getBlockDistanceToStop` to accept a caching context or introduce a `BlockDistanceCalculator` struct to preload required data (`GetBlockIDByTripID`, `GetTripsByBlockID`, etc.).
* Update `getStopDistanceAlongShape` and `getVehicleDistanceAlongShapeContextual` to accept pre-fetched maps of shape points and stop times.

**Stops For Route (`stops_for_route_handler.go`)**
* Refactor the directional loop to batch `GetOrderedStopIDsForTrip` and `GetShapesGroupedByTripHeadSign` to ensure strict adherence to batching standards, even for smaller collections.

AhmedAlian7:
#### Description
- closes #719 

This pull request addresses severe N+1 database query patterns identified REST API handlers, specifically targeting `arrivals_and_departure_for_stop.go` and its associated distance helpers.

Previously, handlers fetched related data iteratively within large loops, resulting in a significant increase in database load per request. By implementing deferred pre-fetch batching and in-memory caching contexts, this PR removes these in-loop database lookups, drastically reducing latency and database contention.

#### Changes

**1. Arrivals and Departures (`arrivals_and_departure_for_stop.go`)**
* **Active Trip & Route Fetching:** Removed loop-based `GtfsDB.Queries.GetTrip` and `GtfsDB.Queries.GetRoute` calls.
* **Deferred Batching:** Introduced a deferred pre-fetch mechanism using `GetTripsByIDs` and `GetRoutesByIDs`. Active trips discovered during loop iterations are now deferred, batched, and resolved simultaneously at the end of the execution block to populate reference sets.
* **Block Context Caching:** Consolidated the block distances calculation. Injected pre-fetched block contexts, allowing distance helpers to compute metrics entirely in memory rather than triggering multiple sequential database lookups per stop iteration.

**2. Distance Helpers (`block_distance_helper.go` / `shape_distance_helpers.go`)**
* **In-Memory Lookups:** Modified distance helper signatures to accept in-memory representations of `StopTimes` and `ShapePoints`.
* **Query Reduction:** These are now pre-fetched by the handlers grouping the trips, preventing multiple sequential trip and stop lookups on `GtfsManager.GtfsDB.Queries`.

**3. Stops For Route (`stops_for_route_handler.go`)**
* **Directional Loop Analysis:** Analyzed the directional fetching loops and confirmed iterations are strictly bounded by the number of unique headsigns/directions on a route (typically N <= 2).
* **Validation:** Verified that standard array grouping and current data access do not trigger an N+1 scaling issue in this context.
* **Test Patching:** Patched related tests to ensure valid `GtfsManager` initialization flows.

### DIFF
--- a/gtfsdb/db.go
+++ b/gtfsdb/db.go
@@ -141,6 +141,9 @@ func Prepare(ctx context.Context, db DBTX) (*Queries, error) {
 	if q.getBlockIDByTripIDStmt, err = db.PrepareContext(ctx, getBlockIDByTripID); err != nil {
 		return nil, fmt.Errorf("error preparing query GetBlockIDByTripID: %w", err)
 	}
+	if q.getBlockIDsByTripIDsStmt, err = db.PrepareContext(ctx, getBlockIDsByTripIDs); err != nil {
+		return nil, fmt.Errorf("error preparing query GetBlockIDsByTripIDs: %w", err)
+	}
 	if q.getBlockTripIndexIDsForBlocksStmt, err = db.PrepareContext(ctx, getBlockTripIndexIDsForBlocks); err != nil {
 		return nil, fmt.Errorf("error preparing query GetBlockTripIndexIDsForBlocks: %w", err)
 	}
@@ -233,6 +236,9 @@ func Prepare(ctx context.Context, db DBTX) (*Queries, error) {
 	}
 	if q.getShapePointsByTripIDStmt, err = db.PrepareContext(ctx, getShapePointsByTripID); err != nil {
 		return nil, fmt.Errorf("error preparing query GetShapePointsByTripID: %w", err)
+	}
+	if q.getShapePointsByTripIDsStmt, err = db.PrepareContext(ctx, getShapePointsByTripIDs); err != nil {
+		return nil, fmt.Errorf("error preparing query GetShapePointsByTripIDs: %w", err)
 	}
 	if q.getShapePointsForTripStmt, err = db.PrepareContext(ctx, getShapePointsForTrip); err != nil {
 		return nil, fmt.Errorf("error preparing query GetShapePointsForTrip: %w", err)
@@ -539,6 +545,11 @@ func (q *Queries) Close() error {
 			err = fmt.Errorf("error closing getBlockIDByTripIDStmt: %w", cerr)
 		}
 	}
+	if q.getBlockIDsByTripIDsStmt != nil {
+		if cerr := q.getBlockIDsByTripIDsStmt.Close(); cerr != nil {
+			err = fmt.Errorf("error closing getBlockIDsByTripIDsStmt: %w", cerr)
+		}
+	}
 	if q.getBlockTripIndexIDsForBlocksStmt != nil {
 		if cerr := q.getBlockTripIndexIDsForBlocksStmt.Close(); cerr != nil {
 			err = fmt.Errorf("error closing getBlockTripIndexIDsForBlocksStmt: %w", cerr)
@@ -692,6 +703,11 @@ func (q *Queries) Close() error {
 	if q.getShapePointsByTripIDStmt != nil {
 		if cerr := q.getShapePointsByTripIDStmt.Close(); cerr != nil {
 			err = fmt.Errorf("error closing getShapePointsByTripIDStmt: %w", cerr)
+		}
+	}
+	if q.getShapePointsByTripIDsStmt != nil {
+		if cerr := q.getShapePointsByTripIDsStmt.Close(); cerr != nil {
+			err = fmt.Errorf("error closing getShapePointsByTripIDsStmt: %w", cerr)
 		}
 	}
 	if q.getShapePointsForTripStmt != nil {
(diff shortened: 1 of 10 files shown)


aaronbrethorst (changes requested):
Ahmed, solid approach to the N+1 problem — pre-fetching stop times and shape points into the `DistanceHelperCache` eliminates the most expensive in-loop queries, and the deferred batch-fetch pattern for active trip/route references is clean. A couple things to address:

## Critical

_(none)_

## Important

1. **Stale `err` variable in the cache-hit path** (`block_distance_helper.go:48-54`). When the cache hit is taken on line 49-50, the `err` on line 54 retains its value from `IsServiceActiveOnDate` on line 43. Today this is safe because line 44 `continue`s on error, but it's correct by coincidence of control flow — any future refactor could break it silently. Use a scoped variable:

   ```go
   var stopTimes []gtfsdb.StopTime
   if cache != nil && cache.stopTimesByTrip[blockTrip.ID] != nil {
       stopTimes = cache.stopTimesByTrip[blockTrip.ID]
   } else {
       var fetchErr error
       stopTimes, fetchErr = api.GtfsManager.GtfsDB.Queries.GetStopTimesForTrip(ctx, blockTrip.ID)
       if fetchErr != nil {
           continue
       }
   }
   if len(stopTimes) == 0 {
       continue
   }
   ```

2. **`getBlockDistanceToStop` still has per-arrival N+1 for block ID and block trip lookups** (`block_distance_helper.go:19,30`). The cache covers stop times and shape points, but every arrival with a positioned vehicle still triggers individual `GetBlockIDByTripID` and `GetTripsByBlockID` calls. Since the handler already batch-fetches block IDs via `GetBlockIDsByTripIDs` (line 280) and block trips via `GetTripsByBlockIDs` (line 307), consider adding these to the `DistanceHelperCache` so `getBlockDistanceToStop` can use them too. This doesn't have to block this PR — filing a follow-up issue is fine — but it's worth noting that the N+1 pattern is only partially resolved.

## Fit and Finish

1. **Missing `len > 1` guard on cache-hit path for shape points** (`block_distance_helper.go:67-69`). The non-cache path guards with `len(shapeRows) > 1` before converting, but the cache-hit path converts unconditionally. The downstream `if len(shapePoints) > 1` catches this, but aligning the guard would be consistent:

   ```go
   if cache != nil && cache.shapePointsByTrip[blockTrip.ID] != nil {
       cachedShapeRows := cache.shapePointsByTrip[blockTrip.ID]
       if len(cachedShapeRows) > 1 {
           shapePoints = shapePointsRowsToPoints(cachedShapeRows)
       }
   }
   ```

2. **`DistanceHelperCache` field alignment** (`shape_distance_helpers.go:29-32`). Minor gofmt style: the struct fields have inconsistent indentation. Run `go fmt` to align.

## Strengths

- The deferred batch-fetch pattern for active trips and routes (collect IDs during the loop, fetch all at once after) is the right design — it avoids both N+1 and speculative over-fetching.
- New SQL queries (`GetBlockIDsByTripIDs`, `GetShapePointsByTripIDs`) are well-structured with proper ordering.
- The `DistanceHelperCache` is nicely nullable — passing `nil` preserves existing behavior for callers that don't pre-fetch.
- Test fixes for missing `defer api.Shutdown()` in `stops_for_route_handler_test.go` are a good drive-by cleanup.

## Verdict

**Request changes** — please fix the stale `err` variable pattern (important #1). The remaining N+1 for block lookups (important #2) can be a follow-up.

aaronbrethorst:
Hi @AhmedAlian7 - I am terribly sorry I completely missed that you had pushed up changes to this branch after my code review! Can I ask you to rebase this branch to fix all of the merge conflicts? Please reopen the PR after that. Again, my sincerest apologies.

PR #725, opened 2026-03-16
fix(openapi): align arrivals-and-departures-for-stop time + arrival m…

### LINKED ISSUE #724: OpenAPI contract mismatch for `arrivals-and-departures-for-stop` (`time` param) + schema drift in `ArrivalAndDeparture`
### Summary
The `/api/where/arrivals-and-departures-for-stop/{stopID}.json` endpoint was not behaving according to the OpenAPI spec. The spec defines the `time` query parameter as an RFC3339 `date-time` string, but the handler was parsing `time` as a Unix timestamp in milliseconds. In addition, the JSON shape of `ArrivalAndDeparture` responses drifted from the OpenAPI schema, and our OpenAPI conformance suite did not include coverage for this endpoint, allowing the mismatch to persist.

### What the OpenAPI spec says
- Endpoint: `/api/where/arrivals-and-departures-for-stop/{stopID}.json`
- `time` parameter:
  - `in: query`
  - `schema.type: string`
  - `schema.format: date-time`
  - i.e. clients should send RFC3339 timestamps (example: `2026-01-01T00:00:00Z`)
- Source: `testdata/openapi.yml:444`

### What the handler was doing
- `internal/restapi/arrivals_and_departure_for_stop.go:73` parsed `time` as **Unix milliseconds** (integer-like string), not RFC3339.
- As a result:
  - Spec-compliant clients sending RFC3339 `date-time` could get `400 Bad Request`.
  - The server behavior was inconsistent with the published contract and generated client expectations.

### Additional drift discovered during investigation
- The `ArrivalAndDeparture` response model (`internal/models/arrival_and_departure.go`) deviated from the OpenAPI schema (`ArrivalDepartureForStop` in `testdata/openapi.yml`):
  - Some fields defined as `string` in OpenAPI were being represented/emitted in ways that could produce `null` or non-string values.
  - Fields intended to be omitted when empty were not consistently aligned with `omitempty` behavior.
- This increased the risk of:
  - OpenAPI schema validation failures.
  - Client deserialization issues for codegen consumers expecting `string` types.

### Missing safety net (tests)
- `internal/restapi/openapi_conformance_test.go` did not include an OpenAPI conformance test for `"arrivals-and-departures-for-stop"`.
- Because the endpoint wasn’t exercised by the conformance suite, the contract mismatch (`time` parsing) and response schema drift weren’t caught automatically.

### Impact
- Incorrect API behavior relative to OpenAPI contract.
- Potential interoperability issues with OpenAPI-generated clients.
- Lack of conformance coverage meant regressions could slip through CI unnoticed.

### Resolution (high level)
- Make the handler parse `time` as RFC3339 `date-time` to match the OpenAPI spec.
- Align `ArrivalAndDeparture` JSON output with the OpenAPI schema (string fields + omit empty optional fields).
- Add missing OpenAPI conformance coverage for the endpoint to prevent future drift.

Adityatorgal17:
## Align `arrivals-and-departures-for-stop` with OpenAPI spec + fix `ArrivalAndDeparture` schema drift

Fixes #724 

### Problem
- Handler parsed `time` as Unix milliseconds instead of RFC3339 `date-time` as the OpenAPI spec requires, breaking spec-compliant requests.
- `ArrivalAndDeparture` model had drifted from the OpenAPI `ArrivalDepartureForStop` schema (`interface{}`/null shapes, missing `omitempty` on optional fields).
- No conformance test coverage existed for this endpoint.

### Changes

**Handler** (`arrivals_and_departure_for_stop.go`)
- Parse `time` as RFC3339 only; return 400 on invalid input.

**Handler Tests** (`arrivals_and_departures_for_stop_handler_test.go`)
- Pass `time` as RFC3339 and URL-escape values so `+05:30` offsets aren't decoded as spaces.

**Model** (`arrival_and_departure.go`)
- Replace `interface{}`/null fields with `string`; add `omitempty` on optional fields to match spec.

**Model Tests** (`arrival_and_departure_test.go`)
- Assert optional fields are absent in marshaled JSON when empty.

**Conformance** (`openapi_conformance_test.go`)
- Add missing conformance test for this endpoint against `testdata/openapi.yml`.

### Notes
- `testdata/openapi.yml` is unchanged — all fixes are implementation-side only.

### Validation Passed
- `make test`

### DIFF
--- a/internal/models/arrival_and_departure.go
+++ b/internal/models/arrival_and_departure.go
@@ -6,23 +6,23 @@ type ArrivalAndDeparture struct {
 	BlockTripSequence          int         `json:"blockTripSequence"`
 	DepartureEnabled           bool        `json:"departureEnabled"`
 	DistanceFromStop           float64     `json:"distanceFromStop"`
-	Frequency                  *Frequency  `json:"frequency"`
+	Frequency                  string      `json:"frequency,omitempty"`
 	HistoricalOccupancy        string      `json:"historicalOccupancy"`
 	LastUpdateTime             *int64      `json:"lastUpdateTime,omitempty"`
 	NumberOfStopsAway          int         `json:"numberOfStopsAway"`
 	OccupancyStatus            string      `json:"occupancyStatus"`
 	Predicted                  bool        `json:"predicted"`
-	PredictedArrivalInterval   interface{} `json:"predictedArrivalInterval"`
+	PredictedArrivalInterval   string      `json:"predictedArrivalInterval,omitempty"`
 	PredictedArrivalTime       int64       `json:"predictedArrivalTime"`
-	PredictedDepartureInterval interface{} `json:"predictedDepartureInterval"`
+	PredictedDepartureInterval string      `json:"predictedDepartureInterval,omitempty"`
 	PredictedDepartureTime     int64       `json:"predictedDepartureTime"`
 	PredictedOccupancy         string      `json:"predictedOccupancy"`
 	RouteID                    string      `json:"routeId"`
 	RouteLongName              string      `json:"routeLongName"`
 	RouteShortName             string      `json:"routeShortName"`
-	ScheduledArrivalInterval   interface{} `json:"scheduledArrivalInterval"`
+	ScheduledArrivalInterval   string      `json:"scheduledArrivalInterval,omitempty"`
 	ScheduledArrivalTime       int64       `json:"scheduledArrivalTime"`
-	ScheduledDepartureInterval interface{} `json:"scheduledDepartureInterval"`
+	ScheduledDepartureInterval string      `json:"scheduledDepartureInterval,omitempty"`
 	ScheduledDepartureTime     int64       `json:"scheduledDepartureTime"`
 	ScheduledTrack             string      `json:"scheduledTrack"`
 	ServiceDate                int64       `json:"serviceDate"`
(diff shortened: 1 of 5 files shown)


aaronbrethorst (changes requested):
Aditya, thanks for digging into the spec and finding the `time` parameter format mismatch and the `interface{}` fields — both are legitimate spec-compliance issues worth fixing. However, there's a conflict with an in-flight PR that needs to be resolved first.

## Critical Issues (1 found)

1. **`Frequency` field change from `*Frequency` to `string` conflicts with open PR #680**

   This PR changes `ArrivalAndDeparture.Frequency` from `*Frequency` (struct pointer) to `string` with `omitempty`. While the OpenAPI spec does say `frequency: type: string`, PR #680 (GTFS Frequencies Phase 3) is actively building on the `*Frequency` type — it populates this field with real frequency data:

   ```go
   // PR #680, arrivals_and_departure_for_stop.go:
   arrival.Frequency = api.lookupTripFrequency(ctx, tripID, serviceMidnight, currentTime)
   // returns *models.Frequency
   ```

   ```go
   // PR #680, arrival_and_departure_for_stop_handler.go:
   arrival.Frequency = &freq  // where freq is models.Frequency
   ```

   If this PR merges first, PR #680 will fail to compile. If #680 merges first, this PR will have merge conflicts. These two PRs need to be coordinated.

   The OpenAPI spec says `type: string`, but this appears to be an under-specification — the production OBA Java server returns frequency as a structured object (`FrequencyBean`) when present, not a string. The spec likely inherited a simplified representation. The real behavior is closer to what #680 implements.

   **Resolution options:**
   - Wait for #680 to merge, then update this PR to skip the `Frequency` field change (or update the spec to use an object schema)
   - Coordinate with the #680 author to agree on the final type

## Important Issues (1 found)

2. **The singular `arrival-and-departure-for-stop` handler still parses `time` as Unix ms**

   This PR fixes the plural handler (`arrivals_and_departure_for_stop.go`) to parse `time` as RFC3339, but the singular handler (`arrival_and_departure_for_stop_handler.go:58`) still uses `strconv.ParseInt`:

   ```go
   // arrival_and_departure_for_stop_handler.go:58 — still Unix ms
   if timeMs, err := strconv.ParseInt(timeStr, 10, 64); err == nil {
   ```

   Interestingly, the OpenAPI spec itself has different types for these two endpoints:
   - Plural endpoint: `time: type: string, format: date-time` (RFC3339)
   - Singular endpoint: `time: type: integer` (Unix ms)

   So the singular handler is actually correct per its own spec entry. However, this inconsistency between the two endpoints is worth noting in the PR description so future readers understand it's intentional, not an oversight.

## Suggestions (1 found)

3. **The `omitempty` changes on interval fields are correct but deserve a note**

   Changing `PredictedArrivalInterval`, `PredictedDepartureInterval`, `ScheduledArrivalInterval`, and `ScheduledDepartureInterval` from `interface{}` (serialized as `null`) to `string` with `omitempty` (omitted entirely) changes the JSON wire format. Previously clients received `"predictedArrivalInterval": null`; now they won't see the field at all when empty. This is spec-compliant (the fields aren't marked `required`), but any clients relying on the key being present may break. Worth noting in the PR description.

## Strengths

- The `time` parameter fix to RFC3339 is correct per the OpenAPI spec for this endpoint
- Good use of `url.QueryEscape` in tests for RFC3339 values containing `+` characters
- The new conformance test for `arrivals-and-departures-for-stop` is a valuable addition
- The `omitempty` test (`TestArrivalAndDepartureJSON`) properly verifies the fields are absent from marshaled JSON

## Recommended Action

1. Resolve the `Frequency` type conflict with PR #680 before merging
2. Add a note to the PR description about the singular vs plural `time` format difference
3. After resolving #680 coordination, this should be good to merge

aaronbrethorst:
Closing for now. please address code review feedback and fix merge conflicts, and then reopen. thank you!

PR #729, opened 2026-03-17
perf(gtfs): build alertIndex for O(1) alert lookups

### LINKED ISSUE #728: perf: build alertIndex for O(1) alert lookups, replacing O(n×m) linear scans
## Problem

Every alert lookup today performs a nested linear scan:

- `GetAlertsByIDs` — O(n) alerts × O(m) entities per alert
- `GetAlertsForStop` — same O(n) scan

These are called inside the per-arrival loop in the hottest handler (`arrivals-and-departures-for-stop`):
- 50+ calls to `GetAlertsForTrip` per request
- Additional calls from `GetAlertsForStop` and `GetAlertsForRoute` in `reference_utils.go`

With 50 arrivals, 200 alerts, and 3 entities per alert: **30,000 comparisons per request**.

Trips and vehicles already have O(1) index maps (`realTimeTripLookup`, `realTimeVehicleLookupByTrip`) built in `rebuildMergedRealtimeLocked`. Alerts have no equivalent.

## Solution

Build an `alertIndex` during `rebuildMergedRealtimeLocked()` under the existing write lock. Alert lookups become O(1) map lookups rebuilt once per feed refresh — zero cost at read time.

## Expected Benefit

30,000 comparisons per request → 1–3 map lookups per call.

reddy-santhu:
Closes #728

## Summary

Replaces O(n×m) linear alert scans with a pre-built `alertIndex` of four hash maps, giving O(1) lookup at read time. The index is rebuilt once per feed refresh under the existing write lock — the same pattern already used for `realTimeTripLookup` and `realTimeVehicleLookupByTrip`.

## Changes

- **`gtfs_manager.go`** — adds `alertIdx alertIndex` field alongside the existing lookup maps
- **`realtime.go`**
  - New `alertIndex` struct with `byTrip`, `byRoute`, `byAgency`, `byStop` maps
  - Index built in `rebuildMergedRealtimeLocked()` via direct two-level feed iteration (no intermediate slice allocation)
  - `GetAlertsByIDs` now does O(1) map lookups with cross-category dedup via `addUnique`
  - `GetAlertsForStop` now does an O(1) map lookup with internal dedup by alert ID, consistent with `GetAlertsByIDs`
  - Per-alert `seen*` maps pre-sized to avoid internal growth
  - Route-entity scoping aligned with Java OBA: `{routeID+tripID}` entities go into `byTrip` only, not `byRoute`
- **`realtime_test.go`** — 4 existing tests updated to use `feedAlerts` + `rebuildMergedRealtimeLocked`; 13 new tests covering each bucket, scoping rules, dedup, and edge cases
- **`realtime_benchmark_test.go`** — benchmarks for `RebuildAlertIndex`, `GetAlertsByIDs`, and `GetAlertsForStop`

## Scoping rules

| Entity shape | Indexed in |
|---|---|
| `{tripID}` | `byTrip` |
| `{routeID}` (no stop, no trip) | `byRoute` |
| `{routeID + tripID}` | `byTrip` only |
| `{routeID + stopID}` | `byStop` only |
| `{agencyID}` (no route, no trip) | `byAgency` |
| `{agencyID + stopID}` | `byAgency` + `byStop` |
| `{stopID}` | `byStop` |

## Benchmarks

Apple M4, 1000 alerts across 100 routes and 50 agencies, `-count=5 -benchmem`:

| Benchmark | Before | After | Speedup |
|---|---|---|---|
| `GetAlertsByIDs` | 13.40 µs/op | 2.43 µs/op | **5.5x** |
| `GetAlertsForStop` | 4.61 µs/op | 102.5 ns/op | **45x** |

`GetAlertsForStop` sees the biggest gain — the old code scanned every alert on every call, so the improvement scales linearly with alert count. `GetAlertsByIDs` picks up 5.5x with a realistic spread of alerts per route/agency. Both are called per arrival in `arrivals-and-departures-for-stop`, so with 50 arrivals per request the savings add up.

### DIFF
--- a/internal/gtfs/gtfs_manager.go
+++ b/internal/gtfs/gtfs_manager.go
@@ -49,10 +49,10 @@ type Manager struct {
 	realTimeTrips                  []gtfs.Trip
 	realTimeVehicles               []gtfs.Vehicle
 	realTimeMutex                  sync.RWMutex
-	realTimeAlerts                 []gtfs.Alert
 	realTimeTripLookup             map[string]int
 	realTimeVehicleLookupByTrip    map[string]int
 	realTimeVehicleLookupByVehicle map[string]int
+	alertIdx                       alertIndex
 	agenciesMap                    map[string]*gtfs.Agency
 	routesMap                      map[string]*gtfs.Route
 	frequencyTripIDs               map[string]struct{}
--- a/internal/gtfs/gtfs_manager_mock.go
+++ b/internal/gtfs/gtfs_manager_mock.go
@@ -135,6 +135,17 @@ func (m *Manager) MockAddTripUpdate(tripID string, delay *time.Duration, stopTim
 	m.realTimeTripLookup[tripID] = len(m.realTimeTrips) - 1
 }
 
+func (m *Manager) MockAddAlert(feedID string, alert gtfs.Alert) {
+	m.realTimeMutex.Lock()
+	defer m.realTimeMutex.Unlock()
+
+	if m.feedAlerts == nil {
+		m.feedAlerts = make(map[string][]gtfs.Alert)
+	}
+	m.feedAlerts[feedID] = append(m.feedAlerts[feedID], alert)
+	m.rebuildMergedRealtimeLocked()
+}
+
 // MockResetRealTimeData clears all mock real-time vehicles and trip updates.
 func (m *Manager) MockResetRealTimeData() {
 	m.realTimeMutex.Lock()
--- a/internal/gtfs/realtime.go
+++ b/internal/gtfs/realtime.go
@@ -17,6 +17,15 @@ import (
 	"maglev.onebusaway.org/internal/logging"
 )
 
+// alertIndex holds pre-built maps for O(1) alert lookups, keyed by trip, route, agency, and stop IDs.
+// It is rebuilt on every call to rebuildMergedRealtimeLocked under the existing write lock.
+type alertIndex struct {
+	byTrip   map[string][]gtfs.Alert
+	byRoute  map[string][]gtfs.Alert
+	byAgency map[string][]gtfs.Alert
+	byStop   map[string][]gtfs.Alert
+}
+
 // staleVehicleTimeout is the duration after which a vehicle is considered stale
 const staleVehicleTimeout = 15 * time.Minute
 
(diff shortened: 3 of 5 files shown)


aaronbrethorst (approved):
Hey Santhosh, this is an impressive performance optimization — replacing O(n×m) linear scans with pre-built hash maps is exactly the right approach, and the 5.5x / 45x speedups on `GetAlertsByIDs` / `GetAlertsForStop` are substantial given these are called per-arrival. The scoping rules are carefully aligned with Java OBA's inverted index behavior, the index builds cleanly inside the existing `rebuildMergedRealtimeLocked` write lock, and the test suite is thorough with 13 new tests covering every bucket, scoping rule, and dedup path.

## Critical Issues (0 found)

None.

## Important Issues (1 found)

### 1. [code-reviewer]: `{routeID+tripID}` scoping is a behavioral change, not just a performance refactor — `realtime.go:670`

The old route-matching condition was `entity.RouteID != nil && entity.StopID == nil` (no TripID check). The new indexing condition adds `entity.TripID == nil`. This means `GetAlertsByIDs("", routeID, "")` previously found alerts with `{routeID+tripID}` entities but now won't — they're filed in `byTrip` only.

This is clearly intentional (tested in `TestAlertIndex_RouteTripEntityIndexedUnderByTrip`, documented in the scoping table), and it aligns with Java OBA's inverted index behavior. But the PR title says "perf" and the description frames this as a performance optimization. A route-level query silently returning fewer alerts than before is a semantic change that operators should be aware of.

**Ask:** Please add a brief note to the PR description (e.g., under a "Behavioral changes" heading) explicitly calling out that `{routeID+tripID}` entities are now excluded from route-level queries, and why this is the correct behavior per Java OBA.

## Suggestions (2 found)

### 2. [code-reviewer]: Unify dedup map types — `realtime.go:127,192`

`GetAlertsByIDs` uses `map[string]bool` while `GetAlertsForStop` uses `map[string]struct{}` for the same dedup purpose. The `map[string]struct{}` idiom is preferred in Go for sets (zero memory per value). Minor, but having two patterns for the same thing in the same file is a small code quality issue.

### 3. [pr-test-analyzer]: Missing positive test for `{routeID+stopID}` in byStop — `realtime_test.go`

`TestGetAlertsByIDs_RouteScoping` verifies that `{routeID+stopID}` does NOT appear in byRoute (negative path), but no test verifies it DOES appear in byStop (positive path). This is the symmetric counterpart to `TestAlertIndex_AgencyStopEntityAlsoIndexedByStop` which covers both sides for `{agencyID+stopID}`. If the stop indexing at line 679-681 regresses, no test catches it.

Both suggestions and the missing test are tracked in #751 — they don't need to block this PR.

## Strengths

- The `alertIndex` struct follows the exact same pattern as the existing `realTimeTripLookup` / `realTimeVehicleLookupByTrip` maps — consistent with the codebase architecture
- Three levels of dedup are tested: intra-alert (repeated entity keys), cross-bucket (same alert in trip+route), and cross-feed (same alert ID from two feeds)
- The `alert.ID == ""` filter at index-build time is correct — without it, dedup by ID would collapse all empty-ID alerts into one entry
- `MockAddAlert` properly calls `rebuildMergedRealtimeLocked`, and there's a test verifying the index is immediately synced
- Edge cases are well-covered: nil entities, empty entity slices, all-empty-args, and non-existent IDs
- Benchmarks provide regression guards for the O(1) claim

## Recommended Action

**Verdict: Merge**

The one important item (documenting the behavioral change in the PR description) is a documentation ask, not a code change. The suggestions and test gap are tracked in #751. The code is correct, well-tested, and follows the existing codebase patterns.

PR #730, opened 2026-03-17
perf(gtfs): resolve lock contention in spatial queries

### LINKED ISSUE #726: Refactor GTFS Manager to Prevent DB I/O Under Read Lock
## Description
The `GetStopsForLocation()` method in `internal/gtfs/gtfs_manager.go` currently holds a read lock (`m.RLock()`) across multiple synchronous database I/O operations. This design pattern creates severe lock contention, blocking other concurrent handlers that require the same read lock and degrading overall system performance during peak traffic.

## Problem
The method acquires an `RLock` to perform a fast in-memory spatial search (`m.rtree.Search(bounds)`), but fails to release the lock before executing subsequent database queries:
* `GetRoutesForStops`
* `GetActiveServiceIDsForDate`
* `GetStopsWithActiveServiceOnDate`

Because these database queries are synchronous and can take 10-50ms each, the lock is held for approximately 50-200ms on busy systems. This unnecessarily blocks 30+ other handler threads waiting for the same lock.

## Impact
* **Severity:** High for location-based queries.
* **Affected Endpoints:** `/stops-for-location`, `/routes-for-location`, `/trips-for-location`
* **Performance Degradation:** P99 latency increases by 2-3x on concurrent location queries.
* **Scalability:** The contention worsens proportionally with larger GTFS datasets (more stops and routes to process).

## Proposed Solution
Refactor the spatial query methods to minimize the scope of the read lock. The `RLock` should only encompass the in-memory R-tree search. All subsequent database I/O operations must be executed outside of the lock's scope.

AhmedAlian7:
## Description
- closes #726 
 
Location-based handlers (`/stops-for-location`, `/routes-for-location`, `/trips-for-location`) were acquiring a broad read lock (`RLock()`) that spanned the entire lifecycle of the request.

This PR mitigates the contention by enforcing narrow-scoped locking. Database operations are now strictly executed outside of the read lock, and handlers no longer wrap the entire request in a locking mechanism.

### `gtfs_manager.go`
* **`GetStopsForLocation()`**: Refactored to manage its own concurrency internally. It now acquires an `RLock()` exclusively for the fast, in-memory R-tree spatial search (`queryStopsInBounds()`) and immediately releases it. Subsequent database queries are executed lock-free. The returned R-tree results are safe to utilize post-unlock, as the spatial index only mutates during static data reloads (which utilize a write lock).

### API Handlers
Removed the encompassing `RLock()` / `defer RUnlock()` pattern from the following handlers, replacing them with precisely scoped locks for specific in-memory accesses:
* **`stopsForLocationHandler`**: Narrowly scoped locks for `GetAgencies()` and `checkIfOutOfBounds()`.
* **`routesForLocationHandler`**: Narrowly scoped locks for `GetAgencies()` and `checkIfOutOfBounds()`.
* **`tripsForLocationHandler`**: Narrowly scoped locks for `GetAgencies()` (within `parseAndValidateRequest`) and `checkIfOutOfBounds()`.

### Test (`arrivals_and_departures_for_stop_handler_test.go`)
* Updated 4 isolated test call sites that were manually wrapping `GetStopsForLocation()` with `RLock()`/`RUnlock()`. This prevents double-locking deadlocks since the method now handles its own internal locking.

### DIFF
--- a/internal/gtfs/gtfs_manager.go
+++ b/internal/gtfs/gtfs_manager.go
@@ -452,7 +452,7 @@ type stopWithDistance struct {
 
 // GetStopsForLocation retrieves stops near a given location using the spatial index.
 // It supports filtering by route types and querying for specific stop codes.
-// IMPORTANT: Caller must hold manager.RLock() before calling this method.
+// This method manages its own locking internally; callers must NOT hold any Manager locks.
 func (manager *Manager) GetStopsForLocation(
 	ctx context.Context,
 	lat, lon, radius, latSpan, lonSpan float64,
@@ -486,7 +486,10 @@ func (manager *Manager) GetStopsForLocation(
 		return []gtfsdb.Stop{}
 	}
 
+	// Acquire the read lock only for the in-memory spatial search.
+	manager.staticMutex.RLock()
 	dbStops := queryStopsInBounds(manager.stopSpatialIndex, bounds)
+	manager.staticMutex.RUnlock()
 
 	for _, dbStop := range dbStops {
 		if ctx.Err() != nil {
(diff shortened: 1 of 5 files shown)


fletcherw:
This lock contention is no longer relevant with change to manager's staticMutex.

PR #737, opened 2026-03-17
fix: Graceful Shutdown Timeout for GTFS Manager

### LINKED ISSUE #736: fix: Add Timeout to Graceful Shutdown in GTFS Manager
## Description
`Manager.Shutdown()` in `internal/gtfs/gtfs_manager.go` calls `m.wg.Wait()` with no timeout, causing the server to hang indefinitely if any goroutine (a stuck real-time feed fetch) never returns.

## Current Behavior
```go
func (m *Manager) Shutdown() error {
    close(m.stopChan)
    m.wg.Wait() // blocks forever if any goroutine is stuck
    return nil
}
```

## Expected Behavior
`Shutdown()` should accept a `context.Context`. If the context deadline is
exceeded before all goroutines finish, it returns a descriptive error and
allows the caller to proceed rather than blocking forever.

## Proposed Fix
```go
func (m *Manager) Shutdown(ctx context.Context) error {
    close(m.stopChan)

    done := make(chan struct{})
    go func() {
        m.wg.Wait()
        close(done)
    }()

    select {
    case <-done:
        return nil
    case <-ctx.Done():
        return fmt.Errorf("shutdown timeout exceeded: %w", ctx.Err())
    }
}
```

The caller constructs the context with an
appropriate deadline before invoking shutdown:
```go
ctx, cancel := context.WithTimeout(context.Background(), 25*time.Second)
defer cancel()
if err := manager.Shutdown(ctx); err != nil {
    log.Warn("shutdown did not complete cleanly", "error", err)
}
```

## Acceptance Criteria
- [ ] `Shutdown` signature updated to `Shutdown(ctx context.Context) error`
- [ ] All call sites updated to pass a context with a timeout
- [ ] Unit test: shutdown completes normally when goroutines finish in time
- [ ] Unit test: shutdown returns an error when context deadline is exceeded
- [ ] `make lint` and `make test` both pass

AhmedAlian7:
## Description
- Closes #736

`Manager.Shutdown()` called `wg.Wait()` with no timeout. A stuck real-time feed goroutine would block shutdown indefinitely.

### Solution
Updated `Shutdown` to accept a `context.Context` and race `wg.Wait()` against `ctx.Done()`. The production call site in `app.go` uses a **25s timeout**, keeping a 5s safety margin inside Kubernetes' default SIGTERM window. Timeouts
log a warning instead of fatally erroring, preserving liveness.

### Changes
- `gtfs_manager.go` new signature `Shutdown(ctx context.Context) error` with context-select pattern
- `app.go` 25s shutdown context with structured warning log on timeout
- 15+ call sites across 7 test files updated; zero uses of old 0-arg signature remain
- 2 new unit tests: clean path (`nil`) and timeout path (`context.DeadlineExceeded`)

### Tests
`go build ./...`  &nbsp; `make test` &nbsp; `make lint` &nbsp; `go fmt ./...`

### DIFF
--- a/cmd/api/app.go
+++ b/cmd/api/app.go
@@ -256,7 +256,11 @@ func Run(ctx context.Context, srv *http.Server, coreApp *app.Application, api *r
 
 	// Then shutdown GTFS manager (stops data fetching - the lowest-level dependency)
 	if coreApp.GtfsManager != nil {
-		coreApp.GtfsManager.Shutdown()
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), 25*time.Second)
+		defer cancel()
+		if err := coreApp.GtfsManager.Shutdown(shutdownCtx); err != nil {
+			logger.Warn("gtfs manager shutdown incomplete", "error", err)
+		}
 	}
 
 	logger.Info("server exited")
--- a/internal/gtfs/advanced_direction_calculator_test.go
+++ b/internal/gtfs/advanced_direction_calculator_test.go
@@ -570,7 +570,7 @@ func TestMain(m *testing.M) {
 	// Global Teardown
 	// If sharedManager was initialized during tests, shut it down now.
 	if sharedManager != nil {
-		sharedManager.Shutdown(context.Background())
+		_ = sharedManager.Shutdown(context.Background())
 	}
 
 	// Exit with the test result code
--- a/internal/gtfs/advanced_direction_calculator_test.go
+++ b/internal/gtfs/advanced_direction_calculator_test.go
@@ -570,7 +570,7 @@ func TestMain(m *testing.M) {
 	// Global Teardown
 	// If sharedManager was initialized during tests, shut it down now.
 	if sharedManager != nil {
-		sharedManager.Shutdown()
+		sharedManager.Shutdown(context.Background())
 	}
 
 	// Exit with the test result code
--- a/internal/gtfs/gtfs_manager.go
+++ b/internal/gtfs/gtfs_manager.go
@@ -64,6 +64,7 @@ type Manager struct {
 	shutdownChan                   chan struct{}
 	wg                             sync.WaitGroup
 	shutdownOnce                   sync.Once
+	dbCloseOnce                    sync.Once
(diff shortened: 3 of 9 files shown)


aaronbrethorst (changes requested):
Ahmed, this is a well-targeted fix for a real problem — an indefinite hang during shutdown if a real-time feed goroutine gets stuck. The context-based timeout with `select` on `wg.Wait()` is the standard Go pattern for this, and the 25s timeout with K8s SIGTERM margin is a thoughtful choice.

## Critical Issues (0 found)

None.

## Important Issues (0 found)

None.

## Suggestions (1 found)

1. **DB close is outside `shutdownOnce`, weakening idempotency** (tracked in #750)

   Moving `GtfsDB.Close()` out of `shutdownOnce.Do()` means a second `Shutdown()` call will spawn an unnecessary goroutine and re-close the already-closed DB. Go's `sql.DB.Close()` handles this gracefully (returns nil), so this isn't a bug — the idempotency test passes correctly. But it would be cleaner to guard the DB close with its own `sync.Once` and extract the duplicated close logic from both `select` branches into a single deferred call. Filed as #750 for follow-up.

## Strengths

- The `select` on `wg.Wait()` channel vs `ctx.Done()` is the correct Go pattern for bounded waits on WaitGroups
- 25s timeout leaves a 5s safety margin inside K8s' default 30s SIGTERM window — good operational thinking
- DB is closed in both the clean and timeout paths, preventing resource leaks either way
- `TestManagerShutdown_TimeoutPath` with a stuck goroutine is a great test — it verifies the exact scenario that motivated this fix
- All 15+ call sites across 7 test files updated consistently; zero old-signature callers remain
- The warning log on timeout (not fatal) is the right severity — shutdown timeout is informational, not something to crash over

## Recommended Action

After fixing the merge conflict, we will merge as-is. Follow up on #750 when convenient.

aaronbrethorst:
I am terribly sorry I completely missed that you had pushed up changes to this branch after my code review! Can I ask you to rebase this branch to fix all of the merge conflicts? Please reopen the PR after that. Again, my sincerest apologies.

PR #753, opened 2026-03-20
fix: trips-for-route logging, test coverage, and performance follow-ups

### LINKED ISSUE #748: trips-for-route: logging and test coverage follow-ups from PR #679
Follow-up items from the review of #679 that do not need to block the merge.

## Logging improvements

1. **DUPLICATED base trip reference fetch error not logged** (`trips_for_route_handler.go:403-409`). When `GetTrip(baseTripID)` fails with a non-`ErrNoRows` error, the error is silently discarded. Add a Warn log for non-`ErrNoRows` errors so operators can see when trip references are degraded.

2. **Stale comment: OccupancyCapacity/OccupancyCount defaults** (`trips_helper.go:46`). Comment says "default to 0" but `NewTripStatus()` initializes both to `-1`. Update or remove.

3. **DUPLICATED trip ID fallback chain lacks logging** (`trips_for_route_handler.go:361-370`). When the suffix-stripping fallback is taken for `ErrNoRows`, there's no log. When the stripped ID also doesn't exist, no warning is emitted. A Debug-level log on suffix-strip and a Warn-level log when neither ID resolves would help troubleshooting.

4. **`GetRouteIDsForStops` error silently discarded** (`trips_for_route_handler.go:588`). If this batch query fails, all stops will have empty `routeIds` in references. Add a Warn log.

5. **`GetShapePointsByTripID` error discarded in `buildScheduleForTrip`** (`trips_for_location_handler.go:406`). Pre-existing issue, but the function was refactored in this PR. A Warn log would surface DB errors that currently cause silent distance degradation.

## Test coverage

6. **No integration test for DUPLICATED trips in handler response**. RABA test data has no DUPLICATED trips, so the end-to-end handler path (lines 340-410) is untested. A test could inject a DUPLICATED vehicle into the Manager mock.

7. **No test for past-midnight / previous-day service lookup**. The time arithmetic for `prevDayNanosSinceMidnight` is untested. A unit test of the time calculation would catch off-by-one errors.

8. **`selectBestTripInBlock` untested with currently-running trip row**. Adding a case where one row spans `now` (min=800, max=1200, now=1000) would document that the function falls to the fallback tier when unexpectedly called with a running trip.

9. **Handler integration test has loose bounds** (`TestTripsForRouteHandler_DifferentRoutes`). The assertion `minExpected: 0, maxExpected: 50` passes even if zero trips are returned. Consider tightening for routes known to have data in RABA.

## Performance

10. **`GetTripsInBlock` query is redundant** (`trips_for_route_handler.go:208`). The result is only used as an emptiness check. `GetTripsInBlockWithTimeBounds` (used on the fallback path) returns the same data. Consider using it directly and checking for empty results instead of making two separate queries per block.

Originated from PR #679.

## Spec compliance

11. **`LastKnownLocation` changed to `omitempty` may diverge from OpenAPI spec** (`trip_details.go:43`). The PR's stated goal was to remove `omitempty` from TripStatus fields to match Java OBA, but `LastKnownLocation` was changed FROM no-omitempty TO `omitempty`. When there's no location data, the field will be absent from JSON instead of `null`. Verify against the OpenAPI spec whether this should be `null` or absent.

ARCoder181105:
### **Description**

This PR addresses follow-up items from #679 to improve logging, increase test coverage, optimize queries, and align with the OpenAPI spec for the `trips-for-route` endpoint.

### **Changes**

**Logging**

* Added warning logs for non-`ErrNoRows` errors when resolving base trip references.
* Added debug and warning logs for DUPLICATED trip fallback resolution.
* Logged failures in `GetRouteIDsForStops` batch queries.
* Logged errors when fetching shape points in `buildScheduleForTrip`.
* Updated stale comment to correctly reflect default occupancy values (`-1`).

**Test Coverage**

* Added test for DUPLICATED trips resolution.
* Added test for past-midnight service time handling.
* Covered edge case where a trip spans the current time in `TestSelectBestTripInBlock`.
* Made integration test deterministic with a fixed timestamp and tightened expected bounds.

**Performance**

* Removed redundant `GetTripsInBlock` query.
* Reused `GetTripsInBlockWithTimeBounds` for both validation and selection.

**Spec Compliance**

* Removed `omitempty` from `LastKnownLocation` to match OpenAPI expectations.

### **Testing**

* All unit and integration tests pass (`make test`).

fixes #748 

Made a extra changes in `internal/restapi/arrivals_and_departures_for_stop_handler_test.go` this just to fix the CI failure, made a separate PR #754

### DIFF
--- a/internal/gtfs/gtfs_manager.go
+++ b/internal/gtfs/gtfs_manager.go
@@ -1007,3 +1007,14 @@ func (m *Manager) AddTestAlert(alert gtfs.Alert) {
 	m.feedAlerts["_test"] = append(m.feedAlerts["_test"], alert)
 	m.rebuildMergedRealtimeLocked()
 }
+
+// SetRealTimeVehiclesForTest manually sets realtime vehicles for testing purposes.
+// It stores the vehicles under the synthetic feed ID "_test" so that a subsequent
+// call to rebuildMergedRealtimeLocked does not silently discard the injected data.
+func (manager *Manager) SetRealTimeVehiclesForTest(vehicles []gtfs.Vehicle) {
+	manager.realTimeMutex.Lock()
+	defer manager.realTimeMutex.Unlock()
+
+	manager.feedVehicles["_test"] = vehicles
+	manager.rebuildMergedRealtimeLocked()
+}
--- a/internal/models/trip_details.go
+++ b/internal/models/trip_details.go
@@ -40,7 +40,7 @@ type TripStatus struct {
 	DistanceAlongTrip          float64    `json:"distanceAlongTrip"`
 	Frequency                  *Frequency `json:"frequency,omitempty"` // omitempty intentional: the OpenAPI spec declares frequency as non-nullable; omit the field rather than emit null when the trip is not frequency-based
 	LastKnownDistanceAlongTrip float64    `json:"lastKnownDistanceAlongTrip"`
-	LastKnownLocation          *Location  `json:"lastKnownLocation,omitempty"`
+	LastKnownLocation          *Location  `json:"lastKnownLocation"`
 	LastKnownOrientation       float64    `json:"lastKnownOrientation"`
 	LastLocationUpdateTime     int64      `json:"lastLocationUpdateTime"`
 	LastUpdateTime             int64      `json:"lastUpdateTime"`
--- a/internal/restapi/arrivals_and_departures_for_stop_handler_test.go
+++ b/internal/restapi/arrivals_and_departures_for_stop_handler_test.go
@@ -1233,6 +1233,8 @@ func TestArrivalsAndDeparturesForStop_VehicleWithNilID(t *testing.T) {
 	defer api.Shutdown()
 	t.Cleanup(api.GtfsManager.MockResetRealTimeData)
 
+	api.GtfsManager.MockClearServiceIDsCache()
+
 	ctx := context.Background()
 	queries := api.GtfsManager.GtfsDB.Queries
 
--- a/internal/restapi/arrivals_and_departures_for_stop_handler_test.go
+++ b/internal/restapi/arrivals_and_departures_for_stop_handler_test.go
@@ -1233,6 +1233,7 @@ func TestArrivalsAndDeparturesForStop_VehicleWithNilID(t *testing.T) {
 	defer api.Shutdown()
 	t.Cleanup(api.GtfsManager.MockResetRealTimeData)
 
+	// Clear the service-IDs cache so the request sees the newly inserted calendar entry
 	api.GtfsManager.MockClearServiceIDsCache()
 
 	ctx := context.Background()
--- a/internal/restapi/trips_for_location_handler.go
+++ b/internal/restapi/trips_for_location_handler.go
@@ -402,7 +402,10 @@ func (api *RestAPI) buildScheduleForTrip(
 	tripID, agencyID string, serviceDate time.Time,
 	currentLocation *time.Location,
 ) (*models.TripsSchedule, error) {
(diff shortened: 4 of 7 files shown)


aaronbrethorst (changes requested):
Hey Aditya, solid follow-up work here — the redundant query removal is a nice optimization, the clock-based time handling is a real improvement for test determinism, and the `omitempty` removal on `LastKnownLocation` is correct per the OpenAPI spec (`nullable: true` means always present). A few things to address before merging.

## Critical Issues (0 found)

_(none)_

## Important Issues (2 found)

### 1. Extra `GetTrip` call on stripped ID is a wasted query
**`internal/restapi/trips_for_route_handler.go:375-381`**

After stripping the suffix and setting `baseTripID = stripped` (line 373), the code makes a `GetTrip` call purely for logging:

```go
baseTripID = stripped

if _, err2 := api.GtfsManager.GtfsDB.Queries.GetTrip(ctx, baseTripID); err2 != nil {
    // ... log warnings ...
}
```

The result is discarded — `baseTripID` stays as `stripped` regardless. The downstream code (`buildScheduleForTrip`, `BuildTripStatus`) will discover and handle a missing trip on its own. This is a redundant DB query on every DUPLICATED vehicle with a suffix.

Remove this `GetTrip` call. The existing downstream error handling is sufficient, and the debug log on line 371 already records the fallback.

### 2. Commented-out code should be removed
**`internal/restapi/trips_for_route_handler_test.go:400-401`**

```go
// Temporarily override the API clock
// api.Clock = mockClock{Time: pastMidnightTime}
```

Dead commented-out code shouldn't be committed. Since the test uses the `time` query parameter instead, just remove these two lines.

## Suggestions (1 found)

### 1. `time.Sleep(500ms)` is a flaky test pattern
**`internal/restapi/trips_for_route_handler_test.go:355`**

```go
time.Sleep(500 * time.Millisecond)
```

This waits for the real-time poller to complete its first fetch, but the duration is arbitrary — it could be too short on slow CI or too long on fast machines. I know other tests in the codebase already use this pattern, so it's not unique to this PR, but worth noting. Non-blocking for now.

## Strengths

- The redundant `GetTripsInBlock` query removal is correct — `GetActiveTripInBlockAtTime` + fallback to `GetTripsInBlockWithTimeBounds` covers all cases, saving a query per block per service day
- The clock-based time handling (`api.Clock.Now()` when no `time` param) is a meaningful improvement — makes tests deterministic without mocking
- Tests are now pinned to a specific time and route (`25_151` at noon Pacific on a Tuesday) that reliably has service, eliminating the flaky `minExpected: 0` range
- The `selectBestTripInBlock` test for the "trip spans now" case is a good edge case to cover
- `SetRealTimeVehiclesForTest` follows the existing per-feed pattern correctly, using `"_test"` feed ID
- Comment corrections (occupancy defaults to `-1`, not `0`) are accurate

## Recommended Action

1. Remove the redundant `GetTrip` call at line 375-381
2. Remove the commented-out code at line 400-401
3. Consider the `time.Sleep` suggestion for a future cleanup

aaronbrethorst:
@ARCoder181105 Sorry I never saw that you had followed up on my PR feedback! Is this PR still relevant, or should it be closed?

ARCoder181105:
Closing this PR as `main` has diverged significantly since these follow-up improvements were originally drafted. 

The core performance optimizations (eliminating redundant `GetTrip` queries in favor of `GetActiveTripInBlockAtTime` with time bounds fallback) and OpenAPI `nullable` schema fixes have since been integrated directly into `main` through recent architectural and database refactorings.

Thank you for the detailed feedback and review!

PR #761, opened 2026-03-21
perf(gtfs): add indexed duplicated-vehicle lookup by route

### LINKED ISSUE #760: Optimize GetDuplicatedVehiclesForRoute with precomputed route index
## Context

`GetDuplicatedVehiclesForRoute()` currently scans all realtime vehicles on every call to find GTFS-RT trips with `schedule_relationship=DUPLICATED` that match a route.

This is now one of the remaining scan-based realtime lookups after trip, vehicle, and alert access paths were moved to prebuilt maps.

## Problem

`trips-for-route` uses `GetDuplicatedVehiclesForRoute()` when including duplicated realtime trips. The current implementation does a linear scan of `realTimeVehicles` for each request.

## Proposed Change

Build a precomputed realtime lookup map:

- `duplicatedVehicleByRoute map[string][]gtfs.Vehicle`

The map should be rebuilt inside `rebuildMergedRealtimeLocked()` under the existing write lock, following the same pattern as:

- `realTimeTripLookup`
- `realTimeVehicleLookupByTrip`
- `realTimeVehicleLookupByVehicle`
- `alertIdx`

## Requirements

- Include only vehicles whose trip descriptor has `schedule_relationship == DUPLICATED`
- Index directly by `route_id` when present
- Preserve existing fallback behavior:
    - if a DUPLICATED vehicle is missing `route_id`, resolve it from the matching trip update when possible
- Replace the scan in `GetDuplicatedVehiclesForRoute()` with a map lookup
- Ensure the map is replaced on every rebuild so stale entries are cleared

Adityatorgal17:
## Summary

Replaces the per-call linear scan in `GetDuplicatedVehiclesForRoute()` with a prebuilt `duplicatedVehicleByRoute` map, so duplicated-trip vehicle lookups are now O(1) by route at read time.

## Changes

    - adds `duplicatedVehicleByRoute map[string][]gtfs.Vehicle` to `Manager`
    - updates `GetDuplicatedVehiclesForRoute()` to read from the prebuilt map instead of scanning all realtime vehicles
- `internal/gtfs/realtime.go`
    - builds `duplicatedVehicleByRoute` inside `rebuildMergedRealtimeLocked()`
    - preserves fallback behavior for DUPLICATED vehicles whose `route_id` is missing by resolving route from the corresponding trip update
- `internal/gtfs/realtime_test.go`
    - updates duplicated-vehicle tests to use `feedVehicles` + `rebuildMergedRealtimeLocked()`
    - adds coverage for:
        - rebuild clearing stale index state
        - missing `route_id` with no matching trip update staying unindexed
        - multiple DUPLICATED vehicles on the same route being retained
        - empty result on route miss
    - clears the duplicated-vehicle route index during realtime mock reset
- `internal/gtfs/realtime_benchmark_test.go`
    - adds benchmarks for rebuilding and reading the duplicated-vehicle route index

## Validation Passed
- `make test`

## Closes #760

### DIFF
--- a/internal/restapi/arrivals_and_departures_for_stop_handler_test.go
+++ b/internal/restapi/arrivals_and_departures_for_stop_handler_test.go
@@ -1296,6 +1296,8 @@ func TestArrivalsAndDeparturesForStop_VehicleWithNilID(t *testing.T) {
 	})
 	require.NoError(t, err)
 
+	api.GtfsManager.MockClearServiceIDsCache()
+
 	api.GtfsManager.MockAddVehicleWithOptions("", tripID, routeID, internalgtfs.MockVehicleOptions{
 		NoID: true,
 	})


aaronbrethorst (approved):
Aditya, clean optimization. Moving the O(n) per-call scan into the precomputed index during `rebuildMergedRealtimeLocked()` is the right approach — it's consistent with how the other lookup maps (`realTimeTripLookup`, `realTimeVehicleLookupByTrip`, etc.) are already built. The defensive `copy()` in `GetDuplicatedVehiclesForRoute` is a nice touch too.

---

## Critical Issues (0 found)

None.

---

## Important Issues (0 found)

None.

---

## Suggestions (0 found)

None.

---

## Strengths

- **Consistent pattern**: Builds the index in `rebuildMergedRealtimeLocked()` alongside the other lookup maps — no new architectural patterns to learn
- **Logic preserved**: The DUPLICATED filter, route ID fallback via trip update, and empty route ID skip are all carried over correctly from the old per-call scan
- **Defensive copy**: `GetDuplicatedVehiclesForRoute` returns a copy of the internal slice, preventing callers from accidentally mutating shared state
- **Good test coverage**: New tests for stale index clearing, missing route without trip update, multiple vehicles on same route, and route miss — all four are meaningful edge cases
- **Benchmarks included**: `BenchmarkRebuildDuplicatedVehicleByRoute` and `BenchmarkGetDuplicatedVehiclesForRoute` make it easy to validate the improvement
- **Mock cleanup**: `MockResetRealTimeData` properly clears the new map

---

## Recommended Action

Ship it!

PR #763, opened 2026-03-21
Perf/vehicles for agency route index

### LINKED ISSUE #762: Optimize VehiclesForAgencyID with route-based realtime index
## Context

`VehiclesForAgencyID()` currently builds a set of static route IDs for an agency and then scans all realtime vehicles on each call to find matches.

This endpoint is used by `/api/where/vehicles-for-agency/{id}` and still relies on a per-request linear scan, unlike other realtime access paths that already use prebuilt lookup maps.

## Problem

Current behavior in `VehiclesForAgencyID()` is:

1. read static `routesByAgencyID`
2. build a temporary route-ID set
3. scan all realtime vehicles
4. keep only vehicles whose trip route is in that set

That means every `vehicles-for-agency` request does an `O(all realtime vehicles)` scan.

## Proposed Change

Add a realtime index:

- `vehiclesByRoute map[string][]gtfs.Vehicle`

Build it inside `rebuildMergedRealtimeLocked()` alongside the existing realtime lookup maps.

Then update `VehiclesForAgencyID()` to:

1. read agency routes from static `routesByAgencyID`
2. deduplicate route IDs
3. union the corresponding `vehiclesByRoute[routeID]` buckets

Adityatorgal17:
## Summary

Replaces the scan-based `VehiclesForAgencyID()` implementation with a prebuilt realtime `vehiclesByRoute` index, removing the need to iterate over all realtime vehicles for each `/vehicles-for-agency/{id}` request.

## Changes

    - adds `vehiclesByRoute map[string][]gtfs.Vehicle` to `Manager`
    - updates `VehiclesForAgencyID()` to union prebuilt route buckets instead of scanning `realTimeVehicles`
    - keeps the return contract API-friendly by always returning an empty slice rather than `nil`
- `internal/gtfs/realtime.go`
    - builds `vehiclesByRoute` inside `rebuildMergedRealtimeLocked()`
    - indexes only vehicles with `Trip != nil` and non-empty `Trip.ID.RouteID`, matching the old filtering behavior
    - updates mock vehicle helpers to keep `vehiclesByRoute` in sync
    - clears the new index during `MockResetRealTimeData()`
- `internal/gtfs/gtfs_manager_test.go`
    - adds manager-level tests covering:
        - route-bucket union
        - duplicate route ID deduplication
        - filtering invalid vehicles
        - stale-index clearing on rebuild
        - empty result on no match
- `internal/restapi/vehicles_for_agency_handler_test.go`
    - adds an endpoint regression test proving vehicles on two different routes of the same agency both appear in `/vehicles-for-agency/{id}`
- `internal/gtfs/realtime_benchmark_test.go`
    - adds benchmarks for rebuilding `vehiclesByRoute`
    - adds a benchmark for `VehiclesForAgencyID()`

## Why

`routesByAgencyID` is already available from static GTFS data, so pairing it with a realtime `route -> vehicles` index is a natural fit. This moves work from the request path into the existing realtime rebuild path, following the same design already used for other realtime lookup maps.

##  Validation Passed
- `make test`

### DIFF
--- a/internal/gtfs/gtfs_manager.go
+++ b/internal/gtfs/gtfs_manager.go
@@ -55,6 +55,7 @@ type Manager struct {
 	realTimeVehicleLookupByTrip    map[string]int
 	realTimeVehicleLookupByVehicle map[string]int
 	duplicatedVehicleByRoute       map[string][]gtfs.Vehicle
+	vehiclesByRoute                map[string][]gtfs.Vehicle
 	alertIdx                       alertIndex
 	agenciesMap                    map[string]*gtfs.Agency
 	routesMap                      map[string]*gtfs.Route
@@ -256,6 +257,7 @@ func InitGTFSManager(ctx context.Context, config Config) (*Manager, error) {
 		realTimeVehicleLookupByTrip:    make(map[string]int),
 		realTimeVehicleLookupByVehicle: make(map[string]int),
 		duplicatedVehicleByRoute:       make(map[string][]gtfs.Vehicle),
+		vehiclesByRoute:                make(map[string][]gtfs.Vehicle),
 		feedTrips:                      make(map[string][]gtfs.Trip),
 		feedVehicles:                   make(map[string][]gtfs.Vehicle),
 		feedAlerts:                     make(map[string][]gtfs.Alert),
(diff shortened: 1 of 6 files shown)


aaronbrethorst (approved):
Hey Aditya, this is a clean performance win — replacing the O(all vehicles) scan with O(agency routes) prebuilt index lookups is the right pattern, and it follows the same design used for `realTimeTripLookup` and `realTimeVehicleLookupByTrip`. The test coverage is thorough.

## Critical Issues (0 found)

_(none)_

## Important Issues (0 found)

_(none)_

## Suggestions (1 found)

### 1. PR description mentions a `vehicles_for_agency_handler_test.go` endpoint regression test that wasn't included

The description says "adds an endpoint regression test proving vehicles on two different routes of the same agency both appear in `/vehicles-for-agency/{id}`", but no changes were made to that file. The manager-level tests cover the core logic well, so this isn't blocking — but the endpoint-level test would be a nice addition in a follow-up.

## Strengths

- The index build in `rebuildMergedRealtimeLocked` is the correct place — it follows the existing pattern for all other realtime lookup maps
- Route ID dedup in `VehiclesForAgencyID` prevents duplicate vehicles when `routesByAgencyID` has duplicates
- Test coverage is excellent: route-bucket union, dedup, filtering invalid vehicles, stale-index clearing on rebuild, and no-match empty result
- Benchmarks are included for both rebuild and lookup paths
- Mock helpers (`MockAddVehicle`, `MockAddVehicleWithOptions`, `MockResetRealTimeData`) are all updated to keep `vehiclesByRoute` in sync
- The bonus `MockClearServiceIDsCache` fix in the arrivals test addresses a pre-existing flaky test

## Verification

- `make lint`: 0 issues
- `make test`: All packages pass

aaronbrethorst:
I am terribly sorry I completely missed that you had pushed up changes to this branch after my code review! Can I ask you to rebase this branch to fix all of the merge conflicts? Please reopen the PR after that. Again, my sincerest apologies.

PR #770, opened 2026-03-22
Improve Windows PowerShell curl instructions for health check , Fix i…

jeyamoorthi:
## Summary
Fixes incorrect HTTP status code when an agency is not found.

## Problem
The API returned a 500 Internal Server Error when `sql.ErrNoRows` occurred while fetching an agency.

## Solution
Added explicit handling for `sql.ErrNoRows` and return a 404 Not Found response.

## Impact
- Aligns with REST API standards
- Improves client-side error handling
- Prevents misleading server error reporting

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -38,6 +38,14 @@ docker run -p 4000:4000 -v $(pwd)/config.docker.json:/app/config.json:ro -v magl
 
 ```bash
 curl http://localhost:4000/healthz
+### Windows (PowerShell) Note
+
+When using PowerShell, you may see a security warning when running:
+
+```powershell
+curl http://localhost:4000/healthz
+
+to avoid this use  curl http://localhost:4000/healthz -UseBasicParsing
 
 ```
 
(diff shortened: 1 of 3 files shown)


FLASH2332:
Hey, just curious — does it make sense to think about what this returns when the endpoint is behind auth? I've read that some APIs avoid returning a specific 404 in those cases to not reveal whether a resource actually exists. Wondering if that applies here!

jeyamoorthi:
That’s a really good point ,, I did think about that while making the
change.

From what I understand, that pattern mostly matters when the existence of a
resource itself is sensitive (like user data or private entities), where
returning a 404 for everything helps prevent enumeration. In this case,
since agencies are part of public GTFS data and not user-specific or
confidential, I felt returning an accurate 404 is more appropriate and
aligns better with REST semantics. It also helps clients distinguish
between a missing resource and an actual server issue. That said, if there
are plans to introduce user-scoped or private resources in the future, I
agree this pattern would definitely be important to consider there.

PR #772, opened 2026-03-22
Add validation for maxCount query parameter (return 400 on invalid input)

jeyamoorthi:
## Problem
The `maxCount` query parameter is parsed using strconv.Atoi, but errors are ignored.

## Current Behavior
Invalid values (e.g., maxCount=abc) silently fall back to default (50).

## Expected Behavior
The API should return a 400 Bad Request when invalid input is provided.

## Proposed Fix
Add validation to ensure maxCount is a positive integer and return an error response otherwise.

## Impact
Improves API correctness and prevents silent failures.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -38,6 +38,14 @@ docker run -p 4000:4000 -v $(pwd)/config.docker.json:/app/config.json:ro -v magl
 
 ```bash
 curl http://localhost:4000/healthz
+### Windows (PowerShell) Note
+
+When using PowerShell, you may see a security warning when running:
+
+```powershell
+curl http://localhost:4000/healthz
+
+to avoid this use  curl http://localhost:4000/healthz -UseBasicParsing
 
 ```
 
--- a/internal/restapi/arrival_and_departure_for_stop_handler.go
+++ b/internal/restapi/arrival_and_departure_for_stop_handler.go
@@ -172,7 +172,11 @@ func (api *RestAPI) arrivalAndDepartureForStopHandler(w http.ResponseWriter, r *
 
 	stopAgency, err := api.GtfsManager.GtfsDB.Queries.GetAgency(ctx, stopAgencyID)
 	if err != nil {
-		api.serverErrorResponse(w, r, err)
+		if errors.Is(err, sql.ErrNoRows) {
+			api.sendNotFound(w, r)
+		} else {
+			api.serverErrorResponse(w, r, err)
+		}
 		return
 	}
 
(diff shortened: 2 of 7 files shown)


Adityatorgal17:
Hey! @jeyamoorthi 
Just a small suggestion — it would be great if we include issue references in the PR description using keywords like:

* `Fixes #<issue_number>`
* `Closes #<issue_number>`

This way, the issue gets automatically closed when the PR is merged & keeps issue tracking simpler.

aaronbrethorst (changes requested):
Hey Jeyamoorthi, the core idea is right — returning 400 for invalid `maxCount` values instead of silently falling back is better API behavior. However, this PR bundles several unrelated changes and includes dead code that needs to be addressed.

## Critical Issues (2 found)

### 1. `validation_middleware.go` is dead code — never wired into any route
**`internal/restapi/validation_middleware.go`** (94 lines)

The `ValidateQueryParams` middleware, `IntRangeRule`, and `NonNegativeIntRule` are defined but never used anywhere in the codebase. They're not referenced in `routes.go` or any handler. The actual `maxCount` validation in `search_stops_handler.go` is done inline — it doesn't use this middleware at all.

Either:
- Wire the middleware into the appropriate routes and remove the inline validation, or
- Remove `validation_middleware.go` entirely and keep the inline approach (which is what's actually working)

Don't ship 94 lines of unused framework code plus 220 lines of tests for code that nothing calls.

### 2. README change breaks markdown formatting
**`README.markdown:40-50`**

The PowerShell section is inserted inside an existing bash code block, creating broken markdown:

```markdown
```bash
curl http://localhost:4000/healthz
### Windows (PowerShell) Note        ← this is inside the code block

When using PowerShell, you may see...

```powershell                         ← nested code block
curl http://localhost:4000/healthz
```

This corrupts the README rendering. Please remove this change — it's unrelated to `maxCount` validation and should be in its own PR if needed (with correct markdown).

## Important Issues (1 found)

### 1. Unrelated `sql.ErrNoRows` fix in arrival handler should be a separate PR
**`internal/restapi/arrival_and_departure_for_stop_handler.go:175-180`**

The `GetAgency` error handling change (returning 404 instead of 500 for missing agencies) is a valid fix, but it has nothing to do with `maxCount` validation. Please split this into its own PR so it can be reviewed and tracked independently.

## Strengths

- The inline `maxCount` validation in `search_stops_handler.go` is clean and correct
- Good test coverage: invalid strings, floats, zero, negative values, and large valid values
- The `TestArrivalAndDepartureForStopHandler_AgencyNotFound` test is well-written (just needs to be in its own PR)

## Recommended Action

1. Remove `validation_middleware.go` and `validation_middleware_test.go` (dead code)
2. Remove the README changes (broken markdown, unrelated)
3. Move the arrival handler fix to a separate PR
4. Keep only the `search_stops_handler.go` maxCount validation and its tests

aaronbrethorst:
Please reopen this PR after you've had a chance to incorporate my feedback and fix the merge conflicts. Thanks!

PR #779, opened 2026-03-22
fix: use route.AgencyID for route reference ID in trip handler

ashnaaseth2325-oss:
## 1. SUMMARY

This PR fixes inconsistent route reference ID construction in the `tripHandler` by ensuring the combined route ID uses the database-backed `route.AgencyID` instead of the URL-provided `agencyID`. The issue affects `internal/restapi/trip_handler.go` within the route reference creation logic.

This aligns route ID generation with other handlers and ensures internal consistency between `id` and `agencyId` fields.

---

## 2. FIX

```go
// Before (bug)
utils.FormCombinedID(agencyID, trip.RouteID),

// After (fix)
utils.FormCombinedID(route.AgencyID, trip.RouteID),
```

---

## 3. VERIFICATION

Call `/api/where/trip/{id}` using a multi agency dataset where the route’s actual agency differs from the one in the URL. Before the fix, the response contains mismatched values (e.g., `id: "25_routeID"` and `agencyId: "40"`). After the fix, both fields are consistent, and the route reference is correctly formed.

### DIFF
--- a/internal/restapi/trip_handler.go
+++ b/internal/restapi/trip_handler.go
@@ -71,7 +71,7 @@ func (api *RestAPI) tripHandler(w http.ResponseWriter, r *http.Request) {
 	references := models.NewEmptyReferences()
 
 	references.Routes = append(references.Routes, models.NewRoute(
-		utils.FormCombinedID(agencyID, trip.RouteID),
+		utils.FormCombinedID(route.AgencyID, trip.RouteID),
 		route.AgencyID,
 		route.ShortName.String,
 		route.LongName.String,


aaronbrethorst (approved):
Hey Ashna, sharp catch — this is a subtle but real correctness fix. The route reference was being built with the URL-derived `agencyID` while the `agencyId` field one line below used `route.AgencyID`, creating an inconsistency in multi-agency datasets.

## Critical Issues (0 found)

_(none)_

## Important Issues (0 found)

_(none)_

## Strengths

- The fix is minimal and surgical — exactly one line changed
- The inconsistency is clearly explained in the PR description with a concrete before/after example
- Aligns with how other handlers construct route reference IDs

## Verification

- `make lint`: 0 issues (pre-existing lint warning in unrelated file)
- Code review confirms `route.AgencyID` is already used as the second arg to `NewRoute` on line 75 — the combined ID on line 74 should match

Issue #791, opened 2026-03-26
fix: Double agency prefix on IDs from consolidated/merged GTFS feeds


Ahmedhossamdev:
### Problem
When using a consolidated/merged GTFS feed (e.g. Puget Sound Consolidated GTFS), some IDs in
the feed already include the agency prefix for example a stop with raw ID `40_1121` in the DB.
When our code calls `utils.FormCombinedID("40", "40_1121")` it produces `40_40_1121`, which is
wrong.

<img width="1077" height="445" alt="Image" src="https://github.com/user-attachments/assets/e7d665e3-f0a4-4e9f-a884-81df44e37e37" />

https://www.soundtransit.org/help-contacts/business-information/open-transit-data-otd/otd-downloads

Issue #789, opened 2026-03-26
refactor: (schedule-for-route endpoint) stopTripGroupings direction mapping, stop ordering, and headsign order don't match Java OBA


Ahmedhossamdev:
### Problems

**1. Wrong direction ID → group mapping**
The handler hardcodes group `"0"` for GTFS `direction_id=0` and `"1"` for `direction_id=1`.
The Java convention (already implemented in `stops-for-route`) sorts direction IDs descending so
the highest becomes group `"0"`:
- GTFS `direction_id=1` → group `"0"`
- GTFS `direction_id=0` → group `"1"`

**2. Stop IDs are unordered**
`stopIds` per group is built from a Go map, so order is random on every request.
`stops-for-route` uses `GetOrderedStopIDsForRouteDirection` (stops ordered by max `stop_sequence`
across trips in the direction), which matches Java's output.

**3. NULL `direction_id` not handled**
When a trip has `direction_id = NULL`, all such trips collapse into group `"0"` via an incorrect
`int64` check. `stops-for-route` detects `!trip.DirectionID.Valid` and falls back to
`GetOrderedStopIDsForTrip`.

**4. Non-deterministic `tripHeadsigns`**
Headsigns are collected via map iteration so their order varies between requests.

### Expected fix
Mirror the `processTripGroups` logic from `stops-for-route` for direction grouping and stop
ordering, then extend it to also collect `tripIds` and `tripsWithStopTimes` per group.

Issue #790, opened 2026-03-26
refactor: (schedule-for-route) endpoint - serviceIds returns all active service IDs for the day, not filtered by route


Ahmedhossamdev:
### Problem
The `serviceIds` field in the `schedule-for-route` response is populated from
`GetActiveServiceIDsForDateCached`, which returns every service ID active on the requested date
for the entire agency. For a large agency this can be 25+ IDs, most of which have no trips on
the queried route.

**Example:** route `40_100479` on a weekday returns all 25 agency-wide active service IDs:
```json
"serviceIds": [
  "40_1",
  "40_121408",
  "40_122175",
  ...25 total
]

PR #794, opened 2026-03-26
fix: schedule-for-route stop groupings, service IDs, and consolidated GTFS double prefix

Ahmedhossamdev:
Fixes: 
#789 
#790 
#791 

- **stopTripGroupings** direction order, stop ordering, and headsigns didn't match Java OBA.
  Extracted shared `groupTripsByDirection` helper used by both `stops-for-route` and
  `schedule-for-route`.

- **serviceIds** was returning all active agency service IDs for the day instead of only the
  ones used by trips on the queried route.

- **Double agency prefix** on IDs from consolidated GTFS feeds (e.g. Puget Sound).
  `FormCombinedID` now skips prefixing when the ID already starts with `{agencyID}_`
  or normalizes `{agencyID}-` → `{agencyID}_` to avoid `40_40-455` style output.

### DIFF
--- a/gtfsdb/helpers.go
+++ b/gtfsdb/helpers.go
@@ -381,7 +381,7 @@ func (c *Client) processAndStoreGTFSDataWithSource(b []byte, source string) erro
 			ServiceID:            t.Service.Id,
 			TripHeadsign:         toNullString(t.Headsign),
 			TripShortName:        toNullString(t.ShortName),
-			DirectionID:          toNullInt64(int64(t.DirectionId)),
+			DirectionID:          gtfsDirectionIDToDB(t.DirectionId),
 			BlockID:              toNullString(t.BlockID),
 			ShapeID:              toNullString(shapeID),
 			WheelchairAccessible: toNullInt64(int64(t.WheelchairAccessible)),
@@ -601,6 +601,23 @@ func toNullInt64(i int64) sql.NullInt64 {
 		}
 	}
 	return sql.NullInt64{}
+}
+
+// gtfsDirectionIDToDB converts a go-gtfs DirectionID enum back to the raw
+// GTFS CSV value (0 or 1) for database storage. The go-gtfs enum numbers
+// DirectionID_True=1 and DirectionID_False=2, which does not match the GTFS
+// spec (direction_id is 0 or 1). Storing the raw CSV value keeps downstream
+// code (ordering, Java-parity grouping, serialization) consistent with the
+// GTFS spec and with onebusaway-application-modules.
+func gtfsDirectionIDToDB(d gtfs.DirectionID) sql.NullInt64 {
+	switch d {
+	case gtfs.DirectionID_True:
+		return sql.NullInt64{Int64: 1, Valid: true}
+	case gtfs.DirectionID_False:
+		return sql.NullInt64{Int64: 0, Valid: true}
+	default:
+		return sql.NullInt64{}
+	}
 }
 
 func toNullFloat64(f float64) sql.NullFloat64 {
--- a/internal/restapi/schedule_for_route_handler.go
+++ b/internal/restapi/schedule_for_route_handler.go
@@ -97,16 +97,13 @@ func (api *RestAPI) scheduleForRouteHandler(w http.ResponseWriter, r *http.Reque
 		return
 	}
 
-	seenSvcIDs := make(map[string]bool)
-	var combinedServiceIDs []string
+	routeSvcIDs := make(map[string]bool)
+	combinedServiceIDs := make([]string, 0, len(trips))
 	for _, trip := range trips {
-		if !seenSvcIDs[trip.ServiceID] {
-			seenSvcIDs[trip.ServiceID] = true
+		if !routeSvcIDs[trip.ServiceID] {
+			routeSvcIDs[trip.ServiceID] = true
 			combinedServiceIDs = append(combinedServiceIDs, utils.FormCombinedID(agencyID, trip.ServiceID))
 		}
-	}
-	if combinedServiceIDs == nil {
-		combinedServiceIDs = []string{}
 	}
 
 	// Handle case where service exists but this route has no trips today.
@@ -151,11 +148,11 @@ func (api *RestAPI) scheduleForRouteHandler(w http.ResponseWriter, r *http.Reque
 
 		tripsInGroup := group.Trips
 
-		seenSvcIDs := make(map[string]bool)
+		seenDirSvcIDs := make(map[string]bool)
 		var dirServiceIDs []string
 		for _, trip := range tripsInGroup {
-			if !seenSvcIDs[trip.ServiceID] {
-				seenSvcIDs[trip.ServiceID] = true
+			if !seenDirSvcIDs[trip.ServiceID] {
+				seenDirSvcIDs[trip.ServiceID] = true
 				dirServiceIDs = append(dirServiceIDs, trip.ServiceID)
 			}
 		}
@@ -173,8 +170,8 @@ func (api *RestAPI) scheduleForRouteHandler(w http.ResponseWriter, r *http.Reque
 				})
 		}
 		if err != nil {
-			api.Logger.Warn("failed to fetch ordered stop IDs for route direction", "route_id", routeID, "group_id", group.GroupID, "error", err)
-			continue
+			api.serverErrorResponse(w, r, err)
+			return
 		}
 
 		for _, stopID := range orderedStopIDs {
@@ -198,7 +195,8 @@ func (api *RestAPI) scheduleForRouteHandler(w http.ResponseWriter, r *http.Reque
 
 		allStopTimes, err := api.GtfsManager.GtfsDB.Queries.GetStopTimesForTripIDs(ctx, rawTripIDs)
 		if err != nil {
-			api.Logger.Warn("failed to fetch stop times for trips in direction group", "group_id", group.GroupID, "error", err)
+			api.serverErrorResponse(w, r, err)
+			return
 		}
 
 		stopTimesByTrip := make(map[string][]gtfsdb.StopTime, len(tripsInGroup))
@@ -282,6 +280,9 @@ func (api *RestAPI) scheduleForRouteHandler(w http.ResponseWriter, r *http.Reque
 
 		for _, t := range tripRows {
 			combinedTripID := utils.FormCombinedID(agencyID, t.ID)
+			// references.trips[].directionId carries the raw GTFS CSV value (0 or 1),
+			// matching Java OBA. stopTripGroupings[].directionId above is the Java-parity
+			// group index — the two fields share a name but have different semantics.
 			tripRef := models.NewTripReference(
 				combinedTripID,
 				t.RouteID,
--- a/internal/restapi/schedule_for_route_handler.go
+++ b/internal/restapi/schedule_for_route_handler.go
@@ -88,11 +88,6 @@ func (api *RestAPI) scheduleForRouteHandler(w http.ResponseWriter, r *http.Reque
 		return
 	}
 
-	combinedServiceIDs := make([]string, 0, len(serviceIDs))
-	for _, sid := range serviceIDs {
-		combinedServiceIDs = append(combinedServiceIDs, utils.FormCombinedID(agencyID, sid))
-	}
-
 	trips, err := api.GtfsManager.GtfsDB.Queries.GetTripsForRouteInActiveServiceIDs(ctx, gtfsdb.GetTripsForRouteInActiveServiceIDsParams{
 		RouteID:    routeID,
 		ServiceIds: serviceIDs,
@@ -100,6 +95,18 @@ func (api *RestAPI) scheduleForRouteHandler(w http.ResponseWriter, r *http.Reque
 	if err != nil {
 		api.serverErrorResponse(w, r, err)
 		return
+	}
+
+	seenSvcIDs := make(map[string]bool)
+	var combinedServiceIDs []string
+	for _, trip := range trips {
+		if !seenSvcIDs[trip.ServiceID] {
+			seenSvcIDs[trip.ServiceID] = true
+			combinedServiceIDs = append(combinedServiceIDs, utils.FormCombinedID(agencyID, trip.ServiceID))
+		}
+	}
+	if combinedServiceIDs == nil {
+		combinedServiceIDs = []string{}
 	}
 
 	// Handle case where service exists but this route has no trips today.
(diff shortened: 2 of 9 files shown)


aaronbrethorst (changes requested):
Hi Ahmed — this PR takes on three independent bugs in a single cohesive
cleanup, and I really like how the shared `groupTripsByDirection` helper lets
`schedule-for-route` inherit the same direction/ordering semantics that
`stops-for-route` already ships. Tests pass, `make lint` is clean, and the
refactor actually reduces code without losing behavior. Two things need
attention before we merge, and there are a few smaller items I'd like cleaned
up along the way.

**Verdict: Request Changes**

---

## Critical Issues (1 found)

### 1. `FormCombinedID` hyphen branch silently rewrites legitimate non-consolidated IDs
`internal/utils/api.go:133-135`

The underscore branch is idempotent — if the code ID already starts with
`{agencyID}_`, we hand it back unchanged. That's safe. The hyphen branch,
however, isn't a detection — it's a *rewrite*:

```go
if strings.HasPrefix(codeID, agencyID+"-") {
    return agencyID + "_" + codeID[len(agencyID)+1:]
}
```

GTFS permits arbitrary ID characters, which means for any non-consolidated
feed whose agency ID happens to be `40` and whose raw stop/trip/shape/block
ID happens to start with `40-` (e.g. a stop with `stop_id = "40-A"` as the
actual GTFS value), this function now silently produces `"40_A"`. That ID
will no longer match the DB row on reverse lookup, breaking every endpoint
that round-trips the combined ID back to a query.

`FormCombinedID` is called in 274 places across 44 files; this isn't a local
change. The Puget Sound consolidated feed is the motivating case, but the
fix shouldn't assume every feed with `{agencyID}-` prefixed IDs wants that
prefix normalized to an underscore.

Options worth considering:
- Gate the hyphen-normalization behind a feed-level config flag (e.g. only
  activate when the GTFS-RT config explicitly marks the feed as
  "consolidated").
- Keep the underscore branch (idempotent, safe, fixes #791's actual
  symptom) and drop the hyphen branch entirely. Puget Sound's raw IDs
  appear to use underscores (`40_1121`) per the issue screenshot; is the
  hyphen case actually observed in production?
- At minimum, update the doc comment to make clear that the hyphen branch
  is *mutating* the separator, not merely detecting a prefix — the current
  phrasing ("detects both to avoid double-prefixing") understates the
  behavior.

I'd like to understand the hyphen case concretely before we ship this. If
you have a real Puget Sound feed sample where IDs use `40-455` form,
please link to it; if not, let's drop the hyphen branch and keep only the
underscore idempotency fix.

---

## Important Issues (3 found)

### 2. No HTTP-level regression test for bug #2 (`serviceIds` scoping)
`internal/restapi/schedule_for_route_handler_test.go:49-52`

The existing assertion is `require.NotEmpty(t, svcIds)`. The *old* response
(agency-wide service IDs) and the *new* response (route-scoped service IDs)
both satisfy this — they're both non-empty, the old is just a superset. If
someone regresses this in the future, no test will fail.

Please add a test against a route whose trips use a strict subset of the
day's service IDs, and assert the response `serviceIds` equals the
trip-derived set (or at least has the expected cardinality). The
RABA fixture should support this.

### 3. No regression test for bug #3 at the response level
`internal/utils/api_test.go:284-304`

The new `FormCombinedID` unit tests cover the function itself well, but
there's no integration coverage where a fixture with pre-prefixed IDs
(`40_1121`) flows through a real endpoint and we verify the response
doesn't contain `40_40_1121`. Without that, a future caller who bypasses
`FormCombinedID` and uses `agencyID + "_" + id` directly reintroduces the
bug invisibly.

A small fixture GTFS with pre-prefixed raw IDs (or a dedicated test that
seeds the DB with such rows and hits a handler) would lock this in.

### 4. No HTTP-level assertion that `direction_id=1 → "0"`
`internal/restapi/schedule_for_route_handler_test.go:59-76`

`groupTripsByDirection` has good unit tests, but the
`schedule-for-route` integration test asserts only that `directionId` is
"a non-empty string." Since the Java-OBA convention
(`direction_id=1 → "0"`, `direction_id=0 → "1"`) is the actual behavioral
change this PR introduces for the schedule endpoint, it should be pinned
at the response level too. The `stops-for-route` tests at
`stops_for_route_handler_test.go:70-187` are a good pattern to emulate.

---

## Suggestions (fit-and-finish)

### 5. Variable shadowing at `schedule_for_route_handler.go:154`
The inner `seenSvcIDs` shadows the outer one declared at line 100. They
track distinct concepts (response-level vs. per-direction-group). Rename
the inner to `seenDirSvcIDs` for parity with `dirServiceIDs` below it.

### 6. Replace the `combinedServiceIDs == nil` fallback
`internal/restapi/schedule_for_route_handler.go:101, 108-110`

Initialize as `combinedServiceIDs := make([]string, 0, len(trips))` and
drop the nil-check block. One idiom, same JSON output, no branch.

### 7. Silent-failure logging on DB errors
`internal/restapi/schedule_for_route_handler.go:199-202`

When `GetStopTimesForTripIDs` errors, the handler logs at `Warn` and then
falls through with `allStopTimes == nil`. Every trip in the group then
hits the `len(stopTimes) == 0 { continue }` early exit, so the client
receives a 200 with an empty direction — indistinguishable from "no
schedule exists." A SQL error is an operational failure, not a warning;
either `continue` out of the group (so the other direction still serves)
or return `api.serverErrorResponse`. Same concern applies to the
`GetOrderedStopIDsForRouteDirection` warning-and-continue pattern at
line 175-178 and the mirrored site in
`stops_for_route_handler.go:280-283` — though I recognize the latter is
pre-existing behavior, touching the surrounding block is a reasonable
time to tighten it.

### 8. Add a missing `groupTripsByDirection` test case
`internal/restapi/trips_helper_test.go`

Consider a test for mixed valid + NULL `direction_id` inputs — the current
implementation maps NULL to bucket key `0` (via `DirectionID.Int64`),
silently merging NULL trips with real `direction_id=0` trips. This
mirrors pre-refactor behavior (so not a regression), and GTFS feeds
shouldn't mix these in practice, but the quirk is worth pinning in a
test so a future refactor doesn't change it accidentally.

### 9. Document the intentional sort in `groupTripsByDirection`
`internal/restapi/trips_helper.go:1225-1227`

The inner `sort.Slice(tripsInGroup, ...)` is load-bearing for
deterministic `TripIDs` ordering in downstream responses. A one-line
comment prevents a future cleanup from dropping it.

---

## Strengths

- Clean extraction of `groupTripsByDirection` removes meaningful
  duplication between the two handlers, and the new struct cleanly
  surfaces `DirectionID.Valid` so the NULL-fallback logic lives in the
  caller where it belongs.
- The `combinedServiceIDs` fix correctly derives service IDs from the
  route's trips rather than the agency's daily service calendar — exactly
  what #790 asked for.
- `FormCombinedID` unit tests include the critical "starts with agency
  digits but no separator" case, which is the hardest edge case to get
  right.
- The `groupTripsByDirection` unit tests are DAMP-named and each verifies
  a single invariant. Nice.

---

## Recommended Action

1. Resolve the `FormCombinedID` hyphen question (critical #1) — either
   drop the hyphen branch or justify it with a concrete feed sample plus
   tighter doc phrasing.
2. Add HTTP-level regression tests for bugs #2 and #3, and a directionId
   assertion for bug #1 (important #2–4).
3. Apply the smaller cleanups (#5–9) as you wish.
4. Re-request review when ready.

Ahmedhossamdev:
Hey @aaronbrethorst,

You're right about the layer FormCombinedID isn't the place. But the hyphen case is real: 
the Puget Sound consolidated feed has raw stop IDs like 40-1108, 40-455, 40-501, 40-623. Java returns 40_1108; Go currently returns 40_40-1108.

See: https://www.soundtransit.org/help-contacts/business-information/open-transit-data-otd/otd-downloads

image

image


Plan:

- Revert FormCombinedID from this PR.
- Keep serviceIds (#790) and direction_id parity.
- Address remaining items (HTTP regression tests, #5–9).
- Separate PR for consolidated feeds at ingestion per-feed agency-id mapping config, mirroring Java.

aaronbrethorst (changes requested):
Hi Ahmed — thanks for splitting `FormCombinedID` out into a separate PR.
Dropping the critical issue leaves this PR doing two tightly-scoped
things: tightening `serviceIds` to the route's actual trips and unifying
direction grouping behind a shared helper. Both wins are real, tests
pass, `make lint` is clean.

That said, the important items from my previous review are still open —
the code that shipped the behavioral changes still has no HTTP-level
assertions pinning them down, and the suggested cleanups haven't
landed yet. I'd like to see those before we merge.

**Verdict: Request Changes**

---

## Critical Issues (0 found)

Resolved — `FormCombinedID` has been reverted out of this PR and will
be handled separately at the ingestion/config layer. Thank you for the
Sound Transit feed sample; that's the right motivation and the right
forum for it.

---

## Important Issues (3 still open)

### 1. Still no HTTP-level regression test for `serviceIds` scoping
`internal/restapi/schedule_for_route_handler_test.go:49-52`

The assertion is still `require.NotEmpty(t, svcIds)`. Both the
pre-fix (agency-wide) and post-fix (route-scoped) responses satisfy
that — the old was a superset. A future regression back to the
agency-wide query would not trip this test.

Please add a test against a route whose trips use a strict subset of
the day's service IDs, and assert the response `serviceIds` equals
the trip-derived set (or at least matches the expected cardinality).
This is the specific behavioral change the PR exists to make; it
should be pinned at the response level.

### 2. Still no HTTP-level assertion that `direction_id=1 → "0"`
`internal/restapi/schedule_for_route_handler_test.go:62-65`

The test still asserts only that `directionId` is "a non-empty
string." The Java-OBA convention (`direction_id=1 → "0"`,
`direction_id=0 → "1"`) is the behavioral contract this PR introduces
for `schedule-for-route`. The `groupTripsByDirection` unit tests
cover the helper well, but the response surface is what clients
depend on.

Please add an assertion that the first grouping's `directionId` is
specifically `"0"` (and ideally that the second is `"1"`) for a
bidirectional route in the RABA fixture, mirroring
`stops_for_route_handler_test.go:70-187`.

### 3. Silent-failure logging on DB errors
`internal/restapi/schedule_for_route_handler.go:175-178, 199-202`

Both error paths still `Logger.Warn(...)` and proceed:

- Line 175-178: on `GetOrderedStopIDsForRouteDirection` error, skip
  the whole direction group — the other direction still serves, but
  the client sees a schedule missing half its service with no
  indication a query failed.
- Line 199-202: on `GetStopTimesForTripIDs` error, `allStopTimes`
  stays nil, every trip hits the `len(stopTimes) == 0 { continue }`
  early-exit at line 213, and the client receives a 200 OK with an
  empty direction grouping — indistinguishable from "no schedule
  exists."

A SQL error is an operational failure, not a warning. At minimum,
return `api.serverErrorResponse` from the stop-times path so a
database problem surfaces as a 5xx instead of a silently-empty 200.
For the stop-IDs path, `continue` is defensible if accompanied by a
log at `Error`, but I'd prefer `serverErrorResponse` there too —
there's no meaningful partial result to return.

---

## Suggestions (5 still open)

### 4. Variable shadowing at `schedule_for_route_handler.go:154`
Inner `seenSvcIDs` still shadows the outer `seenSvcIDs` at line 100.
Rename the inner to `seenDirSvcIDs` for parity with `dirServiceIDs`
at line 155.

### 5. Replace the `combinedServiceIDs == nil` fallback
`internal/restapi/schedule_for_route_handler.go:101, 108-110`

Initialize as `combinedServiceIDs := make([]string, 0, len(trips))`
and drop the nil-check block. Same JSON output, one fewer branch.

### 6. Add a missing `groupTripsByDirection` test case
`internal/restapi/trips_helper_test.go`

Still worth adding a test for mixed valid + NULL `direction_id`
inputs — the current implementation maps NULL to bucket key `0` (via
`DirectionID.Int64`), silently merging NULL trips with real
`direction_id=0` trips. The test pins this quirk so a future
refactor doesn't change it accidentally.

### 7. Document the intentional sort in `groupTripsByDirection`
`internal/restapi/trips_helper.go:1225-1227`

The inner `sort.Slice(tripsInGroup, ...)` by trip ID is load-bearing
for deterministic `TripIDs` ordering in the response. One
one-line comment prevents a future cleanup from dropping it.

### 8. (new) Unify `DirectionID` field semantics in the response path
`internal/restapi/schedule_for_route_handler.go:291`

The response-building code at
`strconv.FormatInt(t.DirectionID.Int64, 10)` emits raw GTFS
`direction_id` values (0 or 1) into the `references.trips` payload,
while `stopTripGroupings[].directionId` now emits the Java-convention
group IDs ("0" or "1" — inverted). That means a client walking
`stopTripGroupings → tripIds → references.trips[tripId].directionId`
sees two different notions of "direction ID" under the same name.

Java OBA uses the raw GTFS values in trip references too, so this
may match upstream — but it's worth an explicit comment (or, better,
a quick check against the Java response) because the dual use of
`directionId` is genuinely confusing to anyone reading the code
cold.

---

## Strengths

- Reverting `FormCombinedID` cleanly was the right call; ingestion-
  layer fix plan is better scoped and will be more robust.
- The `serviceIds` derivation from `trips` instead of the day-wide
  calendar is exactly what #790 asked for, and the loop is tight.
- `groupTripsByDirection` unit tests remain DAMP-named and
  single-invariant — nice.

---

## Recommended Action

1. Add HTTP-level assertions for `serviceIds` scoping and the
   `directionId=1 → "0"` mapping (important #1, #2).
2. Tighten the silent-failure logging paths (important #3).
3. Apply the smaller cleanups (#4–7) and consider the trip-ref
   DirectionID semantics (#8).
4. Re-request review when ready.

Ahmedhossamdev:
Hi @aaronbrethorst,
I'm still working on the other reviews

aaronbrethorst (approved):
Hi Ahmed — third time's the charm. Every item from the previous round
landed cleanly, and you went one step further: the new
`gtfsDirectionIDToDB` helper in `gtfsdb/helpers.go` fixes a real
underlying bug where the go-gtfs `DirectionID_False` enum (value `2`)
was being persisted as `2` in the database instead of the GTFS spec
value `0`. That bug was silently corrupting the `directionId` field on
many endpoints' trip references — the new HTTP test
`TestScheduleForRouteHandler_DirectionIDJavaParity` would not have
been able to assert "0"/"1" parity without it.

Tests pass, behavior changes are pinned at the response level, and the
silent-failure paths are now real `serverErrorResponse` returns. The
only `make lint` failures are three pre-existing `errcheck` warnings in
`gtfsdb/stops_rtree.go`, which this PR doesn't touch.

**Verdict: Merge**

---

## Critical Issues (0 found)

---

## Important Issues (0 found)

All three previous important items are resolved:

- **`serviceIds` HTTP regression test** — `TestScheduleForRouteHandler_ServiceIDsScopedToRoute`
  asserts the response contains exactly `c_868_b_79978_d_31` for
  route `25_1885`, locking in the route-scoped derivation. ✓
- **`directionId` Java-parity HTTP assertion** —
  `TestScheduleForRouteHandler_DirectionIDJavaParity` walks the
  response and asserts the group→CSV mapping
  (`group "0" ↔ direction_id "1"`, `group "1" ↔ direction_id "0"`),
  cross-checked against `references.trips[].directionId`. ✓
- **Silent-failure logging** — both
  `GetOrderedStopIDsForRouteDirection` and `GetStopTimesForTripIDs`
  now `return api.serverErrorResponse(...)` in
  `schedule_for_route_handler.go`, and the same tightening was
  propagated into `processTripGroups` in `stops_for_route_handler.go`
  (returning `error` and bubbling up through `processRouteStops`). ✓

---

## Suggestions (all addressed; one fit-and-finish nit)

All five previous suggestions landed:

- Variable shadowing → renamed to `seenDirSvcIDs`. ✓
- `combinedServiceIDs` initialization → `make([]string, 0, len(trips))`,
  no nil branch. ✓
- NULL `direction_id` test case → `TestGroupTripsByDirection_MixedNullAndValid`
  pins the NULL+0 collision behavior with a load-bearing comment. ✓
- Sort documentation → "Sort by trip ID so tripIds in the response is
  deterministic across runs." ✓
- DirectionID dual-semantics comment →
  `schedule_for_route_handler.go:280-282` makes the difference between
  group ID and reference `directionId` explicit. ✓

### One small fit-and-finish item (post-merge)

`gtfsDirectionIDToDB` in `gtfsdb/helpers.go:612` doesn't have a
dedicated unit test. All three branches are exercised end-to-end by
the new HTTP integration test, so coverage isn't actually missing —
but a 10-line table-driven test would future-proof the helper against
a casual edit (e.g. someone "simplifying" it back to
`toNullInt64(int64(d))` and reintroducing the `2`-instead-of-`0` bug).
Worth doing in a follow-up if you have a spare ten minutes; not
blocking.

---

## Strengths

- The underlying `gtfsDirectionIDToDB` fix is the kind of root-cause
  catch I love: the symptom showed up in `schedule-for-route`, but the
  bug was upstream in how go-gtfs's enum was being narrowed. The
  docstring on the helper explicitly names the `True=1, False=2` quirk
  so the next reader doesn't have to discover it the hard way.
- `TestGroupTripsByDirection_MixedNullAndValid` not only adds coverage
  but also documents *why* the NULL/0 collision happens via
  `sql.NullInt64` zero-value semantics. That's a great use of a test as
  documentation.
- Both error paths in `processTripGroups` and the schedule handler now
  bubble errors up cleanly instead of half-rendering a 200 with missing
  data — a meaningful reliability win.
- The reverted `FormCombinedID` work and the planned ingestion-layer
  follow-up for consolidated feeds (Sound Transit) is the right
  separation of concerns.

---

## Recommended Action

1. Optional follow-up: add a small unit test for `gtfsDirectionIDToDB`.
1. Land the consolidated-feed (Sound Transit) PR separately as
   discussed.

Nice work, Ahmed.

PR #796, opened 2026-03-26
Clarify VehiclesForAgencyID lock ownership and fix test locking

### LINKED ISSUE #795: Clarify static lock ownership for vehicles-for-agency and conformance tests
## Summary

Static GTFS lock ownership around `vehicles-for-agency` was inconsistent, and some tests were calling static manager accessors without the caller-held `RLock()` they require.

## Problem

`vehiclesForAgencyHandler` previously acquired `GtfsManager.RLock()` only for the initial `FindAgency` lookup and released it before continuing with the rest of the handler. `VehiclesForAgencyID` separately acquired the static lock internally.

That split lock ownership made the static read boundary unclear and inconsistent with the rest of the `gtfs.Manager` API, where static accessors generally expect the caller to already hold `RLock()`.

Tests had the same issue:

- Direct calls to `VehiclesForAgencyID` did not consistently hold `RLock()`
- `TestOpenAPIConformance_StaticEndpoints` was reading static manager data without first acquiring `RLock()`

## Expected Behavior

Static manager access should follow one consistent contract:

- Caller holds `manager.RLock()`
- Static accessor reads happen within that caller-owned lock scope

## Issue Solved

- `vehicles-for-agency` now uses one explicit caller-owned static read-lock scope
- `VehiclesForAgencyID` now follows the same caller-locking contract as other static manager accessors
- Missing `RLock()` coverage was added to `TestOpenAPIConformance_StaticEndpoints`

Adityatorgal17:
## Summary

Clarifies static lock ownership for `vehicles-for-agency` and fixes missing test-side locking for static GTFS manager access.

## Changes

- Hold `api.GtfsManager.RLock()` for the full duration of `vehiclesForAgencyHandler`
- Make `VehiclesForAgencyID` require caller-held `manager.RLock()`
- Update direct `VehiclesForAgencyID` test call sites to follow that contract
- Add missing `RLock()` in the OpenAPI conformance tests before static manager reads

## Why

Previously, `vehiclesForAgencyHandler` only held `RLock()` for the initial agency lookup, then released it before continuing with other static-dependent work such as `VehiclesForAgencyID` and DB-backed route/reference reads.

This split the handler's static reads across multiple lock windows instead of one clear snapshot boundary. During `ForceUpdate`, static data, lookup maps, indexes, and `GtfsDB` are swapped under the write lock, so different parts of a single response could otherwise be built from different static snapshots.

This change makes lock ownership explicit:

- The handler owns the static read lock for its full duration
- `VehiclesForAgencyID` assumes the caller already holds it
- Tests now follow the same contract

## Validation Passed
```
~ make test
```

### Fixes #795

### DIFF
--- a/internal/gtfs/concurrency_test.go
+++ b/internal/gtfs/concurrency_test.go
@@ -231,7 +231,9 @@ func TestConcurrentVehicleUpdates(t *testing.T) {
 				case <-done:
 					return
 				default:
+					manager.RLock()
 					_ = manager.VehiclesForAgencyID("test")
+					manager.RUnlock()
 					time.Sleep(time.Microsecond)
 				}
 			}
--- a/internal/gtfs/gtfs_manager.go
+++ b/internal/gtfs/gtfs_manager.go
@@ -629,19 +629,18 @@ func (manager *Manager) GetStopsForLocation(
 }
 
 // VehiclesForAgencyID returns all real-time vehicles serving routes that belong
-// to the given agency. It manages its own locking internally; callers must NOT
-// hold any Manager locks.
+// to the given agency.
+// IMPORTANT: Caller must hold manager.RLock() before calling this method.
+// The method acquires the real-time lock internally while reading vehicles.
 func (manager *Manager) VehiclesForAgencyID(agencyID string) []gtfs.Vehicle {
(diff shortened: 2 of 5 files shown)


fletcherw:
This PR is no longer relevant due to locking changes in GtfsManager.

Issue #747, opened 2026-03-19
Direction calculator: permanent negative cache on transient DB errors


aaronbrethorst:
## Problem

The `AdvancedDirectionCalculator` permanently caches empty-string direction results in `directionResults` (`sync.Map`) without distinguishing between "no data exists" (correct to cache) and "transient database error" (should not be cached).

### Scenario
1. A database hiccup occurs during direction computation for stop "123"
2. `computeFromShapes` returns `""` due to the DB error
3. `""` is stored permanently in `directionResults`
4. All future requests for stop "123" return `""` from cache — even after the DB recovers
5. The stop shows no direction for the lifetime of the application

### Related concerns

1. **Silent error in `computeFromShapes`**: When `calculateOrientationAtStop` returns a non-`sql.ErrNoRows` error (e.g., connection lost, SQLite busy), the error is silently discarded with a bare `continue` at line ~212. No logging, no visibility.

2. **`calculateOrientationAtStop` returns `(0, nil)` for insufficient shape points**: When `len(shapePoints) < 2` but `err == nil`, this returns `(0, nil)` — a zero-radian orientation that maps to "E" (East). The caller treats this as a valid orientation.

3. **Stale `queries` pointer after GTFS reload**: The `DirectionCalculator` is created once at startup with `gtfsManager.GtfsDB.Queries`. During `ForceUpdate`, `manager.GtfsDB` is replaced, but the calculator's `queries` pointer is never updated.

### Suggested fixes

- Have `computeFromShapes` return `(string, error)` to distinguish "no data" from "transient error"
- Only cache results when no transient error occurred
- Log non-`sql.ErrNoRows` errors in the orientation calculation loop
- Return `sql.ErrNoRows` (not `nil`) when shape points are insufficient

Discovered during review of #647.

PR #797, opened 2026-03-26
fix: resolve direction calculator caching bugs and prevent stale queries on hot-swap

3rabiii:
### Description
This PR addresses several reliability and correctness bugs within the `AdvancedDirectionCalculator`. Previously, transient database errors could permanently poison the direction cache, insufficient shape points defaulted to an invalid "East" orientation, and the calculator retained a stale database pointer after a GTFS hot-swap.

### Changes Included
* **Prevent Transient Error Caching:** `computeFromShapes` now returns `(string, error)`. The loop tracks transient DB errors (`lastTransientErr`) and propagates them to the caller, ensuring we only cache valid directions or legitimate "no data" states, avoiding permanent cache poisoning during DB hiccups.
* **Fix Bogus Orientations:** Explicitly return `sql.ErrNoRows` when there are fewer than 2 shape points. This prevents the caller from incorrectly interpreting a `nil` error as a valid `0` radian (East) orientation.
* **Add Error Visibility:** Added `slog.Warn` for non-`ErrNoRows` errors in the orientation calculation loop to surface silent failures like DB connection drops or timeouts.
* **Resolve Stale Queries on Hot-Swap:** - Added a `sync.RWMutex` to protect the `queries` pointer in `AdvancedDirectionCalculator`.
   - Added an `UpdateQueries()` method to atomically swap the DB pointer and safely evict the entire direction result cache.
   - Wired the calculator into `gtfs.Manager` during startup so that `ForceUpdate` can automatically trigger the refresh after a successful GTFS reload.

### Verification
image

fixes: #747

### DIFF
--- a/internal/gtfs/spatial_index_test.go
+++ b/internal/gtfs/spatial_index_test.go
@@ -42,7 +42,7 @@ func TestQueryStopsInBounds(t *testing.T) {
 		name          string
 		bounds        utils.CoordinateBounds
 		expectedCount int
-		expectedIDs   []string 
+		expectedIDs   []string
 	}{
 		{
 			name: "NormalBoundingBox_SomeStops",
--- a/internal/restapi/trips_for_location_handler.go
+++ b/internal/restapi/trips_for_location_handler.go
@@ -708,7 +708,6 @@ func (rb *referenceBuilder) createRoute(route gtfsdb.Route) models.Route {
 
 }
 
-
 func (rb *referenceBuilder) buildTripReferences() error {
 	rb.tripsRefList = make([]models.Trip, 0, len(rb.presentTrips))
 
--- a/internal/restapi/trips_for_route_handler.go
+++ b/internal/restapi/trips_for_route_handler.go
@@ -635,7 +635,7 @@ func buildTripReferences(
 					TripHeadsign:  trip.TripHeadsign,
 					TripShortName: trip.TripShortName,
 					DirectionID:   trip.DirectionID,
-					BlockID: utils.FormCombinedID(currentAgency, trip.BlockID),
+					BlockID:       utils.FormCombinedID(currentAgency, trip.BlockID),
 					ShapeID:       utils.FormCombinedID(currentAgency, trip.ShapeID),
 					PeakOffPeak:   0,
 					TimeZone:      "",


internal/gtfs/advanced_direction_calculator.go
```diff
@@ -45,6 +48,22 @@ func NewAdvancedDirectionCalculator(queries *gtfsdb.Queries) *AdvancedDirectionC
...
+// Call this after a GTFS hot-swap so the calculator queries the new database.
+// It also clears the direction result cache so stale entries from the old database
+// are not served.
+func (adc *AdvancedDirectionCalculator) UpdateQueries(queries *gtfsdb.Queries) {
+	adc.queriesMu.Lock()
+	adc.queries = queries
+	adc.queriesMu.Unlock()
+
+	// Evict all cached directions so they are recomputed against the new DB.
+	adc.directionResults.Range(func(key, _ any) bool {
```
fletcherw: This can be written more simply as adc.directionResults.Clear().

internal/gtfs/advanced_direction_calculator.go
```diff
@@ -100,10 +119,14 @@ func (adc *AdvancedDirectionCalculator) CalculateStopDirection(ctx context.Conte
...
-		// Store in sync.Map for all future requests
-		adc.directionResults.Store(stopID, computedDir)
+		// Only cache when there was no transient error. A transient error (e.g. DB
+		// connection lost) must not permanently poison the cache; omitting it here
+		// means the next request will retry the DB.
+		if err == nil {
+			adc.directionResults.Store(stopID, computedDir)
+		}
 
 		return computedDir, nil
```
fletcherw: here, even if computeFromShapes returned an err, you're returning `computedDir` and a `nil` error. If the intention is to ignore errors, you should leave a comment here explaining why it's safe to do so.

internal/gtfs/advanced_direction_calculator.go
```diff
@@ -25,11 +25,14 @@ const (
...
 	shapeCache                 map[string][]gtfsdb.GetShapePointsWithDistanceRow // Cache of all shape data for bulk operations
 	initialized                atomic.Bool                                       // Tracks whether concurrent operations have started
-	cacheMutex                 sync.RWMutex                                      // Protects map access
+	cacheMutex                 sync.RWMutex                                      // Protects shapeCache map access
 	// directionResults caches computed stop directions.
+	// Only non-error results are cached; transient DB errors are never stored so that
+	// a recovered database will be retried on the next request.
 	// Lifecycle note: This map grows indefinitely for the lifetime of the application.
 	// Unbounded growth is acceptable here because it is strictly bounded by the finite
 	// number of valid real-world stops, and computed directions remain stable across GTFS reloads.
```
fletcherw: This comment is out-of-date now. Either don't clear the directions on reload or update the comment.

internal/gtfs/advanced_direction_calculator_test.go
fletcherw: is there an easy  way you can add a test for the new behavior you've added, verifying that a returned error value isn't cached?

PR #801, opened 2026-03-28
fix: prevent timer leak in GTFS retry loop

ashnaaseth2325-oss:
##  SUMMARY

This PR fixes a timer lifecycle issue in `InitGTFSManager` where retry backoff used `time.After`, which cannot be stopped on context cancellation. It replaces those calls with `time.NewTimer` to ensure proper cleanup.

The changes affect `internal/gtfs/gtfs_manager.go`, specifically the retry sleep logic in the GTFS initialization flow. 

---

## FIX

```go
// Before
select {
case <-ctx.Done():
    return nil, ctx.Err()
case <-time.After(delay):
}

// After
retryTimer := time.NewTimer(delay)
select {
case <-ctx.Done():
    retryTimer.Stop()
    return nil, ctx.Err()
case <-retryTimer.C:
}
```

---

##  VERIFICATION 

Test by starting the service with a failing GTFS endpoint so it enters the retry loop, then cancel the context (e.g., send SIGTERM) while it is sleeping. The function should exit immediately with `ctx.Err()` instead of waiting for the delay. Retry behavior should remain unchanged when the context is not cancelled, with no lingering timers.

### DIFF
--- a/internal/gtfs/gtfs_manager.go
+++ b/internal/gtfs/gtfs_manager.go
@@ -170,10 +170,12 @@ func InitGTFSManager(ctx context.Context, config Config) (*Manager, error) {
 					)
 
 					// Cancellable sleep
+					retryTimer := time.NewTimer(delay)
 					select {
 					case <-ctx.Done():
+						retryTimer.Stop()
 						return nil, ctx.Err()
-					case <-time.After(delay):
+					case <-retryTimer.C:
 					}
 					continue
 				}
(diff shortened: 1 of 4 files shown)


ashnaaseth2325-oss:
Hello @aaronbrethorst 
Kindly review this PR whenever you have time.

aaronbrethorst:
My apologies for not reviewing this earlier. Can you fix the merge conflicts and then reopen?

PR #802, opened 2026-03-28
fix: propagate ctx to gtfstidy subprocess

ashnaaseth2325-oss:
##  SUMMARY

This PR fixes a context propagation issue in the GTFS tidy pipeline where subprocesses could not be canceled, leading to hangs during shutdown or timeout. It updates `tidyGTFSData` to accept a `context.Context` and uses `exec.CommandContext` instead of `exec.Command`.

Primary changes are in `internal/gtfs/tidy.go`, with call-site updates in `internal/gtfs/static.go` and test adjustments in `internal/gtfs/tidy_test.go`.

---

##  FIX

```go
// Before
func tidyGTFSData(inputZip []byte, logger *slog.Logger) ([]byte, error) {
    cmd := exec.Command(gtfstidyPath, "-OscRCSmeD", "-o", outputZipPath, pristineZipPath)
}

// After
func tidyGTFSData(ctx context.Context, inputZip []byte, logger *slog.Logger) ([]byte, error) {
    cmd := exec.CommandContext(ctx, gtfstidyPath, "-OscRCSmeD", "-o", outputZipPath, pristineZipPath)
}
```

---

##  VERIFICATION

Tested by enabling GTFS tidy, triggering a feed update, and sending a SIGTERM while the tidy process was running. The service now shuts down cleanly without hanging, and the subprocess exits immediately on cancellation. Builds and existing tests pass with no regressions.

### DIFF
--- a/internal/gtfs/static.go
+++ b/internal/gtfs/static.go
@@ -69,7 +69,7 @@ func rawGtfsData(ctx context.Context, source string, isLocalFile bool, config Co
 	// Process through gtfstidy if enabled
 	if config.EnableGTFSTidy {
 		logging.LogOperation(logger, "gtfstidy_enabled_processing_gtfs_data")
-		tidiedData, err := tidyGTFSData(b, logger)
+		tidiedData, err := tidyGTFSData(ctx, b, logger)
 		if err != nil {
 			logging.LogError(logger, "Failed to tidy GTFS data, using original data", err)
 		} else {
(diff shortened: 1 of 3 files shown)


ashnaaseth2325-oss:
Hello @fletcherw,
Just a follow up on this PR whenever you have time.
Thanks!

fletcherw (approved):
Thanks for the fix!

PR #803, opened 2026-03-28
fix: propagate context for cancellable imports

ashnaaseth2325-oss:
##  SUMMARY

This PR fixes a context propagation issue in the GTFS import flow that prevented database operations from being cancellable during shutdown. The change ensures the caller’s `context.Context` is properly passed through instead of being overridden.

Primary changes are in `gtfsdb/helpers.go` and `gtfsdb/client.go`, specifically around `processAndStoreGTFSDataWithSource` and its callers.

---

## FIX

```go
// Before 
func (c *Client) processAndStoreGTFSDataWithSource(b []byte, source string) error {
    ctx := context.Background()
}

// After 
func (c *Client) processAndStoreGTFSDataWithSource(ctx context.Context, b []byte, source string) error {
}
```

---

## VERIFICATION

Tested by triggering a GTFS import and sending a SIGTERM while it was still processing. Before this change, the app would hang for minutes waiting for the DB writes to finish. Now, the import cancels immediately and the service shuts down cleanly without delay.

### DIFF
--- a/gtfsdb/conditional_import_test.go
+++ b/gtfsdb/conditional_import_test.go
@@ -52,11 +52,11 @@ func TestConditionalImport_InitialImport(t *testing.T) {
 	require.NoError(t, err, "Failed to create client")
 	defer func() { _ = client.Close() }()
 
-	ctx := context.Background()
+	ctx := t.Context()
 	originalData, _ := createTestData(t)
 
 	// Perform initial import
-	err = client.processAndStoreGTFSDataWithSource(context.Background(), originalData, "test-source")
+	err = client.processAndStoreGTFSDataWithSource(originalData, "test-source")
 	require.NoError(t, err, "Initial import should succeed")
 
 	// Verify metadata was stored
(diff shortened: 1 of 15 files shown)


ashnaaseth2325-oss:
Hello @aaronbrethorst @fletcherw,
Just checking in on this whenever you have time. 
Thanks!

gtfsdb/conditional_import_test.go
```diff
@@ -56,7 +56,7 @@ func TestConditionalImport_InitialImport(t *testing.T) {
 	originalData, _ := createTestData(t)
 
 	// Perform initial import
-	err = client.processAndStoreGTFSDataWithSource(originalData, "test-source")
+	err = client.processAndStoreGTFSDataWithSource(context.Background(), originalData, "test-source")
```
fletcherw: can you use `t.Context()` in this file and all the other test files please instead of `context.Background()`

fletcherw (commented):
Hey Ashnaa, this is a good fix, thanks for making it.

ashnaaseth2325-oss:
Hi @fletcherw! 
I’ve updated the PR to use `t.Context()` across all test files as suggested and resolved the merge conflicts.
I also made an attempt to address the CI failures, but it seems there may still be something I’m overlooking. Could you please take a look when you have a moment?
Thanks.

fletcherw:
It looks like in the process of fixing merge conflicts you dropped the change to processAndStoreGTFSDataWithSource. You've also re-added a few tests that were deleted in other PRs.

ashnaaseth2325-oss:
Hi, thanks for taking the time to review this PR and for the guidance.
I’m going to close this for now. I wasn’t able to get everything aligned cleanly with the latest changes, and I don’t want to leave the PR in a messy state.
I’ll revisit this with a fresh branch later and open a new PR if needed.
Thanks again for your help!

PR #804, opened 2026-03-28
fix: register missing Situation endpoint route (#798)

### LINKED ISSUE #798: [Parity Gap] Orphaned handler: Missing route registration for Situation endpoint (/api/where/situation/{id}.json)
### Describe the bug / gap
While auditing the REST API surface area for parity between the legacy Java server (`onebusaway-api-webapp`) and Maglev, I noticed that a `404 Not Found` is returned when trying to fetch a specific service alert via the standalone Situation endpoint.

Upon further inspection of the codebase, the handler logic actually exists (`internal/restapi/situation_handler.go`) and has tests (`situation_handler_test.go`), but it was never wired up in `internal/restapi/routes.go`. 

### Expected Behavior
A `GET` request to `/api/where/situation/{id}.json` should invoke the existing `situationHandler` and return the specific situation matching the ID, instead of a `404`.

### Proposed Solution
I will register the missing route `GET /api/where/situation/{id}.json` in `routes.go` to point to the existing handler, and verify the output schema against the legacy Java server using `maglev-validator`. I'll open a PR for this shortly! 🚀

tejasva-vardhan:
### Overview
This PR resolves #798 by properly registering the `GET /api/where/situation/{id}` route in the REST API. While the handler logic was present, it was not wired to the router, causing a 404/HTML fallback when requesting real-time service alerts.

### Changes
- **Route Registration:** Registered the missing endpoint in `internal/restapi/routes.go`.
- **Logic Implementation:** Re-implemented the `situationHandler` to correctly fetch and format service alerts using `api.extractAndValidateID`.
- **Data Access:** Added `GetAllAlerts()` in `internal/gtfs/realtime.go` to provide access to the in-memory GTFS-RT alert store.
- **Testing:** Added unit tests in `internal/restapi/situation_handler_test.go` (verified with `go test`).

### Verification (Before vs After)

To ensure full parity with the legacy OBA Java API, I verified the fix using both the `maglev-validator` and manual browser testing.

#### 1. Maglev Validator Comparison
Local Maglev (`localhost:4000`) vs Production Java API (`api.pugetsound.onebusaway.org`).

| Stage | Comparison Result |
| :--- | :--- |
| **Before (Gap)** | Screenshot 2026-03-29 020453 |
| **After (Fixed)** | Screenshot 2026-03-29 021558 |

#### 2. Browser Manual Testing
Endpoint: `/api/where/situation/1_84714.json?key=test`

| Stage | Response Screenshot |
| :--- | :--- |
| **Before (404)** | Screenshot 2026-03-29 013458 |
| **After (200 OK)** | Screenshot 2026-03-29 021511 |

---
**Note for Reviewers:** - Only core logic files for the fix are included in this PR. 
- Windows-specific driver changes (`modernc.org/sqlite`) and local performance bypasses were intentionally excluded to maintain upstream standards.

## Summary by CodeRabbit

* **New Features**
  * Added an API endpoint for retrieving individual service alerts by ID.
  * Alerts are consolidated across realtime feeds, with duplicates and alerts missing IDs excluded.
  * Alert responses include localized summaries, descriptions, references, and metadata.
  * Supports alert IDs with agency prefixes.

* **Bug Fixes**
  * Added validation for invalid alert IDs and API credentials.
  * Requests for unavailable alerts now return a not-found response.


tejasva-vardhan:
@aaronbrethorst @fletcherw @Ahmedhossamdev @burma-shave   would love any review on this!!

aaronbrethorst:
### Code review

Found 2 issues:

1. The branch is stale and no longer compiles against `main` — the green checks are from March and were run against an older base. Two symbols this PR depends on have changed since: `sort` was dropped from `internal/gtfs/realtime.go`'s import block in favor of `slices` (PR #809, "refactor: replace sort package with slices"), so `sort.Strings(feedIDs)` is now `undefined: sort`; and `Manager.AddTestAlert` was renamed to `AddAlertForTest` (`internal/gtfs/gtfs_manager.go:819`), so `situation_handler_test.go:45` no longer resolves. GitHub still reports the PR as MERGEABLE/CLEAN, so merging as-is would break the build on `main`. Needs a rebase plus `slices.Sort(feedIDs)` and `AddAlertForTest`.

https://github.com/OneBusAway/maglev/blob/c9cfa016fed016cb5e4cd487423ee827710f4c38/internal/gtfs/realtime.go#L213-L215

https://github.com/OneBusAway/maglev/blob/c9cfa016fed016cb5e4cd487423ee827710f4c38/internal/restapi/situation_handler_test.go#L44-L46

2. The `len(situations) == 0` guard is unreachable. `BuildSituationReferences` (`internal/restapi/reference_utils.go:136`) appends unconditionally once per input alert, so a one-element input always yields a one-element result. The 500 path can never be taken and has no test. (CONTRIBUTING.md says "Every new branch or condition introduced by a PR needs test coverage — untested code paths and leftover dead code are common review findings.")

https://github.com/OneBusAway/maglev/blob/c9cfa016fed016cb5e4cd487423ee827710f4c38/internal/restapi/situation_handler.go#L32-L38

aaronbrethorst (changes requested):
Thanks for this, and sorry it sat for so long — that's on us, not you. The endpoint design is right: the handler shape matches our other simple-ID real-time endpoints, `extractAndValidateID` handles the `.json` suffix the way the rest of the API does, and putting `GetAllAlerts` on the manager with dedup-by-ID across feeds under the read lock is the correct place for that logic.

The blocker is that the branch has gone stale, and in a way that GitHub is actively hiding from us. The PR shows as mergeable with green checks, but those checks ran on 2026-03-28 against a `main` that no longer exists. There's no textual conflict, so nothing flags it — but merging as-is breaks the build in two places:

**1. `sort` is no longer imported in `internal/gtfs/realtime.go`.**

`GetAllAlerts` calls `sort.Strings(feedIDs)`. Since this branch was opened, #809 ("refactor: replace sort package with slices") removed the `sort` import from that file — `main` now uses `slices.Sort` throughout (three call sites). Because your diff only adds a function body and doesn't touch the import block, the merge result is a file that calls `sort.Strings` with no `sort` import: `undefined: sort`.

Swap it for `slices.Sort(feedIDs)` to match the surrounding code.

**2. `AddTestAlert` no longer exists.**

`situation_handler_test.go` calls `api.GtfsManager.AddTestAlert(alert)`. On `main` that helper is named `AddAlertForTest` (`internal/gtfs/gtfs_manager.go:819`); `AddTestAlert` isn't defined anywhere in the repo now. Renaming the call is the whole fix.

A rebase onto `main` plus those two changes should get CI running against reality again.

Two smaller things while you're in there:

- The `len(situations) == 0` guard in `situationHandler` is unreachable. `BuildSituationReferences` appends exactly one `Situation` per input alert unconditionally, so a one-element input always yields one element. It's dead code and it's the one branch with no test — I'd just drop it.
- `TestSituationHandlerWithSituation` builds the alert with `Header` and `Description` but only asserts on `id`, `reason`, and `severity`. The translated-string fields are the ones most likely to regress silently; asserting `summary` and `description` would make the test earn its keep.

One note that isn't about your code: issue #798 claims the handler and tests already existed and just needed wiring up. That isn't true of `main` — neither `situation_handler.go` nor its test exists there, and this PR creates both. So this is a new endpoint implementation, not a one-line route registration. Worth knowing since it means there's no prior art to match, and also worth flagging that we have no spec for this endpoint: it isn't in the maglev wiki, it isn't in `testdata/openapi.yml`, and the docs repo only describes `situation` as a referenced *element*, not a standalone method. I'm comfortable with the envelope you chose, but if you have a reference response from a production OBA server, that'd be good to capture.

Related: `references` is always empty here, even though a situation's `allAffects` can carry agency/route/stop/trip IDs that a client would want resolved. Not a blocker given there's no spec to point at, but worth a thought.

Rebase and fix the two build breaks and I'll take another look promptly.

tejasva-vardhan:
@aaronbrethorst 
Rebased onto current `main` and addressed the review:

- `slices.Sort(feedIDs)` in `GetAllAlerts`
- `AddAlertForTest` in the handler test
- Dropped the unreachable `len(situations) == 0` guard
- Asserted `summary` and `description` on the 200 path

aaronbrethorst:
### Code review

aaronbrethorst (changes requested):
You cleared every blocker from my July 30 review, and thank you for the patience
on a PR that's been open this long.

Confirming each one:

- **Rebased onto current main** — base is `85548249`, and all CI is green on the
  head SHA, so the "green checks against a dead main" problem is gone.
- **`sort.Strings` → `slices.Sort`** — done, and `slices` was already imported.
- **`AddTestAlert` → `AddAlertForTest`** — done, and it writes into
  `feedAlerts["_test"]`, which is exactly the map `GetAllAlerts` reads.
- **Unreachable empty-situation guard removed** — correct to drop it;
  `BuildSituationReferences` appends exactly one `Situation` per input alert, so
  a one-element input always yields one element and `situations[0]` is safe.
- **`summary`/`description` assertions** — added, including the `lang` sub-fields.

The third commit is the one I want to single out. Noticing that
`route_handler_test.go` already seeds `test-alert-123` into the shared
package-level GTFS manager, and that `GetAllAlerts` keeps the first ID so your
test was asserting against someone else's alert — that's a genuinely subtle
cross-test interaction, and you found and fixed it without being asked. That's
the kind of thing that would have been a confusing flake for whoever hit it next.

The rest holds up too: the route sits in the right block in `routes.go` with
`rateLimitAndValidateAPIKey` and short cache duration matching its
`vehicles-for-agency` neighbour, simple-ID validation is right because situation
IDs are raw `alert.ID` values everywhere else in the codebase, `GetAllAlerts`
takes `RLock` with an immediate `defer`, and the handler reuses
`extractAndValidateID` / `BuildSituationReferences` / `NewEntryResponse` rather
than hand-rolling any of it.

**One blocker, and it's mechanical.** The tip commit `185369a0` carries
`Co-authored-by: Cursor <cursoragent@cursor.com>`. CONTRIBUTING.md prohibits
attributing commits to a coding agent — the rule names Claude "or similar tools",
and Cursor is squarely that. Since we merge rather than squash, it lands in
`main` permanently.

The good news is it's only on the tip, so:

```
git commit --amend
git push --force-with-lease
```

leaves `695da72f` and `f20054e3` untouched — the commits I actually reviewed keep
their SHAs, so this doesn't rewrite reviewed history in the way CONTRIBUTING
warns about.

Heads up that I've asked the same on your #1346 and #1336 today, so it's worth
turning off whatever Cursor setting is adding the trailer rather than fixing it
three times.

Two things I'm explicitly *not* blocking on, same as last time: the `references`
block is still always empty even though `allAffects` carries IDs a client would
want resolved, and there's still no `/situation/{id}` entry in
`testdata/openapi.yml` to validate the envelope against. The second one is mine
to fix upstream, not yours.

Amend the commit and I'll merge this straight away.

internal/restapi/situation_handler.go
```diff
@@ -0,0 +1,36 @@
...
+func (api *RestAPI) situationHandler(w http.ResponseWriter, r *http.Request) {
+	situationID, ok := api.extractAndValidateID(w, r)
+	if !ok {
+		return
+	}
+
+	var alert gtfs.Alert
+	found := false
+	for _, candidate := range api.GtfsManager.GetAllAlerts() {
+		if candidate.ID == situationID {
```
burma-shave: This only matches the alert's raw, unprefixed ID. Legacy Java always looks situations up by the agency-qualified `AgencyAndId` form (`AgencyAndIdLibrary.convertFromString`, splitting the path param on the first `_`), and this codebase already has that exact convention for situations — `situationID()`/`agencyIDForAlert()` in `internal/restapi/reference_utils.go:640-660`. It's what `trip-details` and `trip-for-vehicle` already use to build the `situationIds` a client would plausibly follow into this endpoint. Right now, on a GTFS-RT feed whose alert IDs aren't already self-prefixed with the agency ID, a client following a `situationId` from `trip-details` would 404 here even though the situation exists.

Could this loop also match against `situationID(candidate.ID, agencyIDForAlert(candidate, ""))` alongside the raw ID, so it resolves both conventions? It's a small reuse of existing helpers rather than new logic.

internal/gtfs/realtime.go
```diff
@@ -201,6 +201,36 @@ func (manager *Manager) GetAlertsForStop(stopID string) []gtfs.Alert {
...
+	slices.Sort(feedIDs)
+
+	seen := make(map[string]struct{})
+	alerts := make([]gtfs.Alert, 0)
+	for _, feedID := range feedIDs {
+		for _, alert := range manager.feedAlerts[feedID] {
+			if alert.ID == "" {
+				continue
+			}
+			if _, exists := seen[alert.ID]; exists {
```
burma-shave: Same root cause as the ID-matching comment on `situation_handler.go`: this dedups across feeds by the raw `alert.ID`, so two different `gtfs-rt-feeds` publishing alerts that happen to share a raw ID would silently drop one. Worth keying this the same way once the lookup above moves to the agency-qualified ID — `situationID(alert.ID, agencyIDForAlert(alert, ""))` — so two agencies' alerts can't collide on ID.

internal/restapi/situation_handler.go
```diff
@@ -0,0 +1,36 @@
...
+	"net/http"
+
+	"github.com/OneBusAway/go-gtfs"
+	"maglev.onebusaway.org/internal/models"
+)
+
+// situationHandler serves a single GTFS-RT service alert (OneBusAway "Situation")
+// by its alert id.
+func (api *RestAPI) situationHandler(w http.ResponseWriter, r *http.Request) {
+	situationID, ok := api.extractAndValidateID(w, r)
```
burma-shave: `extractAndValidateID` only checks that the ID is non-empty, within length, and made of allowed characters — it doesn't require the `_` agency separator, so a malformed ID (no underscore) falls through the same not-found path as a well-formed-but-unknown one, and this endpoint always returns 404 for both.

Legacy Java's malformed-ID case is actually broken there (an uncaught `IllegalStateException` crashes to an HTTP 200 null body — documented as DC-1 in maglev.wiki), and several other Maglev endpoints (`stop`, `stops-for-agency`, `schedule-for-stop`, etc.) already fix that same defect class by splitting the two cases: malformed ID → 400, unknown-but-valid ID → 404. Could this endpoint follow the same split rather than collapsing both into 404? I've added `situation` to DC-1's tracker in maglev.wiki with that as the intended behavior.

internal/restapi/situation_handler_test.go
```diff
@@ -0,0 +1,87 @@
...
+	api := createTestApi(t)
+	defer api.Shutdown()
+
+	resp, model := serveApiAndRetrieveEndpoint(t, api, "/api/where/situation/nonexistent-alert.json?key=TEST")
+	assert.Equal(t, http.StatusNotFound, resp.StatusCode)
+	assert.Equal(t, http.StatusNotFound, model.Code)
+	assert.Equal(t, "resource not found", model.Text)
+}
+
+func TestSituationHandlerWithSituation(t *testing.T) {
```
burma-shave, line 62: A couple of coverage gaps worth closing given CONTRIBUTING.md's expectations for new branches: no test exercises `GetAllAlerts()`'s cross-feed dedup (two feeds emitting alerts that share an ID), and none test the PR description's claim that alerts with no ID are excluded — `GetAllAlerts()`'s `if alert.ID == "" { continue }` branch has no coverage. Whatever tests get added for the ID-matching and 400-vs-404 fixes above would be a natural place to also cover these.

burma-shave (changes requested):
Registering this route and filling in the handler/tests is the right shape overall — it follows the pattern of the other simple-ID real-time endpoints, and reuses `extractAndValidateID`/`BuildSituationReferences`/`NewEntryResponse` rather than hand-rolling any of it. A few things worth addressing before merge, left as inline comments: the situation-ID lookup convention, the same convention in `GetAllAlerts()`'s dedup key, splitting malformed-ID (400) from unknown-ID (404) responses, and two test-coverage gaps. None of these are urgent for existing clients — nothing calls this endpoint directly today — but they're straightforward fixes that avoid baking in a gap other tooling or future clients would eventually hit. (I've written up the situation endpoint and the malformed-ID/DC-1 decision in maglev.wiki, since neither existed there before this PR.)

tejasva-vardhan:
@aaronbrethorst @burma-shave pushed the remaining review items.

- Situation lookup matches both the raw GTFS-RT alert ID and `situationID(candidate.ID, agencyIDForAlert(candidate, ""))`, so IDs from trip-details `situationIds` resolve.
- `GetAllAlerts` now dedups on that same agency-qualified ID, so two agencies can share a raw ID without one being dropped.
- Malformed IDs (no `_`) return 400; a well-formed unknown ID returns 404.
- Tests cover prefixed vs unprefixed lookup, 400 vs 404, empty-ID skip, and cross-feed first-seen dedup.

Left empty `references` and the OpenAPI path as discussed — those are upstream / not blocking.

burma-shave:
Thanks for the work on this and for digging into the parity gap.

After looking at it again, I’m going to close this without merging. While the legacy API exposes `/api/where/situation/{id}.json`, this is not a documented public endpoint in the upstream OpenAPI spec, and I don’t see evidence that our current clients need or use it. The alert-producing endpoints already return full situation objects in `data.references.situations` alongside any `situationIds` in the normal/default response, and I couldn’t find usage of the standalone endpoint in Wayfinder, the JS SDK, iOS, or Android.

Given that, I don’t want to increase Maglev’s maintained API surface area for an undocumented endpoint with no demonstrated client demand. 

I’ll close this for now.

PR #806, opened 2026-03-29
chore: automate sqlite driver selection via build tags (#805)

### LINKED ISSUE #805: chore: add pure-Go SQLite support via build tags for Windows compatibility
### Problem
Setting up Maglev on Windows currently requires a complex CGO toolchain (GCC/MinGW) because of the `mattn/go-sqlite3` dependency. This creates a high barrier to entry for new contributors on Windows.

### Proposed Solution
I propose adding a pure-Go SQLite driver (`modernc.org/sqlite`) as an alternative using Go build tags. 
- Create `driver_cgo.go` (with `//go:build !windows || cgo` tag).
- Create `driver_pure.go` (with `//go:build windows && !cgo` tag).

This ensures that:
1. Production/Linux builds remain unaffected (still using CGO).
2. Windows developers can run `go run ./cmd/api` without any extra setup.

I have already verified this locally and would like to submit a PR if this approach aligns with the project goals.

tejasva-vardhan:
## Overview
This PR automates the SQLite driver selection for Maglev using Go build tags, directly addressing and resolving #805. It removes the friction for Windows developers who previously had to manually pass the `-tags purego` flag to build or run the project.

## Changes Made
* **Added `driver_pure.go`:** Uses the `//go:build windows && !cgo` tag to automatically load `modernc.org/sqlite` on Windows machines without CGO.
* **Added `driver_cgo.go`:** Uses the `//go:build !windows || cgo` tag to preserve the high-performance `mattn/go-sqlite3` driver for Linux/macOS and explicit CGO builds.
* **Updated Dependencies:** Ensured `go.mod` and `go.sum` correctly track both drivers.
* *(Optional: If you included `helpers.go`)* **Refactoring:** Replaced hardcoded `"sqlite3"`/`"sqlite"` strings with a cross-platform `DriverName` constant.

## Verification
* [x] **Windows (No CGO):** Verified that running `go run ./cmd/api` works out-of-the-box without GCC/MinGW toolchains.
* [x] **Linux/CGO Default:** Verified that the build still defaults to the CGO driver for non-Windows environments to prevent any performance regressions.

Closes #805

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,14 +5,19 @@ go 1.24.2
 require (
 	github.com/OneBusAway/go-gtfs v1.1.1
 	github.com/davecgh/go-spew v1.1.1
+	github.com/getkin/kin-openapi v0.133.0
 	github.com/google/uuid v1.6.0
+	github.com/hashicorp/golang-lru/v2 v2.0.7
 	github.com/klauspost/compress v1.18.0
 	github.com/mattn/go-sqlite3 v1.14.24
 	github.com/prometheus/client_golang v1.23.2
 	github.com/stretchr/testify v1.11.1
 	github.com/tidwall/rtree v1.10.0
 	github.com/twpayne/go-polyline v1.1.1
+	golang.org/x/sync v0.16.0
 	golang.org/x/time v0.12.0
+	google.golang.org/protobuf v1.36.8
+	modernc.org/sqlite v1.38.2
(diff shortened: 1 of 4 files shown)


aaronbrethorst:
Please rebase against main and fix all merge conflicts, and then reopen this PR.

PR #809, opened 2026-04-03
refactor: replace sort package with slices for improved sorting performance

Not-Dhananjay-Mishra:
### Changes

- Replaced `sort.Ints`, `sort.Slice`, `sort.Float64s` and `sort.Strings` usage with `slices.Sort` and `slices.SortFunc` for sorting operations.

### Improvement

- Adopted `slices.Sort`, which is the idiomatic approach in modern Go (introduced in Go 1.21).
- Reduced heap allocations due to more efficient implementation.
- Aligned with Go 1.22+ behavior, where `sort.Ints` internally delegates to `slices.Sort`  (see https://pkg.go.dev/sort#Ints).
- Improved code readability and future compatibility with the standard library.


fletcherw:
Thanks for the fix, this makes the code a little clearer and potentially faster which is great.

PR #811, opened 2026-04-06
refactor: use db query to fetch routes for agency

fletcherw:
Instead of building an in-memory index use the existing route information in the SQL database as well as the existing index on agency ID to query routes by agency.

### DIFF
--- a/gtfsdb/db.go
+++ b/gtfsdb/db.go
@@ -216,6 +216,9 @@ func Prepare(ctx context.Context, db DBTX) (*Queries, error) {
 	if q.getRoutesByIDsStmt, err = db.PrepareContext(ctx, getRoutesByIDs); err != nil {
 		return nil, fmt.Errorf("error preparing query GetRoutesByIDs: %w", err)
 	}
+	if q.getRoutesForAgencyStmt, err = db.PrepareContext(ctx, getRoutesForAgency); err != nil {
+		return nil, fmt.Errorf("error preparing query GetRoutesForAgency: %w", err)
+	}
 	if q.getRoutesForStopStmt, err = db.PrepareContext(ctx, getRoutesForStop); err != nil {
 		return nil, fmt.Errorf("error preparing query GetRoutesForStop: %w", err)
 	}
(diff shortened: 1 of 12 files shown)


aaronbrethorst (changes requested):
Hey Fletcher, thanks for taking the time to make this refactoring to move route-for-agency lookups from the in-memory index to SQL. Removing the `routesByAgencyID` map and leaning on the existing database index is a nice simplification — it's one less piece of state to rebuild during hot-swaps, and the indentation fixes in the direction calculator tests are appreciated too. Before we can merge this, I will need you to make a couple changes:

## Critical

1. **Race condition in `VehiclesForAgencyID`** — `internal/gtfs/gtfs_manager.go:635`

   The old `VehiclesForAgencyID` explicitly acquired `staticMutex.RLock()` before reading routes, then released it before touching real-time data. The refactored version removes that lock entirely but still accesses `manager.GtfsDB.Queries` (via `RoutesForAgencyID`) — and this PR's own diff updates the `staticMutex` comment to say it "Protects GtfsDB, gtfsData, and lastUpdated." During a `ForceUpdate` hot-swap, `manager.GtfsDB` is reassigned under the write lock (`static.go:364`). Without the read lock, `VehiclesForAgencyID` has a data race on the `GtfsDB` pointer.

   The simplest fix is to snapshot the DB reference under the lock:

   ```go
   manager.staticMutex.RLock()
   routes, err := manager.RoutesForAgencyID(ctx, agencyID)
   manager.staticMutex.RUnlock()
   if err != nil {
       return nil, err
   }
   ```

   Or, if you'd prefer to drop the lock requirement from `RoutesForAgencyID` entirely (since DB access without locks is already a pattern in some handlers), that's a valid approach too — but then `RoutesForAgencyID`'s doc comment needs to be updated and the `staticMutex` comment about protecting `GtfsDB` needs to be reconciled.

## Important

2. **Contradictory doc comments** — `internal/gtfs/gtfs_manager.go:457` and `internal/gtfs/gtfs_manager.go:632-633`

   `RoutesForAgencyID` says "Caller must hold manager.RLock()" while `VehiclesForAgencyID` says "callers must NOT hold any Manager locks" — but `VehiclesForAgencyID` calls `RoutesForAgencyID`. Whichever locking approach you choose for #1, please make these two doc comments consistent.

3. **Removed `TestConcurrentVehicleUpdates` without replacement** — `internal/gtfs/concurrency_test.go`

   This test exercised concurrent reads of `VehiclesForAgencyID` alongside writes to real-time vehicle data. Given the locking change in this PR, a replacement concurrency test for the new implementation would be valuable — especially since `VehiclesForAgencyID` now has a fundamentally different access pattern.

## Fit and Finish

4. **`sql.NullString` accessed without `.Valid` checks** — `internal/restapi/routes_for_agency_handler.go:39-42`

   All nullable fields (`ShortName`, `LongName`, `Desc`, `Url`, `Color`, `TextColor`) are accessed via `.String` directly. Functionally this works (zero value is `""`), but it's inconsistent with the established project pattern visible in `vehicles_for_agency_handler.go:169-191` and documented in CLAUDE.md. Minor, but worth aligning for consistency.

aaronbrethorst (approved):
🔥🚀  great work.

PR #813, opened 2026-04-06
fix(gtfs): honor context cancellation and fix swallowed errors in Adv…

### LINKED ISSUE #812: bug(gtfs): context cancellation ignored and errors swallowed in AdvancedDirectionCalculator
### Describe the bug
Hey team! I was digging into the codebase recently to look at how we handle `r.Context()` across `GtfsManager`, and I stumbled on a couple of edge cases in `AdvancedDirectionCalculator.CalculateStopDirection` that might cause some headaches under heavy load.

**1. Ignored Context Cancellation:** It looks like we're wrapping the request context with `context.WithoutCancel(ctx)` right before passing it down into `computeFromShapes()`. Since `computeFromShapes` kicks off multiple SQLite queries (`GetStopsWithShapeContext`, etc.), this means if a client drops the HTTP connection or hits a timeout, the DB keeps chugging along. We're essentially abandoning those queries and holding onto SQLite connections longer than we need to.

**2. Swallowed Errors in Singleflight:** Inside the `adc.requestGroup.Do` block, the callback is always returning a `nil` error, even if `computeFromShapes` actually fails. Because of this, the `singleflight` waiters just get handed an empty string instead of the actual error. It basically swallows DB timeouts or drops, making it much harder to debug.

### To Reproduce
Check out `internal/gtfs/advanced_direction_calculator.go`:
1. In `CalculateStopDirection`, we do: `computedDir, err := adc.computeFromShapes(context.WithoutCancel(ctx), stopID)`
2. Look at the return of the `Do` callback when `err != nil`—it just returns an empty string without bubbling up the error.

### Expected behavior
* We should just pass `ctx` directly so the SQLite queries naturally respect HTTP client disconnects and server shutdowns.
* The `singleflight` callback should return `("", err)` on failure so we don't hide the error, and so we don't accidentally cache transient DB drops as empty strings.

### Proposed Solution
I've already got a local fix running! 
1. Dropped `context.WithoutCancel(ctx)`.
2. Updated the `Do` callback to return the actual error.
3. Added a quick check after the `Do` block to return an empty string gracefully (so we keep the existing outward behavior without poisoning the cache).

I can toss up a PR for this shortly if this approach makes sense to everyone!

tejasva-vardhan:
### Description

Fixes #812

**The Problem**
While auditing context propagation across the `GtfsManager`, I found a resource leak and a swallowed error bug in `AdvancedDirectionCalculator.CalculateStopDirection`:
1.  **Ignored Cancellation:** The code was wrapping the request context with `context.WithoutCancel(ctx)` before passing it to `computeFromShapes`. Because `computeFromShapes` executes multiple SQLite queries (`GetStopsWithShapeContext`, `GetShapePointsWithDistance`), closing the HTTP connection or hitting a timeout would not cancel the underlying DB work. Under load, these abandoned requests would continue to burn CPU and hold database connections.
2.  **Swallowed Errors in Singleflight:** The `singleflight.Group.Do` callback was implicitly returning a `nil` error even if `computeFromShapes` failed. This meant errors (including standard DB timeouts) were discarded, and the `singleflight` waiters would receive empty strings instead of the actual error.

**The Solution**
* Removed `context.WithoutCancel(ctx)` to ensure SQLite queries correctly honor client cancellation and server shutdowns.
* Updated the `singleflight.Group.Do` callback to return `("", err)` when the underlying computation fails, allowing the caller to see the actual error.
* Added an explicit error check after `requestGroup.Do` to return an empty string gracefully on failure, matching the existing outward behavior but preventing cache poisoning on transient errors.

### Behavioral Comparison
* **Before:** If a client drops their connection, the SQLite `computeFromShapes` query keeps running in the background indefinitely. If the database times out or fails, the `singleflight` group swallows the error and silently caches/returns an empty string `""`.
* **After:** If a client drops their connection, the `ctx` cancellation propagates down and instantly kills the SQLite query, freeing the DB lock. If an error occurs, it correctly bubbles up from the `singleflight` group so it can be handled properly without poisoning the cache.

### Architectural Note / Tradeoff
By honoring `ctx` cancellation inside `singleflight`, there is a known tradeoff: if the "winning" request that initiated the computation is canceled, other concurrent requests waiting on that same `stopID` in the singleflight group will also receive the `context.Canceled` error. Given that this is an HTTP server, favoring resource reclamation (canceling the DB queries) over guaranteeing the background work finishes is the preferred Go idiom here.

### DIFF
--- a/internal/gtfs/advanced_direction_calculator.go
+++ b/internal/gtfs/advanced_direction_calculator.go
@@ -109,26 +109,27 @@ func (adc *AdvancedDirectionCalculator) CalculateStopDirection(ctx context.Conte
 
 	// Fall back to computing from shapes, protected by singleflight
 	// This ensures concurrent requests for the SAME stopID don't hit the DB multiple times.
-	v, _, _ := adc.requestGroup.Do(stopID, func() (interface{}, error) {
+	v, err, _ := adc.requestGroup.Do(stopID, func() (interface{}, error) {
 		// Double-check cache inside the singleflight in case another goroutine just finished it
 		if cached, ok := adc.directionResults.Load(stopID); ok {
 			return cached.(string), nil
 		}
 
 		// Actually compute it (Hits the DB)
-		computedDir, err := adc.computeFromShapes(context.WithoutCancel(ctx), stopID)
+		computedDir, err := adc.computeFromShapes(ctx, stopID)
+		if err != nil {
+			return "", err
+		}
 
 		// Only cache when there was no transient error. A transient error (e.g. DB
 		// connection lost) must not permanently poison the cache; omitting it here
 		// means the next request will retry the DB.
-		if err == nil {
-			adc.directionResults.Store(stopID, computedDir)
-		}
-
-		// Intentionally return nil so singleflight shares the empty fallback result with concurrent callers.
-		// Since we skip caching on error, future requests will safely retry the DB.
+		adc.directionResults.Store(stopID, computedDir)
 		return computedDir, nil
 	})
+	if err != nil {
+		return ""
+	}
 
 	return v.(string)
 }


aaronbrethorst:
Please rebase against main and fix all merge conflicts, and then reopen this PR.

PR #815, opened 2026-04-07
refactor: remove routesMap and migrate route lookups to DB

### LINKED ISSUE #814: Remove in-memory routesMap and migrate route lookups to database
The Manager struct holds a routesMap (map[string]*gtfs.Route) that duplicates data already in SQLite. This adds memory overhead, requires a staticMutex for hot-swap safety, and is rebuilt on every GTFS reload.

Replace it with direct DB queries using the existing GetRoute and GetRoutesByIDs queries.

Ahmedhossamdev:
Fixes: #814 

## Summary
- Removes `routesMap map[string]*gtfs.Route` from the `Manager` struct
- `FindRoute()` now delegates to `GtfsDB.Queries.GetRoute()`
- Agency filter functions (`filterTripsByAgency`, `filterVehiclesByAgency`, `filterAlertsByAgency`) are now pure functions no Manager receiver, no `staticMutex`. A single `buildRouteAgencyMap()` batch-fetches route --> agency data from DB before filtering runs
- `trips_for_route_handler` uses one `GetRoute` DB call instead of `FindRoute`
- `buildLookupMaps` now only builds the agency map

### DIFF
--- a/internal/restapi/trips_for_route_handler.go
+++ b/internal/restapi/trips_for_route_handler.go
@@ -281,9 +281,11 @@ func (api *RestAPI) tripsForRouteHandler(w http.ResponseWriter, r *http.Request)
 	fetchedTrips = fetchedTrips[:n]
 
 	tripAgencyMap := make(map[string]string)
-	for _, trip := range fetchedTrips {
-		if route, err := api.GtfsManager.GtfsDB.Queries.GetRoute(ctx, trip.RouteID); err == nil {
-			tripAgencyMap[trip.ID] = route.AgencyID
+	if len(fetchedTrips) > 0 {
+		if route, err := api.GtfsManager.GtfsDB.Queries.GetRoute(ctx, routeID); err == nil {
+			for _, trip := range fetchedTrips {
+				tripAgencyMap[trip.ID] = route.AgencyID
+			}
 		}
 	}
 


internal/gtfs/gtfs_manager.go
```diff
@@ -441,12 +440,8 @@ func (manager *Manager) FindAgency(id string) *gtfs.Agency {
 	return nil
 }
 
-// IMPORTANT: Caller must hold manager.RLock() before calling this method.
```
fletcherw: I think you still need RLock here as that's what protects GtfsDB during database updates.

fletcherw (changes requested):
I think this is on the right path, it can be simplified by inverting the filtering query (from finding agencies for routes to finding routes for agencies).

internal/gtfs/realtime.go
```diff
@@ -510,81 +530,72 @@ func (manager *Manager) updateFeedRealtime(ctx context.Context, feedCfg RTFeedCo
...
-	defer manager.staticMutex.RUnlock()
+// buildAllowedRouteIDs queries the DB for all route IDs belonging to the allowed
+// agencies and returns them as a set. Returns an empty set if queries is nil.
+func buildAllowedRouteIDs(ctx context.Context, queries *gtfsdb.Queries, allowed map[string]bool) map[string]bool {
+	if queries == nil || len(allowed) == 0 {
+		return map[string]bool{}
+	}
+	routeIDs := make(map[string]bool)
+	for agencyID := range allowed {
+		ids, err := queries.GetRouteIDsForAgency(ctx, agencyID)
```
fletcherw: not for this PR, but we should consider a bulk query for this.

fletcherw (approved):
looks good, one small nit and a rebase is all that's needed.

Ahmedhossamdev:
Hey @fletcherw 
After rebasing against main, the changes described previously (removing routesMap, updating FindRoute, refactoring filter functions) were already incorporated into main. The only remaining change is:
trips_for_route_handler: replace one GetRoute DB call per trip with a single GetRoute call using routeID, since all trips in a trips-for-route response belong to the same route and share the same agency.

PR #817, opened 2026-04-07
refactor: use sql database for FindAgency

fletcherw:
Replace use of in-memory index with queries to the sql database.

### DIFF
--- a/internal/models/agency.go
+++ b/internal/models/agency.go
@@ -1,5 +1,7 @@
 package models
 
+import "maglev.onebusaway.org/gtfsdb"
+
 // AgencyCoverage represents the geographical coverage area of a transit agency
 type AgencyCoverage struct {
 	AgencyID string  `json:"agencyId"`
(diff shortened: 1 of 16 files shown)


internal/models/agency.go
```diff
@@ -48,3 +50,18 @@ func NewAgencyReference(id, name, url, timezone, lang, phone, email, fareUrl, di
...
 	}
 }
+
+func AgencyReferenceFromDatabase(agency *gtfsdb.Agency) AgencyReference {
+	return AgencyReference{
+		ID:             agency.ID,
+		Name:           agency.Name,
+		URL:            agency.Url,
+		Timezone:       agency.Timezone,
+		Lang:           agency.Lang.String,
```
Ahmedhossamdev: Lang:    utils.NullStringOrEmpty(agency.Lang),
Phone:   utils.NullStringOrEmpty(agency.Phone),
Email:   utils.NullStringOrEmpty(agency.Email),
FareUrl: utils.NullStringOrEmpty(agency.FareUrl),

Ahmedhossamdev (dismissed):
Hey @fletcherw, looks great and tested manually! One small thing. Could we use utils.NullStringOrEmpty() in AgencyReferenceFromDatabase for the nullable fields to stay consistent with the rest of the codebase? Not a bug, just consistency. LGTM otherwise!

internal/models/agency.go
fletcherw: I would like to do this, but models can't depend on utils since it causes a cyclic dependency. I duplicated NullStringOrEmpty into `models` with a note explaining why and a TODO to break the cycle.

PR #819, opened 2026-04-07
fix: populate situations in vehicles-for-agency references

### LINKED ISSUE #818: vehicles-for-agency: references.situations is always empty
While comparing our endpoint output against the Java OBA server (tested
against MTA NYCT production), I noticed that `vehicles-for-agency` always
returns an empty `situations` array in the references block, even when there
are active service alerts for the agency.

Every other endpoint that returns a references block arrivals-and-departures,
trip-details, route  already collects and returns alerts. This one just
doesn't. The OpenAPI spec also marks `situations` as required in the
`Reference` schema so it's a spec violation on top of the missing behaviour.

reddy-santhu:
closes #818 

fixes the `vehicles-for-agency` endpoint always returning an empty
`references.situations` array. Checked the Java OBA server (MTA NYCT
production) and it does return situations in the references block for this
endpoint.

The handler was simply never wired up to collect alerts. Every other endpoint
that returns a `references` block already does this.

## Changes

**`internal/restapi/vehicles_for_agency_handler.go`**

After building the vehicle list, collect route-level and agency-wide alerts
and populate `references.situations`. Route IDs are already gathered before
the loop for the existing batch route fetch, so nothing new hits the DB.

```go
alerts := deduplicateAlerts(
    api.collectAlertsForRoutes(routeIDs),
    api.GtfsManager.GetAlertsByIDs("", "", id),
)
references.Situations = append(references.Situations, api.BuildSituationReferences(alerts)...)
```

**`internal/gtfs/gtfs_manager_mock.go`**

`MockResetRealTimeData` wasn't clearing `feedAlerts`, so any test that
injected an alert into the shared manager would leak it into later tests.
Added the missing reset and a `rebuildMergedRealtimeLocked` call.

## Testing

Added two tests:
- route-scoped alert shows up in `references.situations`
- agency-wide alert shows up in `references.situations`

Trip-scoped alerts are intentionally not collected here doing so would
require a DB query per vehicle which doesn't scale over a full fleet. Route
and agency scoping covers what real feeds publish in practice, and matches
how other endpoints handle the same trade-off.

### DIFF
--- a/internal/gtfs/gtfs_manager_mock.go
+++ b/internal/gtfs/gtfs_manager_mock.go
@@ -154,7 +154,7 @@ func (m *Manager) MockAddAlert(feedID string, alert gtfs.Alert) {
 	m.rebuildMergedRealtimeLocked()
 }
 
-// MockResetRealTimeData clears all mock real-time vehicles and trip updates.
+// MockResetRealTimeData clears all mock real-time vehicles, trip updates, and alerts.
 func (m *Manager) MockResetRealTimeData() {
 	m.realTimeMutex.Lock()
 	defer m.realTimeMutex.Unlock()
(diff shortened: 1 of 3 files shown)


fletcherw (approved):
Nice work, thanks for fixing this!

Issue #820, opened 2026-04-07
refactor: Remove gtfsData field and migrate static lookups to database queries


Ahmedhossamdev:
## Overview

`Manager.gtfsData *gtfs.Static` holds the entire parsed GTFS static dataset in memory. All of this data already exists in SQLite. The goal is to remove this field entirely and replace every caller with a DB query.

## Methods to migrate

- [x] `GetAgencies()` → `SELECT * FROM agencies`
- [x] `GetRoutes()` → `SELECT * FROM routes`  
- [x] `GetStops()` → `SELECT * FROM stops`
- [x] `GetTrips()` → replace full-table load with targeted per-handler queries
- [x] `GetStaticData()` → remove entirely (blocked on all above)

## Definition of done
- `gtfsData *gtfs.Static` field removed from `Manager`
- `GetStaticData()` method removed
- No references to the `go-gtfs` static types in production handler code

mann-patwa:
Hey @Ahmedhossamdev , 
Are you going to work on all the sub-issues? can I pick up some?

Ahmedhossamdev:
It's all done now, I'll close this issue.

PR #832, opened 2026-04-08
refactor: migrate GetAgencies() from in-memory to database query

### LINKED ISSUE #821: Replace GetAgencies() with database query
Remove the in-memory GetAgencies() method and replace all callers with a direct SELECT * FROM agencies DB query. Update the 5 production handlers and all test call sites.

Ahmedhossamdev:
Fixes: #821 
Part of: #820 

## Summary

- Replaces `manager.gtfsData.Agencies` (in-memory slice) with `manager.GtfsDB.Queries.ListAgencies(ctx)` in `GetAgencies()`
- Updates all call sites across handler and test files to use the new DB-backed method
- Adds `mustGetAgencies` test helper that wraps `ListAgencies` with `require.NoError` for cleaner test setup

## Notes

- `GetAgencies()` now returns `[]gtfsdb.Agency` instead of `[]gtfs.Agency`, so field access changes from `.Id`/`.Language` to `.ID`/`.Lang.String` etc.

### DIFF
--- a/internal/restapi/trips_for_location_handler.go
+++ b/internal/restapi/trips_for_location_handler.go
@@ -147,13 +147,13 @@ func (api *RestAPI) parseAndValidateRequest(r *http.Request) (
 	includeTrip = queryParams.Get("includeTrip") == "true"
 	includeSchedule = queryParams.Get("includeSchedule") == "true"
 
-	agencies := api.GtfsManager.GetAgencies()
-	if len(agencies) == 0 {
+	agencies, agenciesErr := api.GtfsManager.GetAgencies(r.Context())
+	if agenciesErr != nil || len(agencies) == 0 {
 		return 0, 0, 0, 0, false, false, nil, time.Time{}, time.Time{}, nil, errors.New("no agencies configured in GTFS manager")
 	}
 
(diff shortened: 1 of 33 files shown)


internal/restapi/trips_for_location_handler.go
```diff
@@ -147,13 +147,13 @@ func (api *RestAPI) parseAndValidateRequest(r *http.Request) (
 	includeTrip = queryParams.Get("includeTrip") == "true"
 	includeSchedule = queryParams.Get("includeSchedule") == "true"
 
-	agencies := api.GtfsManager.GetAgencies()
-	if len(agencies) == 0 {
+	agencies, agenciesErr := api.GtfsManager.GetAgencies(r.Context())
+	if agenciesErr != nil || len(agencies) == 0 {
```
fletcherw: if agenciesErr != nil, I think that's a different error case that should be reported differently.

More generally, this is reading the first agency and taking its timezone, I don't know if we want to keep that behavior long term.

fletcherw (approved):
there's some small feedback, feel free to merge yourself after fixing, or just merge as is.

PR #833, opened 2026-04-08
refactor: update queryStopsInBounds to use db

fletcherw:
Replace the in-memory stopSpatialIndex with the database's RTree index on stops.

Update several tests that were implicitly relying on the spatialIndex not being updated after new stops were inserted.


Ahmedhossamdev (approved):
LGTM!

PR #848, opened 2026-04-09
refactor: create RoutesForLocation handler

fletcherw:
Create a new RoutesForLocation database query that encapsulates logic for routes-for-location handler. Splitting out this logic allows simplifying the StopsForLocation function and moving Go logic from the RPC handlers into SQL.

Add a SQLite build flag to support math functions used for calculating haversine distance for latlng.


gtfsdb/stops_rtree.go
```diff
@@ -75,3 +75,86 @@ func (q *Queries) GetActiveStopsWithinBounds(ctx context.Context, arg GetActiveS
 	}
 	return items, nil
 }
+
+const getActiveRoutesWithinBounds = `
```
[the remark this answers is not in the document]
fletcherw: my suspicion was correct, this was picking a random stop to use for the distance ordering. I've fixed this in the next commit.

query plan looks reasonable:
```
--SCAN r VIRTUAL TABLE INDEX 2:D0B1D2B3
|--SEARCH stops USING INTEGER PRIMARY KEY (rowid=?)
|--SEARCH stop_times USING COVERING INDEX idx_stop_times_stop_id_trip_id (stop_id=?)
|--SEARCH trips USING INDEX sqlite_autoindex_trips_1 (id=?)
|--SEARCH routes USING INDEX sqlite_autoindex_routes_1 (id=?)
|--USE TEMP B-TREE FOR GROUP BY
`--USE TEMP B-TREE FOR ORDER BY
```

gtfsdb/stops_rtree.go
```diff
@@ -75,3 +75,92 @@ func (q *Queries) GetActiveStopsWithinBounds(ctx context.Context, arg GetActiveS
...
+FROM
+    stops
+    JOIN stop_times ON stops.id = stop_times.stop_id
+    JOIN stops_rtree r ON r.id = stops.rowid
+    JOIN trips ON stop_times.trip_id = trips.id
+    JOIN routes ON trips.route_id = routes.id
+WHERE
+    r.min_lat >= ?3 AND r.max_lat <= ?4
+    AND r.min_lon >= ?5 AND r.max_lon <= ?6
+    AND (?7 == "" OR routes.short_name == ?7)
```
Ahmedhossamdev: One small thing the handler lowercases the query `query = strings.ToLower(sanitizedQuery)` before passing it to the SQL, but the SQL compares it against routes.short_name as is, so it'll never match routes with mixed-case short names. 
Fix would be LOWER(routes.short_name) == ?7.

Ahmedhossamdev (approved):
LGTM, just we should need to handle the lowercase route short name in the query.

gtfsdb/stops_rtree.go
fletcherw: This was a good catch, I didn't think about this. Fortunately it turns out sqlite is case-insensitive by default if you use LIKE.

```
sqlite> select * from routes where short_name like 'ShuTTle' limit 10;
1-SHUTTLE|40|Shuttle|Link Replacement Shuttle Bus||3||FFB819|000000|1|1
2-SHUTTLE|40|Shuttle|2 Line Shuttle Bus||3||FFB819|000000|1|1
TLINE_S|40|Shuttle|T Line Replacement Shuttle Bus||3||DA7C00|FFFFFF|1|1
```

Added a test for this case as well to make sure this doesn't break in the future.

PR #850, opened 2026-04-10
refactor: replace GetTrips() full-table load with targeted DB query

Ahmedhossamdev:
Fixes: #826 
Part of: #820 

## Summary
- Removes in-memory `GetTrips()` (which loaded all trips from `gtfsData.Trips`) and replaces it with `GetTrips(ctx, limit)` backed by a new `ListTripsWithLimit` SQL query
- Adds `ListTripsWithLimit` to `query.sql` — uses `LIMIT ?` to avoid full-table scans, especially important for large merged feeds
- Updates all test call sites to use `mustGetTrips(t, api)` helper (limit 100) and maps flat `gtfsdb.Trip` fields (`RouteID`, `ServiceID`, `TripHeadsign`, etc.) replacing the old nested `gtfs.ScheduledTrip` fields (`Route.Id`, `Service.Id`, `Headsign`, etc.)
- Fixes pre-existing `errcheck` lint issue in `stops_rtree.go`

### DIFF
--- a/internal/restapi/http_test.go
+++ b/internal/restapi/http_test.go
@@ -109,6 +109,14 @@ func mustGetAgencies(t testing.TB, api *RestAPI) []gtfsdb.Agency {
 	return agencies
 }
 
+// mustGetTrips fetches up to 100 trips from the DB for use in tests.
+func mustGetTrips(t testing.TB, api *RestAPI) []gtfsdb.Trip {
+	t.Helper()
+	trips, err := api.GtfsManager.GetTrips(context.Background(), 100)
+	require.NoError(t, err)
+	return trips
+}
+
 // serveAndRetrieveEndpoint sets up a test server, makes a request to the specified endpoint, and returns the response
 // and decoded model.
 // Accepts testing.TB to support both *testing.T and *testing.B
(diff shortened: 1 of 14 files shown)


internal/restapi/http_test.go
```diff
@@ -109,6 +109,14 @@ func mustGetAgencies(t testing.TB, api *RestAPI) []gtfsdb.Agency {
 	return agencies
 }
 
+// mustGetTrips fetches up to 100 trips from the DB for use in tests.
+func mustGetTrips(t testing.TB, api *RestAPI) []gtfsdb.Trip {
+	t.Helper()
+	trips, err := api.GtfsManager.GetTrips(context.Background(), 100)
```
fletcherw: seems like most of these tests just want a single random trip. You could consider making the helper do that instead of reading 100 trips.

gtfsdb/stops_rtree.go
```diff
@@ -42,8 +42,6 @@ func (q *Queries) GetActiveStopsWithinBounds(ctx context.Context, arg GetActiveS
 	if err != nil {
 		return nil, err
 	}
-	defer rows.Close()
```
fletcherw: where'd you see the error lint for this? was it in presubmits or somewhere else?

fletcherw (approved):

gtfsdb/stops_rtree.go
Ahmedhossamdev: When I run make  lint
image

PR #851, opened 2026-04-11
refactor: simplify GetStopsForLocation

fletcherw:
Remove some special casing that was used when calling this function from routes-for-location. This is no longer needed since that handler uses a separate implementation.


Ahmedhossamdev (approved):
Great!

PR #852, opened 2026-04-11
refactor: use stricter json schema for a test

fletcherw:
Delete a bunch of field casting/validation in the test code by using a stricter response type, so that the json decode handles it for us.

### DIFF
--- a/internal/restapi/routes_for_location_handler_test.go
+++ b/internal/restapi/routes_for_location_handler_test.go
@@ -6,9 +6,24 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
-	"maglev.onebusaway.org/internal/utils"
+	"maglev.onebusaway.org/internal/models"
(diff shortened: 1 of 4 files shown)


internal/restapi/routes_for_location_handler_test.go
```diff
@@ -18,42 +32,42 @@ func TestRoutesForLocationHandlerRequiresValidApiKey(t *testing.T) {
...
-	agency, ok := agencies[0].(map[string]interface{})
-	require.True(t, ok)
-	assert.Contains(t, agency, "id")
-	assert.Contains(t, agency, "name")
-	assert.Contains(t, agency, "url")
-	assert.Contains(t, agency, "timezone")
-	assert.Contains(t, agency, "lang")
-	assert.Contains(t, agency, "phone")
+	assert.ElementsMatch(t, model.Data.List, []models.Route{
+		{
```
fletcherw: specifying this exactly is a little verbose. we could have some constants for these somewhere.

Ahmedhossamdev (approved):
LGTM!

PR #853, opened 2026-04-11
refactor: remove in-memory gtfs.Static, use the DB as single source of truth

Ahmedhossamdev:
Fixes: #827
Part of: #820 

## Summary

This PR completes the migration of static GTFS data away from the in-memory `gtfs.Static` struct, making the SQLite database the single source of truth for all static GTFS data (agencies, routes, stops, trips).


internal/restapi/http_test.go
```diff
@@ -118,6 +118,28 @@ func mustGetTrip(t testing.TB, api *RestAPI) gtfsdb.Trip {
 	return trips[0]
 }
 
+// mustGetTripIDWithBlockID returns the block_id of the first trip that has one set.
+func mustGetTripIDWithBlockID(t testing.TB, api *RestAPI) string {
+	t.Helper()
+	var blockID string
+	err := api.GtfsManager.GtfsDB.DB.QueryRowContext(context.Background(),
+		`SELECT block_id FROM trips WHERE block_id IS NOT NULL AND block_id != '' LIMIT 1`,
```
Ahmedhossamdev, line 126: I used raw SQL here instead of adding a sqlc query to query.sql. This lookup is only needed by tests, so I didn't want to pollute the production query file with query to get one trip with a block or shape, any suggestions?

internal/gtfs/gtfs_manager.go
```diff
@@ -913,17 +896,25 @@ func (manager *Manager) GetActiveServiceIDsForDateCached(ctx context.Context, da
 	return out, nil
 }
 
-// IMPORTANT: Caller must hold manager.RLock() before calling this method.
 func (manager *Manager) PrintStatistics() {
```
fletcherw, line 900: this function is dead code, you can just entirely delete it.

internal/gtfs/realtime.go
```diff
@@ -552,36 +548,33 @@ func (manager *Manager) filterVehiclesByAgency(vehicles []gtfs.Vehicle, allowed
...
+		if alertMatchesAgency(ctx, manager, alert, allowed) {
 			filtered = append(filtered, alert)
 		}
 	}
 	return filtered
 }
 
-// alertMatchesAgencyLocked assumes staticMutex is already held by the caller.
-func alertMatchesAgencyLocked(manager *Manager, alert gtfs.Alert, allowed map[string]bool) bool {
+func alertMatchesAgency(ctx context.Context, manager *Manager, alert gtfs.Alert, allowed map[string]bool) bool {
```
fletcherw: caller still needs to hold staticMutex until we remove the hotswap logic.

internal/restapi/calculate_block_trip_sequence_test.go
```diff
@@ -26,35 +27,39 @@ func TestCalculateBlockTripSequence(t *testing.T) {
 	// Monday within the RABA dataset's active service period (calendar range covers this date)
 	serviceDate := time.Date(2024, 11, 4, 0, 0, 0, 0, time.UTC)
 
-	// Find a block that has multiple active trips so we can verify sequencing
```
fletcherw, line 29: Seems like a reasonable comment to keep, since it explains _why_ the code is doing something.

internal/restapi/http_test.go
fletcherw: what you've done here is fine, especially cause it's short. We could also have a file for test-only queries if it becomes a more common thing where the query code is generated by sqlc.

fletcherw (approved):

internal/gtfs/gtfs_manager.go
Ahmedhossamdev: It should be used to log the current count of gtfs-static. I'll add it when we update the gtfs-static

internal/gtfs/realtime.go
Ahmedhossamdev: Oops, yes, this is true

PR #863, opened 2026-04-13
cleanup: delete query duration logic

fletcherw:
This slow query detector doesn't work. The bulk of query duration happens after the wrapped calls, when fetching rows from the query. Remove this and the query latency histogram since it's only going to give misleading information.


Ahmedhossamdev (approved):
LGTM!

PR #865, opened 2026-04-15
refactor: calculate region bounds from database

fletcherw:
Instead of reading the static data, run a database query to determine the region bounds to store in GtfsManager.

### DIFF
--- a/gtfsdb/db.go
+++ b/gtfsdb/db.go
@@ -270,6 +270,9 @@ func Prepare(ctx context.Context, db DBTX) (*Queries, error) {
 	if q.getStopStmt, err = db.PrepareContext(ctx, getStop); err != nil {
 		return nil, fmt.Errorf("error preparing query GetStop: %w", err)
 	}
+	if q.getStopBoundsPerAgencyStmt, err = db.PrepareContext(ctx, getStopBoundsPerAgency); err != nil {
+		return nil, fmt.Errorf("error preparing query GetStopBoundsPerAgency: %w", err)
+	}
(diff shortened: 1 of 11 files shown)


Ahmedhossamdev (approved):
LGTM! While reviewing I noticed two bugs in how the bounding box is calculated for agencies-with-coverage when comparing our output against the Java reference implementation. If  you would like 'll merge this and open a separate issue + PR to fix them or we can fix this in the same PR

Bug 1: Wrong data source (shape points instead of stop points)

The current implementation uses GetShapeBounds, which queries shape geometry — the full drawn path of every route, including all intermediate curve points between stops. This inflates the bounding box well beyond the actual service area. The Java implementation uses only stop coordinates (the lat/lon of each stop), which is the correct source of truth for coverage area.

Bug 2: One global bounding box applied to all agencies

The current code computes a single bounding box across the entire database and assigns it to every agency. If the dataset contains multiple agencies with different service areas, they all incorrectly report the same bounds. The Java implementation computes bounds per-agency
by walking agency --> routes --> trips --> stop_times --> stops, so each agency's coverage reflects only the stops it actually serves.

The combined effect of both bugs on a multi-agency dataset (Puget Sound) was a latSpan of 3.767 instead of the correct 0.228 roughly 16x too large, covering a multi-state region instead of the actual service area.

Fix: Add a GetStopBoundsForAgency SQL query joining stops --> stop_times --> trips -->routes filtered by agency_id, and call it per-agency in the handler instead of using the global GetRegionBounds().

go vs java

Image

Ahmedhossamdev:
LGTM!

Thanks for the update! @fletcherw

PR #869, opened 2026-04-15
fix: improve performance of routes-for-location

fletcherw:
Fixes some performance issues in routes-for-location so that it's now faster than the old implementation.

  On the same data:
    Old implementation: 850ms
    New implementation: 1000ms
    With fixes: 450ms

Removes some duplicated distance calculation in the SQL query.
Fixes some edge cases in query matching, documents limitations of matching in SQL.
Switches to batch agency query.


Ahmedhossamdev (commented):
LGTM!

PR #870, opened 2026-04-15
Revert "perf: add in-memory cache for GetActiveServiceIDsForDate, inv…

fletcherw:
…alidate on GTFS reload"

This reverts commit b7d6149f6a527580b0f7bfb1624597d638657a81.

Reverting this change because it makes reasoning about correctness significantly harder during GTFS updates, but the query it's caching isn't particularly slow. I tested directly against the database, and this query usually runs in <1ms.

I also tested e2e latency of stops-for-location with and without this change and it's pretty much the same.

Not a clean rollback, also reverted some usages of logic introduced in the change that came after.


Ahmedhossamdev (approved):
Great work!

PR #871, opened 2026-04-15
refactor: simplify stop direction precomputation

fletcherw:
Remove a lot of the parallelism and batch commit complexity, remove shape cache.

On full puget sound GTFS dataset, precomputation time goes from 3.2s to 1.2s.


Ahmedhossamdev:
LGTM!

aaronbrethorst (approved):
great work 👍

PR #873, opened 2026-04-16
fix: match Java's stop distribution behavior in stops-for-location

### LINKED ISSUE #872: stops-for-location: stops clustered in small area instead of distributed across search radius
## Problem

When calling /api/where/stops-for-location.json, the Go server consistently returns stops clustered in a small geographic area rather than distributed across the full search radius. This diverges from the Java reference implementation.

Java behavior: BeanServiceSupport.checkLimitExceeded() shuffles stops randomly before truncating (Collections.shuffle()), producing geographically diverse results. When a routeType filter is active, Java uses ORDERED_BY_CLOSEST mode which skips the shuffle and filters by service date after truncation.

Go before the changes:
Image

Java:

Image

Ahmedhossamdev:
Fix: #872

## Summary

Stops returned by `stops-for-location` were clustered in a small geographic area instead of being distributed across the search radius.

### Root cause
The Go implementation was sorting stops by ID before truncating to `maxCount`.  
In the Puget Sound GTFS dataset, lower stop IDs are geographically clustered (e.g., downtown), which caused biased results.

## Changes

- Added `shuffleBeforeLimit` parameter to `GetStopsForLocation`
  - `stops-for-location` now passes `true` to shuffle results before truncating
  - Matches Java behavior in `BeanServiceSupport.checkLimitExceeded()` (BOUNDS mode)

- When a `routeType` filter is active:
  - Sort stops by distance instead of shuffling
  - Apply truncation after route filtering
  - Matches Java's `ORDERED_BY_CLOSEST` mode

- Fixed `limitExceeded` calculation:
  - Previously computed incorrectly in the handler loop
  - Now returned directly from `GetStopsForLocation`

- Removed unused DB query:
  - `GetStopsWithActiveServiceOnDate`
  - Service-date filtering is already handled in the handler via `GetActiveRouteIDsForStopsOnDate`

### DIFF
--- a/gtfsdb/stops_rtree.go
+++ b/gtfsdb/stops_rtree.go
@@ -73,6 +73,44 @@ func (q *Queries) GetActiveStopsWithinBounds(ctx context.Context, arg GetActiveS
 	return items, nil
 }
 
+const getStopIDsWithinBounds = `
+SELECT s.id
+FROM stops s
+INNER JOIN stops_rtree sr ON sr.id = s.rowid
+WHERE sr.min_lat >= ? AND sr.max_lat <= ?
+  AND sr.min_lon >= ? AND sr.max_lon <= ?
+  AND EXISTS (SELECT 1 FROM stop_times st WHERE st.stop_id = s.id)
+`
+
+type GetStopIDsWithinBoundsParams struct {
+	MinLat float64
+	MaxLat float64
+	MinLon float64
+	MaxLon float64
+}
+
+func (q *Queries) GetStopIDsWithinBounds(ctx context.Context, arg GetStopIDsWithinBoundsParams) ([]string, error) {
+	rows, err := q.db.QueryContext(ctx, getStopIDsWithinBounds,
+		arg.MinLat, arg.MaxLat, arg.MinLon, arg.MaxLon)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+
+	var ids []string
+	for rows.Next() {
+		var id string
+		if err := rows.Scan(&id); err != nil {
+			return nil, err
+		}
+		ids = append(ids, id)
+	}
+	if err := rows.Err(); err != nil {
+		return nil, err
+	}
+	return ids, nil
+}
+
 const getActiveRoutesWithinBounds = `
 -- Calculate stop distance once per stop (not once per stop_time).
 WITH nearby_stops AS (
--- a/gtfsdb/stops_rtree.go
+++ b/gtfsdb/stops_rtree.go
@@ -7,7 +7,7 @@ import (
 // Implemented manually because sqlc doesn't support the virtual tables from the RTree module.
 
 const getActiveStopsWithinBounds = `
-SELECT DISTINCT
+SELECT
     s.id,
     s.code,
     s.name,
@@ -23,10 +23,10 @@ SELECT DISTINCT
     s.direction,
     s.parent_station
 FROM stops s
-INNER JOIN stop_times st ON s.id = st.stop_id
-INNER JOIN stops_rtree r ON r.id = s.rowid
-WHERE r.min_lat >= ? AND r.max_lat <= ?
-  AND r.min_lon >= ? AND r.max_lon <= ?
+INNER JOIN stops_rtree sr ON sr.id = s.rowid
+WHERE sr.min_lat >= ? AND sr.max_lat <= ?
+  AND sr.min_lon >= ? AND sr.max_lon <= ?
+  AND EXISTS (SELECT 1 FROM stop_times st WHERE st.stop_id = s.id)
 `
 
 type GetActiveStopsWithinBoundsParams struct {
--- a/internal/gtfs/gtfs_manager.go
+++ b/internal/gtfs/gtfs_manager.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"math/rand"
 	"os"
 	"slices"
 	"sort"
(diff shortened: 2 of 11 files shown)


internal/gtfs/gtfs_manager.go
```diff
@@ -430,8 +426,8 @@ func (manager *Manager) GetStopsForLocation(
 	stopCodeQuery string,
 	maxCount int,
 	routeTypes []int,
-	queryTime time.Time,
-) []gtfsdb.Stop {
+	shuffleBeforeLimit bool,
```
fletcherw: I don't like having boolean behavior flags like this on functions. It's an indication that it should be split into two functions like I did with GetRoutesForLocation

internal/gtfs/gtfs_manager.go
```diff
@@ -443,117 +439,90 @@ func (manager *Manager) GetStopsForLocation(
...
 
-	// Filter by service date - only include stops with active service on current date
-	if len(stops) > 0 {
-		var currentDate string
-		if !queryTime.IsZero() {
-			currentDate = queryTime.Format("20060102")
-		} else {
-			currentDate = time.Now().Format("20060102")
-		}
+	// Get routes for all stops to filter out stops with no routes
```
fletcherw: If we're going to filter out stops with no routes, why not do that as part of the sql query for queryStopsInBounds?

internal/gtfs/gtfs_manager.go
```diff
@@ -443,117 +439,90 @@ func (manager *Manager) GetStopsForLocation(
...
 	}
 
-	// If the stop does not have any routes actively serving it, don't include it in the results
-	// TODO: move this logic into the first queryStopsInBounds call to avoid 2 db roundtrips. May need
-	// the function split for query logic mentioned above.
-	if len(routeTypes) > 0 {
-		stopIDs := make([]string, 0, len(stops))
-		for _, stop := range stops {
-			stopIDs = append(stopIDs, stop.ID)
+	// Match Java's two query modes:
```
fletcherw: another indication that this should be multiple functions.

fletcherw (approved):

gtfsdb/stops_rtree.go
```diff
@@ -23,10 +23,10 @@ SELECT DISTINCT
...
     s.parent_station
 FROM stops s
-INNER JOIN stop_times st ON s.id = st.stop_id
-INNER JOIN stops_rtree r ON r.id = s.rowid
-WHERE r.min_lat >= ? AND r.max_lat <= ?
-  AND r.min_lon >= ? AND r.max_lon <= ?
+INNER JOIN stops_rtree sr ON sr.id = s.rowid
+WHERE sr.min_lat >= ? AND sr.max_lat <= ?
+  AND sr.min_lon >= ? AND sr.max_lon <= ?
+  AND EXISTS (SELECT 1 FROM stop_times st WHERE st.stop_id = s.id)
```
fletcherw, line 29: very nice!

internal/gtfs/gtfs_manager.go
```diff
@@ -491,69 +487,44 @@ func (manager *Manager) GetStopsForLocation(
...
-					if stopsWithService[stop.ID] {
-						filteredStops = append(filteredStops, stop)
-					}
-				}
-				stops = filteredStops
-			}
-		}
+// GetStopsInBounds returns stops within the given bounds up to maxCount, without shuffling
+// or route-type filtering. Used internally by the arrivals and trips-for-location handlers.
+func (manager *Manager) GetStopsInBounds(
```
fletcherw, line 502: both callers of this function only want stop IDs. How much faster is it if you add a different query that only fetches stop IDs?

internal/gtfs/gtfs_manager.go
```diff
@@ -491,69 +487,44 @@ func (manager *Manager) GetStopsForLocation(
 			}
 			stops = filteredStops
 		}
-	}
 
-	// Filter by service date - only include stops with active service on current date
```
fletcherw, line 496: Isn't this a change in behavior? You check that a stop has a stop_time above, but this is also checking that there's a trip to that stop on this date.

Ahmedhossamdev: Good catch. The date filter had two parts: checking stop_time exists (still there in GetActiveStopsWithinBounds) and filtering by active service date (moved to handler).

For stops-for-location, the service-date filter is still applied via GetActiveRouteIDsForStopsOnDate on line 144 stops with no active routes today get skipped.

For the internal callers (trips-for-location, arrivals), the downstream queries handle the date filtering themselves, so moving it out doesn't change behavior.

I moved it because Java does shuffle + truncate before the service-date filter. Doing the date check inside GetStopsForLocation would bias the random sample away from that.

internal/gtfs/gtfs_manager.go
Ahmedhossamdev: Yep!

I tested it with k6 on the Puget Sound consolidated dataset (13.6k stops). Optimizing the nearby-stop query in arrivals-and-departures-for-stop to fetch only IDs (1 column instead of 14) reduced average latency by ~11% (827 ms → 738 ms) and increased throughput by ~10% (17.4 → 19.1 req/s) under 20 concurrent users.

fletcherw: nice! 10% improvement seems worth it.

internal/gtfs/gtfs_manager.go
```diff
@@ -412,61 +413,58 @@ func (manager *Manager) RoutesForAgencyID(ctx context.Context, agencyID string)
...
+	if len(routeTypes) == 0 {
+		// BOUNDS mode: shuffle then truncate before route filtering.
+		limitExceeded = len(stops) > maxCount
+		if limitExceeded {
+			rand.Shuffle(len(stops), func(i, j int) { stops[i], stops[j] = stops[j], stops[i] })
+			stops = stops[:maxCount]
+		}
+	} else {
+		// ORDERED_BY_CLOSEST mode: sort by distance, filter by route type, then truncate.
+		sort.Slice(stops, func(i, j int) bool {
```
Not-Dhananjay-Mishra, line 463: @Ahmedhossamdev I believe, using `slices.SortFunc` instead of `sort.Slice` as `slices.SortFunc` results in fewer heap allocations and is a more idiomatic approach in modern Go.
Thoughts?

Ahmedhossamdev: Hey @Not-Dhananjay-Mishra 
I agree that slices.SortFunc is more idiomatic and type-safe. It may reduce allocations slightly, but in our case (sorting ~100–250 stops), I don’t expect any meaningful difference since both use similar underlying sorting algorithms.

Issue #861, opened 2026-04-13
Migrate: GET /api/where/trip/{id}.json


burma-shave:
Validate and migrate traffic to this endpoint from the legacy Java OBA API, ensuring behavioural compatibility with existing clients.

## Process

1. **Spec generation** — the legacy Java implementation is analysed to produce a detailed behavioural spec, including any suspected defects identified by comparing the implementation against its inferred intent. This is done using with a llm skill developed for the purpose.
2. **Review** — the spec is reviewed for correctness and each suspected defect classified as preserve (clients may depend on it), fix (correct it in Maglev), or investigate (needs further analysis). Open questions may need to be resolved.
3. **Testing** — the reviewed spec is used to evaluate existing Maglev tests for correctness and coverage; issues are raised for any gaps
4. **Rollout** — the endpoint is cut over in the beta environment and monitored for unexpected behaviour. Any issues arising are lined back to this issue.

burma-shave:
# Use Case: Retrieve a Single Trip

**Endpoint:** `GET /api/where/trip/{id}.json`

---

## 1. Goal in Context

A rider wants to retrieve the schedule and route information for a specific transit trip so they can look up headsign, direction, or block information in a transit application.

---

## 2. Scope

OneBusAway REST API — single-resource read operation.

---

## 3. Level

User goal.

---

## 4. Primary Actor

Rider (via a client application making API requests on their behalf).

---

## 5. Stakeholders and Interests

- **Rider:** Wants the trip's headsign, direction, associated route, service ID, and block identifier so they can correlate the trip with schedule data.

---

## 6. Preconditions

- The caller supplies a valid API key in the `key` query parameter.
- The API key has not exceeded its rate limit.
- The server has a loaded data bundle.

---

## 7. Minimal Guarantees

- Every response carries a top-level `version`, `code`, `currentTime`, and `text` field regardless of outcome.
- A missing or invalid API key always produces a `401` or `429` response before the trip look-up is attempted.

---

## 8. Success Guarantees

- The response body contains a `data` object with an `entry` object holding the trip's fields and a `references` object containing the full record for the route the trip operates on.
- The `code` field is `200` and the HTTP status is `200`.

---

## 9. Trigger

The client sends a GET request to `/api/where/trip/{id}.json` with a `key` query parameter.

---

## 10. Main Success Scenario

1. The API key interceptor extracts the `key` parameter and verifies it is present and authorised.
[...]

PR #879, opened 2026-04-22
Api/where/trip/{id}

FLASH2332:
@burma-shave @fletcherw @aaronbrethorst 

I’ve made these changes:
- malformed trip ID returns 400 (instead of 500)
- `timeZone` removed from trip entry output
- removed an unreachable null/empty-trip branch

One thing is still open: error envelope `version` behavior on failure paths.

Right now we’re mixed:
- some error paths return `version: 2`
- some still return `version: 1` (notably server-error and panic-recovery style responses)

Before I change that, can you confirm intended behavior?

1. For item-2, should error responses use version 2 consistently, or keep some legacy version 1 paths?
2. Is this change expected only for /api/where/trip/{id}.json, or across all endpoints?
3. For 401 and 429 specifically, should we keep version 1 for backward compatibility or move them to version 2 as well?

### DIFF
--- a/internal/restapi/rate_limit_integration_test.go
+++ b/internal/restapi/rate_limit_integration_test.go
@@ -208,8 +208,8 @@ func TestRateLimitingErrorResponse(t *testing.T) {
 		// Once we hit rate limit, check the error response
 		if response.StatusCode == http.StatusTooManyRequests {
 			assert.Equal(t, http.StatusTooManyRequests, model.Code)
-			assert.Contains(t, model.Text, "Rate limit",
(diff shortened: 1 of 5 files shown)


FLASH2332:
Related to #861

burma-shave:
Thanks for submitting these changes.

However, per the process in #861, the spec needs to pass a correctness review before any implementation work begins. That review hasn't happened yet for this endpoint, so we can't validate whether these changes are aligned with the intended behaviour.

Closing this PR for now. Once the spec review is complete, contributions here would be very welcome.

PR #880, opened 2026-04-22
refactor: move block layover index from memory to SQLite

Ahmedhossamdev:
Fixes: #838 

# Summary

Migrates the block-layover index from an in-memory `map[string][]*BlockLayoverIndex`
built on every GTFS load to a persistent `block_layover` table, rebuilt inside the
import transaction. Part of the ongoing effort to remove in-memory GTFS static
state from `gtfs_manager.go` 

## Summary by CodeRabbit

* **Refactor**
  * Layover tracking moved from in-memory indices to a persistent, database-backed model for consistent, scalable layover tracking.

* **New Features**
  * Persistent layover storage with enforced valid intervals and indexes for efficient route/time lookups.
  * DB-backed query to fetch active layover blocks for a route and time window.

* **Improvements**
  * Trip listings use DB-backed layover lookups with stricter time-window filtering and improved handling of shared blocks and agency assignment.
  * Reimports now clear prior layover data to avoid duplicates.

* **Tests**
  * Added integration-style tests validating layover persistence, reimport behavior, and time-window query correctness.


internal/restapi/trips_for_route_handler.go
```diff
@@ -269,23 +273,30 @@ func (api *RestAPI) tripsForRouteHandler(w http.ResponseWriter, r *http.Request)
...
 	}
-	fetchedTrips = fetchedTrips[:n]
 
 	tripAgencyMap := make(map[string]string)
 	if len(fetchedTrips) > 0 {
-		if route, err := api.GtfsManager.GtfsDB.Queries.GetRoute(ctx, routeID); err == nil {
-			for _, trip := range fetchedTrips {
-				tripAgencyMap[trip.ID] = route.AgencyID
+		routeAgencyCache := make(map[string]string)
+		for _, trip := range fetchedTrips {
```
fletcherw, line 282: any reason not to collect all Route IDs in one slice and then call GetRoutesByIDs?

gtfsdb/schema.sql
```diff
@@ -367,6 +367,27 @@ CREATE INDEX IF NOT EXISTS idx_block_trip_entry_trip_id ON block_trip_entry (tri
...
 
+-- migrate
+CREATE TABLE
+    IF NOT EXISTS block_layover (
+        id INTEGER PRIMARY KEY AUTOINCREMENT,
+        block_id TEXT NOT NULL,
+        service_id TEXT NOT NULL,
+        route_id TEXT NOT NULL, -- route of the trip DEPARTING from the layover
+        layover_stop_id TEXT NOT NULL,
+        layover_start INTEGER NOT NULL, -- nanoseconds since service-day midnight
```
fletcherw, line 441: nanoseconds seems a little precise, would probably be fine as seconds. Not a big deal though.

internal/restapi/trips_for_route_handler.go
fletcherw, line 66: all of this interval math using  integers is bug-prone. Ideally you would do all of the time math using Time and Duration, and only covert to an integer for the database query.

gtfsdb/query.sql
```diff
@@ -1051,12 +1080,18 @@ WHERE bte.block_id IN (sqlc.slice('block_ids'))
...
+-- specified BlockTripIndex IDs. Mirrors Java's BlockCalendarServiceImpl.getActiveBlocksInTimeRange,
+-- which binary-searches maxArrivals/minDepartures so "all E blocks" never includes a block
+-- whose trips are hours away from the requested time.
 SELECT DISTINCT bte.block_id
 FROM block_trip_entry bte
-WHERE bte.block_trip_index_id IN (sqlc.slice('index_ids'))
-  AND bte.service_id IN (sqlc.slice('service_ids'))
-  AND bte.block_id IS NOT NULL;
+JOIN trips t ON bte.trip_id = t.id
+WHERE t.max_departure_time >= sqlc.arg('from_time')
```
fletcherw, line 1093: what's the expected behavior if max or min are null? Or is that considered invalid data and out-of-scope?

gtfsdb/query.sql
```diff
@@ -1051,12 +1080,18 @@ WHERE bte.block_id IN (sqlc.slice('block_ids'))
...
+  AND t.min_arrival_time <= sqlc.arg('to_time')
```
fletcherw, line 1094: Does adding this filter slow down the query much? It seems like maybe not because we pass in the block_trip_index_ids to lookup first.

gtfsdb/query.sql
```diff
@@ -1000,6 +1000,35 @@ DELETE FROM block_trip_entry;
...
+-- name: GetActiveLayoverBlockIDsForRoute :many
+-- Return distinct block IDs whose layover overlaps the given time window for the
+-- specified route + active service IDs. Replaces the in-memory
+-- GetBlocksInTimeRange traversal with one indexed range scan per call.
+-- Slice param is last so non-slice param numbering stays contiguous (?1, ?2, ?3)
+-- when the slice is empty and sqlc expands it to NULL.
+SELECT DISTINCT block_id
+FROM block_layover
+WHERE route_id = sqlc.arg('route_id')
+  AND layover_start < sqlc.arg('time_range_end')
```
fletcherw, line 1028: how fast is this query on a large dataset? I know you have an index but the double range query still needs to scan extra rows.

gtfsdb/schema.sql
```diff
@@ -367,6 +367,27 @@ CREATE INDEX IF NOT EXISTS idx_block_trip_entry_trip_id ON block_trip_entry (tri
...
+        layover_end INTEGER NOT NULL,
+        next_trip_id TEXT NOT NULL,
+        FOREIGN KEY (next_trip_id) REFERENCES trips (id)
+    );
+
+-- migrate
+CREATE INDEX IF NOT EXISTS idx_block_layover_route_service_time
+    ON block_layover (route_id, service_id, layover_end, layover_start);
```
fletcherw: I don't know if the layover_start column in this index is used because you have a range query on layover_end. You could potentially just delete it.

fletcherw (commented):
do you think the unit tests for trips-for-route adequately check the blockLayoverIndices logic?

gtfsdb/schema.sql
Ahmedhossamdev: Agree it's overkill for layover windows. I kept ns for uniformity with the rest of the DB: stop_times.arrival_time/departure_time and trips.min_arrival_time/max_departure_time are all nanoseconds-since-midnight, and the handler compares layover bounds against those same columns. Switching layovers alone to seconds would mean converting on every comparison. We can in the feature drop all ns to seconds

internal/restapi/trips_for_route_handler.go
Ahmedhossamdev: Good catch, no reason, I'll refactor it to use GetRoutesByIDs

gtfsdb/query.sql
Ahmedhossamdev: Trips with null min/max have no stop_times, so they can't be active and are correctly filtered out by the SQL three-valued-logic semantics of >=/<=. I'll add a comment making that explicit so the next reader doesn't have to reason about it.

Ahmedhossamdev:
> do you think the unit tests for trips-for-route adequately check the blockLayoverIndices logic?

I think it's still no, we need to add more tests that cover the cases we have

gtfsdb/schema.sql
Ahmedhossamdev: Yeah, I'll remove it

gtfsdb/query.sql
Ahmedhossamdev: I think no, block_trip_index_id is already narrowed to this route upstream, the join to trips is a PK lookup, and the time-window predicate hits idx_trips_time_window.

gtfsdb/query.sql
Ahmedhossamdev: Tried it on our Sound Transit consolidated feed (68k rows in block_layover). ~15µs per query warm, ~200µs cold. Only layover_end is a range scan + layover_start is an in-index filter, not a second range.

gtfsdb/schema.sql
Ahmedhossamdev: Actually it is being used.  When I dropped it the query got noticeably slower (~22µs vs ~15µs warm on our 68k-row Sound Transit feed, ~50%).

gtfsdb/query.sql
fletcherw: nice thanks for checking.

gtfsdb/block_layover_test.go
```diff
@@ -0,0 +1,138 @@
...
+func TestBuildBlockLayoverIndex_PopulatesTable(t *testing.T) {
+	client := newTestClientWithRABA(t)
+	ctx := context.Background()
+
+	var total int
+	err := client.DB.QueryRowContext(ctx, "SELECT COUNT(*) FROM block_layover").Scan(&total)
+	require.NoError(t, err)
+	assert.Greater(t, total, 0, "RABA feed should produce at least one block layover")
+
+	// Invariants: every row has the required fields and layover_start <= layover_end.
```
fletcherw: Are these both enforced by sqlite already? If not, add some CHECK constraints to the table schema to ensure that they are and remove this test.

gtfsdb/block_layover_test.go
```diff
@@ -0,0 +1,138 @@
...
+	rows, err := client.DB.QueryContext(ctx, `
+		SELECT block_id, route_id, service_id, layover_stop_id, next_trip_id,
+		       layover_start, layover_end
+		FROM block_layover`)
```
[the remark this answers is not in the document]
fletcherw: This is fine to ignore. We should update steering doc to specify all prod queries should use sqlc.

internal/restapi/trips_for_route_handler.go
```diff
@@ -269,22 +273,39 @@ func (api *RestAPI) tripsForRouteHandler(w http.ResponseWriter, r *http.Request)
...
+		}
+		routeIDs := make([]string, 0, len(routeIDSet))
+		for id := range routeIDSet {
+			routeIDs = append(routeIDs, id)
+		}
+
+		routes, err := api.GtfsManager.GtfsDB.Queries.GetRoutesByIDs(ctx, routeIDs)
+		if err != nil {
+			api.Logger.Warn("trips-for-route: failed to fetch routes for agency mapping", "error", err)
+		}
```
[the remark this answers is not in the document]
fletcherw: I think this is a good suggestion.

fletcherw (approved):

PR #862, opened 2026-04-13
refactor: use typing in stops-for-location test

fletcherw:
Replace generic json response object with a typed response to eliminate verbose type casting in tests.

PR #883, opened 2026-04-22
fix: use raw GTFS direction_id for schedule/stops-for-route groupings

### LINKED ISSUE #882: fix(schedule-for-route): use raw GTFS direction_id as stopTripGrouping id
## Problem

/schedule-for-route (and /stops-for-route) returned directionId: "0" for trips whose GTFS direction_id=1, and vice versa — stops/trips are paired with the wrong direction label. Confirmed against Java reference on unitrans_E.

Ahmedhossamdev:
Fixes: #882

## Summary
- `groupTripsByDirection` re-indexed groups (descending sort), swapping CSV `direction_id` 0 ↔ 1 in the response.
- Use the raw GTFS `direction_id` string as the group id (Java parity).
- Sort `stops-for-route` stopGroups by headsign name (Java parity).

## Summary by CodeRabbit

* **Bug Fixes**
  * Stop groups are now ordered by stop/route names for more consistent display.
  * Direction groups are ordered and identified by ascending direction ID for more predictable grouping and accurate trip-to-direction matching.

* **Tests**
  * Updated tests to precisely assert direction-to-group and stop-group assignments and reflect the new ordering and ID behavior.

### DIFF
--- a/internal/restapi/schedule_for_route_handler_test.go
+++ b/internal/restapi/schedule_for_route_handler_test.go
@@ -8,8 +8,20 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"maglev.onebusaway.org/internal/clock"
+	"maglev.onebusaway.org/internal/models"
 	"maglev.onebusaway.org/internal/utils"
 )
+
+type scheduleForRouteResponse struct {
+	Code int                  `json:"code"`
+	Data scheduleForRouteData `json:"data"`
+	Text string               `json:"text"`
+}
+
+type scheduleForRouteData struct {
+	Entry      models.ScheduleForRouteEntry `json:"entry"`
+	References models.ReferencesModel       `json:"references"`
+}
 
 func TestScheduleForRouteHandler(t *testing.T) {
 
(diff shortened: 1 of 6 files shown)


internal/restapi/schedule_for_route_handler_test.go
```diff
@@ -218,38 +214,44 @@ func TestScheduleForRouteHandler_DirectionIDJavaParity(t *testing.T) {
 	require.Equal(t, http.StatusOK, resp.StatusCode)
 
 	data, ok := model.Data.(map[string]interface{})
-	require.True(t, ok)
+	require.True(t, ok, "model.Data should be a map")
```
fletcherw: I don't think these extra comments on the assertions are required, this can be inferred from the previous line.

Another option is to convert this test to use typed responses like in #862 to simplify further.

internal/restapi/schedule_for_route_handler_test.go
```diff
@@ -218,38 +214,44 @@ func TestScheduleForRouteHandler_DirectionIDJavaParity(t *testing.T) {
...
 	entry, ok := data["entry"].(map[string]interface{})
-	require.True(t, ok)
-
+	require.True(t, ok, "entry should be a map")
 	groupings, ok := entry["stopTripGroupings"].([]interface{})
-	require.True(t, ok)
-	require.Len(t, groupings, 2, "route 25_1885 has trips in both directions")
```
fletcherw, line 227: why did you delete this comment, I think this is useful context.

internal/restapi/stops_for_route_handler.go
```diff
@@ -342,7 +343,7 @@ func processTripGroups(
 
 	if len(allStopGroups) > 0 {
 		sort.Slice(allStopGroups, func(i, j int) bool {
-			return allStopGroups[i].ID < allStopGroups[j].ID
+			return strings.Join(allStopGroups[i].Name.Names, "") < strings.Join(allStopGroups[j].Name.Names, "")
```
fletcherw: this feels like a strange way to sort, are the stopGroups names in a consistent order?

fletcherw (approved):

internal/restapi/stops_for_route_handler.go
Ahmedhossamdev: Yep. Names is always a single-element slice here (populated once with []string{groupHeadsign}), so the order is consistent. I can simplify it to sort by Name.Name directly.

internal/restapi/schedule_for_route_handler_test.go
Ahmedhossamdev: Fair point, I'll restore it, I was trying with different routeIds (:

PR #889, opened 2026-04-23
refactor: logging cleanup

fletcherw:
* Move some extra verbose logs to be Debug instead of Info
* Remove the Verbose config fields since they were only used for a single log message.
* Set the global logger when building the application so that slog.Default() applies the application's config.

## Summary by CodeRabbit

* **Refactor**
  * Removed the verbose configuration option; logging is now controlled by LogLevel/LogFormat and a single runtime logger.
  * Several informational messages were demoted to debug to reduce noise at normal verbosity; startup warnings now use a dedicated early-start logger.

* **Tests**
  * Test suite updated to match the logging and configuration changes.


Ahmedhossamdev (approved):
LGTM!

PR #893, opened 2026-04-25
refactor: use time types in models

fletcherw:
In models, replace usages of int/int64 to encode duration in seconds or unix timestamp in millis with two wrapper types, ModelTime and ModelDuration.

These wrap time.Time and time.Duration, handle appropriate serialization logic, and help make the intended data in these models clearer.

Also update many places that create these model types to pass go time types directly.

## Summary by CodeRabbit

* **Refactor**
  * Modernized time/duration handling across the service to use native time and duration types instead of raw numeric timestamps.
  * Added typed time/duration wrappers with consistent JSON serialization and clearer zero-value semantics.
  * Updated APIs and responses to emit standard time/duration values, improving consistency of scheduling, vehicle, and realtime timestamps across endpoints and tests.

### DIFF
--- a/internal/restapi/trips_for_route_handler.go
+++ b/internal/restapi/trips_for_route_handler.go
@@ -55,13 +55,9 @@ func (api *RestAPI) tripsForRouteHandler(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
-	// Calculate nanoseconds since midnight of the service day
+	// Time since midnight of the service day, as a duration.
 	serviceDayMidnight := time.Date(currentTime.Year(), currentTime.Month(), currentTime.Day(), 0, 0, 0, 0, currentTime.Location())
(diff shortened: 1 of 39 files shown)


internal/restapi/trips_for_route_handler.go
```diff
@@ -80,7 +75,8 @@ func (api *RestAPI) tripsForRouteHandler(w http.ResponseWriter, r *http.Request)
 		api.Logger.Warn("trips-for-route: failed to fetch previous-day service IDs", "date", prevFormattedDate, "error", err)
 		prevServiceIDs = nil
 	}
-	prevDayNanosSinceMidnight := currentNanosSinceMidnight + oneDayNanos
+	// I'm confused by adding 24 hours to get the previous day here, but that's the existing behavior.
```
fletcherw, line 78: this kind of feels like a bug in the existing code but I don't understand it fully.

Ahmedhossamdev (approved):
LGTM!

PR #903, opened 2026-04-28
refactor: update route and stop IDs response types to match the new response model

### LINKED ISSUE #901: refactor: stop-ids and route-ids for agency tests to use the new response model

Ahmedhossamdev:
Fixes: #901 

## Summary by CodeRabbit

* **New Features**
  * API now provides explicit list response types for route and stop ID lists for an agency, improving consistency of those endpoints.
* **Tests**
  * Enhanced test infrastructure for API handlers with stronger typing, deterministic fixtures, and new cases validating malformed agency IDs (400 responses).

### DIFF
--- a/internal/restapi/response_types.go
+++ b/internal/restapi/response_types.go
@@ -35,4 +35,6 @@ type EntryData[T any] struct {
 type CoverageResponse ListResponse[models.AgencyCoverage]
 type RoutesResponse ListResponse[models.Route]
 type StopsResponse ListResponse[models.Stop]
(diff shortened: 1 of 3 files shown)


fletcherw (approved):
Can you add tests for a misformatted agency id to both of these files, using something that fails utils.ValidateID, and check that we return a 400 BadRequest?

Ahmedhossamdev:
> Can you add tests for a misformatted agency id to both of these files, using something that fails utils.ValidateID, and check that we return a 400 BadRequest?

Sure!

PR #908, opened 2026-04-29
refactor: static meta data from memory to db

### LINKED ISSUE #907: refactor: migrate meta data from memory to db

Ahmedhossamdev:
Fixes: #907 

## Summary by CodeRabbit

* **Refactor**
  * GTFS metadata (feed expiration, system ETag, import/last-update time) is now persisted to the database and served via context-aware accessors, replacing in-memory state.

* **Chores**
  * Database schema extended to include feed expiration and import timestamp; stored import time resolution increased.

* **Tests**
  * Test helpers and test suites updated to use DB-backed metadata for deterministic, context-aware behavior.


internal/gtfs/gtfs_manager.go
```diff
@@ -739,26 +735,46 @@ func (manager *Manager) IsServiceActiveOnDate(ctx context.Context, serviceID str
...
 }
 
-// GetSystemETag retrieves the SystemETag in a thread-safe manner.
-// It acquires the static data read lock to prevent data races during GTFS reloads.
+// GetSystemETag reads the system ETag from the database.
 func (manager *Manager) GetSystemETag() string {
-	manager.staticMutex.RLock()
-	defer manager.staticMutex.RUnlock()
-	return manager.systemETag
+	if manager.GtfsDB == nil {
```
fletcherw: I prefer not having nil checks like this. We should assume that all managers have a valid GtfsDB, and if tests don't, we should fix them.

internal/gtfs/gtfs_manager.go
```diff
@@ -739,26 +735,46 @@ func (manager *Manager) IsServiceActiveOnDate(ctx context.Context, serviceID str
...
+		return ""
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
```
fletcherw: this should either be `context.TODO()`, or you should add a context parameter to GetSystemETag and use it here.

internal/gtfs/gtfs_manager.go
```diff
@@ -739,26 +735,46 @@ func (manager *Manager) IsServiceActiveOnDate(ctx context.Context, serviceID str
...
+	defer cancel()
+	metadata, err := manager.GtfsDB.Queries.GetImportMetadata(ctx)
+	if err != nil || metadata.FileHash == "" {
+		return ""
+	}
+	return fmt.Sprintf(`"%s"`, metadata.FileHash)
```
fletcherw: metadata.FileHash is already a string, why the Sprintf call?

internal/gtfs/gtfs_manager.go
```diff
@@ -739,26 +735,46 @@ func (manager *Manager) IsServiceActiveOnDate(ctx context.Context, serviceID str
...
 }
 
-// FeedExpiresAt returns the parsed feed expiry time.
+// FeedExpiresAt reads the feed expiry time from the database.
 func (manager *Manager) FeedExpiresAt() time.Time {
-	manager.staticMutex.RLock()
-	defer manager.staticMutex.RUnlock()
-	return manager.feedExpiresAt
+	if manager.GtfsDB == nil {
```
fletcherw: same notes about nil checks, timeout, and context in this function and the functions below.

internal/gtfs/gtfs_manager.go
fletcherw: why add a timeout to this operation?

internal/gtfs/gtfs_manager.go
```diff
@@ -809,12 +829,14 @@ func (manager *Manager) SetFeedUpdateTimeForTest(feedID string, t time.Time) {
...
-	manager.staticMutex.Lock()
-	defer manager.staticMutex.Unlock()
-	manager.lastUpdated = t
-	manager.lastUpdatedUnixNanos.Store(t.UnixNano())
+	if manager.GtfsDB == nil {
+		return
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	defer cancel()
+	_ = manager.GtfsDB.Queries.UpdateImportTime(ctx, t.Unix())
```
fletcherw: Other database values use UnixNanos right? should we keep this consistent?

gtfsdb/schema.sql
```diff
@@ -368,7 +368,8 @@ CREATE TABLE
         id INTEGER PRIMARY KEY CHECK (id = 1), -- Only allow one row
         file_hash TEXT NOT NULL,
         import_time INTEGER NOT NULL,
-        file_source TEXT NOT NULL
+        file_source TEXT NOT NULL,
+        feed_expires_at INTEGER
```
fletcherw, line 372: why nullable for this? Do some feeds not expire?

internal/gtfs/gtfs_manager_test.go
```diff
@@ -533,29 +542,39 @@ func TestParseAndLogFeedExpiryLocked(t *testing.T) {
...
 	expectedTime2, _ := time.Parse("20060102150405", "20260405235959")
-	assert.Equal(t, expectedTime2.Unix(), manager.feedExpiresAt.Unix())
+	assert.Equal(t, expectedTime2.Unix(), manager.FeedExpiresAt().Unix())
 }
 
 func TestManager_DataFreshnessTracking(t *testing.T) {
-	manager := &Manager{
-		realTimeMutex: sync.RWMutex{},
-	}
+	// nil DB returns zero
```
fletcherw: this is mostly the existing test, but general advice I like on structuring tests: it's better to have tests focused on a single test case. This is called the arrange-act-assert pattern, description here: https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/

internal/gtfs/static.go
```diff
@@ -262,7 +252,6 @@ func (manager *Manager) ReloadStatic(ctx context.Context) (bool, error) {
 // parseAndLogFeedExpiryLocked checks the GTFS calendar for the last active service date
 // NOTE: Caller must guarantee thread-safety
 func (manager *Manager) parseAndLogFeedExpiryLocked(ctx context.Context, logger *slog.Logger) {
```
fletcherw: is this function relying on locking from the caller anymore? If not you can delete the `Locked` suffix and the Note in the doc comment.

internal/gtfs/gtfs_manager.go
Ahmedhossamdev: Yeah, we don't need the Sprintf, I'll just return metadata.FileHash directly

internal/gtfs/gtfs_manager.go
Ahmedhossamdev: I think it's unnecessary since we use SQLite.

internal/gtfs/gtfs_manager.go
Ahmedhossamdev: The previous implementation used seconds, but I think it makes sense to convert it to Unix nanoseconds to maintain consistency.

gtfsdb/schema.sql
Ahmedhossamdev: I think no, but feeds can be published without expires date

internal/gtfs/gtfs_manager_test.go
Ahmedhossamdev: Cool, thanks for sharing this!

internal/gtfs/static.go
Ahmedhossamdev: yep

internal/gtfs/static.go
```diff
@@ -221,7 +221,12 @@ func (manager *Manager) ReloadStatic(ctx context.Context) (bool, error) {
 			slog.String("source", newData.Source))
 	}
 
-	newRegionBounds := computeRegionBounds(ctx, manager.GtfsDB)
+	// Use a fresh context for all post-import bookkeeping. The caller's ctx may
+	// expire after StoreGtfsData commits, which would silently leave feed_expires_at
+	// and the ETag stale even though the import succeeded.
```
fletcherw: cleaner way to handle this is to just update bookkeeping data in the same transaction as the database import.

fletcherw (approved):

PR #914, opened 2026-05-05
refactor: introduce TripsForRouteResponse type and update related tests

Ahmedhossamdev:
- Added TripsForRouteResponse with type TripsForRouteListEntry.
- Removed verifyTripEntry and verifyRef rhey only checked for the Json keys existed, which is now guaranteed by the typed structs.
- Update the old tests with the new response type.

## Summary by CodeRabbit

* **Tests**
  * Hardened trips-for-route tests with a deterministic fixture and fixed clock; stronger typed assertions for response metadata, per-entry fields, schedule inclusion/exclusion, and malformed-ID handling (400).

* **Refactor**
  * Standardized the trips-for-route response format to improve consistency and predictability of returned data.


internal/restapi/trips_for_route_handler_test.go
```diff
@@ -50,127 +50,32 @@ func TestTripsForRouteHandler_DifferentRoutes(t *testing.T) {
...
-			references := data["references"].(map[string]any)
-			verifyReferences(t, references)
-
-			assert.GreaterOrEqual(t, len(list), tt.minExpected)
-			assert.LessOrEqual(t, len(list), tt.maxExpected)
 		})
 	}
 }
 
-func verifyTripEntry(t *testing.T, trip map[string]any) {
```
fletcherw, line 85: if the handler returned empty json for each response entry, would the new test catch it? You should add some assertions about the response values to improve the coverage for this test.

Not simple assertions like `assert.Contains` but actual validation for the returned data.

internal/restapi/trips_for_route_handler_test.go
```diff
@@ -254,7 +143,9 @@ func TestStripNumericSuffix(t *testing.T) {
 
 func TestCollectStopIDsFromSchedule_NilSchedule(t *testing.T) {
 	stopIDsMap := map[string]bool{}
+
```
fletcherw, line 274: this is a matter of style, so you don't need to change what you've done here, but for short tests you don't _need_ spacing between the Arrange/Act/Assert sections. It helps more as the tests get longer.

internal/restapi/trips_for_route_handler_test.go
Ahmedhossamdev: Done!

burma-shave (commented):
### Consider removing dead types in the models package

Now that `restapi.TripsForRouteResponse` (the new alias) is in place, two types in `internal/models/trips_for_route.go` appear to be unused and can be cleaned up:

| Type | Can remove? | Reason |
|---|---|---|
| `models.TripsForRouteResponse` | ✅ Yes | Shadowed by the new `restapi.TripsForRouteResponse` alias; no non-test callers remain |
| `models.TripsForRouteData` | ✅ Yes | Only referenced by the dead struct above |
| `models.TripsForRouteListEntry` | ❌ No | Still the element type for `ListResponse[models.TripsForRouteListEntry]` throughout the handler and tests |

There's also a subtle naming collision: `models.TripsForRouteResponse` (old struct) and `restapi.TripsForRouteResponse` (new alias) coexist in different packages, which compiles fine but can cause confusion. Removing the models one would resolve that ambiguity.

burma-shave (changes requested):
Just needs merge conflicts resolved and address the comment here:
https://github.com/OneBusAway/maglev/pull/914#pullrequestreview-4348453412

burma-shave (approved):

PR #918, opened 2026-05-08
refactor: arrivals_and_departures tests to use the new response model

Ahmedhossamdev:
## Summary by CodeRabbit

* **New Features**
  * Arrivals/departures responses now include grouped entries with stop context, nearby stop IDs, and situation references.

* **Tests**
  * Handler and benchmark tests refactored for deterministic, typed response validation and a shared URL builder; tests now support injecting a real clock for stable timing.

* **Refactor**
  * Response typing and internal arrivals/departures grouping improved for clearer data representation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OneBusAway/maglev/pull/918)


internal/restapi/arrivals_and_departures_for_stop_handler_test.go
```diff
@@ -18,22 +19,37 @@ import (
...
 
-func TestArrivalsAndDeparturesForStopHandlerRequiresValidApiKey(t *testing.T) {
-	api, cleanup := createTestApiWithRealTimeData(t)
-	defer cleanup()
-
-	time.Sleep(500 * time.Millisecond)
+func arrivalsAndDeparturesURL(stopID string, extras ...string) string {
+	var b strings.Builder
+	b.WriteString("/api/where/arrivals-and-departures-for-stop/")
+	b.WriteString(stopID)
```
fletcherw: optional: golang has support for encoding query parameters that you could use: https://pkg.go.dev/net/url#Values.Encode

internal/restapi/arrivals_and_departures_for_stop_handler_test.go
```diff
@@ -18,22 +19,37 @@ import (
...
+	return utils.FormCombinedID(agency.ID, stops[0].ID)
+}
 
-	if len(stops) == 0 {
-		t.Skip("No stops available for testing")
-	}
+func TestArrivalsAndDeparturesForStopHandlerRequiresValidApiKey(t *testing.T) {
+	api, cleanup := createTestApiWithRealTimeData(t)
+	defer cleanup()
+	time.Sleep(500 * time.Millisecond)
```
fletcherw: why are there sleeps here and elsewhere in these tests?

internal/restapi/arrivals_and_departures_for_stop_handler_test.go
```diff
@@ -18,22 +19,37 @@ import (
...
+		b.WriteByte('&')
+		b.WriteString(e)
+	}
+	return b.String()
+}
 
+// firstStopID returns a combined stop ID built from the first agency and the
+// first active stop in the test fixture. Fails the test if the fixture has no
+// stops (RABA always does).
+func firstStopID(t *testing.T, api *RestAPI) string {
```
fletcherw: this is a broader test pattern that I don't like: randomly picking data from the database and running tests with it. I would rather see explicit agency and stop IDs from our RABA test data. I think we should discuss this more separately from this PR.

For example, below we say `if len(entry.ArrivalsAndDepartures) > 0` meaning that we may not be testing that behavior if the stop we pick doesn't have arrivals and departures.

internal/restapi/arrivals_and_departures_for_stop_handler_test.go
```diff
@@ -44,357 +60,132 @@ func TestArrivalsAndDeparturesForStopHandlerRequiresValidApiKey(t *testing.T) {
...
+	assert.Equal(t, stopID, entry.StopID)
+	assert.NotNil(t, entry.NearbyStopIDs)
+	assert.NotNil(t, entry.SituationIDs)
 
-	time.Sleep(500 * time.Millisecond)
+	require.NotEmpty(t, model.Data.References.Agencies)
 
-	agency := mustGetAgencies(t, api)[0]
-	stops := mustGetStops(t, api)
+	if len(entry.ArrivalsAndDepartures) > 0 {
```
fletcherw: change this to a `require.NotEmpty(entry.ArrivalsAndDepartures)` and then iterate over all arrivals and departures and assert these conditions for each.

internal/restapi/arrivals_and_departures_for_stop_handler_test.go
```diff
@@ -44,357 +60,132 @@ func TestArrivalsAndDeparturesForStopHandlerRequiresValidApiKey(t *testing.T) {
...
-
-	agencies, ok := references["agencies"].([]any)
-	assert.True(t, ok)
-	assert.NotEmpty(t, agencies)
-
-	if len(arrivalsAndDepartures) > 0 {
-		firstArrival, ok := arrivalsAndDepartures[0].(map[string]any)
-		assert.True(t, ok)
-
-		assert.Contains(t, firstArrival, "stopId")
```
fletcherw, line 103: do we need to check that the response has non-empty values for these fields or will the json deserialization ensure that?

internal/restapi/arrivals_and_departures_for_stop_handler_test.go
```diff
@@ -44,357 +60,132 @@ func TestArrivalsAndDeparturesForStopHandlerRequiresValidApiKey(t *testing.T) {
...
-	assert.Equal(t, http.StatusOK, resp.StatusCode)
-	assert.Equal(t, 200, model.Code)
-
-	data, ok := model.Data.(map[string]any)
-	assert.True(t, ok)
-
-	entry, ok := data["entry"].(map[string]any)
-	assert.True(t, ok)
-	assert.Equal(t, stopID, entry["stopId"])
+	specificTimeMs := specificTime.Unix() * 1000
```
fletcherw: nit: specificTime.UnixMillis(). Also happens later in the file.

internal/restapi/arrivals_and_departures_for_stop_handler_test.go
```diff
@@ -44,357 +60,132 @@ func TestArrivalsAndDeparturesForStopHandlerRequiresValidApiKey(t *testing.T) {
...
+
+	tests := []struct {
+		name   string
+		extras []string
+	}{
+		{"minutesAfter and minutesBefore", []string{"minutesAfter=60", "minutesBefore=10"}},
+		{"absolute time param", []string{"time=" + strconv.FormatInt(specificTimeMs, 10)}},
```
fletcherw: nit: `fmt.Sprintf("time=%d", specificTimeMs)` is probably more readable.

internal/restapi/arrivals_and_departures_for_stop_handler_test.go
```diff
@@ -44,357 +60,132 @@ func TestArrivalsAndDeparturesForStopHandlerRequiresValidApiKey(t *testing.T) {
...
-func TestArrivalsAndDeparturesForStopHandlerWithTimeParameters(t *testing.T) {
-	api, cleanup := createTestApiWithRealTimeData(t)
-	defer cleanup()
+	entry := model.Data.Entry
+	assert.Equal(t, stopID, entry.StopID)
+	assert.NotNil(t, entry.NearbyStopIDs)
+	assert.NotNil(t, entry.SituationIDs)
 
-	time.Sleep(500 * time.Millisecond)
+	require.NotEmpty(t, model.Data.References.Agencies)
```
fletcherw: you can make a stronger assertion that
`assert.elementsMatch(t, testdata.Raba, model.Data.References.Agencies)`

here and elsewhere

internal/restapi/arrivals_and_departures_for_stop_handler_test.go
```diff
@@ -44,357 +60,132 @@ func TestArrivalsAndDeparturesForStopHandlerRequiresValidApiKey(t *testing.T) {
...
+		extras []string
+	}{
+		{"minutesAfter and minutesBefore", []string{"minutesAfter=60", "minutesBefore=10"}},
+		{"absolute time param", []string{"time=" + strconv.FormatInt(specificTimeMs, 10)}},
+	}
 
-	assert.Contains(t, entry, "arrivalsAndDepartures")
-	assert.Contains(t, entry, "nearbyStopIds")
-	assert.Contains(t, entry, "situationIds")
+	for _, tt := range tests {
```
fletcherw, line 111: nice. this is more idiomatic go and lets us have more test cases

internal/restapi/arrivals_and_departures_for_stop_handler_test.go
```diff
@@ -44,357 +60,132 @@ func TestArrivalsAndDeparturesForStopHandlerRequiresValidApiKey(t *testing.T) {
...
-	agency := mustGetAgencies(t, api)[0]
-	stops := mustGetStops(t, api)
-
-	if len(stops) == 0 {
-		t.Skip("No stops available for testing")
-	}
-
-	stopID := utils.FormCombinedID(agency.ID, stops[0].ID)
+	stopID := firstStopID(t, api)
+	futureTime := time.Now().AddDate(10, 0, 0)
```
fletcherw: just pick an absolute time rather than 10 days from `now()`

internal/restapi/arrivals_and_departures_for_stop_handler_test.go
```diff
@@ -441,165 +226,112 @@ func TestArrivalsAndDeparturesForStopHandlerWithInvalidParams(t *testing.T) {
...
 	require.NoError(t, err)
-
 	_, err = queries.CreateStopTime(ctx, gtfsdb.CreateStopTimeParams{
-		TripID:        tripB_ID,
-		StopID:        stopID,
-		StopSequence:  1,
-		ArrivalTime:   28800 * 1e9, // 08:00:00 converted to nanoseconds
-		DepartureTime: 29100 * 1e9, // 08:05:00 converted to nanoseconds
+		TripID: tripBID, StopID: stopID, StopSequence: 1,
+		ArrivalTime:   28800 * 1e9, // 08:00:00
```
fletcherw: optional nit: `8 * time.Hour` and `8 * time.Hour + 5 * time.Minute`  are more readable.

internal/restapi/arrivals_and_departures_for_stop_handler_test.go
Ahmedhossamdev: Looks great, I'll use it. 
Thanks for sharing!

internal/restapi/arrivals_and_departures_for_stop_handler_test.go
Ahmedhossamdev: I think it was stale when the initial real-time fetch was async

internal/restapi/arrivals_and_departures_for_stop_handler_test.go
Ahmedhossamdev: Yeah, I agree with that.

internal/restapi/arrivals_and_departures_for_stop_handler_test.go
Ahmedhossamdev: Yes, now we use assert.Equal - assert.NotEmpty

PR #919, opened 2026-05-08
refactor: extract sql null helpers

fletcherw:
Move these helper functions for handling sql null values into a new package with no dependencies so that it can be used across all of the maglev code base.

Delete the existing duplicated helpers from gtfsdb.

## Release Notes

* **Refactor**
  * Consolidated nullable-value helpers into a single internal module and removed duplicate utilities.
  * Updated API response construction and model mapping to use the new null-safe helpers.
  * Added unit tests validating the new nullable-to-value conversions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OneBusAway/maglev/pull/919)


Ahmedhossamdev (approved):
Great work!

PR #920, opened 2026-05-10
feat: add host binding configuration option

### LINKED ISSUE #896: Binding Address Configuration
In addition to the port, it would be great if the binding address could be set via the configuration file. For example to set the binding address for the server to 127.0.0.1:4000 instead of binding to all addresses *:4000.

FLASH2332:
Adds a `host` field to the configuration file to control the server's binding address, allowing operators to restrict the server to a specific interface (e.g. `127.0.0.1`) instead of binding to all interfaces (`0.0.0.0`).

The field is optional and defaults to `0.0.0.0` to preserve existing behavior.

image

</br>

Fixes #896 

@aaronbrethorst     
@chrisls121 

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring the server bind host through configuration files and the command line.
  * Server addresses now correctly support custom hosts, ports, and IPv6 formatting.
  * Updated configuration examples and schema with the new host setting, including a local default example.

* **Tests**
  * Added coverage verifying servers bind to wildcard, IPv4, and IPv6 host-and-port combinations.


FLASH2332:
Shall I update the appropriate comments and documentation for this addition ?

FLASH2332:
@aaronbrethorst @Ahmedhossamdev Should I make further changes ?

aaronbrethorst:
### Code review

Found 4 issues:

1. Defaulting the empty host to `0.0.0.0` is not backward compatible. The previous `Addr: ":port"` makes Go bind the wildcard as a dual-stack IPv6 socket (`[::]:port` with `IPV6_V6ONLY=0`), which accepts both IPv4 and IPv6 clients. `0.0.0.0:port` parses to an IPv4 address and binds `AF_INET` only, so IPv6 clients silently stop being able to reach the server — including anything hitting `localhost` where that resolves to `::1` first (e.g. the container `HEALTHCHECK` in `Dockerfile`). To actually preserve existing behavior the empty default should stay `""`, so the address remains `:port`.

https://github.com/OneBusAway/maglev/blob/229117a88914f23b9eb4730ef04f88d2826d0b27/cmd/api/app.go#L152-L157

2. `fmt.Sprintf("%s:%d", host, cfg.Port)` produces an invalid address for any IPv6 host. `host: "::1"` yields `::1:4000` and `host: "::"` yields `:::4000`; `net.Listen` rejects both with "too many colons in address", so the server fails to start. Since the whole point of the option is picking a bind interface, IPv6 literals are a legitimate value — `net.JoinHostPort(host, strconv.Itoa(cfg.Port))` handles the bracketing and also returns `:4000` for an empty host.

https://github.com/OneBusAway/maglev/blob/229117a88914f23b9eb4730ef04f88d2826d0b27/cmd/api/app.go#L204-L206

3. `config.example.json` sets `"host": "127.0.0.1"`, which changes the bind behavior of the documented onboarding path. Both `README.markdown` ("cp config.example.json config.json") and `CLAUDE.md` ("Copy `config.example.json` to `config.json`") tell operators to copy this file verbatim, so every fresh setup would go from all-interfaces to loopback-only — unreachable from a VM host, a LAN device, or a container. It also contradicts the `"default": "0.0.0.0"` documented one file over in `config.schema.json` and the PR description's "defaults to `0.0.0.0` to preserve existing behavior". Either omit the key from the example or set it to the documented default.

https://github.com/OneBusAway/maglev/blob/229117a88914f23b9eb4730ef04f88d2826d0b27/config.example.json#L3-L5

4. The new conditional has no test coverage for the configured-host path. The only test change updates the existing assertion, which still exercises the empty-host branch; nothing anywhere asserts that a non-empty `host` (from JSON or the `-host` flag) actually reaches `srv.Addr`. CONTRIBUTING.md says "Every new branch or condition introduced by a PR needs test coverage."

https://github.com/OneBusAway/maglev/blob/229117a88914f23b9eb4730ef04f88d2826d0b27/cmd/api/app_test.go#L168-L170

aaronbrethorst (changes requested):
Thanks for this — a configurable bind address is a genuinely useful option to have, and the plumbing is done thoroughly: `JSONConfig`, `ToAppConfig`, the CLI flag, the schema, and the `-f` mutual-exclusion path are all handled consistently with how `port` is threaded through. That's the tedious part and you got it right.

There are three things I need changed before this can land, all in the same area: the defaults change behavior rather than preserve it.

**1. `0.0.0.0` is not the current default behavior.**

Today `Addr` is `fmt.Sprintf(":%d", cfg.Port)`. Go binds a bare `:port` as a dual-stack IPv6 wildcard socket (`[::]:port` with `IPV6_V6ONLY=0`), which accepts both IPv4 and IPv6 connections. `0.0.0.0:port` binds `AF_INET` only, so IPv6 clients silently stop being able to connect. That's the opposite of the PR's stated goal of preserving existing behavior, and it's the kind of regression that won't show up until someone deploys on an IPv6 network.

The fix is to make the empty host stay empty rather than substituting a literal:

```go
srv := &http.Server{
    Addr: net.JoinHostPort(cfg.Host, strconv.Itoa(cfg.Port)),
    ...
}
```

`net.JoinHostPort("", "4000")` yields `":4000"`, exactly matching today's behavior. Please also update the schema `default` and the flag help text away from `0.0.0.0` — describing it as "all interfaces" is fine, but `0.0.0.0` specifically is the IPv4-only wildcard.

**2. `fmt.Sprintf("%s:%d", host, cfg.Port)` can't express an IPv6 address.**

`-host ::1` produces `::1:4000`, which `net.Listen` rejects with "too many colons in address". So the new option works for hostnames and IPv4 literals but not IPv6 ones. `net.JoinHostPort` (above) handles the bracketing and fixes this at the same time. CodeRabbit flagged this one too.

**3. `config.example.json` shouldn't set `"host": "127.0.0.1"`.**

Both README.markdown and CLAUDE.md instruct users to copy `config.example.json` to `config.json` verbatim as the setup step. With this line added, every fresh install becomes loopback-only — a new contributor's first `make run` would work locally but be unreachable from anything else, with nothing obvious pointing at the cause. It also contradicts the schema's own `"default": "0.0.0.0"` in the same PR.

Please drop the key from `config.example.json`. The `examples` block in `config.schema.json` is a fine place to show `"host": "127.0.0.1"` as an illustration, since nobody copies that one.

A couple of smaller notes, not blockers:

- The default would sit more naturally in `JSONConfig.setDefaults()` (`internal/appconf/json_config.go`), which is where `port`, `env`, `rate-limit`, `data-path`, and `log-level` all resolve their defaults. Putting it in `CreateServer` means the resolved value never exists at the config layer. This becomes moot if you go with `net.JoinHostPort` and an empty default, so use your judgment.
- `TestCreateServer` only covers the empty-host path. A case asserting that a configured host actually reaches `srv.Addr` would be worth adding — it's the one branch the feature introduces.
- The `host` block in `config.schema.json` is indented 4 spaces where its siblings use 6.

Happy to re-review as soon as the bind-address defaults are sorted — the rest of this is in good shape.

aaronbrethorst:
Checking in on this one — it's been sitting since my review on July 30.

The only commit since then (`e235aa24`) is a merge of upstream `main`; comparing `229117a8...e235aa24` shows no changes to `cmd/api/`, `internal/appconf/`, `config.example.json`, or `config.schema.json`. So the three blocking items from that review are all still open:

1. **Empty host should stay `""`, not `0.0.0.0`.** `cmd/api/app.go` still defaults to the literal `0.0.0.0`, which gives an IPv4-only `AF_INET` socket; `":port"` gives a dual-stack listener. Our own `Dockerfile` health-checks `http://localhost:4000/healthz`, which resolves `::1` first on dual-stack images — so this regresses in-repo. The schema `default` and the flag help text need to move off `0.0.0.0` too.
2. **Use `net.JoinHostPort(host, strconv.Itoa(cfg.Port))`.** `fmt.Sprintf("%s:%d", ...)` can't express an IPv6 address — `-host ::1` produces `::1:4000` and fails with "too many colons in address". For a feature whose whole purpose is choosing a bind interface, IPv6 values are ordinary input.
3. **Drop `"host": "127.0.0.1"` from `config.example.json`.** Both README.markdown and CLAUDE.md tell users to copy that file verbatim, so every fresh install would come up loopback-only — and it contradicts the schema's own `0.0.0.0` default in the same PR.

The two optional notes (moving the default into `setDefaults()`, and a table-driven `TestCreateServer` case covering the configured-host branch) are still open as well; the second is worth doing regardless, since CONTRIBUTING asks for coverage on every new branch.

No rush on my account, but let me know if you'd rather I take it over or close it — happy either way, and the feature itself is one we want.

FLASH2332:
Hi Aaron, sorry for the delay! I'll get the three blocking changes done — switching to net.JoinHostPort, dropping "host" from config.example.json, and fixing the schema default. I'll also add the test case for the configured-host branch. Will push shortly.

FLASH2332:
Hi, I've addressed all the blocking items and the optional notes. Ready for review

aaronbrethorst:
### Code review

aaronbrethorst (approved):

All four items from my earlier review are addressed, and I checked each one against the code rather than taking the changelog's word for it:

- `net.JoinHostPort(cfg.Host, ...)` with an empty host produces `:8080`, so the default stays the dual-stack wildcard listener. That was the important one — forcing `0.0.0.0` would have broken the Dockerfile healthcheck.
- IPv6 literals bracket correctly now that `fmt.Sprintf("%s:%d", ...)` is gone.
- `config.example.json` is no longer touched, and the schema default is `""` with a matching description.
- The configured-host branch has real coverage via the table-driven `TestCreateServer` (empty / `127.0.0.1` / `::1`).

Declining CodeRabbit's suggestion to default `Host` in `setDefaults()` was the right call — that would have reintroduced the first problem.

Two cosmetic things you can fold into a future PR rather than respinning this one: there are stray blank lines added at the top of `CreateServer` and `TestCreateServer`, and the new `host` block in `config.schema.json` is indented 4 spaces where its siblings use 6.

Thanks for sticking with this one through several rounds.

PR #924, opened 2026-05-11
Remove dead code from internals

Noviv:
This code is unused and likely doesn't need to stick around.


Ahmedhossamdev:
LGTM!
I'll merge it!
Thanks, @Noviv

PR #940, opened 2026-05-14
ci: Drop golangci-lint

burma-shave:
Golangci-lint is causing issues when bumping other dependencies due to its reliance on specific minor versions of go. Linting is already done through go vet and sonar. Drop the dependency to make bumping other dependencies, and the version of go itself easier.

## Summary by CodeRabbit

* **Chores**
  * Simplified the CI/CD pipeline by removing the lint job from automated workflow checks.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OneBusAway/maglev/pull/940)

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -34,29 +34,6 @@ jobs:
             exit 1
           fi
 
-  lint:
-    name: Lint
-    runs-on: ubuntu-latest
-    env:
-      CGO_ENABLED: 1
-    steps:
-      - uses: actions/checkout@v6
-
-      - name: Set up Go
-        uses: actions/setup-go@v6
-        with:
-          go-version-file: 'go.mod'
-          cache: true
-
-      - name: Download dependencies
-        run: go mod download
-
-      - name: Run linter
-        uses: golangci/golangci-lint-action@v3
-        with:
-          version: latest
-          args: --timeout=5m --build-tags='sqlite_fts5 sqlite_math_functions'
-
   test:
     name: Test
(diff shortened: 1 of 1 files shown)


Not-Dhananjay-Mishra:
@burma-shave I think rather than dropping `golangci-lint` we should only enable necessary linters via `.golangci.yml` config file.
Using a minimal and carefully selected linter set would reduce CI/versioning issues while still preserving important static analysis and code quality checks. 
Thoughts?

burma-shave:
Thanks for the suggestion! Unfortunately, limiting the linter set via `.golangci.yml` wouldn't actually solve the problem here.

The breakage happens before any linter runs. The chain looks like this:

1. Dependabot bumps a dependency
2. That dependency's `go.mod` declares a `toolchain` directive (e.g. `toolchain go1.24.3`)
3. Go's toolchain forwarding picks this up and resolves a newer minor version
4. golangci-lint parses Go AST and is tightly coupled to specific Go minor versions — it breaks regardless of which linters are enabled

Which linters are active has no bearing on that. The issue is golangci-lint's sensitivity to Go minor versions at the analysis level, not the behaviour of individual linters. (We also tried `GOTOOLCHAIN=local` — it just changes the failure mode from "download a newer toolchain" to "error out"; the underlying conflict remains.)

Beyond the stability issue, we'd also prefer to keep third-party CI dependencies minimal. `go vet` catches the most important correctness issues natively, and SonarCloud covers additional static analysis — together they give us enough coverage without the fragility golangci-lint introduces.

Issue #895, opened 2026-04-27
Arrivals and Departures for Stop Endpoint numberOfStopsAway


chrisls121:
Using the Java version of OBA the numberOfStopsAway field is being populated with valid data when querying the arrivals-and-departures-for-stop endpoint. Querying the same endpoint, arrivals-and-departures-for-stop/HSR_355573.json for example, the numberOfStopsAway always shows -1.

Below are the GTFS feeds we are using.

Static Feed
https://opendata.hamilton.ca/GTFS-Static/google_transit.zip

Realtime Feeds
https://opendata.hamilton.ca/GTFS-RT/GTFS_TripUpdates.pb
https://opendata.hamilton.ca/GTFS-RT/GTFS_VehiclePositions.pb
https://opendata.hamilton.ca/GTFS-RT/GTFS_ServiceAlerts.pb

Ahmedhossamdev:
Hi @chrisls121,
I'll take a look about this.

Ahmedhossamdev:
Hi @chrisls121 ,

I’ve been working on a fix for this issue and I believe I found the cause of why `numberOfStopsAway` is always returning `-1` in Maglev for the `arrivals-and-departures-for-stop` endpoint.

I’m currently validating the behavior against the Java OBA implementation and the HSR GTFS/GTFS-RT feeds you shared. I should have a PR up soon for review.

Would also appreciate your feedback once the PR is open, especially to help confirm the behavior matches what you’re seeing in production.

chrisls121:
Thanks @Ahmedhossamdev, I appreciate your work on this. I would be happy to test once the PR is up.

Ahmedhossamdev:
Hey @chrisls121,
Do you have an OBA deployment that consumes the HSR feeds and that you use for testing?

chrisls121:
Hi @Ahmedhossamdev, we do. We are using Maglev as a backend to a small proxy service that adds authentication, caching, and the ability to query arrivals/departures for multiple stops in one call. This data is used to drive digital signage departure displays.

Ahmedhossamdev:
Hi @chrisls121,

Thanks for the context!

I've opened a draft PR for the fix: https://github.com/OneBusAway/maglev/pull/1000

The PR focuses on bringing Maglev's trip status calculations in line with the Java OBA implementation, including numberOfStopsAway, distanceFromStop, and tripStatus. I've been validating the behavior against both the Java server and the HSR GTFS & GTFS-RT feeds.

[...]

PR #962, opened 2026-05-18
Fix/arrival-and-departure-for-stop endpoint

3rabiii:
### Description
This PR addresses several critical functional parity gaps in the `arrival-and-departure-for-stop` API endpoint to ensure it matches the legacy OneBusAway implementation and strictly follows the API specification. 

### Key Changes
- Gap 1 `vehicleId`: Formats the vehicle ID into the combined `{agencyId}_{vehicleId}` structure instead of returning the raw GTFS ID.

- Gap 2: Sparse Sequence & Arrival/Departure Fix`: Computed a 0-based stop_ordinal in SQL to handle sparse GTFS sequences. This stable ordinal dynamically controls the arrival/departureEnabled flags and prevents 404 errors by safely matching client echo-requests.

- Gap 3 `routeShortName`: Implements the 3-step narrative fallback for route short names.
- Gap 4 `tripHeadsign`: Implements the stop-level narrative override for trip headsigns.
- Gap 5 `lastUpdateTime`: Fixes the JSON serialization tag so it correctly emits `0` when there is no real-time vehicle data, instead of omitting the field entirely.
- Gap 6 `status` & Cancelled Trips: Properly handles `CANCELED` schedule relationships by setting the entry `status` to `"CANCELED"`, zeroing out predictions, stripping the vehicle ID, and omitting the `tripStatus` object.
- Gap 7 Dynamic `status`: Maps non-scheduled GTFS-RT statuses (like `ADDED` or `DUPLICATED`) directly to the top-level entry `status` instead of unconditionally returning `"default"`.
- Gap 8 Prediction Zeroing: Ensures `predictedArrivalTime` and `predictedDepartureTime` are reset to `0` when `predicted` is `false`, rather than erroneously falling back to the scheduled times.
- Gap 9 serviceDate Parsing: Supports YYYY-MM-DD date strings in the serviceDate parameter and correctly localizes dates to the agency's timezone to prevent calendar-day shifts.
- Gap 10 time Parsing: Supports yyyy-MM-dd_HH-mm-ss datetime strings in the time parameter and localizes it to the agency's timezone.
- Gap 11 includeReferences: Supports includeReferences=false for the singular endpoint to omit reference details (Agencies, Routes, Stops, Trips) from the response when requested.
- Gap 12 stopSequence Search: Implements the expand-outward search algorithm for stopSequence matching, permitting tolerance to shifted indices from schedule modifications.
- Gap 13 Closest Stop Visit: Resolves loop route ambiguity by selecting the closest scheduled stop visit using SQLite-driven absolute time difference calculations when stopSequence is absent.
-  Gap 14 blockTripSequence Fix: Accurately computes the 0-based trip sequence within a block by filtering active services. Updates the sentinel value to -1 for blockless trips and ensures the first trip (index 0) is correctly sent to the client.
-  Gap 15 Fix numberOfStopsAway: The numberOfStopsAway field in the arrival-and-departure-for-stop API was incorrectly returning 0 in Maglev. The previous implementation strictly relied on the GTFS-RT CurrentStopSequence to locate the vehicle, which is frequently unpopulated in many feeds 
     - Changes:
        - Missing Data Fallback: Updated getNumberOfStopsAway to accept the tripStatus object.
        - New Block Sequence Helper: Introduced getBlockSequenceForStopID in block_sequence_helper.go to seamlessly translate a raw stopID into its global block sequence ordinal.

        - Distance-Based Lookup Fallback: Refactored buildTripStatusCore in trips_helper.go to compute shape points and distances earlier.
        - Formula Correction: Fixed the calculation logic to use a direct ordinal subtraction (target - vehicle) instead of the previous off-by-one formula (target - vehicle - 1).
        - Testing: Added new test cases (TestGetNumberOfStopsAway_Valid and TestGetNumberOfStopsAway_UsesTripStatusNextStop) to verify the correct ordinal differences and ensure both data paths resolve accurately.

## Summary by CodeRabbit

* **Bug Fixes & Improvements**
  * Vehicle ID formatting and prediction/cancellation handling improved; canceled trips clear prediction and vehicle info.
  * Stop indexing/ordinals are now zero‑based; nearest-stop selection favors time/distance proximity; numberOfStopsAway uses trip status when available and defaults to 0.
  * Request time/serviceDate parsing accepts Unix ms or formatted strings and localizes inputs more accurately.
  * Block/sequence fallbacks now return -1 for missing/invalid cases.

* **Chores**
  * Minor JSON tag and cursor-close error handling fixes.

* **Tests**
  * Expanded/reshaped tests for stop selection, trip/block indexing, DST edge cases, includeReferences behavior, and stop-sequence handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OneBusAway/maglev/pull/962?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


3rabiii:
@coderabbitai full review

Ahmedhossamdev:
Hey @3rabiii 
Just a heads-up. I've been working on the numberOfStopsAway and distance-related fields while addressing issue https://github.com/OneBusAway/maglev/issues/895.

I ended up changing the logic to handle both cases: when GTFS-RT data exists and when it doesn't. I'm planning to open my PR later today after I finish a bit more testing, so there might be some overlap/conflicts with your changes around numberOfStopsAway.

I'll take a look through your PR as well once I get a chance 🙂

3rabiii:
take your time, @Ahmedhossamdev  I also wanted to share the results after my recent changes. Things are looking much better now. However, there is a minor, rare issue I noticed. Sometimes, when testing with different trip_ids, there is a slight delay in Maglev, the value is off by 1 compared to the Java server, If I refresh after a few seconds, it updates and matches the Java server perfectly. I'm not entirely sure what causes this slight sync delay, but in most cases, the values match exactly 
42d0631a-f30b-41f8-a030-c4955cd5c4cf

Ahmedhossamdev:
Good work, @3rabiii 

As you can see, it might display the correct numberOfStopsAway, but if you add something like minutesBefore or minutesAfter (for example 200 minutes), it may show different values because Maglev doesn’t handle this correctly at the moment.

Also, the distanceFromStops is inaccurate in Maglev. From your screenshot, it shows 20 km in Maglev, while it is 2.5 km in Java.

So we need to make numberOfStops and distanceFromStops consistent. I’ve made good progress on this, but I’m currently fixing some corner cases related to delay calculation, etc.

I’ll open a PR soon after I finish.

Issue #972, opened 2026-05-28
spec-gap: Enforce explicit version validation in route-ids-for-agency handler


3rabiii:
**Description**
During the specification review for `route-ids-for-agency`, we identified a gap in how the API version is handled. 

The original OneBusAway specification dictates returning an error for invalid versions with the message `"unknown version: {n}"`. While the legacy spec suggested returning a `500` error, returning an **HTTP 400 Bad Request** is the semantically correct behavior since passing an unsupported version is a client-side fault, not an internal server failure.

Currently, the Go implementation ignores the `version` query parameter entirely and proceeds to return a `200 OK` successful response regardless of the provided version number.

**Current Behavior**
* Requesting `/api/where/route-ids-for-agency/1.json?key=TEST&version=99` ignores the invalid version and returns HTTP `200 OK`.

**Expected Behavior (Spec Compliance & REST Standard)**
* The handler should explicitly parse the `version` query parameter.
* If the `version` is provided and is not `"2"`, the handler must immediately return an **HTTP 400 Bad Request** error with the exact text `unknown version: {n}`.

**Tests**
* Add a unit test (e.g., `TestRouteIdsForAgencyInvalidVersion`) to verify that passing `version=99` correctly triggers the `400 Bad Request` error and returns the expected text.
### Proof

<img width="1870" height="376" alt="Image" src="https://github.com/user-attachments/assets/dd98292c-c584-43b1-b5dc-482de498bb5c" />

Issue #971, opened 2026-05-28
spec-gap: Add missing test assertions and edge cases for route-ids-for-agency


3rabiii:
### Description
The `route-ids-for-agency` endpoint implementation is solid, but the test suite is missing coverage for several specific requirements detailed in the specification. We need to expand the test coverage to ensure full spec compliance and prevent future regressions.

### Missing Coverage (Spec Gaps)
1. **Extension 4b (Zero Routes):** There is no test verifying that an agency existing in the database but having zero routes returns an HTTP 200 with an empty list `[]`.
2. **limitExceeded:** The success tests do not verify that `limitExceeded` is strictly `false`.
3. **References Block:** The tests only verify that `references.agencies` is empty. The spec requires asserting that **all six** sub-lists (`agencies`, `routes`, `stops`, `trips`, `situations`, `stopTimes`) are empty.
4. **Envelope Metadata:** Success tests lack assertions for `version == 2` and `currentTime > 0`.

PR #976, opened 2026-05-28
Fix/route-ids-for-agency endpoint

### LINKED ISSUE #970: spec-gap: Fix unknown agency returning 200 with null instead of 404 in route-ids-for-agency
### Description
While reviewing the `route-ids-for-agency` endpoint against the OneBusAway specification, we found that the Go implementation replicates a legacy Java defect. 

When a requested agency is not found, the handler currently returns an HTTP 200 OK with a `null` body. According to the spec (Extension 4a / Suspected Defects), the intended behavior is to return an HTTP 404 error.

Additionally, the existing unit test `TestInvalidAgencyIdForRouteIds` is incorrectly written to assert this defective `200 + null` behavior instead of the specified `404` behavior.

### Current Behavior
- Handler uses `api.sendNull(w, r)` resulting in a `200` status with `null`.
- `TestInvalidAgencyIdForRouteIds` asserts `http.StatusOK`.

### Expected Behavior (Spec Compliance)
- Handler should use `api.sendNotFound(w, r)` to return a standard 404 envelope.
- The test should be updated to assert `http.StatusNotFound` and the appropriate error code/text.
### Proof

Image

3rabiii:
Fix: #970, #971, #972 

## Summary by CodeRabbit

* **New Features**
  * Added support for deleting a single agency record.

* **Bug Fixes**
  * Route IDs for Agency endpoint returns 404 when the agency is not found.
  * Endpoint validates the version query parameter and returns an error for unsupported versions.
  * Improved robustness of result-set cleanup to avoid unhandled close errors.

* **Tests**
  * Expanded tests for the Route IDs endpoint, including invalid-version and empty-agency scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OneBusAway/maglev/pull/976?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


burma-shave (commented):
This looks like a good start. But the diff mixes quite a few different concerns. Please open a PR for each separate issue addressed. This'll make each change easier to review independently.

3rabiii:
Thanks @burma-shave, I will open now a separate pr for every issue

3rabiii:
Closing this, I already divided it for three prs, one for every issue

PR #1000, opened 2026-06-02
Java-parity for arrivals/departures: deviation + distance + tripStatus

### LINKED ISSUE #895: Arrivals and Departures for Stop Endpoint numberOfStopsAway

Ahmedhossamdev:
# Java parity for arrival / departure trip status

Fixes: #895 

## Summary

This PR makes Maglev's arrival/departure responses follow Java OneBusAway's
block-location algorithm for:

- `distanceFromStop`
- `numberOfStopsAway`
- `tripStatus.scheduleDeviation`
- `tripStatus.scheduledDistanceAlongTrip`
- scheduled-only `tripStatus` position/orientation fields

The previous implementation mixed GPS projection, per-trip GTFS-RT delay,
and publisher-specific `current_stop_sequence` semantics. That produced
zeros for scheduled-only rows, large loop-route distance errors, unit
mismatches when `shape_dist_traveled` was not metres, and schedule-deviation
differences on multi-trip blocks.

The new path mirrors Java's model: compute a block-wide schedule deviation,
interpolate the scheduled block at `currentTime - scheduleDeviation`, then
derive stop metrics from block distance and block stop sequence.

When GTFS-RT is unavailable for an arrival, Maglev now still computes the bus's
scheduled position by interpolating the static GTFS block at the request time.
That gives scheduled-only rows meaningful `distanceFromStop`,
`numberOfStopsAway`, and `tripStatus` fields instead of leaving them at zero.

### Scheduled interpolation math

With GTFS-RT, the scheduled snapshot is evaluated at:

```text
effectiveTime = currentTime - scheduleDeviation
```

Without GTFS-RT, `effectiveTime` is just the request time.

The snapshot builder finds the two scheduled block stops around that time:

```text
from = previous scheduled block stop
to   = next scheduled block stop
```

Then it computes how far the effective time is through that scheduled segment:

```text
ratio =
  (effectiveSeconds - from.effectiveStopSeconds)
  / (to.effectiveStopSeconds - from.effectiveStopSeconds)
```

Finally, it linearly interpolates the block distance:

```text
snapshot.distanceAlongBlock =
  from.distanceAlongBlock
  + ratio * (to.distanceAlongBlock - from.distanceAlongBlock)
```

If the request time is halfway between two scheduled stops, the snapshot
position is halfway along the scheduled block distance between those stops.
Before the first stop it clamps to the first stop distance; after the last stop
it clamps to the last stop distance.

The arrival metrics are then:

```text
distanceFromStop =
  targetStop.distanceAlongBlock - snapshot.distanceAlongBlock
```

Positive means the target stop is still ahead; negative means the scheduled
position has already passed it.

```text
numberOfStopsAway =
  targetStop.blockSequence - nextStop.blockSequence
```

Positive means the target stop is that many stops ahead; zero means it is the
next/current stop; negative means it is behind.

### Worked example

Assume the scheduled block has these stops:

| Stop | Time | Distance along block | Block sequence |
| --- | ---: | ---: | ---: |
| A | 12:00:00 | 1,000 m | 10 |
| B | 12:10:00 | 2,000 m | 11 |
| C | 12:20:00 | 3,500 m | 12 |

If the request time is 12:05:00 and there is no GTFS-RT, then:

```text
effectiveTime = 12:05:00
from = A
to   = B

ratio =
  (12:05 - 12:00)
  / (12:10 - 12:00)
  = 5 minutes / 10 minutes
  = 0.5

snapshot.distanceAlongBlock =
  1,000 + 0.5 * (2,000 - 1,000)
  = 1,500 m
```

For target stop B:

```text
distanceFromStop =
  2,000 - 1,500
  = 500 m

numberOfStopsAway =
  11 - 11
  = 0
```

The bus's scheduled position is 500 m before B, and B is the next scheduled
stop.

For target stop C:

```text
distanceFromStop =
  3,500 - 1,500
  = 2,000 m

numberOfStopsAway =
  12 - 11
  = 1
```

C is 2,000 m ahead of the scheduled position and one stop after the next stop.

If GTFS-RT says the bus is 2 minutes late at 12:05:00, then:

```text
effectiveTime = 12:05:00 - 2 minutes = 12:03:00
ratio = 3 minutes / 10 minutes = 0.3

snapshot.distanceAlongBlock =
  1,000 + 0.3 * (2,000 - 1,000)
  = 1,300 m
```

The same formulas still apply; the only difference is that the scheduled
snapshot is evaluated 2 minutes earlier because the bus is late.

## What Changed

- Added `internal/restapi/scheduled_block_helper.go`.
  - Resolves the active block/shift for a trip.
  - Projects stops monotonically through route shapes, including loop routes.
  - Builds a scheduled block snapshot with `DistanceAlongBlock`,
    `NextStopIndex`, active trip, and active-trip scheduled distance.
  - Computes Java-equivalent `distanceFromStop` and `numberOfStopsAway`.

- Reworked `GetScheduleDeviationForBlock`.
  - Iterates all TripUpdates for trips in the block, in block start order.
  - Preserves Java's "trip-level delay wins" behavior.
  - Chooses the closest StopTimeUpdate only when there is no trip-level delay.
  - Applies Java's `blockNotActive` guard by discarding deviations over 1 hour.

- Updated both arrivals handlers.
  - Scheduled-only rows now get a `tripStatus`.
  - Stop distance/count metrics now come from the scheduled block snapshot.
  - `minutesBefore` now caps at 240 minutes, matching `minutesAfter`.
  - Unknown/wrong-case agencies return 404 instead of 500.

- Updated `BuildTripStatus`.
  - Uses block-wide deviation.
  - Fills `scheduledDistanceAlongTrip` from the active block trip.
  - Emits raw GTFS-RT `vehicleId` instead of double-prefixing it.
  - Uses the shared scheduled-position fallback when no RT vehicle is present.

- Removed superseded helpers.
  - `block_distance_helper.go`
  - `block_sequence_helper.go`
  - dead vehicle current-stop/active-trip helpers

- Refreshed `testdata/openapi.yml` from upstream.

## Java References

The implementation was checked against these Java paths:

- `GtfsRealtimeTripLibrary.applyTripUpdatesToRecord`
  - block trip iteration
  - trip-level delay short-circuit
  - per-stop closest-in-time deviation selection
- `GtfsRealtimeSource.blockNotActive`
  - ignores records whose schedule deviation is over 1 hour
- `ArrivalsAndDeparturesBeanServiceImpl.applyBlockLocationToBean`
  - `distanceFromStop = stopTime.distanceAlongBlock - blockLocation.distanceAlongBlock`
  - `numberOfStopsAway = stopTime.blockSequence - nextStop.blockSequence`
- `TripStatusBeanServiceImpl`
  - `scheduledDistanceAlongTrip = scheduledDistanceAlongBlock - activeBlockTrip.distanceAlongBlock`
- `DistanceAlongShapeLibrary.computeBestAssignment`
  - monotonic stop-to-shape assignment for loop routes

## Validation

- Added focused tests for:
  - scheduled block snapshot interpolation
  - stop metrics
  - active-trip scheduled position
  - monotonic loop-route projection
  - shape-distance unit scaling
  - block-wide schedule deviation rules
  - Java's 1-hour stale-deviation guard

- Live comparison against Java showed exact schedule-deviation parity on the
  sampled Unitrans arrivals, with remaining differences limited to stop-boundary
  and independent-feed-snapshot timing effects.

## Follow-Ups

- Cache scheduled block snapshots per request / RT update to reduce DB query
  amplification.
- Decide whether Maglev should keep smooth per-request interpolation or match
  Java's stepped GTFS-RT-poll cadence.
- Make GTFS-RT `agency-ids` filtering case-insensitive so config casing cannot
  silently drop feeds.
- Work in TO-DOS

## Summary by CodeRabbit

* **New Features**
  * Improved trip status, vehicle position, distance, and stops-away calculations across multi-trip blocks.
  * Enhanced arrival and departure predictions using real-time delays and schedule data.
  * Expanded arrival/departure searches to cover a full 24-hour window.
  * Improved nearby-stop results with more precise proximity matching.
  * Vehicle identifiers now include agency context.

* **Bug Fixes**
  * Improved handling of inactive trips, cancellations, excessive delays, schedule boundaries, and missing stops.
  * Corrected block layover timing and route-shape distance calculations.
  * Requests for unavailable stops, trips, vehicles, or agencies now return clearer not-found responses.
  * Improved future-service detection for route schedules.


Ahmedhossamdev:
Hey @aaronbrethorst,
If you get some time today, could you take a look at this PR?
Thanks!

internal/restapi/arrival_and_departure_for_stop_handler.go
```diff
@@ -590,12 +614,15 @@ func (api *RestAPI) getPredictedTimes(
...
 	targetStopSequence int64,
 	scheduledArrivalTime, scheduledDepartureTime time.Time,
+	currentTime time.Time,
 ) (predictedArrivalTime, predictedDepartureTime time.Time, predicted bool) {
 	realTimeTrip, _ := api.GtfsManager.GetTripUpdateByID(tripID)
-	// trip-level delay exists but StopTimeUpdates is empty
 	if realTimeTrip == nil || (len(realTimeTrip.StopTimeUpdates) == 0) && realTimeTrip.Delay == nil {
 		return time.Time{}, time.Time{}, false
 	}
+	if !isTripActive(*realTimeTrip, currentTime) {
```
aaronbrethorst: **`isTripActive` gate silently disables predictions for delay-only feeds**

**Correctness — highest severity finding in this review.**

This new gate rejects any TripUpdate whose StopTimeUpdates lack absolute `time` values, and any TripUpdate carrying a trip-level `Delay` with zero StopTimeUpdates. Delay-only GTFS-RT feeds (STUs with only `delay` set, no `arrival.time`/`departure.time`) are extremely common, and for them `isTripActive` computes `firstPrediction = lastPrediction = -1` and returns false. `getPredictedTimes` then returns `(zero, zero, false)` — so **every arrival comes back `predicted: false` with `predictedArrivalTime: 0`**, reporting schedule-only data even though live predictions were fetched successfully.

**This is a parity break, not parity.** Java uses `isTripActive` *only* to classify matched/unmatched trip IDs for the `MonitoredResult` metrics counter (`GtfsRealtimeTripLibrary.java:570-590`). It never gates prediction generation. Porting a metrics predicate into the prediction path inverts its meaning.

Knock-on effect: the gate makes the trip-level-`Delay` fallback at lines 682-689 (and the `realTimeTrip.Delay != nil` allowance at line 620) **unreachable dead code** — a TripUpdate with `Delay` but no STUs hits `len(tu.StopTimeUpdates) == 0 -> return false` inside `isTripActive` and is dropped before the fallback can run.

Same root cause at `trip_updates_helper.go:216` and `trip_updates_helper.go:422` (`GetStopDelaysFromTripUpdates`).

Suggest dropping the gate from the prediction path entirely, and keeping `isTripActive` for metrics if you want it at all.

internal/restapi/trips_helper.go
```diff
@@ -48,7 +49,10 @@ func (api *RestAPI) BuildTripStatus(
 
 	if vehicle != nil {
 		if vehicle.ID != nil {
-			status.VehicleID = utils.FormCombinedID(agencyID, vehicle.ID.ID)
+			// Emit the GTFS-RT vehicle id verbatim — Java keeps it un-prefixed
+			// here and the top-level ArrivalAndDeparture.VehicleID already does
+			// the same, so wrapping it with the agency was a double-prefix bug.
+			status.VehicleID = vehicle.ID.ID
```
aaronbrethorst: **tripStatus.vehicleId lost its agency prefix — breaks the trip-for-vehicle round-trip**

**Correctness.** `status.VehicleID` is now emitted as the raw GTFS-RT vehicle id with no agency prefix, and the comment justifies it with *"Java keeps it un-prefixed here."*

**Java does the opposite.** `TripStatusBeanServiceImpl.java:364` calls `ApplicationBeanLibrary.getId(vid)`, and `ArrivalsAndDeparturesBeanServiceImpl.applyBlockLocationToBean` calls `AgencyAndIdLibrary.convertToString(blockLocation.getVehicleId())` — both produce `agencyId_vehicleId`. So the "double-prefix bug" this change claims to fix doesn't appear to exist upstream.

Maglev's own API contract requires the prefix *on the way in*: `/api/where/trip-for-vehicle/{id}` calls `extractAndValidateAgencyCodeID` (`trip_for_vehicle_handler.go:16`), and the `vehicleId` query param on arrival-and-departure-for-stop calls `utils.ExtractAgencyIDAndCodeID` (`arrival_and_departure_for_stop_handler.go:317`). A client that reads `tripStatus.vehicleId` = `"4001"` from an arrivals response and feeds it to `/api/where/trip-for-vehicle/4001.json` now gets a 400/404 — there's no `agency_` prefix to split on.

This leaks beyond the endpoints in this PR: it affects **every** endpoint that emits a TripStatus (trip-details, trips-for-location, arrivals-and-departures).

gtfsdb/helpers.go
```diff
@@ -1415,13 +1415,19 @@ func (c *Client) buildBlockLayoverIndex(ctx context.Context, staticData *gtfs.St
 					continue
 				}
 
+				layoverStart := int64(lastStopCurrent.ArrivalTime)
```
aaronbrethorst: **Block layover bounds inverted relative to Java**

**Correctness.** Java's `BlockTripLayoverTimeComparator` defines:
- `layoverStart` = previous trip's **last stop `departureTime`**
- `layoverEnd`   = next trip's **first stop `arrivalTime`**

which is exactly the code this PR deleted. The new code uses `arrivalTime` for the start and `departureTime` for the end — inverted — widening the layover window by the dwell time at *both* ends.

Concrete: trip A's terminal stop arrives 10:00 / departs 10:20 (scheduled recovery dwell); trip B's first stop arrives 10:35 / departs 10:40. Java and the old code record the layover as **[10:20, 10:35]**. The new code records **[10:00, 10:40]**.

`GetActiveLayoverBlockIDsForRoute` (`gtfsdb/query.sql:1036`, used by `trips-for-route`) matches on `layover_start < range_end AND layover_end > range_start`. So at 10:05 — while the bus is still standing at trip A's last scheduled stop, mid-trip — the block is now reported as *in layover* and its trip gets injected into the `trips-for-route` response.

The added `layoverStart > layoverEnd` skip guard also becomes near-dead under the widened bounds: `start <= end` now holds in essentially every case where the original bounds were valid.

internal/restapi/trip_updates_helper.go
```diff
@@ -1,49 +1,428 @@
...
+		if t.tu.Delay != nil {
+			deviation = int(t.tu.Delay.Seconds())
+			found = true
+		}
+	}
+	if !found {
+		return 0, false, false
 	}
+	const javaBlockNotActiveThreshold = 60 * 60
+	if deviation > javaBlockNotActiveThreshold || deviation < -javaBlockNotActiveThreshold {
```
aaronbrethorst: **`blockNotActive` >1h discard applied to only one of two consumers — self-contradictory responses**

**Correctness.** `pickTripLevelDeviation` discards any trip-level delay over 1 hour and returns `(0, false)`, but the *prediction* half of the pipeline still reads the same `tu.Delay` and applies it. The two halves of the response now contradict each other.

A bus running 70 minutes late (`delay=4200` — routine during snow/breakdowns) previously reported `tripStatus.scheduleDeviation = 4200`. Now `GetScheduleDeviationForBlock` returns `(0, false)`, so `BuildTripStatus` leaves `scheduleDeviation: 0, predicted: false` — while `getPredictedTimes` independently emits a `predictedArrivalTime` 70 minutes past scheduled. **The client renders "on time" next to an arrival time 70 minutes late.**

Java's `blockNotActive` (`GtfsRealtimeSource.java:728-732`) discards the entire `VehicleLocationRecord` with `continue`, so the predictions vanish too and Java stays self-consistent. This port applies the guard to only one of the two consumers, which is worse than not applying it at all.

internal/restapi/trip_updates_helper.go
```diff
@@ -1,49 +1,428 @@
...
+		return -v
+	}
+	return v
+}
+
+// pickReverseWalkSTUDelay is the final fallback when schedule lookup failed
+// (e.g. test fixtures without static stop times) but the feed still carries
+// per-stop delays. Returns the latest stop's delay in block-iteration order.
+func pickReverseWalkSTUDelay(tripUpdates []tripUpdateForTrip) (int, bool) {
+	for i := len(tripUpdates) - 1; i >= 0; i-- {
```
aaronbrethorst: **`pickReverseWalkSTUDelay` returns the terminal STU's delay, not the closest-in-time one**

**Correctness.** This returns the delay of the **last** StopTimeUpdate of the **last** block trip (both loops walk backward). The old code returned the *first* STU's delay, and Java's `updateBestScheduleDeviation` picks the STU whose predicted time is **closest to now** — never the terminal one.

This compounds with the `isTripActive` finding: because `pickClosestSTUDeviation` filters through `isTripActive` (which rejects delay-only feeds), the reverse-walk is the **only** path that ever fires for those feeds. So the "Java parity" closest-in-time picker is dead code exactly where it matters most.

Failure: a delay-only feed publishes a bus currently 10 min late at the next stop, whose predicted end-of-line delay is 0 (schedule recovery built into the last leg). `pickTripLevelDeviation` finds nothing; `pickClosestSTUDeviation` is short-circuited by `isTripActive`; `pickReverseWalkSTUDelay` walks back from the last STU and returns **delay = 0**. `tripStatus.scheduleDeviation` publishes 0 for a bus that is 10 minutes late — and `computeScheduledBlockSnapshot` is then evaluated at an unshifted `effectiveTime`, so `distanceFromStop` / `numberOfStopsAway` place the bus ~10 minutes further along the block than it actually is.

internal/restapi/scheduled_block_helper.go
```diff
@@ -0,0 +1,594 @@
...
+// (km/miles/...) come out as metres. Identical to calculateBatchStopDistances.
+func computeShapeDistScale(stopTimes []gtfsdb.StopTime, cumulativeDistances []float64) float64 {
+	totalShapeDist := cumulativeDistances[len(cumulativeDistances)-1]
+	var maxStopDist float64
+	for _, st := range stopTimes {
+		if st.ShapeDistTraveled.Valid && st.ShapeDistTraveled.Float64 > maxStopDist {
+			maxStopDist = st.ShapeDistTraveled.Float64
+		}
+	}
+	if maxStopDist > 0 && totalShapeDist > 0 {
```
aaronbrethorst: **`computeShapeDistScale` corrupts distances when the shape extends past the last stop**

**Correctness.** `computeShapeDistScale` rescales `shape_dist_traveled` by `geometricShapeTotal / maxStopShapeDist`, which assumes the last stop's `shape_dist_traveled` equals the shape's end. GTFS shapes routinely extend past the final stop (deadhead tail into a garage, loop closure), so **the ratio is not 1.0 even when the feed already publishes metres** — and the scale is then applied to every stop.

Concrete: a feed publishes `shape_dist_traveled` in metres; the trip's last stop sits at 10,000 m but the polyline continues 2,000 m into the layover/garage, giving a geometric total of 12,000 m. The scale comes out to 1.2, so **every** stop's `DistanceAlongBlock` / `DistanceAlongTrip` is inflated 20%. `metricsForStop` then reports `distanceFromStop` ~20% too large (1,200 m instead of 1,000 m), and `calculateBatchStopDistances` publishes the same inflated `distanceAlongTrip` on schedule/trip-details responses.

This is a regression: the pre-PR code used `shape_dist_traveled` verbatim. If the goal is unit normalization, it needs actual unit detection, not a ratio against the last stop.

Same block duplicated at `trips_helper.go:836-849`.

internal/restapi/arrivals_and_departures_for_stop_handler.go
```diff
@@ -335,61 +356,66 @@ func (api *RestAPI) arrivalsAndDeparturesForStopHandler(w http.ResponseWriter, r
...
+			// See GetScheduleDeviationForBlock for the deviation rules.
+			effectiveTime := params.Time
+			if vehicle != nil && vehicle.Trip != nil && vehicle.Trip.ID.ID != "" {
+				blockTripIDs := api.blockTripIDsSortedByStartTime(ctx,
+					api.blockTripIDsForServiceDate(ctx, st.TripID, serviceMidnight))
+				if dev, hasRT := api.GetScheduleDeviationForBlock(ctx, blockTripIDs, serviceMidnight, params.Time); hasRT {
+					effectiveTime = params.Time.Add(-time.Duration(dev) * time.Second)
 				}
+			}
+			if snapshot := api.computeScheduledBlockSnapshot(ctx, st.TripID, effectiveTime, serviceMidnight); snapshot != nil {
```
aaronbrethorst: **Block snapshot built twice per arrival row, while the window caps were raised 60/240 → 1440/1440**

**Efficiency — this one is likely to take the endpoint down under load.**

Three changes in this PR compound:

1. **Caps widened**: `maxBefore` 60 min → 24 h, `maxAfter` 240 min → 24 h (lines 32-33).
2. **`tripStatus` is now unconditional** for every row — the old code skipped all of this unless `vehicle != nil && vehicle.Position != nil`.
3. **The snapshot is computed twice per row** — once inside `BuildTripStatus` (line 361), once again here (line 392).

Per row that's roughly `blockTripIDsForServiceDate` (3) + `blockTripIDsSortedByStartTime` (N) + `GetScheduleDeviationForBlock` (up to N) + `computeScheduledBlockSnapshot` (3 + 2N + stop batches) — **doubled**. For a busy downtown stop with ~300 stop_times in a `minutesBefore=1440&minutesAfter=1440` window and 5-trip blocks, that's on the order of **10⁴ SQLite round-trips for a single HTTP request**, where the pre-PR handler returned promptly.

The PR's own `TODO(perf)` at `scheduled_block_helper.go:71-72` acknowledges the amplification — but the caps were widened anyway, which turns a known hot spot into a much bigger one. At minimum: compute the snapshot once and thread it through, and reconsider the 1440/1440 caps.

internal/restapi/scheduled_block_helper.go
```diff
@@ -0,0 +1,594 @@
...
+func (api *RestAPI) loadBlockTripData(ctx context.Context, tripIDs []string) []blockTripData {
+	out := make([]blockTripData, 0, len(tripIDs))
+	for _, id := range tripIDs {
+		stopTimes, err := api.GtfsManager.GtfsDB.Queries.GetStopTimesForTrip(ctx, id)
+		if err != nil || len(stopTimes) == 0 {
+			continue
+		}
+		// TODO(correctness): shape errors leave totalDist=0 — trip is still
+		// appended so block_sequence stays consistent, but its DistanceAlongTrip
+		// values are zero. Stop-only Haversine fallback would fix this.
```
aaronbrethorst: **Shapeless block trips silently get totalDist=0, collapsing block distances (shipped behind a TODO)**

**Correctness / bandaid.** `loadBlockTripData` appends the trip with `cumDistances = nil` and `totalDist = 0` when `GetShapePointsByTripID` errors or returns < 2 points. `emitBlockStops` then adds 0 to `cumulativeBlockDist`, so:

- every stop on the shapeless trip gets the **same** `distanceAlongBlock` → `distanceFromStop = 0` ("the bus is at your stop") for every stop on it, and
- **every subsequent trip in the block** starts at the same `distanceAlongBlock` as the shapeless one, so their distances are short by the missing trip's entire length.

This is left as a `TODO(correctness)` in shipped code. The TODO itself names the fix (a stop-only Haversine fallback) — shipping the known-wrong path instead of the fallback is a bandaid at the wrong altitude. Either implement the fallback or make the failure explicit rather than silently emitting zeros.

internal/restapi/scheduled_block_helper.go
```diff
@@ -0,0 +1,594 @@
...
+		return fallback
+	}
+	return ids
+}
+
+// loadBlockTripData fetches stop_times + shape for each tripID and bundles them.
+// Skips trips with no stop times or unusable shape data.
+//
+// TODO(perf): 2N round-trips; batch via GetStopTimesForTrips / GetShapePointsForTrips.
+func (api *RestAPI) loadBlockTripData(ctx context.Context, tripIDs []string) []blockTripData {
```
aaronbrethorst, line 526: **`loadBlockTripData` is a textbook N+1 — batch queries for exactly this already exist**

**Efficiency.** `loadBlockTripData` issues **2 DB queries per block trip** in a loop: `GetStopTimesForTrip` (line 344) and `GetShapePointsByTripID` (line 351). For a 6-trip block that's 12 round-trips — repeated for every arrival row, and *twice* per row given the duplicate-snapshot finding above.

`gtfsdb/query.sql` already ships the batch versions for precisely this: `GetStopTimesForTripIDs :many` (line 1164) and `GetShapePointsByIDs :many` (line 1158). CLAUDE.md lists both under **"Batch Queries (N+1 prevention)"**. The hand-rolled loop is the exact pattern those queries exist to avoid.

Same shape at `trip_updates_helper.go:371` (`blockTripIDsSortedByStartTime` re-queries stop times per trip just to read `sts[0].ArrivalTime` — and `min_arrival_time` is already a cached column in `gtfsdb/schema.sql:294`) and `trip_updates_helper.go:114`.

internal/restapi/scheduled_block_helper.go
```diff
@@ -0,0 +1,594 @@
...
+// projectStopsInSequence returns each stop's distance-along-trip in metres,
+// projecting monotonically through the shape so loop routes (where the same
+// lat/lon appears at multiple shape segments) get distinct distances per
+// occurrence. Mirrors Java's DistanceAlongShapeLibrary.computeBestAssignment;
+// a naive global-minimum search picks the same segment for both occurrences,
+// producing the catastrophic distanceFromStop outliers we saw on the Q route.
+//
+// Prefers GTFS shape_dist_traveled (scaled to metres) when present, falls
+// back to forward-walking geometric projection.
+func projectStopsInSequence(
```
aaronbrethorst, line 636: **Stop→shape projection now exists in two divergent copies**

**Reuse / drift.** `projectStopsInSequence` / `computeShapeDistScale` / `projectStopGeometric` re-implement `calculateBatchStopDistances` (`trips_helper.go:797`). The new code's own comment at line 451 says *"Identical to calculateBatchStopDistances"* — but **the copies already differ**: `projectStopGeometric` adds the `ratio > 0.95` cursor advance and `advanceCursorThroughCumDist`, which `calculateBatchStopDistances` lacks.

So the same stop on the same trip gets one `distanceAlongTrip` in the trip's schedule `stopTimes` (built via `calculateBatchStopDistances`) and a different distance-along-block basis in `distanceFromStop` (built via `projectStopsInSequence`). A client comparing trip-details `stopTimes` against arrivals `distanceFromStop` sees inconsistent metres for the same stop.

Any future fix to the projection now has to land in two places or the drift widens. Worth collapsing to one implementation before this merges.

aaronbrethorst (changes requested):
Hey Ahmed, I've left ten inline comments with the specifics; this note is about the parts that don't fit on a single line.

## Critical

The parity claims need to be checked against the Java before they go in a comment. Several places in this PR cite Java as the justification for behavior Java doesn't actually have — `isTripActive` gating predictions, the un-prefixed `vehicleId`, the layover bounds. In each case the code comment asserts parity and the Java in `app-modules` says the opposite. That's more concerning than any single bug here, because a wrong comment that claims authority is what the next person will trust instead of re-deriving it.

So: for every "Java does X" in this diff, cite the Java `file:line` you're porting from. Where you're deliberately diverging from Java, say that instead — divergence is fine, but it should be a decision, not an accident. The inline comments cover the individual cases.

## Important

The tests bent around the new behavior rather than checking it. `arrivals_and_departures_for_stop_handler_test.go:462` sets a fixture time with the comment "must be > mock clock 08:02 for isTripActive" — the test was adjusted to satisfy the gate instead of asking whether the gate belonged there. And nothing in the suite builds a TripUpdate carrying only `delay` with no absolute stop times, which is the exact feed shape that the gate silently breaks. The suite passes because the RABA fixtures all publish absolute times. Please add a delay-only case; it would have caught most of what I flagged inline.

Don't ship the `TODO(correctness)` and `TODO(perf)` markers. A TODO on a path that emits wrong distances is a known-wrong path with a note attached, and once it merges nobody is coming back for it. Either implement the fallback the TODO names or make the case fail loudly.

This is 2,400 added lines across fifteen files, and it changes deviation picking, distance projection, tripStatus construction, layover bounds, and shape-distance scaling all at once. Any one of those is a reviewable PR. Bundled together, a regression in one is hard to isolate from the others — which is roughly what happened here. Please look at CONTRIBUTING.md on scope, and consider landing the block snapshot on its own first.

## Fit and Finish

The green perf smoke test on this PR isn't clearance for the widened `minutesBefore`/`minutesAfter` caps. It runs 5 VUs for 30s against default params, so it never exercises the window you opened up. Worth a k6 scenario that actually requests the 1440-minute window against a busy stop.

Thanks — the hard part of this is done and the structure is sound. Once the Java claims are verified and the delay-only case is covered, I think this lands.

internal/restapi/trips_helper.go
Ahmedhossamdev: In consolidated feeds, Java adds the agency prefix to the vehicle ID. In a single feed like Unitrans Java leaves the vehicle ID exactly as it comes from the GTFS-RT feed.

So I was thinking of keeping it without the prefix for now, since Maglev doesn't support consolidated feeds yet.

What do you think? @aaronbrethorst

internal/restapi/arrival_and_departure_for_stop_handler.go
Ahmedhossamdev: Nice catch! I haven't encountered a feed that omits the arrival time yet, but this is a real issue. If the GTFS-RT feed doesn't include an arrival time and only provides a delay, we'll end up ignoring the real-time data completely.

I'm working on a fix for this right now. I think I'll remove that check, since we don't actually need it.

Thanks, @aaronbrethorst

gtfsdb/helpers.go
Ahmedhossamdev: OOPS! 🥸🥸. Good catch, you're right. Java's `BlockTripLayoverTimeComparator` is pretty clear here (lines 44 and 51): `layoverStart` should use the previous trip's last-stop **departure**, and `layoverEnd` should use the next trip's first-stop **arrival**. I had those two swapped, which effectively widened the layover window by including the in-service dwell time on both ends.

One interesting thing I noticed is that a lot of GTFS feeds have `arrival_time == departure_time` for every stop. Unitrans is one of them (`SELECT COUNT(*) FROM stop_times WHERE arrival_time != departure_time` returns `0`). On feeds like that, this bug is basically invisible because both values are identical. It only shows up on feeds that actually record dwell times (major terminals, rail stations, layover/recovery time), which is probably why we didn't notice it in our test data. -_-

I'm pushing a fix that:

* Swaps the two fields in `buildBlockLayoverIndex` to match Java.
* Keeps the `layoverStart > layoverEnd` guard. With the corrected bounds, it now catches overlapping-trip anomalies (where trip B arrives before trip A departs at the shared stop), which would otherwise violate the `CHECK (layover_start <= layover_end)` constraint and fail the import.
* Adds `TestBuildBlockLayoverIndex_BoundsMatchJava`, which reconstructs the expected bounds from `stop_times` and verifies that each generated layover row matches.

Thanks for catching this!

burma-shave:
Here's the result of running an LLM analysis across our specs and clients:

## OBA API Review: PR #1000 — `arrival-and-departure-for-stop` / `arrivals-and-departures-for-stop`

**Input**: [PR #1000](https://github.com/OneBusAway/maglev/pull/1000), "Java-parity for arrivals/departures: deviation + distance + tripStatus"
**Stated goal**: Fixes #895 — `numberOfStopsAway` always returns `-1` on Hamilton's (HSR) feed
**Changes**: Both production and test (9 production files, 6 test files touched; 2 helper files deleted)

### Overview

**What this change does**

Issue #895 reports that on Hamilton's (HSR) feed, `arrivals-and-departures-for-stop` always returns `numberOfStopsAway: -1`. The old implementation computed this by reading `current_stop_sequence` off the GTFS-RT `VehiclePosition` message, finding the vehicle's "global" block sequence number, and subtracting it from the target stop's. If a publisher's feed doesn't populate `current_stop_sequence` (as Hamilton's apparently doesn't), the lookup returns `nil` and the handler hard-codes `-1`. The same GPS/vehicle-position dependency also caused large distance errors on loop routes and unit mismatches when `shape_dist_traveled` wasn't already in metres.

The fix (mirroring Java OneBusAway) stops depending on GTFS-RT position/sequence fields for this calculation entirely:
1. Computes a single **block-wide schedule deviation** from whatever trip updates exist for any trip in the vehicle's block, using Java's exact precedence rules (latest trip-level delay wins; otherwise closest-in-time per-stop deviation; deviations over 1 hour are discarded as "block not active").
2. Builds a **scheduled block snapshot**: walks every trip in the block in order, projects stops onto their shape (preferring `shape_dist_traveled`, falling back to a monotonic-forward geometric projection that fixes the loop-route bug), and assigns cumulative distance/sequence numbers.
3. Evaluates that snapshot at `currentTime − scheduleDeviation` to get the bus's interpolated position — purely from the schedule, no GPS.
4. Derives `distanceFromStop`/`numberOfStopsAway` from that snapshot, mirroring Java's `applyBlockLocationToBean`.

Because it no longer needs `current_stop_sequence` or a live position at all, this fixes the bug directly, and as a side effect gives scheduled-only rows (no GTFS-RT at all) a real `tripStatus` instead of zeros. Concretely: a rider at a stop, bus 2 minutes late, feed without `current_stop_sequence` — before, always `-1`; after, the schedule-implied position adjusted for the 2-minute delay. The PR also bundles: `minutesBefore`/`minutesAfter` caps widened to 1440/1440 (both endpoints), unknown/wrong-case agency (and singular's stop/trip) now 404 instead of 500, and a delay-propagation fix in `getPredictedTimes`.

**Domain background**: A GTFS **block** groups trips served back-to-back by one physical vehicle; Java reasons about position at the block level because a trip's delay also implies position relative to the *next* trip's stops. **`shape_dist_traveled`** is the authoritative (if often absent or wrong-unit) way to place a stop along a shape; without it, lat/lon projection is ambiguous on loops. **`current_stop_sequence`** is an optional GTFS-RT `VehiclePosition` field many publishers simply omit — which is the entire root cause here.

### Goal check

Stated goal (#895): `numberOfStopsAway` always `-1` on Hamilton's feed.

- ✓ **Root cause structurally removed.** `getNumberOfStopsAway`/`getCurrentVehicleStopSequence` and their `-1` fallback are deleted from both handlers. `current_stop_sequence` is no longer read anywhere in the new computation — a feed omitting it is no longer a special case.
- ✓ **Replacement matches spec**: schedule-derived fields are now populated even when no RT vehicle exists, correcting a pre-existing spec violation as a side effect.
- ⚠ **Test coverage: solid at unit level, gap at the end-to-end level.** `scheduled_block_helper_test.go` (753 new lines) thoroughly covers the new math. But the old test that literally encoded the bug (`TestGetNumberOfStopsAway_NilCurrentSequence`) is simply deleted, and nothing replaces it with a handler-level test asserting `numberOfStopsAway` is a real, non‑`-1` value for a vehicle with no position/sequence data — the closest test (`TestPluralArrivals_BlockNotActiveDoesNotShiftEffectiveTime`) only checks subject == baseline, which would still pass if both were silently wrong.

**Overall: fully closed.**

### Spec check

- **Consistent** — schedule-derived `distanceFromStop`/`numberOfStopsAway`/`tripStatus` now match both wiki specs' "present whenever a block location exists, including schedule-derived" language (fixes a pre-existing violation).
- **Consistent** — 404-not-500 changes extend the singular endpoint's already-recorded Implementation Decision; the plural endpoint's change matches its existing extension-3a text (though the plural spec has no Implementation Decisions section at all yet).
- **Spec gap (pre-existing)** — the singular endpoint's `minutesBefore`/`minutesAfter` aren't documented in its wiki spec at all, and are parsed/validated/capped but **never actually used anywhere else in the handler** — dead code the PR's cap change touches without fixing.
- **Conflict found** — both wiki specs document `tripStatus.vehicleId` as "Combined `{agencyId}_{vehicleId}}`," but the PR changes it to emit the raw, un-prefixed GTFS-RT ID (matching the top-level `entry.vehicleId`, which was *already* raw pre-PR despite the same spec wording). Either the spec is wrong for both fields, or this is a longstanding bug the PR just extended.

**Overall: spec-consistent, with one conflict surfaced** (the `vehicleId` format vs. documented "combined ID" contract).

### Client impact

All four client repos resolved cleanly, no staleness.

| Behaviour                                       | Wayfinder/SDK                                        | iOS                  | Android                   |
| ----------------------------------------------- | ---------------------------------------------------- | -------------------- | ------------------------- |
| Schedule-derived distance/stops-away/tripStatus | none                                                 | none                 | none                      |
| `tripStatus.vehicleId` raw vs. combined         | none                                                 | none (opaque string) | **impact — user-visible** |
| `minutesBefore`/`minutesAfter` cap widening     | none                                                 | none                 | none                      |
| 500→404 on unknown agency/stop/trip             | none (Wayfinder collapses all non-200 to 500 anyway) | none                 | none                      |

- **Distance/stops-away fix**: no client breaks. Notably, **Android doesn't even model `distanceFromStop`/`numberOfStopsAway`** in `ObaApiModels.kt` — the fix has zero observable effect there today.
- **`vehicleId` format change — real finding**: `BuildTripStatus` is a **shared helper used by five other endpoint handlers** beyond the two named in this PR (`trip-details`, `trip-for-vehicle`, `trips-for-route`, `trips-for-location`), so this format change lands everywhere `tripStatus.vehicleId` appears, not just the arrivals endpoints. Android's `TripDetailsRepository.kt` reads it straight into `TripHeader.vehicleId`, which `TripDetailsScreen.kt` **renders on screen** — riders will see a differently-formatted vehicle ID in the trip-details header.
- Cap widening and 500→404 changes: no client depends on either boundary.

### Summary

The core fix is solid and structural: the exact mechanism behind issue #895 (`current_stop_sequence` dependency) is deleted outright, not just patched, and the replacement is well-covered by unit tests and spec-consistent — including fixing a pre-existing spec violation (schedule-only rows previously got zeroed-out fields). The main things worth knowing before merge:

1. **No end-to-end regression test** reproduces issue #895's literal scenario (vehicle present, no position/sequence data, assert `numberOfStopsAway != -1`) — the fix's correctness rests on the structural code removal and unit tests, not an explicit issue-shaped test.
2. **`tripStatus.vehicleId` now emits a raw GTFS-RT ID instead of a combined `{agencyId}_{vehicleId}`**, contradicting both wiki specs' documented format for that field (and for `entry.vehicleId`, which was already this way pre-PR). This change's blast radius is wider than the PR's title suggests — it touches `trip-details`, `trip-for-vehicle`, `trips-for-route`, and `trips-for-location` too, and is confirmed to change a **user-visible string in Android's trip-details screen**.
3. The PR's `minutesBefore`/`minutesAfter` cap change on the **singular** endpoint touches parameters that are parsed and validated but never actually consumed anywhere in that handler, and aren't part of the documented spec or the JS SDK's generated types for that endpoint — pre-existing dead code, not a regression, but worth a look while the PR is already in this area.
4. The PR bundles a large amount beyond the stated `numberOfStopsAway` fix (loop-route projection, schedule-deviation precedence, cap widening, error-code changes, delay-propagation logic) across ~19 commits — all reviewed above, none found to conflict with spec, but worth knowing the diff is much broader than issue #895 alone.

Ahmedhossamdev:
Thanks, Eric!  @burma-shave

internal/restapi/trip_updates_helper.go
Ahmedhossamdev: Nice catch. It turned out to be a bit bigger than that though. The `blockNotActive` check was only applied to the trip-level `Delay` path, while the two STU-based deviation paths and `getPredictedTimes` weren't checking it at all. That meant we could still return predictions that Java would discard.

Pushing a fix now:

* Move the threshold into a shared helper.
* Apply it to all three deviation sources.
* Apply the same check in `getPredictedTimes` after the final offsets are calculated, so every prediction path behaves consistently.
* Added regression tests for trip-level delay, STU delay, and the 60-minute boundary to match Java's behavior.

internal/restapi/trip_updates_helper.go
Ahmedhossamdev: Verified, you're right (: The walk direction got flipped during a refactor, so we ended up returning the last STU instead of the first one. Java picks the closest scheduled update, but when the static schedule lookup fails we can't do that, so returning the first available STU is a much better fallback than the terminal one.

Pushed a fix:

* Renamed `pickReverseWalkSTUDelay` to `pickFirstAvailableSTUDelay` and switched both loops to walk forward.
* Added regression tests for the `[600, 300, 0]` case and the forward walk across block trips.

Thanks for catching this one.

internal/restapi/scheduled_block_helper.go
Ahmedhossamdev: Ended up going quite a bit deeper on this one.

**Kept the intent, fixed the mechanism.** `computeShapeDistScale` really was broken for shape overhang — you were right. But a straight revert to publisher-verbatim regressed non-metre feeds by 25× on Unitrans. Java doesn't trust the publisher's `shape_dist_traveled` at all — `StopTimeEntriesFactory.ensureStopTimesHaveShapeDistanceTraveledSet` overwrites every value at load time with a projection-derived metre value. Ported that: dropped the scale, dropped the `ShapeDistTraveled` fast-paths, always project via `projectStopGeometric`. Distances are now real metres from our own geometry, regardless of publisher unit.

**Bonus catches while investigating:**
- `getVehicleDistanceAlongShapeContextual` was using a naive global-minimum for prev/next stop bounds, so on Q-route loops the vehicle could land 20 km off (wrong loop pass). Fixed to use sequence-aware `projectStopsInSequence` for bounds.
- `scheduledDistanceAlongTrip` was spiking by a full trip length at trip-start boundaries — has-vehicle branch subtracted the wrong active-trip offset. Made it match Java's `TripStatusBeanServiceImpl:283-292`: `activeTripId`, `scheduledDistanceAlongTrip`, and `totalDistanceAlongTrip` all come from the snapshot's active trip together.
- Scheduled-only arrivals (no vehicle assigned) were clamping to trip length when currentTime fell outside the block's shift range. Ported Java's null-BlockLocation semantics from `ScheduledBlockLocationServiceImpl:241-244` — added an `InRange` flag on the snapshot, callers skip setting position fields when out of range instead of emitting misleading trip-length distances.

**Live A/B against Java on ~460 Unitrans arrival rows (fresh capture, tight timing):**

| metric | mean | median | p95 | max |
|---|---|---|---|---|
| distanceFromStop | 156 m | 136 m | 428 m | 484 m |
| scheduledDistanceAlongTrip | 132 m | 98 m | 407 m | 447 m |
| totalDistanceAlongTrip | 0 m | 0 m | 0 m | 0 m |

Per-route `scheduledDistanceAlongTrip` mean error goes from up to **22 km before** these fixes down to **26–408 m** on every route. Regression test `TestComputeScheduledBlockSnapshot_TripStartBoundary` pins the boundary invariant. Design writeup at `docs/scheduled_distance_boundary_bug.md`.

Full test suite green.

internal/restapi/arrivals_and_departures_for_stop_handler.go
Ahmedhossamdev: I changed BuildTripStatus to return the scheduled block snapshot it builds, so the handler can reuse it for metricsForStop instead of recomputing it.

This cut the median response time on stop=22001 unitrans feed (±1440 min) from 3.43s → 1.61s (~2.1× faster), with no Java parity regressions on the 48-stop probe.

I kept the 24h caps for now since the real issue is fixed, but I'm happy to lower them if you think that's better.

internal/restapi/arrivals_and_departures_for_stop_handler.go
```diff
@@ -335,61 +355,61 @@ func (api *RestAPI) arrivalsAndDeparturesForStopHandler(w http.ResponseWriter, r
...
-			status, statusErr := api.BuildTripStatus(ctx, route.AgencyID, st.TripID, nil, serviceMidnight, params.Time)
-			if statusErr != nil {
-				api.Logger.Warn("BuildTripStatus failed for arrival",
-					"tripID", st.TripID, "error", statusErr)
-			}
-			if status != nil {
-				tripStatus = status
+		// Always built — Java attaches a BlockLocation (real-time or scheduled) to
+		// every arrival, so tripStatus is always non-null.
+		status, snapshot, statusErr := api.BuildTripStatus(ctx, route.AgencyID, st.TripID, nil, serviceMidnight, params.Time)
```
aaronbrethorst, line 369: **`BuildTripStatus` now runs for every arrival row, and the block snapshot inside it is an N+1 — measured 22–34× latency regression**

**Efficiency — this is the one likely to take the endpoint down under load.** I raised a version of this earlier and you fixed the *double* build by returning the snapshot from `BuildTripStatus`. Re-measured at `9b3bd71`, the regression is still there, because the underlying shape of the work didn't change:

1. The `vehicle != nil` gate is gone — this call is now unconditional, so every stop time in the window pays for a full `BuildTripStatus` (~12 queries) *plus* a `computeScheduledBlockSnapshot`.
2. `computeScheduledBlockSnapshot` is not the "~3 DB round-trips per call" its doc comment claims: `blockTripIDsForServiceDate` is 3, `loadBlockTripData` is 2, and `emitBlockStops` issues **one `GetStopsByIDs` per block trip** (`scheduled_block_helper.go:197`).
3. `calculateBlockTripSequence` is then recomputed at line 427 even though `BuildTripStatus` already computed it at `trips_helper.go:297`.
4. Meanwhile the caps went from 60/240 to 1440/1440.

**Failure scenario:** `GET /api/where/arrivals-and-departures-for-stop/{id}.json?minutesBefore=1440&minutesAfter=1440` on a busy downtown stop with ~400 stop times in the window. That's many thousands of SQLite round-trips in a single request — well past client timeouts, and one such request starves the shared connection for every other caller.

Two things need to happen: memoize the snapshot per block+serviceDate for the lifetime of the request (all rows on the same block share one), and batch the `GetStopsByIDs` calls in `emitBlockStops` into a single query across the block. Related open thread: `scheduled_block_helper.go:397`.

internal/restapi/trips_helper.go
```diff
@@ -194,12 +212,84 @@ func (api *RestAPI) BuildTripStatus(
...
+			// setScheduledDistanceAlongTrip(blockScheduledDist − activeBlockTrip.distanceAlongBlock).
+			//
+			// Snapshot.InRange guard: matches Java's null-BlockLocation semantics
+			// (ScheduledBlockLocationServiceImpl.java:241-244). When currentTime
+			// falls outside the shift's [firstStop, lastStop] range, Java returns
+			// null and the arrivals bean leaves tripStatus position fields at
+			// their defaults. Without this guard, our interpolation clamps to a
+			// block boundary and emits misleading trip-length distances for
+			// scheduled-only arrivals of past or future trips.
+			snap = api.computeScheduledBlockSnapshot(ctx, dbTripID, currentTime, serviceDate)
```
aaronbrethorst: **No-vehicle branch builds the snapshot at raw wall time, contradicting the `scheduleDeviation` in the same response**

**Correctness.** The live-vehicle branch correctly shifts time by the deviation (line 225: `effectiveTime := currentTime.Add(-time.Duration(scheduleDeviation) * time.Second)`), but this branch passes `currentTime` straight through.

**Failure scenario:** an agency publishes trip updates but no vehicle positions — very common. The block has `scheduleDeviation = +900`, so `status.ScheduleDeviation` is set to 900 at line 118 and `Predicted` is true. `vehicle == nil`, so we land here and build the snapshot as if the bus were on time. The returned `position` / `scheduledDistanceAlongTrip` place the bus 15 minutes further along the route than it is, and because this same snapshot feeds `metricsForStop` in both arrivals handlers, `distanceFromStop` and `numberOfStopsAway` are on-time values too. The client shows "2 stops away" for a bus that is really 6 stops back — in a response that simultaneously reports `scheduleDeviation: 900`.

Use the same `effectiveTime` here that the live-vehicle branch uses.

internal/restapi/trips_helper.go
```diff
@@ -166,17 +185,16 @@ func (api *RestAPI) BuildTripStatus(
 			slog.String("trip_id", dbTripID),
 			slog.String("error", shapeErr.Error()))
 	}
+	// snap survives to the return so callers (notably the plural arrivals
+	// handler) can reuse it for per-stop metrics without paying for
+	// computeScheduledBlockSnapshot a second time.
+	var snap *scheduledBlockSnapshot
 	if shapeErr == nil && len(shapeRows) > 1 {
```
aaronbrethorst, line 306: **Gating the whole snapshot on the target trip's shape means shapeless feeds lose `numberOfStopsAway` entirely**

**Correctness — a regression against the deleted code.** `snap` is only ever assigned inside this `if shapeErr == nil && len(shapeRows) > 1` block (both `computeScheduledBlockSnapshot` calls, lines 226 and 268), so a trip with a missing or unresolvable `shape_id` returns `snapshot = nil`. Both arrivals handlers then skip `metricsForStop`, leaving the literal initializers at `arrivals_and_departures_for_stop_handler.go:325-326` — `distanceFromStop = 0.0`, `numberOfStopsAway = 0`.

The `getNumberOfStopsAway` / `getBlockSequenceForStopSequence` path this PR deletes needed no shape at all — it counted block stop sequences — so a shapeless feed used to return a correct "7 stops away". Now every arrival on such a feed reports `numberOfStopsAway: 0`, which OBA clients render as "arriving now / at the stop" for a bus seven stops out.

Note the irony: this PR *added* `haversineStopDistances` specifically to handle shapeless block trips inside `emitBlockStops` — but that fallback is unreachable when the **target** trip is the shapeless one, because the caller gates on the target's shape. Move the gate inward so the snapshot is still built (with the haversine fallback) when the target trip has no shape.

internal/restapi/arrivals_and_departures_for_stop_handler.go
```diff
@@ -335,61 +355,61 @@ func (api *RestAPI) arrivalsAndDeparturesForStopHandler(w http.ResponseWriter, r
...
-					} else {
-						numberOfStopsAway = -1
-					}
+			// Reuse the snapshot BuildTripStatus already computed for this trip.
+			// BuildTripStatus applies the same schedule-deviation shift internally,
+			// so recomputing here just to run metricsForStop was doubling every
+			// per-arrival snapshot cost — a real problem on the plural handler
+			// where minutesBefore/minutesAfter can be 24h in each direction.
+			if snapshot != nil {
+				if d, n, ok := snapshot.metricsForStop(st.TripID, int(st.StopSequence)); ok {
```
aaronbrethorst, line 396: **`metricsForStop` is called without the `InRange` gate that `BuildTripStatus` itself applies**

**Correctness.** `BuildTripStatus` deliberately leaves the tripStatus distance fields at their defaults when `snap.InRange` is false. This call site only checks `snapshot != nil`, so it consumes exactly the out-of-range snapshots `BuildTripStatus` refused to trust.

**Failure scenario:** with `minutesAfter` now allowed up to 1440, a request at 08:00 returns a stop time for a trip departing 23:00 the same service day. `currentSeconds < firstStopSec`, so `InRange` is false — but `DistanceAlongBlock` is clamped to `stops[0]` (0) and `NextStopIndex` is 0, so `metricsForStop` still returns `ok = true`, with `distanceFromStop` = the stop's full distance along the block (say 6000 m) and `numberOfStopsAway` = its block sequence (say 14). The API claims a bus that won't leave the depot for another 15 hours is "14 stops / 6 km away", which clients render as an imminent arrival badge.

Same unguarded call at `arrival_and_departure_for_stop_handler.go:376`. Gate both on `snapshot.InRange`.

internal/restapi/arrival_and_departure_for_stop_handler.go
```diff
@@ -662,13 +678,34 @@ func (api *RestAPI) getPredictedTimes(
...
+	} else if departureOffset == nil {
 		departureOffset = &propagatedDelay
 	}
 
+	// Java's blockNotActive filter applies here too: if the resolved offset
+	// exceeds ±1h, the entire RT record would be discarded upstream in Java.
+	// Without this guard, BuildTripStatus (via GetScheduleDeviationForBlock)
+	// reports "no RT" while we would emit a predictedArrivalTime derived from
+	// the same delay, producing self-contradictory responses.
+	if exceedsBlockNotActiveThreshold(int(arrivalOffset.Seconds())) ||
```
aaronbrethorst: **The ±1h guard, now applied here too, silently drops predictions for genuinely very late buses**

**Correctness.** You extended `blockNotActive` to all the deviation paths after my earlier note — but applying it to `getPredictedTimes` goes the wrong direction. This is not a "block isn't active" signal; it's the actual prediction for the stop being asked about.

**Failure scenario:** a feed reports a bus 70 minutes late (arrival delay 4200s) for stop S on trip T. `getPredictedTimes` resolves `arrivalOffset = 4200s`, `exceedsBlockNotActiveThreshold(4200)` is true, and it returns `(time.Time{}, time.Time{}, false)`. Both handlers then set `predicted = false` and blank `predictedArrivalTime` / `predictedDepartureTime`. The rider sees a severely delayed bus rendered as an ordinary on-time scheduled trip — the delay vanishes from the response at exactly the moment it matters most.

This also isn't the parity the PR is chasing: Java's 1h filter lives in `GtfsRealtimeSource` and applies to `VehicleLocationRecord`s, not to STU-derived arrival predictions in `ArrivalAndDepartureServiceImpl`. The guard belongs on the block-activity determination, not on the prediction itself.

internal/restapi/trips_helper.go
```diff
@@ -93,7 +106,13 @@ func (api *RestAPI) BuildTripStatus(
 		}
 	}
 
-	scheduleDeviation, hasRealtimeTripUpdate := api.GetScheduleDeviation(dbTripID)
+	// Mirror Java's applyTripUpdatesToRecord: iterate all block trips in
+	// start-time order; last trip-level delay wins (Java overwrites
+	// best.scheduleDeviation unconditionally each time a trip-level delay is
+	// seen). This is what makes the published `scheduleDeviation` match Java's.
+	blockTripIDs := api.blockTripIDsSortedByStartTime(ctx,
```
aaronbrethorst: **Deviation lookup gets the unfiltered block trip list — skips the shift split the snapshot path applies**

**Correctness.** `blockTripIDsForServiceDate` returns every trip sharing the `block_id` on that service date. `computeScheduledBlockSnapshot` then narrows that with `keepShiftContainingTrip` (`scheduled_block_helper.go:126`) precisely because some feeds reuse one `block_id` for an entire day across multiple physical vehicles. This path skips that step and hands the raw list to `GetScheduleDeviationForBlock`.

**Failure scenario:** a feed reuses `block_id "B1"` for every bus on a route all day. The unfiltered list is all 60 trips; `pickTripLevelDeviation` takes the **last** trip-level delay in start-time order — the delay published for the final evening trip, operated by a completely different bus. That value lands in `status.ScheduleDeviation`, flows into `calculateOffsetForStop` / `findStopsByScheduleDeviation`, and into `effectiveTime` at line 225. A morning arrival reports the evening bus's lateness, and its nextStop/closestStop offsets are shifted minutes in the wrong direction.

Run the same `keepShiftContainingTrip` filter here (or factor the shift resolution out so both callers share one result — which would also help the N+1 problem).

internal/restapi/trips_helper.go
```diff
@@ -194,12 +212,84 @@ func (api *RestAPI) BuildTripStatus(
...
+				// different BlockInstance and doesn't count.
+				//
+				// Also require the arrival's trip to be in the shift: snap is
+				// built from tripID's shift, so this is normally true, but
+				// guards against edge cases where the mapping went sideways.
+				if vehicle.Trip != nil && snap.ShiftTripIDs != nil {
+					_, sameShift := snap.ShiftTripIDs[vehicle.Trip.ID.ID]
+					_, arrivalInShift := snap.ShiftTripIDs[tripID]
+					if sameShift && arrivalInShift {
+						status.DistanceAlongTrip = snap.ActiveTripScheduledDistance
```
aaronbrethorst: **`distanceAlongTrip` is now the *scheduled* distance, and is 0 for exactly the vehicles riders care about**

**Correctness.** The PR deletes:

```go
actualDistance := api.getVehicleDistanceAlongShapeContextual(ctx, activeTripRawID, vehicle)
status.DistanceAlongTrip = actualDistance
```

which ran for every live vehicle with a position. The replacement assigns `snap.ActiveTripScheduledDistance` — a schedule-interpolated value, not the GPS-derived one — and only when `snap != nil && snap.ActiveTripID != "" && snap.InRange` **and** `vehicle.Trip != nil` **and** both the vehicle's declared trip and the arrival's trip are in `snap.ShiftTripIDs`.

**Failure scenario:** a bus running 25 minutes late, still transmitting GPS past its block's last scheduled stop time. `currentSeconds > lastStopSec`, so `InRange` is false, and `/api/where/trip-details/{id}?includeStatus=true` returns `tripStatus.distanceAlongTrip: 0` while `position` and `lastKnownLocation` show the bus mid-route. Clients that place the vehicle marker from `distanceAlongTrip`, or compute `progress = distanceAlongTrip / totalDistanceAlongTrip`, draw the bus at the start of the route or show 0% progress. Same nulling whenever the RT feed omits the trip descriptor (`vehicle.Trip == nil`) — e.g. the DUPLICATED-trip path in `trips_for_route_handler.go` that passes a vehicle in explicitly.

If the intent is to expose the scheduled distance, that's a different field. `distanceAlongTrip` should keep meaning "where the vehicle actually is", with the GPS projection preserved whenever a position is available.

internal/restapi/trips_helper.go
```diff
@@ -811,115 +854,73 @@ func (api *RestAPI) findNextStopAfter(
...
+	// missing / too short, projectStopsInSequence returns zeros — the response
+	// still emits every stop with distanceAlongTrip=0 (same as before).
+	var cumulativeDistances []float64
+	var distances []float64
+	if len(shapePoints) >= 2 {
+		cumulativeDistances = preCalculateCumulativeDistances(shapePoints)
+	}
+	if len(cumulativeDistances) == len(shapePoints) {
+		stopByID := stopByIDFromCoords(timeStops, stopCoords)
+		distances = projectStopsInSequence(timeStops, stopByID, shapePoints, cumulativeDistances)
```
aaronbrethorst, line 938: **The feed's authoritative `shape_dist_traveled` is now discarded in favor of geometric projection**

**Correctness / regression.** The deleted branch was:

```go
// Prefer shape_dist_traveled from the GTFS feed — it is authoritative and
// handles stops that are off the main shape geometry (e.g. loop ends, branching
// routes) correctly. Only fall back to geometric projection when missing.
if stopTime.ShapeDistTraveled.Valid {
    distanceAlongTrip = stopTime.ShapeDistTraveled.Float64
}
```

`projectStopsInSequence` never reads `ShapeDistTraveled` — it always runs `projectStopGeometric`.

**Failure scenario:** a route with a signed deviation (a school-tripper leg, a hospital loop) where a stop sits several hundred metres off the published polyline. Nearest-segment projection snaps it back onto the main line, so `/api/where/trip-details/{id}` returns `stopTimes[].distanceAlongTrip` values that are out of order or identical for consecutive stops — where the feed had correct, monotonically increasing `shape_dist_traveled`. This silently changes the field for *every* feed that publishes the column, and OpenAPI-driven clients treat it as metres along the trip.

The doc-comment justification — "Java overwrites the publisher's `shape_dist_traveled` at load time" — is asserted but nothing in this repo's import path does that overwrite, so the feed-supplied value is simply thrown away. If the claim is load-time normalization, that belongs in the importer, not in a read-path projection that ignores the column.

internal/restapi/arrivals_and_departures_for_stop_handler.go
Ahmedhossamdev: It was part of my follow-up plan, but I'll include the refactor in this PR instead.

Thanks for the review! ❤️

internal/restapi/trips_helper.go
Ahmedhossamdev: Hey @aaronbrethorst  . I tried the obvious fix (project GPS onto the target trip's shape) and it made things much worse same shared shape across sibling trips in a block, so a vehicle on trip A shows up at position X for arrivals on trips A/B/C alike. Measured 41 big-drift rows vs Java for the 14 unitrans stops I test with; revert brings that to 4.

The correct fix needs a block-cumulative shape (concat every block-trip's polyline), project onto that, then subtract the active-trip block offset. Java's blockLocation.distanceAlongBlock - activeBlockTrip.distanceAlongBlock. Bigger change than I want in this PR. Filed a followup doc with the plan.

Shipping the shift-based snap value for now, which still zeros out the two cases you called out. Not ideal but at least not worse than the schedule-only path Java also falls through to when there's no blockLocation.

internal/restapi/trips_helper.go
burma-shave: I can't seem to find a path in the legacy impl where a single rt feed is handled differently. Also, I thought that a consolidated feed is something different to supporting multiple rt feeds. In any case, it looks like maglev always expects an agency prefixed vehicle id, so we should probably be returning that way. That would also keep things in line with the spec.

aaronbrethorst:
### Code review

Found 3 issues:

1. `tripStatus.vehicleId` is emitted without the agency prefix, breaking the round-trip into `/api/where/trip-for-vehicle/{id}` and `/api/where/trip-details/{id}?vehicleId=`. `main` had `status.VehicleID = utils.FormCombinedID(agencyID, vehicle.ID.ID)`; this changes it to the raw GTFS-RT id. The wiki spec is explicit that this is the combined form — `arrivals-and-departures-for-stop.md` ("**`tripStatus.vehicleId`** — Combined `{agencyId}_{vehicleId}`"), `arrival-and-departure-for-stop.md` ("**`data.entry.tripStatus.vehicleId`** — Combined `{agencyId}_{vehicleId}` of the vehicle"), and `trip-details.md` ("**`data.entry.status.vehicleId`** — Combined `{agencyId}_{entityId}` identifier"). Both consumers require the prefix on the way in: `trip_details_handler.go:202` calls `utils.ExtractAgencyIDAndCodeID(params.VehicleID)` and 404s when it fails, and `trip_for_vehicle_handler.go` does the same. This leaks to every endpoint that emits a TripStatus, not just the two changed here. The justifying comment ("Java keeps it un-prefixed here") is the one Java claim in this diff still carrying no `file:line` citation, and `TestBuildTripStatus_VehicleIDFormat` was flipped from asserting `FormCombinedID(agencyID, vehicleID)` to asserting the bare id rather than the claim being verified.

https://github.com/OneBusAway/maglev/blob/713ecf15e227f4b58c0f5e184635a5ba903ca8d8/internal/restapi/trips_helper.go#L67-L73

2. `tripStatus.distanceAlongTrip` is now assigned the same value as `scheduledDistanceAlongTrip`, so the two fields are always identical and the GPS-derived value is gone. Line 230 sets `status.ScheduledDistanceAlongTrip = snap.ActiveTripScheduledDistance` and line 248 sets `status.DistanceAlongTrip = snap.ActiveTripScheduledDistance` — the same schedule-interpolated number, in the same `snap.InRange` block, with nothing downstream overwriting it. `testdata/openapi.yml` defines them as distinct: `distanceAlongTrip` is "Distance, in meters, the transit vehicle **has progressed** along the active trip" vs `scheduledDistanceAlongTrip` "…is **scheduled to have progressed**"; `trip-details.md` says `distanceAlongTrip` is "how far the vehicle has actually travelled along the active trip". `main` computed this from the vehicle position (`getVehicleDistanceAlongShapeContextual`), which this PR deletes. It also drops to `0` for a live vehicle whenever `snap.InRange` is false or the RT feed omits the trip descriptor, so clients computing `distanceAlongTrip / totalDistanceAlongTrip` show 0% progress for a bus visibly mid-route.

https://github.com/OneBusAway/maglev/blob/713ecf15e227f4b58c0f5e184635a5ba903ca8d8/internal/restapi/trips_helper.go#L228-L250

3. The two arrivals handlers now disagree on the unresolved value of `numberOfStopsAway`. The singular handler declares `numberOfStopsAway = -1` (line 311) while the plural handler declares `numberOfStopsAway = 0` (`arrivals_and_departures_for_stop_handler.go:331`). Both are only overwritten when `snapshot.metricsForStop(...)` returns `ok`, which is false for the same conditions in both — `snapshot == nil` (CANCELED trip, no stop times) and `!snap.InRange` / `NextStopIndex < 0`. With `minutesAfter` now allowed up to 1440, `!InRange` is a routine outcome, so the same stop/trip pair reports `-1` from `/arrival-and-departure-for-stop` and `0` from `/arrivals-and-departures-for-stop`. The spec picks zero: `arrivals-and-departures-for-stop.md` §11a says that with no block location available "`distanceFromStop`, `numberOfStopsAway`, `vehicleId`, `predicted`, `lastUpdateTime`, and `tripStatus` are absent or zero." Before this PR both handlers defaulted to `0`.

https://github.com/OneBusAway/maglev/blob/713ecf15e227f4b58c0f5e184635a5ba903ca8d8/internal/restapi/arrival_and_departure_for_stop_handler.go#L305-L312

aaronbrethorst (dismissed):
Thanks for continuing to push on this — it's a big piece of work and the schedule-deviation and block-distance machinery is genuinely hard to get right. I re-reviewed the current head rather than re-reading the old conversation, and I want to be specific about what's still blocking rather than leaving you to guess.

The good news first: I went looking for the classic failure modes in this area and didn't find them. The nanosecond/second/millisecond conversions are consistent throughout the new code, the deviation sign convention is right, the span and segment guards mean there's no divide-by-zero on the distance ratios, there's no off-by-one in the block-sequence formula, and `getNearbyStopIDs(..., 0)` correctly means uncapped. That's the part I'd have expected to be buggy, and it isn't.

Three things are still wrong, and all three are spec conformance rather than crashes — which makes them easy to miss and expensive to ship, because clients will quietly compute wrong numbers.

**1. `vehicleId` lost its agency prefix, and this reaches well beyond this endpoint.**

```go
-  status.VehicleID = utils.FormCombinedID(agencyID, vehicle.ID.ID)
+  status.VehicleID = vehicle.ID.ID
```

The spec requires the combined form in three separate places: `vehicleId` is "Combined `{agencyId}_{vehicleId}` of the vehicle currently assigned", `tripStatus.vehicleId` is "Combined `{agencyId}_{vehicleId}`", and over in trip-details, `data.entry.status.vehicleId` is the "Combined `{agencyId}_{entityId}` identifier".

The change comment says this matches Java, but our spec documents Java's behavior and says the opposite. If you have a captured production response showing a bare ID, I'll happily reopen the question — but I'd want to see it, because this is load-bearing for clients that key vehicles across agencies.

The wider problem is that `BuildTripStatus` is shared. This doesn't just change arrivals-and-departures; it changes `tripStatus.vehicleId` for trip-details, trip-for-vehicle, and trips-for-route too. That's a breaking response change across four endpoints riding along in a PR about arrival deviations.

I'd also ask you to revert the two test changes that were flipped to assert the bare ID (`TestBuildTripStatus_VehicleIDFormat` and the pre-resolved-vehicle assertion). Those tests were correct before; changing an assertion to match new behavior is exactly how a regression becomes permanent.

**2. `distanceAlongTrip` collapsed onto `scheduledDistanceAlongTrip`.**

These are defined as different quantities: `scheduledDistanceAlongTrip` is "distance in metres from the start of the active trip to where the vehicle is *expected* to be at the query time, based on the schedule," while `distanceAlongTrip` is "metres from the start of the active trip to the vehicle's *real-time* position. Present only when real-time data is available."

With both set to the schedule-derived value, a client differencing them to get distance-based deviation always gets zero, and there's no longer any way to tell whether real-time position data was present at all. The removal of the `getVehicleDistanceAlongShapeContextual` call is where the real-time value went. `distanceAlongTrip` should be populated from real-time position and left absent when there's none.

**3. `numberOfStopsAway` sentinel differs between the singular and plural handlers.**

The two paths disagree on what to emit when the value can't be determined — one uses `-1`, the other `0`. Per the spec, `numberOfStopsAway` is a real signed quantity ("Negative when the vehicle has already passed"), so `-1` is a legitimate value meaning "one stop behind," not a sentinel. `arrival-and-departure-for-stop` and `arrivals-and-departures-for-stop` need to agree, and the no-data case should be the absent/zero behavior the spec describes for the no-real-time path.

Two smaller correctness notes on lines this PR adds, both narrow enough that I won't hold the merge on them, but worth fixing while you're in here:

- `InRange` is derived positionally from `stops[0]` / `stops[len-1]` while the same function computes and threads a `monotonic` flag into `findNextStopIndex` and `interpolateBlockDistance`. On a block with out-of-order stop times, `InRange` can come out false while `NextStopIndex >= 0`, silently suppressing all `distanceFromStop` and `numberOfStopsAway` output. `keepShiftContainingTrip` is supposed to prevent that state, so it's narrow — but it's an internal inconsistency.
- When a stop can't be resolved in `stopByID`, the code writes `distances[i] = 0`. That isn't a neutral default: `emitBlockStops` then records `DistanceAlongBlock = cumulativeBlockDist + 0`, placing a mid-trip stop back at the trip's start offset and feeding a bogus anchor into `interpolateBlockDistance`. The same happens wholesale when `fetchStopCoordsForStopTimes` returns nil after a `GetStopsByIDs` error — it's logged, but the response still looks fully populated while carrying wrong numbers. A sentinel that suppresses the row would be safer than a plausible-looking zero.

One thing I want to flag that isn't a defect: the per-row query cost. Only `computeScheduledBlockSnapshot` is memoized, so each arrival row still drives roughly a dozen uncached queries — `calculateBlockTripSequence`, `GetSituationIDsForTrip`, `blockShiftTripIDsSortedByStartTime`, `GetStopTimesForTrip`, `GetShapePointsByTripID` — plus a full shape load. Against a wide time window that's a lot of work per request. I don't want to expand this PR's scope to fix it, but let's not lose track of it.

Last, a process point, offered as a reason this has been slow rather than a criticism of your work. At ~4,100 added and ~900 deleted lines across 27 files, this is past the size where I can review it well in one pass, and it's why we've been round-tripping for so long. If you're willing, splitting it would get the good parts landed much faster — something like: (a) the block-distance/scheduled-snapshot helpers with their unit tests, (b) schedule deviation, (c) `tripStatus` population. Each of those is independently reviewable and independently useful. If you'd rather push this through as one unit, that's fine too and I'll keep reviewing it — but the three items above need fixing either way, and the `vehicleId` change should come out regardless since it affects four endpoints that aren't this PR's subject.

Happy to look again as soon as those are addressed.

Ahmedhossamdev:
Hey @aaronbrethorst,

For issue num 1:

> 1. `vehicleId` lost its agency prefix, and this reaches well beyond this endpoint.

In Java, when using a single GTFS-Realtime feed, the agency prefix is not added at all. Since we don't currently support multiple GTFS-Static feeds, I kept the behavior the same for now.

That said, I'll refactor it to always include the agency prefix. This will differ slightly from the current Java production behavior for single-feed deployments like Unitrans, but it will make the behavior consistent and better prepare us for supporting multiple feeds in the future.

burma-shave (approved):
Thanks for your work on this @Ahmedhossamdev!

Ahmedhossamdev:
Thanks, guys, for taking the time to review this! 🙏
@burma-shave  @aaronbrethorst

PR #1002, opened 2026-06-03
chore: fix errcheck and staticcheck linting errors

3rabiii:
### Changes Included:
* **`errcheck` Fixes:** Addressed unhandled errors from `rows.Close()` in `gtfsdb/block_layover_test.go` and `gtfsdb/stops_rtree.go` by wrapping them in anonymous `defer` functions.
* **`staticcheck` (QF1008) Fixes:** Removed redundant embedded field selectors (e.g., calling `.Time.Time` or `.ServiceDate.Time`) in `internal/models/current_time_test.go` and `internal/models/trip_details_test.go` to utilize Go's promoted methods for cleaner syntax.


gtfsdb/stops_rtree.go
```diff
@@ -42,7 +42,7 @@ func (q *Queries) GetActiveStopsWithinBounds(ctx context.Context, arg GetActiveS
 	if err != nil {
 		return nil, err
 	}
-	defer rows.Close()
+	defer func() { _ = rows.Close() }()
```
burma-shave: The four `defer rows.Close()` changes should be reverted. The pattern doesn't need changing — in `stops_rtree.go` it intentionally mirrors sqlc-generated code, where plain `defer rows.Close()` is correct because any error `Close()` could return is already surfaced by `rows.Err()`. The same applies in `block_layover_test.go`.

burma-shave (changes requested):

gtfsdb/stops_rtree.go
3rabiii: @burma-shave Ok, I made this because when I run make lint on my local it gives me an errcheck. I will revert to the past pattern, but should I comment with //nolint:errcheck to bypass the local linter or ignore this errcheck?
image

burma-shave: Since we're not using golangci-lint, I think it's better to avoid including comments for that linter. I believe you can define your own config for golanci-lint in a parent directory suppress the errors. We're relying on go vet and sonar for linting in CI.

burma-shave (approved):

PR #1019, opened 2026-06-05
fix(route): support includeReferences query parameter

### LINKED ISSUE #1007: spec-gap: support includeReferences query parameter in route endpoint
### Description:
The GET /api/where/route/{id} endpoint currently fetches and appends the agency reference unconditionally. Per the spec, if a client passes ?includeReferences=false, the server must skip the agency database lookup and return an empty array for data.references.agencies.

### Expected Behavior:
* includeReferences=true (or omitted): Fetch and include the agency.
* includeReferences=false: Skip GetAgency and return [] for agencies.

### Tests 
* Add tests for both includeReferences=true and false cases.

3rabiii:
### Description
This PR implements the `includeReferences` query parameter for the route endpoint, bringing the Maglev implementation into alignment with the OBA API specification. 

Previously, the route endpoint fetched and appended the agency reference unconditionally. Now, when a client passes `?includeReferences=false`, the server skips the agency database lookup entirely, saving a database call and returning an empty array for `data.references.agencies`.

### Changes Made
* **`route_handler.go`**: Added parsing for the `includeReferences` query parameter (defaulting to `true`). 
* **`route_handler_test.go`**: Added table-driven tests to verify both the explicit `false` behavior (empty agencies array) and the default/explicit `true` behavior (agency included).

fixes: #1007

## Summary by CodeRabbit

* **New Features**
  * Added an includeReferences query parameter for route responses. When false, agency references are omitted; included by default.

* **Tests**
  * Added test coverage validating includeReferences behavior (default/included, explicit false, and invalid values handled as default).

### DIFF
--- a/internal/restapi/reference_utils.go
+++ b/internal/restapi/reference_utils.go
@@ -2,6 +2,8 @@ package restapi
 
 import (
 	"context"
+	"net/http"
+	"strconv"
 	"time"
 
 	"github.com/OneBusAway/go-gtfs"
@@ -253,4 +255,20 @@ func (api *RestAPI) collectAlertsForStopsAndRoutes(stopIDs, routeIDs []string) [
 		alerts = append(alerts, api.GtfsManager.GetAlertsForRoute(routeID)...)
 	}
 	return deduplicateAlerts(alerts)
+}
+
+// ShouldIncludeReferences parses the "includeReferences" query parameter from the request.
+// It defaults to true if the parameter is absent or if it fails to parse as a boolean.
+func ShouldIncludeReferences(r *http.Request) bool {
+	val := r.URL.Query().Get("includeReferences")
+	if val == "" {
+		return true
+	}
+
+	parsed, err := strconv.ParseBool(val)
+	if err != nil {
+		return true
+	}
+
+	return parsed
 }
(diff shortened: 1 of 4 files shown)


ARCoder181105:
Hi @3rabiii ,

Nice work on this! But just to flag — `includeReferences` support doesn't exist anywhere in the codebase yet, and there are 20 handlers that build and populate references. 

So instead of doing this handler by handler, can we make it central?

I did a quick analysis and suggest adding a shared helper in `internal/restapi/reference_utils.go`:

```go
func shouldIncludeReferences(r *http.Request) bool {
    val := r.URL.Query().Get("includeReferences")
    if val == "" { return true }
    b, err := strconv.ParseBool(val)
    if err != nil { return true }
    return b
}
```

Each handler then just calls this instead of reimplementing the same logic. One thing to keep in mind — some endpoints have references that are always empty regardless of this parameter, so each handler still needs to handle its own spec rules on top of this.

3rabiii:
Hey @ARCoder181105, I just noticed after pushing this pr about this info that all handlers have the same bug, so i totally agree with you about this approach, so Have you already implemented this approach? if not i can update this pr with the new approach and also i will add a quick test for shouldIncludeReferences

ARCoder181105:
Great !, 
Please go ahead

PR #1035, opened 2026-06-06
Fix: Support Unix millisecond timestamps for schedule-for-stop date parameter

### LINKED ISSUE #1029: schedule-for-stop: Missing Unix millisecond support for date parameter
### Description
The `date` parameter in the `schedule-for-stop` endpoint only supports `YYYY-MM-DD` formatted strings. Passing a valid Unix timestamp results in a `400 Bad Request`.

### Spec Requirement
The `date` parameter must accept either a `YYYY-MM-DD` string OR a Unix millisecond integer.

### Implementation Gap
The handler only attempts to parse the parameter using `time.ParseInLocation("2006-01-02", ...)`. It needs to be updated to detect and parse integer Unix timestamps.

### Acceptance Criteria
- The `date` parameter is successfully parsed if provided as a Unix millisecond timestamp.
- The `date` parameter continues to parse successfully if provided as a `YYYY-MM-DD` string.
- Unit tests verify both date string parsing and Unix integer millisecond parsing.

ARCoder181105:
### Description
Updates the `schedule-for-stop` endpoint to accept Unix millisecond integers for the `date` parameter, resolving an issue where only `YYYY-MM-DD` formatted strings were supported. This brings the endpoint into full compliance with the API specification.

### Changes Made
- **Validation Update:** Modified `utils.ValidateDate` to accept valid 64-bit integers (Unix milliseconds) in addition to the standard `YYYY-MM-DD` format.
- **Handler Update:** Modified `scheduleForStopHandler` to parse the incoming integer timestamp, convert it to a timezone-aware date using the agency's location, and echo it back correctly in the response.
- **Testing:** Added new unit tests in both `validation_test.go` and `schedule_for_stop_handler_test.go` to verify integer parsing, error handling for invalid strings, and correct JSON property mapping.

Closes #1029 

## Release Notes

* **New Features**
  * The schedule API endpoint now accepts dates as Unix milliseconds in addition to the YYYY-MM-DD format.
  * Date responses now include precise millisecond timestamps in timezone-aware format.

* **Bug Fixes**
  * Improved date parsing and timezone handling for schedule queries.

### DIFF
--- a/internal/restapi/schedule_for_stop_handler.go
+++ b/internal/restapi/schedule_for_stop_handler.go
@@ -2,6 +2,7 @@ package restapi
 
 import (
 	"net/http"
+	"strconv"
 	"strings"
 	"time"
 
@@ -24,17 +25,7 @@ func (api *RestAPI) scheduleForStopHandler(w http.ResponseWriter, r *http.Reques
 	// Get the date parameter or use current date
 	dateParam := r.URL.Query().Get("date")
 
-	// Validate date parameter
-	if err := utils.ValidateDate(dateParam); err != nil {
-		fieldErrors := map[string][]string{
-			"date": {err.Error()},
-		}
-		api.validationErrorResponse(w, r, fieldErrors)
-		return
-	}
-
 	agency, err := api.GtfsManager.GtfsDB.Queries.GetAgency(ctx, agencyID)
-
 	if err != nil {
 		api.sendNotFound(w, r)
 		return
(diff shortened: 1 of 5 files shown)


ARCoder181105:
@coderabbitai full review

internal/restapi/schedule_for_stop_handler.go
burma-shave: This is not correctly initialised when date is passed in a unix timestamp. Need to set the time of day to midnight. Also it doesn't need to be in the loop. When no date is supplied, startOfDay is set to a default value further up the handler.

burma-shave: Lines 54 onward effectively do the same thing as utils.ValidateDate(). ValidateDate parses the date internally but discards the result. I'd suggest adding a func along the lines of `ParseDate` and return an error on an invalid date string .

burma-shave (changes requested):

internal/restapi/schedule_for_stop_handler.go
ARCoder181105: Done made the changes

ARCoder181105: Done

internal/utils/validation.go
```diff
@@ -167,3 +168,34 @@ func ValidateAndSanitizeQuery(query string) (string, error) {
...
 }
+
+const (
+	minUnixMillis = int64(0)
+	maxUnixMillis = int64(32503680000000) // year 3000
+)
+
+// ParseDate parses date strings in YYYY-MM-DD format or as a Unix millisecond integer.
+// It returns a time.Time set to midnight (start of day) in the provided location.
+func ParseDate(date string, loc *time.Location) (time.Time, error) {
```
burma-shave: Every other func in this file is called Validate*. This would be better placed in api.go where there are a number of other param parsing funcs.

burma-shave (changes requested):

internal/utils/validation.go
ARCoder181105: Done..

burma-shave (approved):

PR #1036, opened 2026-06-06
fix(routes): return 500 on DB failure instead of 404 during route lookup

### LINKED ISSUE #1009: bug: route lookup DB errors silently return 404 instead of 500
### Description:
In route_handler.go, the handler treats all errors returned by GetRoute as a 404 Not Found (e.g., if err != nil || route.ID == ""). This swallows actual database or connection errors.

### Expected Behavior:
The handler should distinguish between a missing record and a system failure:
* sql.ErrNoRows → Return 404 Not Found.
* Any other DB error → Return 500 Internal Server Error

### Tests:
* Add a test to simulate a DB failure and verify it returns a 500 error

3rabiii:
### Description
Previously, the `routeHandler` treated all errors returned by `GetRoute` as a `404 Not Found`. This was actively swallowing underlying database or connection errors and disguising them as missing records, making it difficult to debug actual system failures in production. 

This PR updates the handler to properly distinguish between a missing record and a genuine internal server error.

**Changes Made:**
* **Error Handling (`route_handler.go`):** * Refactored the error checking logic using `errors.Is(err, sql.ErrNoRows)`.
  * `sql.ErrNoRows` continues to correctly return a `404 Not Found`.
  * Any other database or connection error now properly triggers a `500 Internal Server Error`.
  * Preserved the `route.ID == ""` fallback check as an additional safety net.

**Tests:**
* **Added `TestRouteHandler_DatabaseError`:** Created a new integration test to verify the 500 response behavior. 

fixes: #1009 

## Summary by CodeRabbit

* **New Features**
  * Service date parameter now accepts calendar date format (yyyy-MM-dd) in addition to Unix timestamp format.

* **Bug Fixes**
  * Improved error handling to properly distinguish between missing routes and server errors.


internal/restapi/route_handler_test.go
```diff
@@ -108,6 +113,38 @@ func TestRouteHandlerWithSituations(t *testing.T) {
 	assert.Equal(t, alertID, model.Data.References.Situations[0].ID)
 }
 
+func TestRouteHandler_DatabaseError(t *testing.T) {
```
burma-shave: We don't currently test paths that return 500 errors anywhere else. So we should probably stick with that convention. It's very close to simple error propagation so doesn't really warrant a test.

burma-shave (changes requested):
L

burma-shave (approved):

Issue #1062, opened 2026-06-12
spec-gap: trip endpoint — routeShortName sourced from route, includeReferences ignored


Ahmedhossamdev:
## Summary

Reviewing `/api/where/trip/{id}` against the wiki spec and the Java OBA reference
(`onebusaway-application-modules`) surfaced two gaps. Both were confirmed against the live
Java server (unitrans feed) and the Java source.

---

### Spec

> `data.entry.routeShortName` — Short public name for the route **specific to this trip** (e.g. a
> variant name). May be an empty string if the trip does not carry its own route short name;
> clients **should fall back to the route's shortName in the references block**.

### Java source (the field comes from the trip, never the route)

- `GenerateNarrativesTask.java:590` — `builder.setRouteShortName(deduplicate(trip.getRouteShortName()))`
- `TripBeanServiceImpl.java:88` — `tripBean.setRouteShortName(tripNarrative.getRouteShortName())`
- `BeanFactoryV2.java:515` — `bean.setRouteShortName(trip.getRouteShortName())`

`trip.getRouteShortName()` is the GTFS `trips.txt` `route_short_name` extension column. unitrans
and RABA do not supply it, so Java returns `""` for every trip.

### Live comparison

```
GET /api/where/trip/unitrans_A_33_outbound_0755.json
  JAVA: "routeShortName": ""
  GO:   "routeShortName": "A"
```

Verified across 7 unitrans routes — Java returns `""` every time, while the route reference
carries the real short name.

### Cause

`internal/restapi/trip_handler.go` set `RouteShortName: route.ShortName.String`, reading from the
**routes** table. Maglev's `trips` table has no `route_short_name` column (only `trip_short_name`),
so the correct, parity-accurate value is `""`.

### Fix

Emit `""` for `entry.routeShortName`; clients fall back to `references.routes[].shortName` as the
spec instructs. Test assertion at `trip_handler_test.go` updated accordingly.

---

### Spec (Extension 6a)

> `includeReferences=false`: The references block is returned but all its sub-arrays are empty.
> The route and agency are not populated.

### Live comparison

```
GET /api/where/trip/{id}.json?includeReferences=false
  JAVA: references = { agencies:[], routes:[], stops:[], trips:[], situations:[] }
  GO:   references still fully populated with route + agency
```

### Cause

The handler never read the `includeReferences` query parameter.

### Fix

Parse `includeReferences` (default `true`) and only populate the references block when `true`.
The entry is always returned in full.

---

[...]

PR #1063, opened 2026-06-12
fix: trip endpoint spec gaps — routeShortName source and includeReferences

### LINKED ISSUE #987: Spec review: trip
## What to review
Read the [wiki spec](https://github.com/OneBusAway/maglev/wiki/trip) and compare the full Maglev implementation of this endpoint against it. Look for:

- **Request parameters**: all params supported with correct types, defaults, and validation
- **Response structure**: all fields present with correct names, types, and nesting
- **Error handling**: error cases match the spec (404, 400, etc.)
- **Behaviour**: the endpoint behaves as the spec describes, including edge cases
- **Test coverage**: main scenarios and edge cases from the spec are covered by tests

## Workflow
1. Compare the implementation against the wiki spec. If no gaps found, move this card to **Done**.
2. For each gap or bug, file a GitHub issue labelled `spec-gap`, link it to this card, and assign it to yourself.
3. Move this card to **Issues Filed**.
4. Work on the issues you filed.
5. Once all linked issues are closed, move this card back to **In Review** and repeat from step 1.

## Deeper analysis
The skills at https://github.com/burma-shave/maglev-workspace can help when the spec or implementation behaviour is not clear. They analyse the Java OBA reference implementation by downloading, building, and running the server — enabling both static and runtime analysis.

Install them into your coding agent (e.g. Claude Code). Key capabilities:
- **Q&A mode**: ask arbitrary questions about this endpoint and get answers grounded in the Java source and live server behaviour
- **Spec generation**: regenerate or cross-check the spec directly from the Java source

**Requirements**: Docker and Git
**Note**: the skills are token-intensive and trigger many permission prompts — a VM or sandbox environment is recommended

Ahmedhossamdev:
Fixes: #987, #1062

Fixes two spec gaps in `/api/where/trip/{id}`, both verified byte-for-byte against the Java reference server.

**1. `entry.routeShortName` sourced from the wrong table**
It read the route's short name; the spec/Java define it as the trip's own per-trip route short name (GTFS `trips.txt route_short_name`). Maglev doesn't store that column, so the parity-correct value is `""`. Clients fall back to `references.routes[].shortName` as the spec instructs.

**2. `includeReferences=false` ignored**
The handler never read the parameter and always populated references. Now parses it (default `true`) and returns empty reference arrays when `false`.

## Summary by CodeRabbit

* **New Features**
  * Added `includeReferences` query parameter to the trip endpoint. When set to `false`, reference data (routes, agencies, stops, trips, situations) are excluded from responses. References are included by default.

* **Bug Fixes**
  * Updated route short name field to reflect correct stored values.

* **Tests**
  * Added test coverage for the new `includeReferences` parameter functionality.

### DIFF
--- a/internal/restapi/trip_handler.go
+++ b/internal/restapi/trip_handler.go
@@ -74,14 +74,7 @@ func (api *RestAPI) tripHandler(w http.ResponseWriter, r *http.Request) {
 
 	// includeReferences defaults to true; when explicitly false the references
 	// block is returned with all sub-arrays empty (matches the Java reference).
-	includeReferences := true
-	if v := r.URL.Query().Get("includeReferences"); v != "" {
-		if parsed, err := strconv.ParseBool(v); err == nil {
-			includeReferences = parsed
-		}
-	}
-
-	if includeReferences {
+	if ShouldIncludeReferences(r) {
 		references.Routes = append(references.Routes, models.NewRoute(
 			utils.FormCombinedID(route.AgencyID, trip.RouteID),
 			route.AgencyID,
--- a/internal/restapi/trip_handler.go
+++ b/internal/restapi/trip_handler.go
@@ -49,15 +49,20 @@ func (api *RestAPI) tripHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	tripModel := &models.Trip{
-		ID:             utils.FormCombinedID(agencyID, trip.ID),
-		RouteID:        utils.FormCombinedID(agencyID, trip.RouteID),
-		ServiceID:      utils.FormCombinedID(agencyID, trip.ServiceID),
-		DirectionID:    strconv.FormatInt(trip.DirectionID.Int64, 10),
-		BlockID:        blockID,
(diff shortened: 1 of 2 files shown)


internal/restapi/trip_handler.go
```diff
@@ -67,29 +72,40 @@ func (api *RestAPI) tripHandler(w http.ResponseWriter, r *http.Request) {
...
-		route.LongName.String,
-		route.Desc.String,
-		models.RouteType(route.Type),
-		route.Url.String,
-		route.Color.String,
-		route.TextColor.String))
+	// includeReferences defaults to true; when explicitly false the references
+	// block is returned with all sub-arrays empty (matches the Java reference).
+	includeReferences := true
+	if v := r.URL.Query().Get("includeReferences"); v != "" {
```
burma-shave: There is a ShouldIncludeReferences() in reference_utils.go. This is a good place to use it.

burma-shave (changes requested):

burma-shave (approved):

PR #1082, opened 2026-06-18
fix: schedule-for-route returns 400 for invalid date param (#1042)

### LINKED ISSUE #1042: spec-gap: schedule-for-route invalid date param returns 510 instead of 400
## Summary

The `schedule-for-route` endpoint returns HTTP 200 / code 510 (`ServiceDateOutOfRange`) when
the `date` parameter cannot be parsed. The spec requires **HTTP 400 / code 400**.

## Spec reference

> **Extension 4a:** If the date value is not a valid YYYY-MM-DD string and not a pure decimal
> integer, the system returns HTTP 400 with code: 400 and a validation-error body.

## Current behaviour

```
GET /api/where/schedule-for-route/25_1.json?key=TEST&date=2025/06/12
```

```json
{ "code": 510, "text": "ServiceDateOutOfRange", ... }
```

HTTP status: **200**

## Expected behaviour

HTTP status: **400**, body `code: 400`.

## Affected code

- `internal/restapi/schedule_for_route_handler.go` lines 47–50 — falls through to 510 when
  epoch-ms parse also fails instead of returning 400
- `internal/restapi/schedule_for_route_handler_test.go` lines 122–125 and 233–235 — tests
  currently assert 510 and need to be updated to assert 400

Ahmedhossamdev:
Fixes #1042

## What
Unparseable `date` params on `schedule-for-route` now return **HTTP 400** with a
`{"fieldErrors":{"date":[...]}}` body, matching the Java OBA reference. Previously
they returned HTTP 200 with `code: 510` (`ServiceDateOutOfRange`).

## Changes
- Add `sendFieldError` helper for Spring-style `fieldErrors` 400 responses
- Return it from the handler when `date` is neither `YYYY-MM-DD` nor epoch-ms
- Update the two date tests to expect 400 + `fieldErrors`
- Remove an unused helper (`collectAlertsForStopsAndRoutes`) to clear a pre-existing lint failure on `main`

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid date parameters in API requests now return proper validation error responses with clearer field-level feedback instead of generic service errors.

### DIFF
--- a/internal/restapi/responses.go
+++ b/internal/restapi/responses.go
@@ -80,3 +80,20 @@ func (api *RestAPI) sendError(w http.ResponseWriter, r *http.Request, code int, 
 		api.serverErrorResponse(w, r, err)
 	}
 }
+
+// sendFieldError returns a Spring-framework-compatible 400 response with a
+// fieldErrors body, matching the Java OBA reference implementation.
+func (api *RestAPI) sendFieldError(w http.ResponseWriter, r *http.Request, field, message string) {
+	setJSONResponseType(&w)
+	w.WriteHeader(http.StatusBadRequest)
+
+	body := map[string]any{
+		"fieldErrors": map[string][]string{
+			field: {message},
+		},
+	}
+
+	if err := json.NewEncoder(w).Encode(body); err != nil {
+		api.serverErrorResponse(w, r, err)
+	}
+}
--- a/internal/restapi/responses.go
+++ b/internal/restapi/responses.go
@@ -80,20 +80,3 @@ func (api *RestAPI) sendError(w http.ResponseWriter, r *http.Request, code int, 
 		api.serverErrorResponse(w, r, err)
 	}
 }
-
-// sendFieldError returns a Spring-framework-compatible 400 response with a
-// fieldErrors body, matching the Java OBA reference implementation.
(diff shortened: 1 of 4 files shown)


internal/restapi/responses.go
```diff
@@ -80,3 +80,20 @@ func (api *RestAPI) sendError(w http.ResponseWriter, r *http.Request, code int,
 		api.serverErrorResponse(w, r, err)
 	}
 }
+
+// sendFieldError returns a Spring-framework-compatible 400 response with a
+// fieldErrors body, matching the Java OBA reference implementation.
+func (api *RestAPI) sendFieldError(w http.ResponseWriter, r *http.Request, field, message string) {
```
burma-shave: This is replicating the legacy error response format whereas the rest of maglev uses a different one. Should probably generate the response via `sendValidationErrorResponse`.

burma-shave (changes requested):

internal/restapi/responses.go
Ahmedhossamdev: Yeah, I thought we should match the Java implementation, but I'll revert it to use the standard error message instead.

burma-shave: I updated the spec to reflect that maglev should use the maglev style response body

burma-shave (approved):

PR #1083, opened 2026-06-18
fix: schedule-for-route headsign fallback to last stop name (#1041)

### LINKED ISSUE #1041: spec-gap: schedule-for-route missing headsign fallback to last stop name
## Summary

When a trip has no recorded headsign, the spec requires falling back to the name of the trip's
last stop. The implementation silently skips those trips, leaving `tripHeadsigns` incomplete for
routes that have trips without a headsign.

## Spec reference

> **Step 5:** Collects the trip's headsign (from narrative data; falls back to the name of the
> last stop if no headsign is recorded).

## Current behaviour

Trips where `trip_headsign` is empty are skipped entirely in the headsign collection loop.
If all trips in a direction group have no headsign, `tripHeadsigns` is an empty array.

## Expected behaviour

When `trip_headsign` is blank, look up the name of the last stop in the trip's stop sequence
and use that as the headsign string.

## Affected code

- `internal/restapi/schedule_for_route_handler.go` lines 184–192 — headsign collection loop
  has no fallback

## Notes

No existing test exercises this path. A test should be added alongside the fix.

Ahmedhossamdev:
Fixes #1041

When a trip has no `trip_headsign`, the name of its last stop is now used as the
headsign (matching the Java reference). Previously headsign-less trips were
silently skipped, leaving `tripHeadsigns` incomplete.

Note: RABA/unitrans fixtures populate all headsigns, so this fallback path isn't
covered by existing test data.

### DIFF
--- a/internal/restapi/schedule_for_route_handler.go
+++ b/internal/restapi/schedule_for_route_handler.go
@@ -181,16 +181,6 @@ func (api *RestAPI) scheduleForRouteHandler(w http.ResponseWriter, r *http.Reque
 			globalStopIDSet[stopID] = struct{}{}
 		}
 
-		seenHeadsigns := make(map[string]bool)
-		var headsigns []string
-		for _, trip := range tripsInGroup {
-			hs := trip.TripHeadsign.String
-			if hs != "" && !seenHeadsigns[hs] {
-				seenHeadsigns[hs] = true
-				headsigns = append(headsigns, hs)
-			}
-		}
-
 		rawTripIDs := make([]string, 0, len(tripsInGroup))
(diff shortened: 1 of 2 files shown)


burma-shave (changes requested):
Any thoughts on how to add a test for this?

Ahmedhossamdev:
> Any thoughts on how to add a test for this?


Yep!. The RABA and Unitrans fixtures already have headsigns, so we can insert a synthetic route with a headsign-less trip (same approach as the shapes test) and make sure tripHeadsigns falls back to the last stop's name.

burma-shave (approved):

PR #1084, opened 2026-06-18
fix: schedule-for-route route-scoped ServiceDateOutOfRange (#1043)

### LINKED ISSUE #1043: spec-gap: schedule-for-route ServiceDateOutOfRange uses feed end date instead of route end date
## Summary

The endpoint distinguishes `ServiceDateOutOfRange` (no future service for *this route*) from
`NoServiceThatDay` (no service today but future service exists for *this route*). The current
implementation compares against the feed's overall max end date rather than checking whether
this specific route has any future trips, so routes that end before the feed does return the
wrong 510 variant.

## Spec reference

> **Extension 3b (ServiceDateOutOfRange):** After step 5, if no block trips have any scheduled
> service on the requested date **and no service is scheduled for any later date**, the system
> returns HTTP 200 with code: 510, text: "ServiceDateOutOfRange".
>
> **Extension 3c (NoServiceThatDay):** After step 5, if no trips are active on the requested
> date **but the route does have service scheduled on later dates**, the system returns HTTP 200
> with code: 510, text: "NoServiceThatDay".

## Current behaviour

`GetFeedEndDate` returns `MAX(end_date)` across the entire calendar table (all routes, all
agencies). The handler returns `ServiceDateOutOfRange` if `targetDate > feedEndDate`, and
`NoServiceThatDay` for everything else with no active trips.

Scenario that produces the wrong result:
- Feed runs until 2026-06-30
- Route X last runs on 2025-12-31
- Query date: 2026-01-15 → returns `NoServiceThatDay` (wrong) — should be `ServiceDateOutOfRange`

## Expected behaviour

Check whether the queried route has any trips whose calendar is active on a date **after** the
requested date. If none exist, return `ServiceDateOutOfRange`; otherwise return `NoServiceThatDay`.

## Affected code

- `internal/restapi/schedule_for_route_handler.go` lines 68–77 — `ServiceDateOutOfRange` check
- `gtfsdb/query.sql` — likely needs a new query such as
  `HasFutureServiceForRoute(routeID, afterDate)` to drive the decision

## Notes

A new SQL query and a regression test covering a route whose service ends before the feed are
needed alongside the fix.

Ahmedhossamdev:
Fixes #1043

`ServiceDateOutOfRange` vs `NoServiceThatDay` now depends on whether the **queried route** has service after the requested date, via a new
`RouteHasFutureService` query not the feed-wide max end date. Routes that end before the feed does now return the correct 510 variant.

Also pins the conformance test to a fixed RABA service date, removing a time-dependent flake.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the schedule-for-route endpoint to return HTTP 200 instead of 510 when a route has no service on the requested date but may have service in the future.
  * Improved error messaging to distinguish between "no service on that day" and "route has no future service beyond the requested date."


internal/restapi/schedule_for_route_handler.go
```diff
@@ -105,8 +94,32 @@ func (api *RestAPI) scheduleForRouteHandler(w http.ResponseWriter, r *http.Reque
 		return
 	}
 
+	// noTripsResponse distinguishes the two 510 cases for this route: if the route
+	// has no service on any date after the requested one, it is ServiceDateOutOfRange;
+	// otherwise the route simply has no service that day (NoServiceThatDay).
+	noTripsResponse := func() (models.ResponseModel, error) {
```
burma-shave, line 103: I've updated the specs to reflect that we decided to fix this in the maglev implmentation.


The `schedule-for-route` spec has two Implementation Decisions that affect the `ServiceDateOutOfRange` response:

* Both "no service" responses should use `code: 200` in the body, not `code: 510`. The route exists and the request succeeded — the schedule for that date is just empty. The `text` field (`"ServiceDateOutOfRange"` / `"NoServiceThatDay"`) still carries the reason.
* `ServiceDateOutOfRange` should include the same partial data body as `NoServiceThatDay` — route ID, schedule date, empty `serviceIds` and `stopTripGroupings` arrays, plus agency and route references.

The `NoServiceThatDay` path already does the right thing via `buildNoServiceThatDayResponse()`. To bring `ServiceDateOutOfRange` in line, the `hasFuture == 0` branch in `noTripsResponse()` needs two changes:

1. Response code: `510` → `200`
2. Response body: `nil` → same partial-body structure as `NoServiceThatDay` (with `text: "ServiceDateOutOfRange"`)

The route-scoped future service check itself is correct — this is just about the response envelope.

Spec references: [schedule-for-route.md — Implementation Decisions](https://github.com/OneBusAway/maglev/wiki/schedule-for-route#implementation-decisions), [OBA-API-Defect-Workshop.md — DC-2, DC-3](https://github.com/OneBusAway/maglev/wiki/OBA-API-Defect-Workshop#dc-2-wrong-status-code-for-no-service-conditions)

burma-shave (changes requested):
The `noTripsResponse()` branch introduced here has two paths — `ServiceDateOutOfRange` and `NoServiceThatDay` but neither is exercised by any test. Adding test cases that hit both would be good coverage for this change.

internal/restapi/schedule_for_route_handler.go
Ahmedhossamdev: Cool, thanks, Eric!

burma-shave (approved):

PR #1092, opened 2026-06-20
fix: stop direction parity with Java reference

### LINKED ISSUE #1091: spec-gap: stop direction diverges from Java reference (~41% of stops)
## Summary

Stop `direction` is computed from shape geometry (it is not a GTFS field). maglev's bearing
diverged from the Java OBA reference on **112/275 unitrans stops** — a symmetric ±45° spread.
This is in `internal/gtfs/advanced_direction_calculator.go` and affects **every** endpoint that
returns a stop direction (stop, stops-for-agency, stops-for-route, arrivals, …), so it is split
out from the stops-for-agency response fixes.

## Root cause

1. **Wide chord vs single segment.** maglev computed the orientation over a wide **±5-point
   chord** around the nearest shape point. Java
   (`DistanceTraveledShapePointIndex.getPointAndOrientation`) computes it over the **single shape
   segment the stop lies on** (`computeOrientation(index-1, index)`). On curves, the wide chord
   rotates the bearing into an adjacent compass bucket.
2. **cos(latitude) scaling.** maglev applied an equirectangular `cos(latitude)` correction to the
   longitude delta; Java uses raw lon/lat deltas (`SphericalGeometryLibrary.getOrientation`).

## Fix

- Compute the bearing over the **single bracketing segment** (the adjacent segment the stop
  projects onto), not a ±5-point chord.
- Drop the `cos(latitude)` scaling to match Java's raw-delta angle.

## Result

112 → **3** mismatches (98.9% parity with Java on unitrans).

The remaining 3 (`22036`, `22169`, `22294`) are precision-floor boundary cases: maglev selects
the segment geometrically, while Java projects via `shape_dist_traveled` (which it synthesizes at
bundle build — the feed has 0/97,740 stop distances). At these 3 stops the segment choice differs
by one, tipping the consistency `stdDev` across the 0.7 cutoff (22036, 22169) or the median angle
across a 45° compass boundary (22294). Closing them would require porting Java's per-stop distance
projection — disproportionate for 1.1% of stops on threshold/bucket edges.

## Notes

- Directions are **precomputed at GTFS load**; verifying the fix live requires rebuilding the DB
  (delete `gtfs.db`) and restarting so directions recompute.
- Existing orientation tests only assert the result is within `[-π, π]`, so nothing pinned the
  old behavior.

Ahmedhossamdev:
Fixes: #1091

Direction is computed from shape geometry. maglev diverged from Java on 41% of stops because it measured the bearing over a **wide ±5-point chord**; Java uses the **single shape segment the stop lies on**. Also dropped a `cos(latitude)` longitude correction that Java doesn't apply. Now matches Java on 272/275 stops; the remaining 3 are threshold/compass-boundary edge cases (Java projects via synthesized `shape_dist_traveled`, which the feed lacks). This fix improves every stop-direction-bearing endpoint, not just this one.

### DIFF
--- a/internal/gtfs/advanced_direction_calculator.go
+++ b/internal/gtfs/advanced_direction_calculator.go
@@ -19,7 +19,6 @@ import (
 
 const (
 	defaultStandardDeviationThreshold = 0.7
-	shapePointWindow                  = 5
 )
 
 // AdvancedDirectionCalculator implements the OneBusAway Java algorithm for stop direction calculation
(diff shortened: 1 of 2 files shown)


burma-shave (changes requested):
Just needs some tests

Ahmedhossamdev:
Done @burma-shave

burma-shave (approved):

PR #1093, opened 2026-06-20
fix: stops-for-agency response gaps — 404, code fallback, parent stations

### LINKED ISSUE #1090: spec-gap: stops-for-agency — 404 handling, code fallback, parent station references
## Gap 1 — Unknown agency returns HTTP 200 + `null` instead of 404

The handler called `sendNull` for an unknown agency. The spec's Implementation Decisions section
states maglev should return **404** (intentionally correcting the legacy Java defect), and the
sibling `stop-ids-for-agency` already does. The two endpoints were inconsistent, and a test
(`TestStopsForAgencyInvalidAgency`) asserted the wrong 200.

**Fix:** return 404 via `sendNotFound`; update the test.

## Gap 2 — `code` did not fall back to the entity ID

The handler set `Code: stop.Code.String` (empty when the feed defines no code). Spec: *"Falls back
to the entity portion of the combined stop ID when the feed does not define a code."* The
single-stop handler already does this.

**Fix:** `nulls.StringOrDefault(stop.Code, stop.ID)`.

## Gap 3 — `parent` never populated; parent stations missing from `references.stops`

The handler never set `Parent`, and never added parent stations to `references.stops`
(spec Extension 3a). The single-stop handler does both.

**Fix:** populate `Parent` from `stop.ParentStation`; resolve unique parent stations into
`references.stops` via a new `buildParentStationReferences` helper.

Ahmedhossamdev:
Fixes: #1090

## Summary
- Unknown agency → 404 (was 200 + null), matching stop-ids-for-agency and the spec
- code falls back to the entity ID when the feed defines none
- parent populated + parent stations resolved into references.stops (Extension 3a)

## Summary by CodeRabbit

* **New Features**
  * Stop and trip responses now include richer parent-station references, including wheelchair boarding.
  * Route and static route IDs are populated per stop with deterministic ordering and de-duplicated references.

* **Bug Fixes**
  * Unknown agencies now return a proper “not found” response (404).
  * Empty stop inputs return clean, empty results.
  * Duplicate stop IDs are collapsed to avoid repeated stop entries.
  * Parent-station reference lookup failures now fail the request instead of being silently ignored.

* **Refactor**
  * Route name sorting is now more natural and consistent.

### DIFF
--- a/internal/restapi/reference_utils.go
+++ b/internal/restapi/reference_utils.go
@@ -347,7 +347,7 @@ func groupRoutesByStop(agencyID string, allRoutes []gtfsdb.GetRoutesForStopsRow)
 func (api *RestAPI) combinedRouteIDsForStop(agencyID string, routesForStop []gtfsdb.Route) []string {
 	// Sort naturally by ShortName (falling back to LongName, then AgencyID, then ID) so the
 	// route IDs are returned in a stable, human-friendly order.
-	utils.SortRoutesByName(routesForStop, utils.DBRouteSortKey)
+	utils.SortRoutesByName(routesForStop)
 	combinedRouteIDs := make([]string, len(routesForStop))
 	for i, rt := range routesForStop {
 		combinedRouteIDs[i] = utils.FormCombinedID(agencyID, rt.ID)
--- a/internal/restapi/reference_utils.go
+++ b/internal/restapi/reference_utils.go
@@ -347,7 +347,7 @@ func groupRoutesByStop(agencyID string, allRoutes []gtfsdb.GetRoutesForStopsRow)
 func (api *RestAPI) combinedRouteIDsForStop(agencyID string, routesForStop []gtfsdb.Route) []string {
 	// Sort naturally by ShortName (falling back to LongName, then AgencyID, then ID) so the
 	// route IDs are returned in a stable, human-friendly order.
-	utils.SortDBRoutesByName(routesForStop)
+	utils.SortRoutesByName(routesForStop, utils.DBRouteSortKey)
 	combinedRouteIDs := make([]string, len(routesForStop))
 	for i, rt := range routesForStop {
 		combinedRouteIDs[i] = utils.FormCombinedID(agencyID, rt.ID)
--- a/internal/restapi/reference_utils.go
+++ b/internal/restapi/reference_utils.go
@@ -1,10 +1,8 @@
 package restapi
 
 import (
-	"cmp"
 	"context"
 	"net/http"
-	"slices"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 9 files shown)


internal/restapi/stops_for_agency_handler.go
```diff
@@ -61,10 +61,66 @@ func (api *RestAPI) stopsForAgencyHandler(w http.ResponseWriter, r *http.Request
...
+	}
+	references.Stops = parentRefs
+
 	response := models.NewListResponse(stopsList, *references, false, api.Clock)
 	api.sendResponse(w, r, response)
 }
 
+// buildParentStationReferences resolves the unique parent stations referenced by
+// stopsList into minimal Stop records for the references block.
+func (api *RestAPI) buildParentStationReferences(ctx context.Context, agencyID string, stopsList []models.Stop) ([]models.Stop, error) {
```
burma-shave, line 78: `buildParentStationReferences` leaves `RouteIDs` and `StaticRouteIDs` as empty slices on the parent station records it writes to `references.stops`. The spec said the stop records should be 'minimal' but no other spec that does this says the same, the specify full records. I've update the spec so that it's consistent with the others and now requires a full recoed.

There is already a shared helper that does what this function needs: `BuildStopReferencesAndRouteIDsForStops` in `trip_for_vehicle_handler.go`. It should be made a shared function in `reference_utils.go` rather than a handler file.

Suggested approach: move `BuildStopReferencesAndRouteIDsForStops` to `reference_utils.go` and use it in:

- `stops_for_agency_handler.go` (this PR) — replacing `buildParentStationReferences`
- `stop_handler.go` — replacing the inline parent-station block
- `trip_details_handler.go` — replacing the duplicate `buildStopReferences` method

`trip_for_vehicle_handler.go` and `schedule_for_route_handler.go` already call the function and would just benefit from the file move.

burma-shave (changes requested):

internal/restapi/reference_utils.go
```diff
@@ -267,3 +270,96 @@ func ShouldIncludeReferences(r *http.Request) bool {
...
+	modelStops := make([]models.Stop, 0, len(uniqueStopIDs))
+	for _, stopID := range uniqueStopIDs {
+		stop, exists := stopMap[stopID]
+		if !exists {
+			continue
+		}
+		routesForStop := routesByStop[stopID]
+		// Sort naturally by ShortName (falling back to LongName, then ID) so the
+		// route IDs are returned in a stable, human-friendly order.
+		slices.SortFunc(routesForStop, func(a, b gtfsdb.Route) int {
```
burma-shave: There's a similar sort function already defined in sort.go. This one could be extracted and placed along side the other. This would help address the sonar flagged complexity issue.

internal/restapi/reference_utils.go
```diff
@@ -267,3 +270,96 @@ func ShouldIncludeReferences(r *http.Request) bool {
 
 	return parsed
 }
+
+// BuildStopReferencesAndRouteIDsForStops builds full stop references and collects unique routes for the given stop IDs.
+func BuildStopReferencesAndRouteIDsForStops(api *RestAPI, ctx context.Context, agencyID string, stopIDs []string) ([]models.Stop, map[string]gtfsdb.GetRoutesForStopsRow, error) {
```
burma-shave, line 273: It's be worth addressing the sonar flagged complexity issue and also adding tests.

burma-shave (changes requested):

internal/utils/sort.go
```diff
@@ -7,45 +7,54 @@ import (
...
-		return cmp.Compare(agencyIDA, agencyIDB)
+	if a.agencyID != b.agencyID {
+		return cmp.Compare(a.agencyID, b.agencyID)
 	}
-	return cmp.Compare(idA, idB)
+	return cmp.Compare(a.id, b.id)
 }
 
 // SortRoutesByName sorts routes naturally by ShortName, falling back to LongName, then by AgencyID and ID.
 func SortRoutesByName(routes []gtfsdb.GetRoutesForStopRow) {
```
burma-shave: These names now need to be tidied up. I believe the convention in go is name the type in the name, so Sort*RoutesForStopRows*ByName and Sort*Routes*ByName. But an even better solution would be to *use generics* and have a single func.

burma-shave (changes requested):

Ahmedhossamdev:
Done @burma-shave

burma-shave (approved):
Looks great. I partially undid the use of generics as once you'd implemented them I could see that the type param wasn't really doing anything.

PR #1099, opened 2026-06-20
fix: validate agency namespace for stops 

### LINKED ISSUE #1094: spec-gap: Missing Agency Namespace Validation
### Description
The `/stop` endpoint parses the requested combined ID (e.g., `Agency_Stop123`) to extract the agency ID, but fails to actually use it when querying the database.

### Spec Requirement
* "The system parses the ID by splitting on the first underscore: everything before it is the agency ID; everything after is the stop's entity ID."

### Implementation Gap
* The `stopHandler` extracts the `agencyID`, but the underlying SQL call `GetStop` queries solely by the entity ID. 
* This allows a caller to request `WrongAgency_123` and successfully retrieve the stop if entity `123` exists in *any* agency's data.

### Acceptance Criteria
* The handler (or SQL query) validates that the retrieved stop is actually served by the requested `agencyID` (or is the default agency).
* A request with an invalid agency namespace for a valid stop ID correctly returns a `404 Not Found`.
* A new test case `TestStopHandler_WrongAgency` asserts this behavior.

ARCoder181105:
### Description
Resolves #1094 by ensuring the `/stop` endpoint strictly validates the requested agency namespace. 

### Changes Made
* **Validation Added:** The handler now verifies that the requested `agencyID` actually serves the retrieved stop. If the agency does not have any routes serving the stop, it returns a `404 Not Found` instead of falsely returning the stop.
* **Orphaned Stop Fallback:** Because Maglev's `stops` table lacks an explicit `agency_id` column, if a stop has *no routes at all*, the handler allows retrieval as long as the requested `agencyID` exists in the database.
* **Test Coverage:** Added `TestStopHandler_WrongAgency` to guarantee that requesting a valid stop under an incorrect or non-existent agency namespace correctly triggers a 404.

### Testing
* Run `make test` to verify the new `TestStopHandler_WrongAgency` assertions pass successfully.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the stop endpoint to enforce agency namespace ownership: requests with a mismatched or non-existent agency namespace now return **404 Not Found**.
  * Stops without associated routes follow existence-based agency validation, and reference population respects the same ownership rules.
* **Tests**
  * Added coverage for wrong-namespace scenarios, verifying both routed and orphan stop cases (including behavior when reference inclusion is disabled).

### DIFF
--- a/internal/restapi/stop_handler.go
+++ b/internal/restapi/stop_handler.go
@@ -36,10 +36,35 @@ func (api *RestAPI) stopHandler(w http.ResponseWriter, r *http.Request) {
 	utils.SortRoutesByName(routes)
 
 	combinedRouteIDs := make([]string, len(routes))
+	uniqueAgencyIDs := make(map[string]bool)
+
 	for i, route := range routes {
 		// Use route.AgencyID, not the stop's agencyID.
 		// A stop can be served by routes from other agencies.
 		combinedRouteIDs[i] = utils.FormCombinedID(route.AgencyID, route.ID)
+		uniqueAgencyIDs[route.AgencyID] = true
+	}
+
+	// Validate the requested agency namespace
+	if len(routes) > 0 {
+		if !uniqueAgencyIDs[agencyID] {
+			// Stop exists, but is not served by the requested agency namespace.
+			api.sendNotFound(w, r)
+			return
+		}
+	} else {
+		// If the stop has no routes, we allow it to be retrieved under any valid agency namespace
+		// because Maglev stops do not have a dedicated agency_id column in the schema.
+		// Just ensure the requested agency actually exists.
+		agency, err := api.GtfsManager.FindAgency(ctx, agencyID)
+		if err != nil {
+			api.serverErrorResponse(w, r, err)
+			return
+		}
+		if agency == nil {
+			api.sendNotFound(w, r)
+			return
+		}
 	}
 
 	parentID := ""
@@ -66,7 +91,6 @@ func (api *RestAPI) stopHandler(w http.ResponseWriter, r *http.Request) {
 
 	// Only populate references if the query parameter is absent or true
 	if ShouldIncludeReferences(r) {
-		uniqueAgencyIDs := make(map[string]bool)
 
 		// Add routes to references and collect unique agency IDs
 		for _, route := range routes {
--- a/internal/restapi/stop_handler.go
+++ b/internal/restapi/stop_handler.go
@@ -90,7 +90,9 @@ func (api *RestAPI) stopHandler(w http.ResponseWriter, r *http.Request) {
 	references := models.NewEmptyReferences()
 
 	// Only populate references if the query parameter is absent or true
-	if ShouldIncludeReferences(r) {
+	includeReferences := ShouldIncludeReferences(r)
+
+	if includeReferences {
 
 		// Add routes to references and collect unique agency IDs
 		for _, route := range routes {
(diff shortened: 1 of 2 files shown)


ARCoder181105:
@coderabbitai full review

internal/restapi/stop_handler.go
Ahmedhossamdev: Nice work on this! The namespace validation logic looks right and the 404 vs 200 behavior is spot on. Left a few small notes below. Only the first one really needs changing, the rest are just optional stuff, Great work overall! @ARCoder181105

Ahmedhossamdev: I think this line is a leftover. You already set uniqueAgencyIDs[route.AgencyID] = true up in the first loop (line 45), so setting it again here doesn't do anything. Can we drop it? Also keeps things from breaking quietly if someone adds route filtering later.

internal/restapi/stop_handler_test.go
```diff
@@ -512,3 +512,89 @@ func TestStopHandler_IncludeReferences(t *testing.T) {
...
+	})
+	require.NoError(t, err)
+	_, err = q.CreateStopTime(ctx, gtfsdb.CreateStopTimeParams{
+		TripID: tripID, StopID: stopID, StopSequence: 1, ArrivalTime: 36000, DepartureTime: 36000,
+	})
+	require.NoError(t, err)
+
+	// Test 1: Request with the valid agency namespace -> 200 OK
+	respValid, modelValid := callAPIHandler[StopEntryResponse](t, api, stopURL(utils.FormCombinedID(validAgency, stopID)))
+	require.Equal(t, http.StatusOK, respValid.StatusCode)
```
Ahmedhossamdev, line 578: Tests look good! Only thing, the happy path ones just check the status code. Could add a quick check on the body so a shape regression doesn't sneak by:

assert.Equal(t, utils.FormCombinedID(validAgency, stopID), modelValid.Data.Entry.ID)

internal/restapi/stop_handler_test.go
```diff
@@ -512,3 +512,89 @@ func TestStopHandler_IncludeReferences(t *testing.T) {
...
+	assert.Equal(t, http.StatusOK, modelValid.Code)
+
+	// Test 2: Request with the invalid agency namespace -> 404 Not Found
+	respInvalid, modelInvalid := callAPIHandler[StopEntryResponse](t, api, stopURL(utils.FormCombinedID(invalidAgency, stopID)))
```
Ahmedhossamdev, line 583: Good coverage here. Since the validation runs before the ShouldIncludeReferences check, might be nice to add a ?includeReferences=false version of the wrong agency case just to lock that in. Low priority, feel free to skip.

burma-shave (changes requested):
Just needs Ahmed's comments to be addressed

internal/restapi/stop_handler.go
ARCoder181105: Thanks.. I addressed the requested changes

internal/restapi/stop_handler_test.go
ARCoder181105: Done...

internal/restapi/stop_handler_test.go
ARCoder181105: Done...

internal/restapi/stop_handler.go
ARCoder181105: I dropped the map...
Done..

Ahmedhossamdev (commented):
Great work @ARCoder181105 
I think it's ready to merge now.

burma-shave (approved):
Looks great, but keep commit messages relevant. "address PR review comments" is not helpful in a commit message. Describe the changes the commit makes so they can be understood independently of the PR.

PR #1100, opened 2026-06-20
fix(trip-details): omit status key when tracking is absent

### LINKED ISSUE #1086: trip-details: status key is present when block has no current tracking record
### Description
According to the specification (Extension 4e), if the block has no current tracking record (e.g., the trip hasn't started, has finished, or lacks location data), the `status` key must be **absent entirely** from the response entry.

### Expected Behavior
The `"status"` field should be omitted from the JSON output when there is no active vehicle tracking data.

### Actual Behavior
Even without real-time tracking, `BuildTripStatus` instantiates and returns a default `TripStatus` object populated with `"status": "default"` and `"phase": "scheduled"`. Because the struct pointer is not `nil`, the `omitempty` tag on the `Status` field in `internal/models/trip_details.go` fails to trigger, and the field is incorrectly emitted in the JSON response.

3rabiii:
## Description
This PR addresses **Extension 4e** of the `trip-details` specification, which requires the `status` key to be entirely absent from the response if the block has no current tracking record.

Previously, `BuildTripStatus` returned a default non-nil placeholder (`Status="default"`, `Predicted=false`) when no vehicle was tracked, which prevented the `omitempty` JSON tag from dropping the field. 

**Changes made:**
* Explicitly set the `status` pointer to `nil` in `tripDetailsHandler` when the returned status is just the default placeholder.
* Added `TestTripDetailsHandlerStatusOmittedWhenNoTracking` to explicitly verify compliance with Extension 4e.
* Updated `TestServiceDateTimezoneRegression_BlockTripSequence` to inject a mock vehicle, ensuring the timezone calculation still has valid tracking data to test against.

Closes: #1086 

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the trip-details API to omit the `status` field entirely when no vehicle tracking data is available (instead of returning an untracked placeholder).  
  * Added/updated coverage to ensure the endpoint doesn’t include `status` in the raw JSON response when tracking is missing, and added unit tests for the untracked status detection logic.

### DIFF
--- a/internal/restapi/trip_details_handler.go
+++ b/internal/restapi/trip_details_handler.go
@@ -200,7 +200,7 @@ func (api *RestAPI) tripDetailsHandler(w http.ResponseWriter, r *http.Request) {
 
 		// Extension 4e: Explicitly nil out the status if there is no actual tracking record.
 		// BuildTripStatus returns a default placeholder when tracking is absent, so we nil it to trigger JSON omitempty.
-		if status != nil && status.Status == "default" && !status.Predicted {
+		if status != nil && status.IsUntracked() {
 			status = nil
 		}
 	}
--- a/internal/restapi/trip_details_handler.go
+++ b/internal/restapi/trip_details_handler.go
@@ -197,6 +197,12 @@ func (api *RestAPI) tripDetailsHandler(w http.ResponseWriter, r *http.Request) {
 				"error", statusErr.Error())
 			status = nil
 		}
+
+		// Extension 4e: Explicitly nil out the status if there is no actual tracking record.
+		// BuildTripStatus returns a default placeholder when tracking is absent, so we nil it to trigger JSON omitempty.
+		if status != nil && status.Status == "default" && !status.Predicted {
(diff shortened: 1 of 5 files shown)


internal/restapi/trip_details_handler.go
```diff
@@ -197,6 +197,12 @@ func (api *RestAPI) tripDetailsHandler(w http.ResponseWriter, r *http.Request) {
 				"error", statusErr.Error())
 			status = nil
 		}
+
+		// Extension 4e: Explicitly nil out the status if there is no actual tracking record.
+		// BuildTripStatus returns a default placeholder when tracking is absent, so we nil it to trigger JSON omitempty.
+		if status != nil && status.Status == "default" && !status.Predicted {
```
burma-shave: A helper func like status.IsUntracked() would make the intent clearer here.

burma-shave (changes requested):

burma-shave (approved):

PR #1106, opened 2026-06-25
fix: honor time request param in vehicles-for-agency API

### LINKED ISSUE #1102: [spec-gap] vehicles-for-agency: time request parameter is ignored
### Description
Per the spec (Main Success Scenario step 2; Request Parameters), the `time` parameter sets the reference time used for two purposes: determining which trip each vehicle is serving at that moment, and computing the cutoff for the `ageInSeconds` staleness filter. The parameter accepts a Unix timestamp or a `yyyy-MM-dd_HH-mm-ss` string.

### Expected Behavior
When the `time` parameter is supplied in the request, the server should parse it and use it as the effective reference time for the response. When absent, it should default to the server's current time.

### Actual Behavior
In `vehicles_for_agency_handler.go`, the handler never reads the `time` query parameter. It unconditionally uses `api.Clock.Now()` for the reference time. A caller supplying a specific `time` gets the same response as if they had omitted it, and the parameter is silently ignored.

3rabiii:
### Summary

Fixes a spec gap in the `vehicles-for-agency` endpoint where the `time` request parameter was previously ignored. The handler now properly parses the supplied time to use as the reference time, falling back to the server clock if absent, as required by the OpenAPI spec.

### Problem

According to the specification, the `time` parameter should set the reference time used for trip resolution. However, the handler was unconditionally using `api.Clock.Now()`. This meant that historical or future time parameters sent by clients were silently ignored.

### Changes

- Parse the `time` query parameter using `utils.ParseTimeParameter`.
- Fall back to `api.Clock.Now()` if the `time` parameter is absent.
- Return an HTTP 400 validation error if the provided time is unparseable.
- Use the parsed reference time instead of the server clock to populate `currentTime` for entry/trip-status timestamps.
- Add three new tests to verify behavior with epoch-ms input, absent input, and invalid string input.

Closes #1102 

## Summary by CodeRabbit

* **Bug Fixes**
  * `/vehicles-for-agency` now uses a consistent reference time derived from the agency’s location (and the provided time, when present) for vehicle and trip timestamps.
  * Invalid `time` query values now return a clear validation error (HTTP 400) instead of producing an inconsistent response.
* **New Features**
  * Added support for an optional `time` query parameter to generate results for a specific moment in time.
  * The `time` parameter can now be provided in additional datetime-like formats when it includes a dash-separated timestamp.
* **Tests**
  * Expanded coverage for `time` parsing and endpoint behavior.


internal/restapi/vehicles_for_agency_handler.go
```diff
@@ -29,6 +29,22 @@ func (api *RestAPI) vehiclesForAgencyHandler(w http.ResponseWriter, r *http.Requ
...
 
+	// Parse requested reference time for status entries, falling back to server clock if absent.
+	referenceTime := api.Clock.Now()
+	if timeParam := r.URL.Query().Get("time"); timeParam != "" {
+		loc, err := loadAgencyLocation(agency.ID, agency.Timezone)
+		if err != nil {
+			api.serverErrorResponse(w, r, err)
+			return
+		}
+		_, parsedTime, fieldErrors, ok := utils.ParseTimeParameter(timeParam, loc)
```
burma-shave, line 40: The spec lists two valid formats for `time`: epoch ms and `yyyy-MM-dd_HH-mm-ss`. This PR handles epoch ms but `ParseTimeParameter` only tries `"2006-01-02"`, so `"2024-03-15_12-00-00"` still returns 400. Update ParseTimeParameter to handler the format in the spec

burma-shave (changes requested):

burma-shave (approved):

PR #1116, opened 2026-06-30
fix: default missing update times to 0 instead of server clock for vehicles-for-agency

### LINKED ISSUE #1115: [spec-gap] vehicles-for-agency: lastUpdateTime/lastLocationUpdateTime not "absent when zero"
### Description
Per maglev.wiki/vehicles-for-agency.md, lastUpdateTime and lastLocationUpdateTime must be absent when there's no update — on both the outer data.list[] entry and the nested tripStatus. Maglev always emits them as 0 instead.
Scope to vehicles-for-agency only: the shared models.TripStatus is reused by endpoints whose specs require these fields to stay 0.
### Expected behaviour
- No update → lastUpdateTime / lastLocationUpdateTime omitted (outer entry + tripStatus).
- Has update → present as Unix-ms integers.
- No impact on other endpoints sharing TripStatus.
### Actual behaviour
- Both fields always present, emitted as 0 even with no update.
- Handler defaults them to the current clock time, so "no update" looks like a real update.

3rabiii:
### Summary

This PR aligns the `vehicles-for-agency` endpoint with the updated API spec. It removes the previous behavior of falling back to a fabricated server time (`api.Clock.Now()`) when no real-time update has been received. Instead, missing timestamps now default to `0`, which correctly reflects the absence of real-time data.

### Changes

- **Handler:** Removed the fallback to `api.Clock.Now()` when a vehicle lacks a timestamp.
- **Cleanup:** Removed obsolete `omitzero` tags and custom wrappers, allowing `0` to be explicitly emitted per the updated spec.

### Why this approach?

As documented by the maintainers in the wiki, standardizing on emitting `0` across endpoints that share the `TripStatus` model simplifies the marshaling logic and provides an unambiguous sentinel value for clients, avoiding the complexity of per-endpoint omission wrappers.

### Testing

- Kept the `NoTimestamp` mock option.
- Updated integration tests to assert that update times are `0` (rather than absent) when no update is received.
- Verified that existing tests for other endpoints sharing `TripStatus` remain unaffected.

Closes: #1115 

## Summary by CodeRabbit

* **Bug Fixes**
  * Vehicle and trip status timestamps now remain unset when the vehicle update timestamp is not provided, instead of being filled with a default server time.
  * Nested trip status time fields now follow the same “present only when expected” behavior.

* **Tests**
  * Added end-to-end coverage for `vehicles-for-agency` responses to verify `lastUpdateTime` and `lastLocationUpdateTime` are included and non-zero only when expected.

### DIFF
--- a/internal/models/vehicle.go
+++ b/internal/models/vehicle.go
@@ -2,8 +2,8 @@ package models
 
 type VehicleStatus struct {
 	VehicleID              string      `json:"vehicleId"`
-	LastLocationUpdateTime ModelTime   `json:"lastLocationUpdateTime"`
-	LastUpdateTime         ModelTime   `json:"lastUpdateTime"`
+	LastLocationUpdateTime ModelTime   `json:"lastLocationUpdateTime,omitzero"`
+	LastUpdateTime         ModelTime   `json:"lastUpdateTime,omitzero"`
 	Location               *Location   `json:"location"`
 	TripID                 string      `json:"tripId"`
 	TripStatus             *TripStatus `json:"tripStatus"`
--- a/internal/models/vehicle.go
+++ b/internal/models/vehicle.go
@@ -2,8 +2,8 @@ package models
 
 type VehicleStatus struct {
 	VehicleID              string      `json:"vehicleId"`
-	LastLocationUpdateTime ModelTime   `json:"lastLocationUpdateTime,omitzero"`
-	LastUpdateTime         ModelTime   `json:"lastUpdateTime,omitzero"`
+	LastLocationUpdateTime ModelTime   `json:"lastLocationUpdateTime"`
+	LastUpdateTime         ModelTime   `json:"lastUpdateTime"`
 	Location               *Location   `json:"location"`
 	TripID                 string      `json:"tripId"`
 	TripStatus             *TripStatus `json:"tripStatus"`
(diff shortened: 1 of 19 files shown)


internal/models/vehicle.go
```diff
@@ -2,8 +2,8 @@ package models
 
 type VehicleStatus struct {
 	VehicleID              string      `json:"vehicleId"`
-	LastLocationUpdateTime ModelTime   `json:"lastLocationUpdateTime"`
-	LastUpdateTime         ModelTime   `json:"lastUpdateTime"`
+	LastLocationUpdateTime ModelTime   `json:"lastLocationUpdateTime,omitzero"`
```
burma-shave: The absent-vs-zero distinction is more complexity than it's worth. Let's standardise on emitting `0` across all endpoints that share `TripStatus`. I've documented the deviation from legacy behaviour in the wiki specs.

**Keep:** the `NoTimestamp` mock option, the two new integration tests (updated to assert `0` not absence), and the removal of the six `= currentTime` fallback lines.

**Drop:** the `omitzero` tags on `VehicleStatus`, the `vehicleStatusResponse` wrapper and `MarshalJSON`, and `wrapVehicleStatusesForResponse`.

burma-shave (changes requested):

burma-shave (approved):
Looks great, but the commit message could have been better: "Merge main, resolve conflicts, and apply mentor review". Could have been two commits, and referring to 'mentor review' is unhelpful.

Issue #1124, opened 2026-07-01
stops-for-route: build polylines with shape-set merge, not headsign grouping


Ahmedhossamdev:
# Summary

`processTripGroups` currently builds group polylines by deduplicating trips by **trip headsign** and calling `GetShapesGroupedByTripHeadSign`, which selects a single `shape_id` per headsign using `LIMIT 1`. This produces one merged polyline per unique headsign.

This behavior does not match the Java implementation.

## Expected Behavior

Java (`ShapeBeanServiceImpl.getMergedPolylinesForShapeIds`) builds both group-level and entry-level polylines from a **set of distinct `shape_id`s**.

### Group-level polylines

Group-level polylines are built from the distinct `shape_id`s of the trips within a direction (`getShapeIdsForStopSequenceBlock`).

### Entry-level polylines

Entry-level polylines are built independently from the distinct `shape_id`s of **all qualifying trips** on the route. They are **not** constructed by concatenating or reusing the group-level polylines.

## Merge Algorithm

Java merges shapes by iterating through each shape's points while maintaining a shared **undirected edge set**.

During the merge:

* Consecutive duplicate points are removed.
* If an edge (in either direction) has already been visited, the current line is flushed as a completed polyline and a new line is started.
* This de-overlaps shared track between shapes.

Each resulting line is encoded using `PolylineEncoder` (equivalent to `utils.EncodePolyline`) with:

* `length` equal to the merged line's point count.
* `levels` set to `null`.

## Current Impact

For routes with a single shape per direction, Maglev matches Java after Issue 1 (polyline encoding), producing byte-identical output.

For routes with multiple shapes or loop patterns, the implementations diverge. For example, the **Q** route produces:

* **Java:** 9 de-overlapped polylines.
* **Maglev:** 2 headsign-merged polylines.

This issue is independent of Issue 1. The encoded coordinates are correct, but the **number of generated polylines and their segmentation** do not match the Java implementation.

Issue #1122, opened 2026-07-01
stops-for-route: order entry polylines to match sorted groups


Ahmedhossamdev:
### Summary
`processTripGroups` appends group polylines to the entry-level `allPolylines` in
unsorted `dirGroups` (insertion) order, while stop groups are later sorted
alphabetically by destination name. Java orders entry polylines to match the sorted
groups. The *set* of polylines matches; only the order differs.

Issue #1119, opened 2026-07-01
stops-for-route: includePolylines=false leaves group polylines populated


Ahmedhossamdev:
### Summary
`processRouteStops` clears only the **entry-level** polylines (`allPolylines = []models.Polyline{}`). Each `stopGroup.Polylines` is still built
and returned. The spec requires both entry-level and group-level `polylines` to be
present but empty when `includePolylines=false`.

PR #1125, opened 2026-07-01
stops-for-route: match Java polyline output

### LINKED ISSUE #1118: stops-for-route: use Java floor polyline encoder, not rounding encoder
### Summary
`stops-for-route` encodes polylines with `github.com/twpayne/go-polyline`
(`polyline.EncodeCoords`), which **rounds** coordinates. The repo already ships a
Java-compatible encoder, `utils.EncodePolyline` in
[`internal/utils/polyline.go`](../maglev/internal/utils/polyline.go), which
**floors** (`floor(coord * 1e5)`) to match the Java reference `PolylineEncoder`
byte-for-byte. The handler never uses it, so every polyline drifts from Java by up
to one unit in the last coordinate digit.

Ahmedhossamdev:
Fixes: #1118 #1119 #1122 #1124

# stops-for-route: Match the Java reference implementation for polyline output

This PR brings the `stops-for-route` polyline output—and the trip set used to build it—in line with the Java OneBusAway reference implementation.

The implementation has been verified against a live Java server. All 20 Unitrans routes now return the same:

* Stop IDs
* Direction groups
* Polyline counts
* Set of coordinate points

Each logical change is contained in its own commit.

## 1. Use the Java floor polyline encoder (#1118)

`generatePolylines` previously encoded coordinates using `github.com/twpayne/go-polyline`, which **rounds** coordinates. The Java reference (`PolylineEncoder`) instead **floors** them (`floor(coord * 1e5)`), causing encoded polylines to differ by up to one unit in the final digit.

This PR switches to the in-repo `utils.EncodePolyline`, which matches the Java implementation byte-for-byte. For single-shape routes, the encoded `points` are now identical to Java.

## 2. Clear group polylines when `includePolylines=false` (#1119)

When `includePolylines=false`, only the entry-level polylines were cleared. Direction groups still contained their polylines.

The Java implementation (and the API spec) expects both entry-level and group-level polylines to be present but empty.

Group polylines are now omitted when the flag is disabled and serialized as non-nil empty slices (`[]`) instead of `null`.

## 3. Build polylines using Java's shape-set merge (#1124)

Previously:

* Group polylines were built by deduplicating trips by **trip headsign** and selecting one `shape_id` per headsign.
* Entry-level polylines were created by concatenating the group polylines.

Neither behavior matches the Java implementation and produces incorrect results for routes with multiple shapes or loops (for example, route **Q** produced 2 polylines instead of Java's 9).

This PR ports the logic from `ShapeBeanServiceImpl.getMergedPolylinesForShapeIds`.

The new implementation:

* Builds group polylines from the distinct `shape_id`s of trips within each direction.
* Builds entry-level polylines independently from the distinct `shape_id`s across all qualifying trips for the route (rather than concatenating the group polylines, addressing #1122).
* Merges shapes by walking each shape's ordered points while maintaining a shared **undirected edge set**.
* Removes consecutive duplicate points.
* Starts a new polyline whenever an edge has already been visited, effectively de-overlapping shared track.
* Encodes each merged line using `utils.EncodePolyline`.

Because Java iterates `HashSet`s in an unspecified order, exact split boundaries cannot be reproduced byte-for-byte. However, the resulting shape set, polyline count, and rendered coverage are identical.

## 4. Default to all service dates when `time` is omitted

While validating Java parity, another behavioral difference surfaced.

Maglev previously defaulted the service date to **today**, meaning only trips running on the current day contributed to the merged polylines. For routes such as **Q**, this excluded valid shapes and produced only 2 polylines instead of 9.

The Java implementation defaults `display.serviceDateFiltering` to `false`. When no `time` parameter is supplied, stops and shapes are built from trips across **all service dates**.

This PR now matches that behavior:

* No `time` parameter → include trips from all service dates.
* `time` provided → filter trips to the requested service date.

This completes parity for multi-shape routes such as **Q**.

## 5. Tests

* Updated the end-to-end golden outputs for the new merged polyline behavior.
* Added tests covering:

  * the default `includePolylines` behavior,
  * `includePolylines=false`, ensuring both entry-level and group-level polylines are returned as empty arrays.

## Verification

Compared Maglev against a live Java server for every Unitrans route (`A`–`Z`) without a `time` parameter.

For all 20 routes, the following now match the Java implementation:

* Stop IDs
* Direction groups
* Polyline counts
* Complete set of coordinate points

## Summary by CodeRabbit

* **Bug Fixes**
  * Route-stop results now apply the `time` filter only when `time` is provided, with validation and server errors handled accordingly.
  * Removed the prior “no matches” fallback behavior for `time`, ensuring trip selection stays consistent.
  * Improved and stabilized polyline generation and merging across route entries and direction groups.

* **New Features**
  * Enhanced `includePolylines` behavior: omitted/enabled returns polylines; `includePolylines=false` returns empty (non-null) polyline arrays.

* **Tests**
  * Updated polyline expectations and added coverage for default behavior and `includePolylines=false` across entry and direction stop groups.

### DIFF
--- a/internal/restapi/stops_for_route_handler.go
+++ b/internal/restapi/stops_for_route_handler.go
@@ -7,7 +7,6 @@ import (
 	"slices"
 	"time"
 
-	"github.com/twpayne/go-polyline"
 	"maglev.onebusaway.org/gtfsdb"
 	"maglev.onebusaway.org/internal/models"
 	"maglev.onebusaway.org/internal/nulls"
@@ -98,7 +97,6 @@ func (api *RestAPI) stopsForRouteHandler(w http.ResponseWriter, r *http.Request)
 
 func (api *RestAPI) processRouteStops(ctx context.Context, agencyID string, routeID string, serviceIDs []string, includePolylines bool) (models.RouteEntry, []models.Stop, error) {
 	allStops := make(map[string]bool)
-	allPolylines := make([]models.Polyline, 0, 100)
 	var stopGroupings []models.StopGrouping
 
 	// Get trips for route that are active on the service date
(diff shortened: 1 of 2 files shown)


burma-shave (approved):
<sub>Approved in [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/OneBusAway/maglev/pull/1125)</sub>

PR #1129, opened 2026-07-01
Return trip-less vehicles for agency

### LINKED ISSUE #1128: [spec-gap] vehicles-for-agency: trip-less vehicles dropped instead of included
### Description
Per maglev.wiki/vehicles-for-agency.md (Extension 5a), a vehicle with no matching trip must still be included in data.list, with tripId and tripStatus absent. Maglev drops these vehicles entirely: VehiclesForAgencyID (internal/gtfs/gtfs_manager.go) requires v.Trip != nil and resolves agency only via trip→route→agency, so trip-less vehicles are never returned.
### Expected behaviour
- Trip-less vehicles belonging to the agency are included in data.list.
- Their tripId and tripStatus are absent.
- Agency membership for trip-less vehicles resolved without a trip (via feed→agency mapping).
### Actual behaviour
- Trip-less vehicles are filtered out and never appear in the response.

3rabiii:
### Summary
This PR aligns the `vehicles-for-agency` endpoint with Spec Extension 5a by including vehicles that currently lack a trip assignment, rather than dropping them. For these vehicles, `tripId` and `tripStatus` are explicitly omitted from the JSON response.

### Changes 
- **Agency Resolution (`gtfs_manager.go`):** Updated the loop to resolve trip-less vehicles using the `feedAgencyFilter` (mapping the feed to the agency) instead of relying solely on the `trip->route->agency` chain.
- **Handler (`vehicles_for_agency_handler.go`):** Removed the fallback logic that generated fabricated `TripStatus` objects for trip-less vehicles.
- **Models (`vehicle.go`):** Added `omitempty` to `TripID` and `TripStatus` in the exclusive `VehicleStatus` struct so they drop cleanly when nil/empty.
- **Mocks (`gtfs_manager_mock.go`):** Upgraded `MockAddVehicleWithOptions` to support `FeedID` and `NoTrip` flags, and added `MockSetFeedAgencyFilter` to accurately simulate production feed-to-agency resolution in tests.

### Testing
- Rewrote `TestVehiclesForAgencyHandler_VehicleWithoutTrip` to assert that trip-less vehicles are successfully returned and that trip fields are entirely absent.
- Added a negative cover test (`TestVehiclesForAgencyHandler_VehicleWithoutTripOtherAgency`) to ensure trip-less vehicles on feeds serving other agencies are correctly filtered out.

Closes: #1128 

## Summary by CodeRabbit

* **New Features**
  * Trip-less vehicles can now appear in the “vehicles for agency” response when they match the configured feed-to-agency mapping (instead of being excluded).
* **Bug Fixes**
  * `tripId` and `tripStatus` are no longer returned as placeholder values for vehicles without trip data.
  * “Vehicles for agency” selection now respects per-feed agency filtering and route ID matching for more accurate results.
* **Documentation / Tests**
  * Updated JSON behavior and API/JSON tests to confirm `tripId`/`tripStatus` are omitted when unavailable.

### DIFF
--- a/internal/gtfs/gtfs_manager.go
+++ b/internal/gtfs/gtfs_manager.go
@@ -62,8 +62,7 @@ type Manager struct {
 	feedVehicles map[string][]gtfs.Vehicle
 	feedAlerts   map[string][]gtfs.Alert
 	// Per-feed agency filter: feedID -> set of allowed agency IDs.
-	// Populated once during InitGTFSManager before goroutines start; read-only thereafter.
-	// No lock is required for reads.
+	// Populated during InitGTFSManager; reads/writes are guarded by realTimeMutex.
 	feedAgencyFilter map[string]map[string]bool
 	// Per-feed, per-vehicle last-seen timestamps for stale vehicle expiry
 	feedVehicleLastSeen map[string]map[string]time.Time // feedID -> vehicleID -> lastSeen
--- a/internal/gtfs/gtfs_manager.go
+++ b/internal/gtfs/gtfs_manager.go
@@ -527,13 +527,23 @@ func (manager *Manager) VehiclesForAgencyID(ctx context.Context, agencyID string
 		routeIDs[route.ID] = true
 	}
 
-	// Step 2: Acquire real-time lock independently to read vehicles.
-	rtVehicles := manager.GetRealTimeVehicles()
+	manager.realTimeMutex.RLock()
+	defer manager.realTimeMutex.RUnlock()
 
 	var vehicles []gtfs.Vehicle
-	for _, v := range rtVehicles {
-		if v.Trip != nil && routeIDs[v.Trip.ID.RouteID] {
-			vehicles = append(vehicles, v)
+	for feedID, feedVehicles := range manager.feedVehicles {
+		for _, v := range feedVehicles {
+			if v.Trip != nil {
+				// Trip-bound vehicle: match via trip -> route -> agency.
+				if routeIDs[v.Trip.ID.RouteID] {
+					vehicles = append(vehicles, v)
+				}
+				continue
+			}
+			// Trip-less vehicle (spec Extension 5a): match via the feed's agency filter.
+			if filter := manager.feedAgencyFilter[feedID]; filter[agencyID] {
+				vehicles = append(vehicles, v)
+			}
 		}
 	}
 
--- a/internal/gtfs/gtfs_manager_mock.go
+++ b/internal/gtfs/gtfs_manager_mock.go
@@ -36,31 +36,7 @@ func (m *Manager) MockAddRoute(id, agencyID, name string) {
 	})
 }
 func (m *Manager) MockAddVehicle(vehicleID, tripID, routeID string) {
-	m.realTimeMutex.Lock()
-	defer m.realTimeMutex.Unlock()
-
-	for _, v := range m.realTimeVehicles {
-		if v.ID != nil && v.ID.ID == vehicleID {
-			return
-		}
-	}
-	now := time.Now()
-	m.realTimeVehicles = append(m.realTimeVehicles, gtfs.Vehicle{
-		ID:        &gtfs.VehicleID{ID: vehicleID},
-		Timestamp: &now,
-		Trip: &gtfs.Trip{
-			ID: gtfs.TripID{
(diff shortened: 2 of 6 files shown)


burma-shave:
Closing this one rather than continuing to iterate — a few issues with the current approach that go beyond a quick fix:

**Feed access doesn't belong in this code path.** `VehiclesForAgencyID` is a domain-level query — "give me the vehicles for agency X" — but this PR has it reaching into `manager.feedVehicles` and `manager.feedAgencyFilter` directly, and holding `realTimeMutex.RLock()` across the whole per-feed filtering loop to do it. "Feed" isn't a domain concept (it never appears in the OBA API surface or the GTFS spec — it's purely how this deployment is configured to poll upstream sources), and it shouldn't need to leak into a method answering a question about agencies. It's also a step back from the pattern the rest of `gtfs_manager.go` follows: lock briefly, copy out, release, then compute outside the lock (see `GetRealTimeVehicles`, `GetFeedUpdateTimes`).

**The fix doesn't actually work.** `filterVehiclesByAgency` (`internal/gtfs/realtime.go`, unmodified by this PR) runs in `refreshFeed` whenever a feed has `agency-ids` configured, and unconditionally drops any vehicle with `Trip == nil` before it's ever written to `feedVehicles`. That's exactly the condition under which `feedAgencyFilter[feedID]` would be non-empty and the new query-time lookup could otherwise succeed — so for a feed with `agency-ids` set, trip-less vehicles never survive to be found. For a feed *without* `agency-ids` set, `feedAgencyFilter[feedID]` is nil and the lookup fails anyway. There's no configuration where a real, ingested trip-less vehicle reaches the response. The PR's own tests pass only because `MockAddVehicleWithOptions` writes straight into `feedVehicles`, bypassing `refreshFeed` and this filter entirely — nothing in the test suite exercises the real ingestion path where the conflict actually lives.

**Multi-agency feeds make "attribute to one agency" underspecified.** A single feed can list multiple `agency-ids`. For a trip-bound vehicle that's fine — the route resolves it exactly. For a trip-less vehicle on a shared feed, there's no data-driven way to pick one agency over another; this PR doesn't address that case at all (`filter[agencyID]` returns true for a trip-less vehicle against *every* agency configured on its feed, which as currently structured means it'd show up under all of them — untested, undocumented, and not a deliberate decision anywhere in the diff).

**Low value for the primary use case.** The spec's primary actor for this endpoint is the rider, not dispatch — and a vehicle with no trip has no route, no stop, no arrival prediction. It's not answerable to "where's my bus." The operational-visibility argument for including it is real but secondary, and doesn't justify shipping a mechanism this tangled for a case that's edge-of-edge for the endpoint's main consumer.

### If we pick this up again

Resolve agency once, at ingestion, and stop re-deriving it per query:

- Add a derived index alongside the others `rebuildMergedRealtimeLocked()` already builds (`realTimeVehicleLookupByTrip`, etc.): `vehiclesByAgencyMap map[string][]gtfs.Vehicle`, keyed by resolved agency ID.
- Trip-bound vehicles: resolve via `route.AgencyID`, same lookup `filterVehiclesByAgency` already does today — currently that result gets thrown away after the keep/drop decision; keep it instead.
- Trip-less vehicles: resolve via the feed's `agency-ids`. If more than one agency is configured for that feed, append the vehicle to every one of them rather than arbitrarily picking one — worth an explicit Implementation Decisions note in the spec, since legacy doesn't actually solve this cleanly either (it silently assigns everything from a multi-agency source to `agencyIds.get(0)` — see `GtfsRealtimeSource.cacheVehicleLocations`, a real, currently-undocumented legacy defect, probably worth its own spec-gap issue independent of this one).
- `filterVehiclesByAgency` needs to change so it stops dropping trip-less vehicles outright — it should either skip them (leave attribution to the ingestion-time resolution step above) rather than filtering them out before that step runs.
- `VehiclesForAgencyID` then collapses to a lock-copy-return read of `vehiclesByAgencyMap[agencyID]` — no feed awareness, no per-request trip/route chasing, consistent with every other accessor in the file.
- Needs a test that goes through `refreshFeed` end-to-end with a trip-less `VehiclePosition` and a feed configured with `agency-ids`, not just direct injection into `feedVehicles` — that's the path this PR's tests didn't cover, and where the bug actually lives.

PR #1134, opened 2026-07-05
Set blockTripSequence to real value or -1

### LINKED ISSUE #1130: [spec-gap] vehicles-for-agency: blockTripSequence is hardcoded to 0
### Description:
Per the spec (maglev.wiki/vehicles-for-agency.md), tripStatus.blockTripSequence must represent the zero-based position of the active trip in the block, or -1 if unavailable.

Currently, in vehicles_for_agency_handler.go, blockTripSequence is hardcoded to 0 for all vehicles.

### Expected behavior:

* blockTripSequence accurately reflects the active trip's block position.

* Defaults to -1 when the sequence is unavailable.

### Actual behavior:

* blockTripSequence is always emitted as 0, inaccurately suggesting every vehicle is on its first block trip.

3rabiii:
### Summary
Fixes the hardcoded `blockTripSequence` in the `vehicles-for-agency` endpoint to accurately reflect the vehicle's position, aligning with the API specification.

### Changes
* **trips_helper.go:** Extracted logic into a new `blockTripSequence` function returning `(int, bool)` to clearly differentiate between a valid zero-index and missing data.
* **vehicles_for_agency_handler.go:** Replaced the hardcoded `0` with the actual sequence. Defaults to `-1` if unavailable.
* **Tests:** Added tests for both resolved (`>= 0`) and unavailable (`-1`) block sequences.

### Why this approach?
Returning a boolean flag from the core helper allows the handler to safely emit `-1` for missing data, while the legacy wrapper continues to return `0` to avoid breaking backward compatibility for other endpoints.

Closes: #1130

## Summary by CodeRabbit

* **Bug Fixes**
  * Vehicle trip status now resolves block trip sequence for each trip when possible; unavailable sequences are reported as `-1`.
  * Block sequence resolution now consistently uses the agency’s timezone/local calendar boundaries, including with a `time` query parameter.
  * Vehicles-for-agency requests now fail with a server error if the agency timezone can’t be loaded.
* **Tests**
  * Added REST API test coverage for resolvable vs unavailable sequences, requested-time behavior, and timezone boundary handling.
* **Chores**
  * Updated the debug web UI to generate proper escaped HTML output and render strings/JSON more reliably.

### DIFF
--- a/internal/restapi/trips_helper.go
+++ b/internal/restapi/trips_helper.go
@@ -558,31 +558,42 @@ func getDistanceAlongShapeInRange(lat, lon float64, shape []gtfs.ShapePoint, min
 
 // calculateBlockTripSequence calculates the index of a trip within its block's ordered trip sequence
 // for trips that are active on the given service date.
-// Uses GetBlockTripSequence with ROW_NUMBER() window function instead of fetching all trips and looping.
+// Returns 0 when the sequence is unavailable, for callers that treat 0 as "no data".
 func (api *RestAPI) calculateBlockTripSequence(ctx context.Context, tripID string, serviceDate time.Time) int {
+	seq, ok := api.blockTripSequence(ctx, tripID, serviceDate)
+	if !ok {
+		return 0
+	}
+	return seq
+}
+
+// blockTripSequence returns the zero-based index of a trip within its block's
+// ordered sequence for the given service date, and whether it was resolved.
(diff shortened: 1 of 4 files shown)


burma-shave:
@coderabbitai review

burma-shave:
@coderabbitai review

burma-shave (approved):

PR #1143, opened 2026-07-07
fix(api): implement silent bounds clamping for trips-for-location endpoint

### LINKED ISSUE #1139: Clamp search radius and span bounds instead of returning HTTP validation errors
**Description:**

In `internal/utils/validation.go`, `ValidateRadius` enforces a hard limit of 10,000 meters and returns an error if exceeded, triggering an HTTP 400 Bad Request.

Similarly, `ValidateSpan` returns an error if latitude or longitude spans exceed 5.0 degrees.

According to the OneBusAway specification, the maximum allowed search radius is 20,000 meters (20 km), and requests exceeding maximum radius or span bounds should be clamped to the maximum allowed area rather than rejected with HTTP validation errors.

**Expected Behavior:**

Increase the maximum radius limit to 20,000 meters. When supplied radius or span parameters exceed maximum thresholds, silently clamp the bounding box to the maximum allowed dimensions instead of returning HTTP 400 validation errors.

**Affected Components:**

- `internal/utils/validation.go`
- `internal/gtfs/location_params.go`
- `internal/utils/validation_test.go`

ARCoder181105:
### Summary
Increases the maximum search radius limit to 20,000 meters and replaces hard HTTP 400 validation failures with silent bounds clamping across location-based endpoints, bringing Maglev into full specification parity with OneBusAway.

### Problem
In `internal/utils/validation.go`, `ValidateRadius` previously enforced a strict limit of 10,000 meters, returning an error if exceeded that triggered an `HTTP 400 Bad Request`. Similarly, `ValidateSpan` rejected queries where latitude or longitude spans exceeded 5.0 degrees. 

According to the OneBusAway REST API specification, the maximum allowed search radius is 20,000 meters (20 km). Furthermore, requests exceeding the maximum radius or coordinate span bounds must be silently clamped to the maximum allowed area rather than rejected with HTTP validation errors.

### Changes
* **`internal/models/constants.go`:** Defined `MaxSearchRadiusInMeters` (20,000m) and `MaxSearchSpanInDegrees` (5.0°) constants to centralize search boundary rules.
* **`internal/utils/validation.go`:** Removed strict upper-bound error returns in `ValidateRadius` and `ValidateSpan`. Negative checks remain intact, but upper bounds now pass validation so they can be clamped downstream.
* **`internal/restapi/location_params.go`:** Updated `parseLocationParams` to silently clamp requested `radius`, `latSpan`, and `lonSpan` parameters to their maximum thresholds before constructing query models.
* **`internal/gtfs/location_params.go`:** Added defensive clamping inside `BoundsFromParams` to guarantee that bounding boxes calculated for spatial database queries never exceed specification boundaries.
* **Test Suites:** 
  - Updated `validation_test.go` and `input_validation_integration_test.go` to assert that large inputs (`radius=50000`, `span=10.0`) are allowed and return `HTTP 200 OK`.
  - Added dedicated unit tests (`location_params_test.go`) on GTFS location parameter helpers and on REST API parameter parsing.

### Why this approach?
Silently clamping search dimensions instead of returning validation errors improves client resilience by allowing mobile apps and web integrations with large default map viewports to receive bounded transit data without failing. Implementing clamping at both the HTTP parameter parsing layer and the GTFS spatial boundary layer ensures both spec compliance and safe database query limits.

### Verification
* **Unit & Integration Tests:** Ran `make test`

closes #1139 

## Summary by CodeRabbit

* **New Features**
  * Trips and stop lookups now automatically clamp oversized `radius` and `latSpan`/`lonSpan` to a safe maximum.
  * When both `radius` and span parameters are provided, `radius` takes precedence.
  * If `radius` is missing or non-positive, a default radius is used.
* **Bug Fixes**
  * Requests with overly large values are no longer rejected as “too large”; results are computed using clamped bounds.
* **Tests**
  * Added and updated unit/integration tests covering clamping, precedence, defaulting, and REST handler behavior.

### DIFF
--- a/internal/gtfs/location_params.go
+++ b/internal/gtfs/location_params.go
@@ -17,12 +17,23 @@ type LocationParams struct {
 // If LatSpan and LonSpan are both positive, they define the box; otherwise the
 // box is computed from Radius (defaulting to DefaultSearchRadiusInMeters).
 func BoundsFromParams(loc *LocationParams) utils.CoordinateBounds {
-	if loc.LatSpan > 0 && loc.LonSpan > 0 {
-		return utils.CalculateBoundsFromSpan(loc.Lat, loc.Lon, loc.LatSpan/2, loc.LonSpan/2)
+	latSpan := loc.LatSpan
+	lonSpan := loc.LonSpan
+	if latSpan > models.MaxSearchSpanInDegrees {
+		latSpan = models.MaxSearchSpanInDegrees
+	}
+	if lonSpan > models.MaxSearchSpanInDegrees {
+		lonSpan = models.MaxSearchSpanInDegrees
+	}
+	if latSpan > 0 && lonSpan > 0 {
+		return utils.CalculateBoundsFromSpan(loc.Lat, loc.Lon, latSpan/2, lonSpan/2)
 	}
 	radius := loc.Radius
 	if radius == 0 {
 		radius = models.DefaultSearchRadiusInMeters
+	}
+	if radius > models.MaxSearchRadiusInMeters {
+		radius = models.MaxSearchRadiusInMeters
 	}
 	return utils.CalculateBounds(loc.Lat, loc.Lon, radius)
 }
--- a/internal/gtfs/location_params.go
+++ b/internal/gtfs/location_params.go
@@ -17,14 +17,8 @@ type LocationParams struct {
 // If LatSpan and LonSpan are both positive, they define the box; otherwise the
 // box is computed from Radius (defaulting to DefaultSearchRadiusInMeters).
 func BoundsFromParams(loc *LocationParams) utils.CoordinateBounds {
-	latSpan := loc.LatSpan
-	lonSpan := loc.LonSpan
-	if latSpan > models.MaxSearchSpanInDegrees {
-		latSpan = models.MaxSearchSpanInDegrees
-	}
-	if lonSpan > models.MaxSearchSpanInDegrees {
-		lonSpan = models.MaxSearchSpanInDegrees
-	}
+	latSpan := utils.ClampSpan(loc.LatSpan)
+	lonSpan := utils.ClampSpan(loc.LonSpan)
 	if latSpan > 0 && lonSpan > 0 {
 		return utils.CalculateBoundsFromSpan(loc.Lat, loc.Lon, latSpan/2, lonSpan/2)
 	}
@@ -32,9 +26,7 @@ func BoundsFromParams(loc *LocationParams) utils.CoordinateBounds {
 	if radius == 0 {
 		radius = models.DefaultSearchRadiusInMeters
 	}
-	if radius > models.MaxSearchRadiusInMeters {
-		radius = models.MaxSearchRadiusInMeters
-	}
+	radius = utils.ClampRadius(radius)
 	return utils.CalculateBounds(loc.Lat, loc.Lon, radius)
 }
 
--- a/internal/gtfs/location_params.go
+++ b/internal/gtfs/location_params.go
@@ -14,20 +14,41 @@ type LocationParams struct {
 }
 
 // BoundsFromParams converts LocationParams into a CoordinateBounds bounding box.
-// If LatSpan and LonSpan are both positive, they define the box; otherwise the
-// box is computed from Radius (defaulting to DefaultSearchRadiusInMeters).
-func BoundsFromParams(loc *LocationParams) utils.CoordinateBounds {
-	latSpan := utils.ClampSpan(loc.LatSpan)
-	lonSpan := utils.ClampSpan(loc.LonSpan)
-	if latSpan > 0 && lonSpan > 0 {
-		return utils.CalculateBoundsFromSpan(loc.Lat, loc.Lon, latSpan/2, lonSpan/2)
+// If Radius is positive (or when neither Radius nor valid Spans are provided),
+// the box is computed from Radius (defaulting to DefaultSearchRadiusInMeters).
+// If both Radius and LatSpan/LonSpan are provided, Radius takes precedence.
+// If clamp is true, dimensions exceeding the maximum allowed search radius (20km)
+// are clamped to the maximum circle bounds.
+func BoundsFromParams(loc *LocationParams, clamp ...bool) utils.CoordinateBounds {
+	shouldClamp := len(clamp) > 0 && clamp[0]
+
+	// If Radius is specified (>0) OR neither Radius nor both Spans are provided (>0), use radius calculation.
+	// This ensures radius takes precedence when both radius and span are supplied per OBA spec.
+	if loc.Radius > 0 || !(loc.LatSpan > 0 && loc.LonSpan > 0) {
+		radius := loc.Radius
+		if radius <= 0 {
+			radius = models.DefaultSearchRadiusInMeters
+		}
+		if shouldClamp && radius > models.MaxSearchRadiusInMeters {
+			radius = models.MaxSearchRadiusInMeters
+		}
+		return utils.CalculateBounds(loc.Lat, loc.Lon, radius)
 	}
-	radius := loc.Radius
-	if radius == 0 {
-		radius = models.DefaultSearchRadiusInMeters
+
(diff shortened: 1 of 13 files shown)


internal/restapi/trips_for_location_handler_test.go
```diff
@@ -206,3 +206,29 @@ func TestTripsForLocationHandler_MissingParameters(t *testing.T) {
...
+		{"Radius exceeding max 20km (25km)", "/api/where/trips-for-location.json?key=TEST&lat=40.583321&lon=-122.426966&radius=25000"},
+		{"Spans exceeding max 5 degrees (6.0)", "/api/where/trips-for-location.json?key=TEST&lat=40.583321&lon=-122.426966&latSpan=6.0&lonSpan=6.0"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			resp, model := callAPIHandler[TripsForLocationResponse](t, api, tt.url)
+
+			assert.Equal(t, http.StatusOK, resp.StatusCode)
+			assert.Equal(t, http.StatusOK, model.Code)
```
Ahmedhossamdev, line 328: Nice cleanup, swapping the hard 400 for silent clamping matches how real OBA behaves and drops a spec-divergent error. Only left a couple of small optional notes on the test. Nothing blocking. Did you confirm the OpenAPI spec doesn't expect a 400 for oversized radius? Pretty sure it doesn't, just worth a glance.

Also note:
Small one, this proves the request doesn't error but not that clamping still returns a usable search. Mind adding an assertion on the body (some trip data, or the expected count) so we know the clamp didn't silently empty the result? Optional though.

internal/restapi/trips_for_location_handler_test.go
```diff
@@ -206,3 +206,29 @@ func TestTripsForLocationHandler_MissingParameters(t *testing.T) {
 		})
 	}
 }
+
+func TestTripsForLocationHandler_BoundsClamping(t *testing.T) {
+	api, cleanup := createTestApiWithRealTimeData(t, clock.RealClock{})
+	defer cleanup()
+
+	time.Sleep(500 * time.Millisecond)
```
Ahmedhossamdev, line 312: The time.Sleep(500ms) matches the other RT tests so I'm fine leaving it, just flagging it adds half a second per run. Feel free to skip.

ARCoder181105: Done...
Kept `time.Sleep(500 * time.Millisecond)` to maintain consistency

internal/restapi/trips_for_location_handler_test.go
ARCoder181105: I checked `testdata/openapi.yml` for `/api/where/trips-for-location.json` (lines 648–713), and it only defines a 200 successful response there is no expectation for a 400 on oversized search boundaries.

Also addressed your test suggestion! 

I added a non-empty assertion (`assert.NotEmpty(t, model.Data.List)`) inside `TestTripsForLocationHandler_BoundsClamping`

ARCoder181105:
@coderabbitai full review

internal/restapi/trips_for_location_handler_test.go
Ahmedhossamdev: Nice!

Ahmedhossamdev (approved):
I think it's ready to merge now

burma-shave:
@coderabbitai full review

ARCoder181105:
@coderabbitai full review

internal/restapi/location_params.go
```diff
@@ -31,6 +31,10 @@ func (api *RestAPI) parseLocationParams(r *http.Request, fieldErrors map[string]
 		return nil, fieldErrors
 	}
 
+	radius = utils.ClampRadius(radius)
```
burma-shave: I don't think we should be clamping the bounds in here. This funcs responsibility is to parse params. I don't think a caller would expect it to do any clamping. In any case, this func is used in other handlers where we don't want the search area to be clamped.

internal/gtfs/location_params.go
```diff
@@ -17,13 +17,16 @@ type LocationParams struct {
 // If LatSpan and LonSpan are both positive, they define the box; otherwise the
 // box is computed from Radius (defaulting to DefaultSearchRadiusInMeters).
 func BoundsFromParams(loc *LocationParams) utils.CoordinateBounds {
```
burma-shave: This is probably where it makes sense to introduce an optional clamping parameter. It needs to be a parameter because this func is shared and we only want to clamp in the case of trips-for-location. The call-sites in trips_for_location_handler.go would need to be updated to use this parameter.

internal/models/constants.go
```diff
@@ -13,6 +13,8 @@ const (
 const (
 	DefaultSearchRadiusInMeters = 600
 	QuerySearchRadiusInMeters   = 10000
+	MaxSearchRadiusInMeters     = 20000
+	MaxSearchSpanInDegrees      = 5.0
```
burma-shave: I don't think the specs refer to a span in degrees, only 20km radius. These two limits result in very different search areas.

internal/gtfs/location_params.go
```diff
@@ -17,13 +17,16 @@ type LocationParams struct {
...
-	if loc.LatSpan > 0 && loc.LonSpan > 0 {
-		return utils.CalculateBoundsFromSpan(loc.Lat, loc.Lon, loc.LatSpan/2, loc.LonSpan/2)
+	latSpan := utils.ClampSpan(loc.LatSpan)
+	lonSpan := utils.ClampSpan(loc.LonSpan)
+	if latSpan > 0 && lonSpan > 0 {
+		return utils.CalculateBoundsFromSpan(loc.Lat, loc.Lon, latSpan/2, lonSpan/2)
 	}
 	radius := loc.Radius
 	if radius == 0 {
```
burma-shave: It's not really is scope, but might as well fix it anyway. According to the spec, radius should take precedence over lat/lon.

burma-shave (changes requested):

internal/restapi/location_params.go
ARCoder181105: Removed global clamping (ClampRadius and ClampSpan) from parseLocationParams. Shared parameter parsing now strictly validates parameters without modifying or clamping them, ensuring other endpoints remain unclamped.

internal/gtfs/location_params.go
ARCoder181105: Centralized all bounds calculation and optional clamping directly inside `BoundsFromParams(loc *LocationParams, clamp ...bool)`. This eliminates duplicated/drifted logic across packages.

internal/models/constants.go
ARCoder181105: Removed MaxSearchSpanInDegrees = 5.0 and ClampSpan. Spans exceeding the maximum 20km circle (MaxSearchRadiusInMeters) are now dynamically clamped to the bounding dimensions (maxLatSpan / maxLonSpan) derived directly from that 20km circle, matching the OBA spec and SearchBoundsFactory.

internal/gtfs/location_params.go
```diff
@@ -14,17 +14,41 @@ type LocationParams struct {
...
+	if shouldClamp {
+		maxBounds := utils.CalculateBounds(loc.Lat, loc.Lon, models.MaxSearchRadiusInMeters)
+		maxLatSpan := maxBounds.MaxLat - maxBounds.MinLat
+		maxLonSpan := maxBounds.MaxLon - maxBounds.MinLon
+		if latSpan > maxLatSpan {
+			latSpan = maxLatSpan
+		}
+		if lonSpan > maxLonSpan {
+			lonSpan = maxLonSpan
+		}
```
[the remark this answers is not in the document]
ARCoder181105: A 5.0-degree span is approximately 550km, which severely violates the 20km maximum radius spec and will cause massive database performance regressions. My original code correctly calculated the exact span for 20km dynamically, so I will revert to that.

ARCoder181105:
@coderabbitai full review

burma-shave (approved):

PR #1149, opened 2026-07-09
fix(restapi): honor time query parameter in trips-for-location

### LINKED ISSUE #1137: Honor time query parameter for vehicle status calculations in trips-for-location endpoint
**Description:**

When a `time` query parameter is supplied to `/api/where/trips-for-location.json`, the handler parses the timestamp to determine the service date but discards it when evaluating vehicle positions and real-time status. 

Instead, the handler passes the server's live wall-clock time (`api.Clock.Now()`) to `BuildTripStatus`. 

Consequently, querying historical or future vehicle positions evaluates schedule deviations and extrapolations against live server time rather than the requested query time.

**Expected Behavior:**

When the `time` parameter is provided, the server should use the parsed target timestamp as the current evaluation time across all vehicle position calculations, schedule deviations, and status builders.

**Affected Components:**

- `internal/restapi/trips_for_location_handler.go`

ARCoder181105:
## Summary
Resolves a specification gap where the `/api/where/trips-for-location.json` endpoint parsed the optional `time` query parameter but ignored it during vehicle status and position calculations, defaulting to live server wall-clock time.

## Changes
- **Parameter Propagation**: Updated `parseAndValidateRequest` to return the target `currentTime` (and derived `serviceDate` / `todayMidnight`) when `time` is provided.
- **Handler Fix**: Removed the wall-clock override (`api.Clock.Now()`) in `tripsForLocationHandler` so `buildTripsForLocationEntries` and `BuildTripStatus` evaluate vehicle offsets and stale-data checks against the requested timestamp.
- **Unit Testing**: Added `TestTripsForLocationHandler_TimeParameter` to assert accurate `ServiceDate` calculations and verify `HTTP 400 Bad Request` responses for malformed time parameters.

## Verification
- Ran `make test`

closes #1137 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the trips-for-location endpoint’s handling of the `time` query so trip service dates correctly match the requested local day (midnight).
  * Invalid `time` values now return a consistent HTTP 400 response.
  * When status details are included, the status service-date fields also align with the computed local midnight.
* **Tests**
  * Added coverage to validate correct behavior for both valid and invalid `time` query values on the trips-for-location endpoint.

### DIFF
--- a/internal/restapi/trips_for_location_handler.go
+++ b/internal/restapi/trips_for_location_handler.go
@@ -137,7 +137,9 @@ func (api *RestAPI) parseAndValidateRequest(r *http.Request) (
 	fieldErrors map[string][]string,
 	serverErr error,
 ) {
-	loc, fieldErrors := api.parseLocationParams(r, nil)
+
+	var loc *internalgtfs.LocationParams
+	loc, fieldErrors = api.parseLocationParams(r, nil)
 
 	queryParams := r.URL.Query()
 
--- a/internal/restapi/trips_for_location_handler.go
+++ b/internal/restapi/trips_for_location_handler.go
@@ -22,7 +22,7 @@ import (
 func (api *RestAPI) tripsForLocationHandler(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 
-	locationParams, includeTrip, includeSchedule, currentLocation, todayMidnight, serviceDate, fieldErrors, err := api.parseAndValidateRequest(r)
+	locationParams, includeTrip, includeSchedule, currentLocation, currentTime, todayMidnight, serviceDate, fieldErrors, err := api.parseAndValidateRequest(r)
 	if len(fieldErrors) > 0 {
 		api.validationErrorResponse(w, r, fieldErrors)
 		return
@@ -35,8 +35,6 @@ func (api *RestAPI) tripsForLocationHandler(w http.ResponseWriter, r *http.Reque
 	// Intentionally defaulting includeStatus to false to align with includeSchedule
 	// behavior for this endpoint, even though trips-for-route defaults to true.
 	includeStatus := r.URL.Query().Get("includeStatus") == "true"
-	// Note: re-deriving currentTime here rather than returning it from parseAndValidateRequest(line: 150)
-	currentTime := api.Clock.Now().In(currentLocation)
 
 	stops := api.GtfsManager.GetStopsInBounds(ctx, locationParams, 100)
 	stopIDs := extractStopIDs(stops)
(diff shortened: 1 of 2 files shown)


ARCoder181105:
@coderabbitai full review

ARCoder181105:
@coderabbitai full review

Ahmedhossamdev (commented):
Nice, this makes time actually do something. I checked it against the Java TripsForLocationAction and the spec, and both say time should set the query moment and default to now, which is exactly what this does. Bonus: it quietly fixes a mismatch where todayMidnight used now while serviceDate used the param, so those two now agree.

Just small optional notes inline, nothing blocking. One thing that's not in this diff so I couldn't inline it: the active-trip set still comes from live vehicles (getActiveTrips at line 49), so a past/future time changes the timestamps but not which trips come back. Java uses time to pick the active-trip moment too. Pre-existing and out of scope here, just worth tracking so nobody assumes time fully back-dates results.

ARCoder181105:
> Nice, this makes time actually do something. I checked it against the Java TripsForLocationAction and the spec, and both say time should set the query moment and default to now, which is exactly what this does. Bonus: it quietly fixes a mismatch where todayMidnight used now while serviceDate used the param, so those two now agree.
> 
> Just small optional notes inline, nothing blocking. One thing that's not in this diff so I couldn't inline it: the active-trip set still comes from live vehicles (getActiveTrips at line 49), so a past/future time changes the timestamps but not which trips come back. Java uses time to pick the active-trip moment too. Pre-existing and out of scope here, just worth tracking so nobody assumes time fully back-dates results.

Great then I had opened a new issue to track this, will implement the fixes in the future when this PR get merged

thanks @Ahmedhossamdev

ARCoder181105:
the PR is failing becuz of openapi #1152 has the fix and it is ready to merge

burma-shave (approved):

internal/restapi/trips_for_location_handler.go
```diff
@@ -147,35 +146,39 @@ func (api *RestAPI) parseAndValidateRequest(r *http.Request) (
...
+			if fieldErrors == nil {
+				fieldErrors = make(map[string][]string)
+			}
+			for k, v := range timeFieldErrors {
+				fieldErrors[k] = append(fieldErrors[k], v...)
+			}
 		}
 	}
 
+	todayMidnight = time.Date(currentTime.Year(), currentTime.Month(), currentTime.Day(), 0, 0, 0, 0, currentLocation)
```
burma-shave: The is already a helper to calculate service date and midnight in api.go `ServiceDateMidnight`
https://github.com/OneBusAway/maglev/blob/0807405f3280dbdd0ef817994a7c88baf85e4739/internal/utils/api.go#L21

burma-shave (changes requested):

internal/restapi/trips_for_location_handler.go
ARCoder181105: Done

burma-shave (approved):

PR #1152, opened 2026-07-10
Make latitude and longitude optional for location searches

### LINKED ISSUE #1138: Default absent latitude and longitude parameters to (0, 0) in location searches
**Description:**

In `parseLocationParams`, missing `lat` and `lon` query parameters cause the server to return an HTTP 400 Bad Request validation error.

According to Section 3c of the OneBusAway `trips-for-location` specification, when latitude and longitude are absent, the server should treat both as `0.0`, construct a bounding box centered at `(0, 0)`, and apply normal processing.

If `(0, 0)` falls outside the loaded transit data service area, the server should return HTTP 200 OK with `outOfRange: true`.

**Expected Behavior:**

Make `lat` and `lon` optional in location parameter parsing. If omitted, default both coordinates to `0.0` and allow normal service area boundary checks to handle the request.

**Affected Components:**

- `internal/restapi/location_params.go`

ARCoder181105:
### Description

This PR aligns `maglev`'s location parameter parsing with the legacy Java application's behavior. Previously, omitting `lat` or `lon` resulted in a `400 Bad Request`. Now, they are treated as optional parameters that default to `0.0`. 

Because `(0, 0)` falls outside standard transit service areas, the existing bounds-checking logic securely catches the request and returns a standard `200 OK` response with an empty list and `outOfRange: true`.

### Changes Made
* **`internal/restapi/location_params.go`**: Swapped `utils.ParseRequiredFloatParam` to `utils.ParseFloatParam` for `lat` and `lon`.
* **Tests**: Updated `trips_for_location`, `stops_for_location`, and `routes_for_location` handler tests. The "Missing parameters" tests now accurately assert a `200 OK` and `outOfRange: true` instead of a `400` validation error.

### Testing
- [x] Passed all unit tests.
- [x] Verified locally via `curl` that endpoints return `outOfRange: true` when `lat`/`lon` are omitted.

*Note: A corresponding issue has been opened in `sdk-config` to update the OpenAPI specification.*

Closes #1138. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Location-based endpoints now return `200 OK` when latitude and/or longitude are missing.
  * Missing coordinates are handled as “out of range,” producing an empty results list for a consistent response shape across location routes.
* **Tests**
  * Updated route/stop/trip handler tests to validate `200 OK`, `model.Code == OK`, `model.Data.OutOfRange == true`, and an empty `model.Data.List` instead of expecting `400 Bad Request`.

### DIFF
--- a/internal/restapi/location_params.go
+++ b/internal/restapi/location_params.go
@@ -10,8 +10,8 @@ import (
 func (api *RestAPI) parseLocationParams(r *http.Request, fieldErrors map[string][]string) (*gtfs.LocationParams, map[string][]string) {
 	queryParams := r.URL.Query()
 
-	lat, fieldErrors := utils.ParseRequiredFloatParam(queryParams, "lat", fieldErrors)
-	lon, fieldErrors := utils.ParseRequiredFloatParam(queryParams, "lon", fieldErrors)
+	lat, fieldErrors := utils.ParseFloatParam(queryParams, "lat", fieldErrors)
+	lon, fieldErrors := utils.ParseFloatParam(queryParams, "lon", fieldErrors)
 	radius, fieldErrors := utils.ParseFloatParam(queryParams, "radius", fieldErrors)
 	latSpan, fieldErrors := utils.ParseFloatParam(queryParams, "latSpan", fieldErrors)
 	lonSpan, fieldErrors := utils.ParseFloatParam(queryParams, "lonSpan", fieldErrors)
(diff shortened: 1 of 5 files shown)


Ahmedhossamdev (approved):
Great work @ARCoder181105!
I have updated the openapi.yml file to match the new changes. 
https://github.com/OneBusAway/sdk-config/pull/19

ARCoder181105:
updated the openapi.yaml to match the latest config

PR #1170, opened 2026-07-14
fix: resolve FTS5 tokenization and multi-word search issues 

### LINKED ISSUE #1155: Refine FTS5 tokenization and multi-word query matching
### Description
**Spec Requirement:**
The search index must match prefixes of whitespace/hyphen/slash/parenthesis/ampersand-delimited tokens exact character by character without stemming. For multi-word queries like `pine st & 9th`, keys should match across token boundaries (`Pine St & 9th Ave`) (`BundleSearchServiceImpl.java:169-201`).

**Current Maglev Behavior:**
1. `stops_fts` (`gtfsdb/schema.sql`) uses `tokenize = 'porter'`, which applies Porter stemming (e.g., folding words like `stations` to `station` and `running` to `run`), violating exact prefix matching.
2. `sanitizeFTS5Query` removes special characters and formats multi-word queries into phrase queries (`"\"pine 9th*\""`). Double-quoted FTS5 phrase queries require tokens to be strictly adjacent, failing to match `Pine St & 9th Ave` because `St` and `&` sit between `Pine` and `9th`.

**Proposed Solution:**
- Evaluate removing or replacing `tokenize = 'porter'` on `stops_fts` with standard unicode/exact tokenization (`tokenize = 'unicode61'`).
- Update `sanitizeFTS5Query` / query construction to format multi-word queries using token prefix conjunctions (e.g., `pine* AND 9th*` or `NEAR`) so non-adjacent tokens match accurately.

ARCoder181105:
## Summary

This PR improves multi-word stop search behavior by fixing FTS5 query parsing, tokenizer configuration, and fallback handling.

## Changes

* **Schema Update**: Switched the `stops_fts` virtual table tokenizer from `porter` to `unicode61` to enable exact prefix matching and eliminate incorrect suffix stemming.
* **Query Construction**: Updated `search_stops_handler` to tokenize multi-word inputs and join them using the `AND` operator. This allows non-adjacent tokens to match, ensuring queries such as `Capitol & Broadway` return the expected results.
* **Error Handling**: Refined the FTS5 fallback logic to use `AND` intersections, ensuring that if the initial wildcard query fails, the fallback query remains accurate.
* **Test Coverage**: Added a new test case in `search_stops_handler_test.go` to verify that multi-word inputs containing special characters are parsed safely and handled correctly by the FTS engine.

## Testing

* Verified via `curl` that queries such as `Capitol & Broadway` now return the expected results instead of FTS syntax errors.
* Ran the existing test suite to ensure there were no regressions in search behavior.

**Closes #1155**

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stop search matching for multi-word queries by generating safer term-based full-text searches rather than relying on the full sanitized input.
  * Enhanced handling of whitespace-only and punctuation-only queries to reliably return an empty result set when no meaningful terms are present.
  * Updated the stop text index to use a more Unicode-aware tokenizer for more consistent matching.
* **Tests**
  * Added coverage for multi-word inputs containing special symbols.
  * Added coverage confirming punctuation-only queries return an empty stop list.


ARCoder181105:
@coderabbitai full review

Ahmedhossamdev:
Hey @ARCoder181105 
Great work on this!

I was testing it with unitrans data
When I type the full name of the stop it does not return anything. Can you check this?


Input: First St & D St / Downtown


Note: it works with other stops name


image

internal/restapi/search_stops_handler.go
```diff
@@ -82,7 +82,12 @@ func (api *RestAPI) searchStopsHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	searchQuery := `"` + sanitizedQuery + `*"`
+	terms := strings.Fields(sanitizedQuery)
+	var queryTerms []string
+	for _, term := range terms {
+		queryTerms = append(queryTerms, `"`+term+`"*`)
+	}
+	searchQuery := strings.Join(queryTerms, " AND ")
```
Ahmedhossamdev, line 86: Found an edge case here while testing against unitrans. If the input has a stray punctuation token that the sanitizer doesn't strip (like `/` or `-` with spaces around it), the `AND` join breaks the whole query. Example: `First St & D St / Downtown` returns 0 results on maglev but the Java server returns the stop. The `&` gets stripped fine, but the `/` survives as its own token, so we build `... AND "/"* AND ...`, FTS5 throws a syntax error on the empty `"/"*` phrase, and the fallback then does `... AND "/" AND ...` which intersects to nothing.
>
> Simplest fix I think is to drop tokens that have no actual word characters right after splitting, so stray punctuation just gets ignored instead of forcing an empty match:
> ```suggestion
> 	terms := strings.Fields(sanitizedQuery)
> 	queryTerms := make([]string, 0, len(terms))
> 	for _, term := range terms {
> 		if !strings.ContainsFunc(term, func(r rune) bool {
> 			return unicode.IsLetter(r) || unicode.IsDigit(r)
> 		}) {
> 			continue // skip tokens FTS5 tokenizes to nothing (e.g. "/", "-")
> 		}
> 		queryTerms = append(queryTerms, `"`+term+`"*`)
> 	}
> ```

gtfsdb/schema.sql
```diff
@@ -219,7 +219,14 @@ END;
 -- FTS5 external content table for full-text stop search.
 -- Data lives in 'stops' table; only the search index is stored here.
 -- migrate
-CREATE VIRTUAL TABLE IF NOT EXISTS stops_fts USING fts5 (id UNINDEXED, stop_name, tokenize = 'porter');
+DROP TABLE IF EXISTS stops_fts;
```
Ahmedhossamdev: Small thing, not a blocker. The migration runner re-runs schema.sql on every startup and the DB is a persistent file (./gtfs.db), so this DROP TABLE + repopulate rebuilds the whole FTS index every boot, not just once. It's actually needed to get the tokenizer swap onto an existing DB, so I'm fine keeping it. Could we add a one-line comment saying it's intentional so nobody "optimizes" the drop away later?

Ahmedhossamdev (changes requested):

internal/restapi/search_stops_handler.go
ARCoder181105: Hey @Ahmedhossamdev, thanks for testing and pointing this out!

I looked into `First St & D St / Downtown` with the Unitrans dataset: the root cause was that SQLite's `unicode61` tokenizer treats `/` as a delimiter and evaluates `"/"*` as an empty phrase, causing the FTS5 query `MATCH` to fail and return `0` results.

I've updated the handler with an `extractFTS5Terms` helper function that filters out tokens without letters or digits (e.g. `/` and `-`) before constructing the FTS5 query (`"First"* AND "St"* AND "D"* AND "St"* AND "Downtown"*`). I tested this locally against the Unitrans feed via `curl`, and `"First St & C St / Downtown"` now returns cleanly without errors.

gtfsdb/schema.sql
ARCoder181105: Done

ARCoder181105:
@coderabbitai full review

internal/restapi/search_stops_handler.go
Ahmedhossamdev: Great work!

Ahmedhossamdev (approved):
LGTM, could you please resolve the merge conflicts?

ARCoder181105:
Done resolved the merged conflicts...

internal/restapi/search_stops_handler.go
```diff
@@ -56,37 +72,23 @@ func (api *RestAPI) searchStopsHandler(w http.ResponseWriter, r *http.Request) {
...
+	queryTerms := make([]string, len(terms))
+	for i, term := range terms {
+		queryTerms[i] = `"` + term + `"*`
+	}
+	searchQuery := strings.Join(queryTerms, " AND ")
 
 	searchParams := gtfsdb.SearchStopsByNameParams{
 		SearchQuery: searchQuery,
-		Limit:       int64(limit + 1), // Request limit + 1 to accurately determine if pagination boundaries are exceeded.
+		Limit:       int64(limit) + 1, // Request limit + 1 to accurately determine if pagination boundaries are exceeded.
```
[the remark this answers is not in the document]
ARCoder181105: While theoretically `maxCount` could be parsed up to `math.MaxInt` causing `int64(limit) + 1` to overflow on 64-bit systems, in our domain context (transit feeds containing at most tens/hundreds of thousands of stops), `maxCount` values anywhere near `math.MaxInt64` never occur in legitimate traffic. We have simplified the calculation to `int64(limit + 1)` and will rely on standard pagination bounds/limits (`Limit <= 1000`) across handlers rather than adding arbitrary `math.MaxInt` clamping here. It can be fixed in future if needed for now it is out of conext

gtfsdb/schema.sql
burma-shave: I don't think we need worry about migrating users given we're pre version 1.0. It doesn't make sense to maintain compatibility at this stage in development. Especially for a file that is just a persistent cache. If it causes an issue for anyone, we can tell them to delete the file and restart. Removing the drop saves the cost of a fts rebuild on every start.

burma-shave (changes requested):

burma-shave (approved):
Looks great, thanks! But watch those commit messages. Instead of "removed drop table command of stop_fts" it should be "*Remove* drop table command of stop_fts" and then a body explaining why the change was made. Have a look at CONTRIBUTING.md.

PR #1172, opened 2026-07-14
Add oba-api-review skill for API change review

burma-shave:
## Summary

- Add four Claude Code skills (`oba-api-review` plus `oba-api-verify`, `oba-api-client-impact`, `oba-api-spec-check`) that review a maglev API change against the `maglev.wiki` behavioural specs and assess impact on the three OBA clients (Wayfinder + JS SDK, iOS, Android).
- Add `.claude/skills/lib/resolve-oba-repo.sh`, a shared script that resolves each dependent repo (`wayfinder`, `js-sdk`, `onebusaway-ios`, `onebusaway-android`, `maglev.wiki`) on demand — using a nearby checkout or `OBA_WORKSPACE` override if present, otherwise cloning into a local cache. Checkouts it didn't create itself are never mutated; a read-only freshness check warns if one looks stale instead of silently trusting it.
- Narrow `.gitignore`'s `.claude` entry down to just `settings.local.json`, so this config and any future project skills can be tracked.

This makes the skill set usable with only `maglev` checked out, cloning whatever else it needs the first time it's used.

## Test plan

- [x] Verified the resolver script's three resolution paths (`OBA_WORKSPACE` override, sibling checkout, cache-dir clone) in isolation, confirming stdout only ever contains the resolved path.
- [x] Ran a cold-start test in an isolated checkout with no sibling repos and an empty cache: all 5 repos resolved and cloned correctly, and every documented file path (JS SDK, Wayfinder route, iOS service/models, Android contract/models, maglev handler) was checked against the real repos and corrected where stale.
- [x] Verified the staleness-warning path against known out-of-date checkouts: warnings fire with the correct behind-count, stdout stays clean, and the checkouts' working trees are provably untouched (`git status`/HEAD unchanged before and after).

## Summary by CodeRabbit

- **New Features**
  - Added guided workflows for reviewing OBA API changes: verify goals, check spec consistency, and assess client impact across supported platforms.
  - Added automated repository resolution to map requested OBA repositories to local checkout paths for the workflows.

- **Documentation**
  - Added new skill guides with argument rules, diff-source handling (PR/branch/working tree), per-client checklists, and standardized report templates.

- **Chores**
  - Updated ignore rules to exclude only the local settings file rather than the entire configuration directory.

### DIFF
--- a/.claude/skills/lib/resolve-oba-repo.sh
+++ b/.claude/skills/lib/resolve-oba-repo.sh
@@ -78,8 +78,10 @@ warn_if_stale() {
   remote_rev="$(git -C "$path" rev-parse "$upstream_ref")"
   if [[ "$local_rev" != "$remote_rev" ]]; then
     behind="$(git -C "$path" rev-list --count "HEAD..$upstream_ref" 2>/dev/null || echo "?")"
-    last_date="$(git -C "$path" log -1 --format=%ad --date=short)"
-    echo "Warning: $REPO at $path (branch $branch) is $behind commit(s) behind origin/$branch (last local commit: $last_date). This checkout was found independently, not cloned by this script, so it was not updated automatically - results below may reflect stale code. Run 'git -C $path pull' to update it, or unset OBA_WORKSPACE / move it aside to let this script manage a fresh copy in its cache instead." >&2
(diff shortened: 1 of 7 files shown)


burma-shave:
@coderabbitai review

PR #1184, opened 2026-07-17
test(restapi): add edge case unit tests and specification coverage for trips-for-location 

### LINKED ISSUE #1142: Add comprehensive unit test coverage for trips-for-location specification scenarios
**Description:**

In `internal/restapi/trips_for_location_handler_test.go`, there is currently zero test coverage for several critical specification requirements and edge cases:

1. Toggling `includeTrip` (verifying default `true` when absent, and omitting when `false`).
2. Toggling `includeReferences` (verifying references block is omitted or empty when `false`).
3. Overriding evaluation time via the `time` query parameter (verifying historical/future vehicle status calculations use query time rather than `api.Clock.Now()`).
4. Omitting `lat` and `lon` parameters (verifying fallback to `(0, 0)` and `outOfRange: true`).
5. Bounding box clamping for large `radius` (>20km) and `latSpan`/`lonSpan` (>5°).

**Note** : More test edge cases will also be added if found during the covering up the gaps

**Expected Behavior:**

Add unit tests covering these 5 core specification scenarios to ensure regression prevention and full OneBusAway specification compliance.

**Affected Components:**

- `internal/models/trips_for_location_test.go`

ARCoder181105:
## Summary
Enhances unit test coverage for the `/api/where/trips-for-location.json` endpoint to ensure complete alignment with the official OneBusAway specification and protect against regressions. 

Additionally, this PR includes a handler improvement that splits the agencies database error check, ensuring that genuine database failures are surfaced correctly (HTTP 500) rather than being masked as `"no agencies configured"`.

## Key Changes & Test Coverage
- **Handler Fix (Agencies Error)**: Split the `api.GtfsManager.GetAgencies` error handling to explicitly return DB failures, preventing real outages from being masked as a missing agencies configuration.
- **Spatial Bounds & Precedence**: Validates OBA specification rules ensuring `radius` takes precedence when both `radius` and `latSpan`/`lonSpan` are provided (strictly asserted by verifying a 1m radius filters out trips caught by a large span). Confirms the default `10km` radius fallback when spatial spans are omitted.
- **Parameter Validation (`400 Bad Request`)**: Verifies strict parameter validation and error responses (`fieldErrors`) for out-of-range coordinates (`lat=95`, `lon=-185`), non-numeric inputs, negative dimensions (`radius`, `latSpan`, `lonSpan`), and malformed `time` strings.
- **Time Propagation (`ServiceDate`)**: Verifies that custom `time` query parameters (both `YYYY-MM-DD` date strings and epoch millisecond timestamps) correctly propagate to `ServiceDate` across all returned trips and vehicle statuses.
- **Static-Only & Reference Filtering**: Confirms proper handling when no real-time vehicle positions exist (`list=[]`), and verifies `includeTrip=false` omits `References.Trips` while maintaining `Stops`, `Routes`, and `Agencies`.
- **Schedule Distances & Reference IDs**: Asserts that `DistanceAlongTrip` values are non-negative and monotonically non-decreasing along the stop sequence, and enforces `{agencyID}_{rawID}` formatting for adjacent trip IDs (`NextTripId`/`PreviousTripId`).
- **Context Cancellation**: Directly verifies `clientCanceledResponse` behavior during client disconnects (`context.Canceled`) and gateway timeouts (`context.DeadlineExceeded` -> `504 Gateway Timeout`).

closes #1142

## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for trips-by-location requests, including radius and time handling.
  * Added validation for missing or invalid parameters, zero coordinates, and static-only data.
  * Verified reference data, schedule details, stop-time distances, and trip identifiers.
  * Added coverage for request cancellation, deadlines, and expected timeout responses.


ARCoder181105:
@coderabbitai full review

Ahmedhossamdev (commented):
Really nice breadth here, the monotonic DistanceAlongTrip check and the context cancellation cases are the kind of invariants that actually catch regressions. I traced the cancellation ones to make sure they weren't passing for free, and they're genuinely hitting clientCanceledResponse since the ctx.Err() check sits at the top level of the handler rather than inside the vehicle loop. Good.

One thing I'd like to see before merge, in RadiusAndPrecedence. The precedence subtest can't currently fail. It only asserts 200 plus a non-nil list, so a handler that ignored radius and used the 2 degree span would pass it just the same. I checked and the rule itself is right, Java does if (_radius > 0) ... else if (_latSpan > 0 && _lonSpan > 0) in SearchBoundsFactory, and we match that, so this is purely about the assertion not pinning it down. Details in the inline comment.

Also worth adding a line to the description about the handler change. It splits the agencies error so a real DB failure isn't reported as "no agencies configured", which is a good fix, it's just easy to miss in something labelled as test coverage.

ARCoder181105:
Thanks @Ahmedhossamdev  for the thorough review 

 I've pushed an update to address both points:

- **Radius Precedence Assertion**: I tightened the `Radius Precedence over Spans` subtest. It now uses a tiny 1-meter radius combined with a large 2-degree span and asserts that an empty list is returned. This strictly proves the radius is taking precedence and preventing the span from catching any trips. 
- **PR Description Updated**: I've updated the PR description to explicitly highlight the handler change that splits the agencies error to ensure real DB failures aren't masked.

Let me know if anything else is needed before merge!

Ahmedhossamdev (approved):
LGTM!  @ARCoder181105

burma-shave (approved):

PR #1196, opened 2026-07-18
Return HTTP 404 when no stops match prefix query in search-stop

### LINKED ISSUE #1156: Enforce HTTP 404 response when no stops match prefix query (Pre-filter check)
### Description
**Spec Requirement:**
If no stops match the prefix query in the database index (before any route-type or revenue service filtering), the server MUST return HTTP 404 with `code: 404` and `text: "resource not found"` in the standard response envelope (*Minimal Guarantees* & *Extensions 2a*). (Note: if stops match pre-filter but are subsequently excluded by route-type filtering, HTTP 200 with `list: []` is returned per *Extensions 2b*).

**Current Maglev Behavior:**
`searchStopsHandler` does not check if pre-filter `len(stops) == 0`. If `SearchStopsByName` returns 0 rows (or if `sanitizedQuery == ""`), Maglev returns HTTP 200 OK with an empty `list: []`. `TestSearchStopsHandlerNoResults` currently tests and enforces this non-conformant 200 OK behavior.

**Proposed Solution:**
- After executing `SearchStopsByName`, check `if len(stops) == 0`.
- If zero matches exist pre-filter, invoke a 404 standard envelope error response (`api.notFoundErrorResponse(w, r)` with `text: "resource not found"`).
- Update `TestSearchStopsHandlerNoResults` to assert HTTP 404 and `code: 404`.

ARCoder181105:
## Problem
`searchStopsHandler` returns HTTP 200 with an empty list when no stops match
the prefix query. The OBA spec requires HTTP 404 with `code: 404` and
`text: "resource not found"` (Minimal Guarantees & Extensions 2a).

## Changes
- Return 404 via `sendNotFound` when the sanitized query is empty
- Return 404 via `sendNotFound` when `SearchStopsByName` returns zero rows (pre-filter)
- Update affected tests to assert HTTP 404 and the standard error envelope

Closes #1156

## Summary by CodeRabbit

* **Bug Fixes**
  * Stop searches with no matching results now return **404 Not Found** instead of a successful response with an empty list.
  * Invalid, blank, punctuation-only, and sanitized search queries now consistently return a not-found response.

### DIFF
--- a/internal/restapi/search_stops_handler.go
+++ b/internal/restapi/search_stops_handler.go
@@ -74,8 +74,7 @@ func (api *RestAPI) searchStopsHandler(w http.ResponseWriter, r *http.Request) {
 	terms := extractFTS5Terms(sanitizedQuery)
 
 	if len(terms) == 0 {
-		response := models.NewListResponseWithRange([]models.Stop{}, *models.NewEmptyReferences(), false, api.Clock, false)
-		api.sendResponse(w, r, response)
+		api.sendNotFound(w, r)
 		return
 	}
 
@@ -129,6 +128,13 @@ func (api *RestAPI) searchStopsHandler(w http.ResponseWriter, r *http.Request) {
 			)
 			return
 		}
+	}
+
+	// Pre-filter check: if no stops matched the prefix query at all, return 404
+	// per OBA spec (Minimal Guarantees & Extensions 2a).
+	if len(stops) == 0 {
+		api.sendNotFound(w, r)
+		return
 	}
 
 	stops, isLimitExceeded := utils.PaginateSlice(stops, 0, limit)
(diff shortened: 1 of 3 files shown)


ARCoder181105:
Switched assert.NoError to require.NoError in the `trips_for_location` tests. This prevents a nil pointer panic if a transient DB timeout occurs during CI test runs, allowing the test to fail gracefully instead of crashing.

ARCoder181105:
@coderabbitai full review

Ahmedhossamdev (approved):
LGTM

burma-shave:
Thanks for the PR, but I've updated the specs to document that maglev intentionally deviates from the legacy behaviour and instead returns 200 and an empty list for no search results. This is the current behaviour and what would normally be expected from an endpoint like this.

PR #1203, opened 2026-07-19
Refactor route search parameters and enforce specification parity

### LINKED ISSUE #1197: search-route: return 404 when no routes match
### Description

**Current behavior:**
When `input` matches no routes, the handler returns HTTP 200 with an empty `data.list`.

> If the input matches no routes, the server returns HTTP 404 with the standard error envelope (`code: 404`, `text: "resource not found"`).
>
> — [wiki/search-route § Extension 3a](https://github.com/OneBusAway/maglev/wiki/search-route)

**What to change:**
After calling `SearchRoutes()` in `route_search_handler.go`, check if the result slice is empty and call `api.sendNotFound(w, r)` — same pattern already used by the sibling `search-stop` handler in `search_stops_handler.go`.

### LINKED ISSUE #1199: search-route: fix limitExceeded always being false
### Description

**Current behavior:**
`limitExceeded` is hardcoded to `false` in the `NewListResponseWithRange()` call. The handler requests exactly `maxCount` rows from the database, so it has no way to determine whether additional results exist.

> If the index contains more than `maxCount` routes for that key, the list is truncated to `maxCount` and `limitExceeded` is set to `true`; otherwise `limitExceeded` is `false`.
>
> — [wiki/search-route § Step 4](https://github.com/OneBusAway/maglev/wiki/search-route)

**What to change:**
1. In `route_search_handler.go`, pass `maxCount + 1` to `SearchRoutes()` (overfetch by one).
2. Use `utils.PaginateSlice()` to truncate back to `maxCount` and derive the `isLimitExceeded` boolean.
3. Pass the boolean to `NewListResponseWithRange()`.

This is the same pattern used by `search_stops_handler.go`.

### LINKED ISSUE #1201: search-route: use standardized parameter parsing (ParseRequiredStringParam / ParseMaxCount)
### Description

**Current behavior:**
The handler uses ad-hoc parameter parsing that differs from other search endpoints:
- `input` is validated via `utils.ValidateAndSanitizeQuery()` + a manual whitespace check.
- `maxCount` is parsed via `utils.ParseFloatParam()`, which accepts float values (e.g. `"3.7"` silently truncates to `3`) and caps at a hardcoded `100`.

**Expected behavior:**
Use the standardized utilities (`utils.ParseRequiredStringParam` + `utils.ParseMaxCount`) already used by the sibling `search-stop` handler. This also aligns the `maxCount` upper bound with the project-wide `MaxAllowedCount` of 250 instead of the ad-hoc cap of 100.

**What to change:**
Refactor the parameter parsing block at the top of `route_search_handler.go` to use `ParseRequiredStringParam()` and `ParseMaxCount()`, matching the pattern in `search_stops_handler.go`.

ARCoder181105:
### Description
This PR standardizes the parameter parsing in the `search-route` endpoint to align with sibling endpoints and ensures full compliance with the OneBusAway specifications regarding pagination and error handling.

**Key Changes:**
- **Standardized Parsing ( #1201 ):** Swapped out ad-hoc validation for `utils.ParseRequiredStringParam` and `utils.ParseMaxCount`. Aligned the `maxCount` upper boundary with the project-wide limit of 250 (up from the hardcoded 100).
- **Spec Compliance (404s) (#1197):** Enforced HTTP 404 (Not Found) responses with the standard OBA error envelope when a query yields no routes (including whitespace-only inputs).
- **Accurate Pagination (#1199):** Updated the database query to fetch `maxCount + 1` routes internally to accurately determine if the results were truncated. The `limitExceeded` flag is now properly set dynamically via `utils.PaginateSlice` instead of being hardcoded to `false`.
- **Test Coverage:** Updated unit tests to assert against the new 250 limit boundaries and the corrected 404 HTTP statuses. Added `TestRouteSearchHandlerLimitExceeded` to explicitly verify that the truncation flag correctly fires when results are artificially constrained.

closes #1201
closes #1197 
closes #1199

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved route search request validation and error reporting.
  * Empty or whitespace-only searches now return a clear **“resource not found”** (404) response.
  * Route search results now correctly flag when additional matches exist beyond the requested limit.
  * Updated maximum result-count validation to allow up to **250** results (and reject higher values).  
* **Tests**
  * Updated and added coverage for no-results, whitespace input, maxCount boundaries, and limit-exceeded behavior.


ARCoder181105:
@coderabbitai full review

Ahmedhossamdev (commented):
Nice cleanup, this reads a lot better than the ad-hoc parsing it replaces. I checked all three behavior changes against the Java implementation and they line up exactly: the 404 on empty matches RouteAction.index() (which does if (routeSuggestions == null || ...isEmpty()) return setResourceNotFoundResponse()), the default of 20 matches ApiSearchAction, and the maxCount + 1 plus PaginateSlice trick is the right equivalent of Java's if (size() > maxCount) in BundleSearchServiceImpl.getRouteSuggestions(). I also diffed it live against a running Java server and the 404 envelopes are identical. Full gate passes locally (lint, fmt, tests, check-openapi). No blockers from me.

One thing that isn't on a changed line so I'll put it here: dropping ValidateAndSanitizeQuery here also drops the 200 character cap on input for this endpoint. Java has no cap at all so the behavior is fine, but routes-for-location and stops-for-location still run their query param through that helper, so the three endpoints now disagree on whether long input is allowed. Might be worth deciding one way or the other, no need to block this PR on it.

ARCoder181105:
Thanks @Ahmedhossamdev  for the thorough review and for testing

Great catch on the 200-character query cap. Since Java doesn't enforce a limit, I agree we should drop it everywhere for consistency. 

I just pushed a follow-up commit that replaces `ValidateAndSanitizeQuery` with `SanitizeInput` in both the `routes-for-location` and `stops-for-location` handlers. All three endpoints now safely strip HTML tags without throwing 400 errors for long queries. I also removed the outdated integration tests for those 400 errors, so everything is green and aligned!

Ahmedhossamdev (commented):
LGTM!

internal/restapi/route_search_handler.go
```diff
@@ -58,12 +27,19 @@ func (api *RestAPI) routeSearchHandler(w http.ResponseWriter, r *http.Request) {
...
 
-	routes, err := api.GtfsManager.SearchRoutes(ctx, sanitizedInput, maxCount)
+	routes, err := api.GtfsManager.SearchRoutes(ctx, query, maxCount+1)
 	if err != nil {
 		api.serverErrorResponse(w, r, err)
 		return
 	}
 
+	if len(routes) == 0 {
+		api.sendNotFound(w, r)
```
burma-shave: This is odd behaviour for a search endpoint. I've updated the specs to document that maglev deliberately deviates from this and instead of returning 404 for no results, returns 200 and an empty list which is what you would expect from this sort of endpoint.

https://github.com/OneBusAway/maglev/wiki/search-route#implementation-decisions

burma-shave (changes requested):

burma-shave (approved):
Rebased on main and updated route_search_handler to return HTTP 200 with an empty list instead of 404 when a search yields no matches, mirroring search/stop.json and the wiki decision at https://github.com/OneBusAway/maglev/wiki/search-route#implementation-decisions. Also resolved the merge conflict with main (which had independently picked up `includeReferences` gating for this endpoint).

Everything else in this PR — standardized parsing, 250 maxCount cap, accurate pagination/limitExceeded, and the SanitizeInput alignment across the three search endpoints — looks good. Thanks for the work here!

PR #1217, opened 2026-07-21
Add includeTrip parameter to trips-for-route

### LINKED ISSUE #1208: `includeTrip` parameter is not implemented
### Current behavior
The handler never reads an `includeTrip` query parameter. Instead it passes
`includeSchedule` into `buildTripReferences` as the "include trip beans" flag
(`internal/restapi/trips_for_route_handler.go`), conflating two independent
parameters. Consequences:
- `includeTrip=false` has no effect at all.
- `includeSchedule=false` wrongly drops trip beans from `data.references.trips`.

### Spec requirement
- `includeTrip` (default `true`) controls whether trip
  beans are populated in `data.references.trips`.
- Extension 5a: with `includeTrip=false`, trip beans are not
  added to `data.references.trips`; the `tripId` field in each list entry still
  identifies the trip.

3rabiii:
### Summary
Implements the `includeTrip` query parameter for the `trips-for-route` endpoint. Previously, the handler conflated `includeTrip` with `includeSchedule`, passing the wrong flag to `buildTripReferences`. This caused `includeTrip=false` to be silently ignored and `includeSchedule=false` to incorrectly drop trip beans from `data.references.trips`.

### Changes
- **`internal/restapi/trips_for_route_handler.go`:**
  - Added `includeTrip` parsing alongside the existing `includeSchedule` and `includeStatus` parameters (defaults to `true`).
  - Updated the `buildTripReferences` calls in both the early empty-list return path and the main response path to use `includeTrip` instead of `includeSchedule`.
- **`internal/restapi/trips_for_route_handler_test.go`:**
  - Added a new table-driven test `TestTripsForRouteHandler_TripInclusion` with four cases: default inclusion, explicit inclusion, explicit exclusion, and a regression guard (excluding schedule but keeping trips).
 
Closes: #1208 

## Summary by CodeRabbit

* **New Features**
  * Added an `includeTrip` option for controlling trip references in trips-for-route responses.
  * Trip references are included by default and support explicit true/false values.

* **Bug Fixes**
  * Standardized trip-reference handling across response formats.
  * Preserved trip identifiers on returned trip entries regardless of reference inclusion settings.
  * Improved handling of uppercase and numeric option values.


aaronbrethorst:
@3rabiii this has merge conflicts

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Approved on the merits — this is a good catch and a clean fix.

The bug is a genuinely sneaky one: `buildTripReferences`'s third parameter is already named `includeTrip` and gates `references.trips`, but the handler was passing `includeSchedule` into it. So `includeSchedule=false` was silently also emptying the trip references, and `includeTrip` did nothing at all. Two conflated flags in one argument position is exactly the kind of thing that survives review by looking correct at the call site.

I checked the behavior against the spec and it lines up:

- `includeTrip` defaults to `true` when the parameter is omitted, which is what the spec requires.
- With `includeTrip=false`, `references.trips` is empty but `data.list[].tripId` is still populated — that's spec extension 5a verbatim ("the trip ID is still present in the list entry but no detail is available in references"). Your test asserts both halves, which is the part that matters.
- Reusing the existing `parseIncludeTrip` helper instead of hand-rolling a string comparison was the right call.

The table-driven test covering omitted / `true` / `false` / `FALSE` / `0` plus an `includeSchedule=false` regression guard is more thorough than I'd have asked for.

**One thing before this can land: it has merge conflicts.** Since you opened it, `main` wrapped both `buildTripReferences` call sites in an `if includeReferences { ... }` block. The conflict is purely textual — the same one-argument swap re-applies cleanly on the other side of that guard — so no re-review needed once it's rebased. Please merge `main` in and resolve, and I'll merge it.

One non-blocking observation for whenever you or someone else is next in this area: `parseIncludeTrip` returns `false` for a present-but-malformed value (`includeTrip=yes`, or an empty `includeTrip=`), whereas the analogous `ShouldIncludeReferences` falls back to the documented default on a parse failure. Since the documented default here is `true`, malformed input currently drops all trip references silently. That's pre-existing behavior in the shared helper and reusing it was correct, so I don't want it in this PR — just worth knowing it's there.

burma-shave (approved):

PR #1219, opened 2026-07-22
Fix: Revert pagination in agencies-with-coverage to match legacy spec

### LINKED ISSUE #1212: agencies-with-coverage: `maxCount` and pagination behavior deviation
**Description:**

The legacy OneBusAway specification explicitly notes a defect for this endpoint where `maxCount` was accepted by the router but *never applied*, meaning `limitExceeded` was always `false`. 

**Current Behavior**
The Maglev handler utilizes a generic pagination utility that parses `maxCount` (along with `offset` and `limit`), actively truncates the response list, and correctly flags `limitExceeded = true`.

**Action Required**
Maintainer/team consensus is required to decide if Maglev should maintain bug-for-bug compatibility with the legacy server (by removing the pagination logic for this specific handler) or intentionally preserve this pagination fix. If we decide to keep the fix, we must document it explicitly as a deliberate improvement over the reference implementation.

ARCoder181105:
### Description
This PR resolves a discrepancy regarding how pagination is handled in the `/api/where/agencies-with-coverage.json` endpoint.

According to the OneBusAway specification, this endpoint accepts the `maxCount` parameter but does not apply it, and `limitExceeded` is always returned as `false`.

This PR removes the pagination logic from this specific handler and ensures `limitExceeded` is always returned as `false`.

### Changes Made
- Removed `utils.ParsePaginationParams` and `utils.PaginateSlice` calls from `agencies_with_coverage_handler.go`.
- Hardcoded `limitExceeded := false` for the response payload to match the spec.
- Removed the `TestAgenciesWithCoverageHandlerPagination` test since pagination is no longer supported for this endpoint.

### Related Issues
Closes #1212

## Summary by CodeRabbit

* **Behavior Changes**
  * The agencies-with-coverage endpoint now returns all matching agencies and their regional coverage in a single response.
  * Pagination parameters no longer limit or offset the results.
  * Responses consistently indicate that the result set was not truncated.


Ahmedhossamdev (commented):
Nice!

burma-shave:
Thanks for the PR, but in this case I've updated to the spec to reflect that maglev deviates from the spec and honors the maxCount param. I wasn't able to see any negative client impact in preserving the current and correct behaviour.

PR #1223, opened 2026-07-22
Add maxCount validation to trips-for-route

### LINKED ISSUE #1210: `maxCount` parameter is not accepted or validated
**Current behavior**

The handler never reads or validates `maxCount`. The reusable `utils.ParseMaxCount` (`internal/utils/api.go`) exists but is unused here.

**Spec requirement**

- `maxCount` is accepted and validated (even though it is ultimately ignored for result-limiting — the documented Java defect).
- Java accepts any integer silently (e.g., `0`, `-1`, `300`) without returning a `400` error and without applying any limits.
- Java returns `400 Bad Request` only for non-numeric inputs (e.g., `maxCount=abc`) due to type conversion failures.

3rabiii:
### Summary

Adds `maxCount` query parameter validation to the `trips-for-route` endpoint. The parameter is validated to strictly match the legacy Java OBA behavior: it silently accepts any integer (including 0, negative values, and large numbers) and only rejects non-numeric inputs. The parsed value is intentionally not applied to the result list, preserving the documented upstream defect.

### Changes

*   **`internal/restapi/trips_for_route_handler.go`**: Added a simple `strconv.Atoi` check to reject non-numeric inputs (which cause a `400 Bad Request` upstream via Struts2 type conversion). Any valid integer is parsed but intentionally discarded via the blank identifier `_` so it does not truncate the result set.
*   **`internal/restapi/trips_for_route_handler_test.go`**: 
    * Updated `TestTripsForRouteHandler_MaxCountValidation` to expect `200 OK` for default/omitted, valid value, zero, negative, and out-of-bounds inputs. Only non-numeric inputs (`abc`) expect a `400 Bad Request`.
    * Added a new test `TestTripsForRouteHandler_MaxCountDoesNotTruncate` using a fixture with multiple concurrently active trips to explicitly assert that passing `maxCount=1` returns the exact same baseline count as a request without `maxCount`, proving the parameter is correctly ignored.

Closes: #1210 

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for the `maxCount` query parameter when retrieving trips for a route.
  * Invalid values now return a clear validation error identifying the affected field.
  * Valid and omitted values continue to be accepted.


aaronbrethorst:
### Code review

Found 1 issue:

1. `maxCount` validation is stricter than the reference implementation and turns requests that upstream answers with `200 OK` into `400 Bad Request`. `TripsForRouteAction` builds its `MaxCountSupport` with the no-arg constructor (`_maxCount = _absoluteMaxCount = Integer.MAX_VALUE`), `setMaxCount` carries no `@IntRangeFieldValidator`, and there is no `TripsForRouteAction-validation.xml` — so `maxCount=0`, `maxCount=-1` and `maxCount=300` are all accepted and simply never applied. The only value that produces a `400` upstream is a non-integer, via Struts2 type conversion. The wiki spec agrees: `trips-for-route.md` lists no `maxCount` error path under Extensions, the request schema says only "Accepted but ignored — has no effect on result count.", and Minimal Guarantees says the response is always `200 OK`. Where a ceiling does exist upstream (e.g. `routes-for-location`), the documented behaviour is "The server silently clamps `maxCount` ... No error is returned" — never a `400`. Using `utils.ParseMaxCount` (reject-above-max) here therefore imports both a `<= 0` rejection this action does not have and a 250 ceiling this action does not have. The corresponding assertions in `TestTripsForRouteHandler_MaxCountValidation` for `0`, `-1` and `300` encode the wrong expectation. (Note: the premise in issue #1210 that "Java returns 400 for maxCount <= 0 or maxCount > 250" does not hold for this action.)

https://github.com/OneBusAway/maglev/blob/48b14ef8e36bc3e9fd9a76270c4d1511cb5ad6e6/internal/restapi/trips_for_route_handler.go#L32-L39

aaronbrethorst (changes requested):
You got the hard part exactly right, and I want to call it out before the criticism: you discarded the parsed value with `_` so `maxCount` never truncates the result list, and left `limitExceeded` hard-coded `false`. That's the central spec point for this endpoint and it's counterintuitive — the obvious "fix" would have been to apply the limit, which would have been wrong. The spec is explicit that `TripStatusBeanServiceImpl.getTripsForRoute` never reads `query.getMaxCount()`, so all matching trips are always returned. You matched that.

The problem is the validation half.

**Returning 400 for `maxCount <= 0` or `maxCount > 250` isn't upstream behavior for this endpoint.**

`utils.ParseMaxCount` produces a `maxCount` field error for values `<= 0` and rejects values above the cap. But `TripsForRouteAction` has no range validation at all: `_maxCount` is a bare `new MaxCountSupport()`, whose no-arg constructor sets both the default and the absolute max to `Integer.MAX_VALUE`, and `setMaxCount(int)` carries no validator annotation. So upstream returns `200 OK` for `maxCount=0`, `-1`, and `300` alike.

The wiki backs this up from the other direction. The Minimal Guarantees say the response is always `200 OK`, and the Extensions section enumerates exactly one 400 case for this endpoint — a missing `id` path parameter. There is no `maxCount` error listed. The parameter's own schema entry reads simply "Accepted but ignored — has no effect on result count."

It's worth contrasting `routes-for-location`, where a ceiling *does* exist: even there the spec notes the server "silently clamps `maxCount` … No error is returned", and the `<= 0` → 400 behavior comes from explicit code in `RoutesForLocationAction.index` that `TripsForRouteAction` simply doesn't have. Neither `ParseMaxCount` nor `ParseMaxCountClamped` matches an action with no ceiling whatsoever.

So as written, this converts requests that currently succeed into 400s. What I'd like instead: keep rejecting non-numeric input (`maxCount=abc` genuinely is a 400 upstream, via Struts2 type conversion) and accept any integer value silently.

I should flag the root cause so it doesn't bite again: linked issue #1210 states "Java returns 400 for `maxCount <= 0` or `maxCount > 250`". That's accurate for some endpoints but not for `trips-for-route`, and it's what sent this PR in the wrong direction. Worth correcting the issue too — that's not on you to have caught.

One more thing while you're in here: since the whole point is that `maxCount` must *not* truncate, the test should assert that. Right now the `200` cases only check the status code, so if someone later wires the limit in, these tests still pass. Asserting that the result count is identical with `maxCount` absent and with `maxCount=1` would lock in the behavior you're deliberately preserving.

This also has merge conflicts — its base predates the `includeReferences := ShouldIncludeReferences(r)` addition on `main`. The inserted block itself stays correct after a rebase since it sits before the `GetAgency` call, so that part is mechanical.

Drop the range validation, keep the non-numeric check, add the count-unchanged assertion, and rebase — then I'll merge it.

burma-shave:
Closing this one out without merging — the validation logic and tests are good, this is about whether `maxCount` should apply to this endpoint at all. I don't think it should.

- **The result set isn't open-ended.** `maxCount` matters on endpoints like `stops-for-location` or `route-search`, where the result set can genuinely be large. `trips-for-route`'s list is one entry per currently-active vehicle on the route in a ±30/+10 min window — bounded by how many vehicles a route can physically have in service, not something that needs capping.
- **There's nothing meaningful to keep when truncating.** Per the wiki's "Ordering note," results come out of a hash-set collection in non-deterministic order. There's no "closest" or "soonest" to sort by, so honoring `maxCount` just returns an arbitrary N of whatever's active — not a usable "top N."
- **Truncating actively misleads the clients that consume this.** Checked how Wayfinder and Android use this response: Wayfinder's marker sweep (`vehicleUtils.js`) treats "not in this response" as "vehicle no longer active" and deletes its map marker; Android's extrapolation layer silently drops observations for anything missing. This isn't pagination where "there's more, fetch page 2" is fine — it reads as vehicles going offline that are still running.

So the "Suspected Defect" framing in the spec is probably wrong for this endpoint — Java isn't failing to apply `maxCount`, the parameter likely doesn't belong here at all, inherited from a shared request path rather than deliberately supported. Accept-and-validate-but-never-apply (your original approach, before I pushed toward enforcing it) may be the correct behavior, not a bug. Let me know if you see it differently.

I'll update spec accordingly.

PR #1240, opened 2026-07-24
align reference stop IDs with schedule combined IDs

### LINKED ISSUE #1237: Stop ID format mismatch between references and schedule stop times in trips-for-route
### Description
There is a data inconsistency in the `trips-for-route` endpoint response. The Stop IDs provided in the `data.references.stops` array do not match the format of the Stop IDs used in `data.list[].schedule.stopTimes[]`. 

This mismatch breaks the reference model, as clients cannot look up a stop in the `references` dictionary using the `stopId` provided in the trip schedules.

### Current Behavior
- In `internal/restapi/trips_for_route_handler.go`, the `collectStopIDsFromSchedule` function strips the agency prefix from the combined stop IDs (e.g., stripping `"25_"` from `"25_tfr-stop1"` to store `"tfr-stop1"`).
- Consequently, the stops populated in `data.references.stops` use these bare IDs.
- Meanwhile, the schedule stop times generated (e.g., via `calculateBatchStopDistances` in `trips_helper.go`) retain the **combined** stop IDs (`AgencyID_StopID`).

### Expected Behavior (Spec Requirement)
Stop IDs must be strictly consistent throughout the response envelope. Since the schedule stop times follow the combined ID convention (matching standard OBA specifications), the reference dictionary must use the exact same combined IDs to allow for successful client-side lookups.

3rabiii:
### Description
This PR resolves a data inconsistency in the `trips-for-route` response where the Stop IDs provided in `data.references.stops` did not match the combined ID format (`AgencyID_StopID`) used within the trip schedules. This mismatch prevented successful client-side lookups.

### Changes Made
- **Translation Map Strategy:** Updated `collectStopIDsFromSchedule` to build a `map[string]string` acting as a translation dictionary, mapping bare database IDs to their fully combined schedule IDs.
- **Reference Serialization:** Modified `buildTripReferences` to accept this translation map, applying the combined IDs when constructing the `models.Stop` references.
- **Database Compatibility:** Ensured that `GetStopsByIDs` queries continue to use bare IDs, maintaining compatibility with the database schema.
- **Handler Signature:** Updated the early return path in the main handler to pass `nil` for the new map parameter to satisfy the Go compiler.
- **Test Coverage:** Updated all relevant `collectStopIDsFromSchedule` unit tests to reflect the new map type. Enhanced the integration test (`TestTripsForRouteHandler_DifferentRoutes`) to strictly assert that reference stop IDs are generated in the combined format (verifying the presence of `_`).

Closes: #1237 

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected trip route references so stop entries retain their complete stop identifiers.
  * Improved handling of stop references across routes and schedules.
  * Added validation to ensure referenced stop IDs are returned in the expected format.


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Approved on the merits. This is a real bug and the fix is well-chosen.

The spec says `data.list[].schedule.stopTimes[].stopId` is a combined stop ID and that full detail for it appears in `data.references.stops` — which only works if the two strings match. They didn't: the schedule emitted `utils.FormCombinedID(agencyID, stopTime.StopID)` while `references.stops[].id` carried the bare feed ID straight off the DB row. So a client doing the lookup the spec describes found nothing.

What I like about the approach: you carry the *exact* combined string the schedule already produced through the map, rather than re-deriving it. `stopID = combinedID` is a plain assignment, not a re-wrap, so there's no way to double-prefix — which is the obvious failure mode for a change like this. And keeping the DB lookup keyed on bare IDs is right; `GetStopsByIDs` needs the feed ID.

I also checked this leaves the endpoint consistent with its siblings — `stops_for_route_handler.go`, `reference_utils.go`, and `trips_for_location_handler.go` all emit `utils.FormCombinedID(agencyID, stop.ID)` for reference stop IDs, so this brings `trips-for-route` in line rather than inventing a local convention. And `references.stops[].routeIds` were already combined via the SQL, so there's no residual mismatch left inside the stop object.

Tightening the integration test to exact set equality between `references.stops[].id` and the deduped schedule stop IDs is the right assertion — much better than checking for the presence of an underscore.

**Blocked only on the merge conflict.** `main` has since wrapped both `buildTripReferences` call sites in an `includeReferences` guard (from #1216), which shifts them. The resolution is mechanical and doesn't change behavior, so no re-review needed — merge `main` in, resolve, and I'll merge this.

Two small notes for later, neither blocking:

- `collectStopIDsFromSchedule` is first-wins per bare stop ID. If two trips in the same result set belong to different agencies and share a stop, only one combined ID lands in references and the other agency's `stopTimes[].stopId` still won't resolve. Blocks are agency-scoped in practice so this is rare, and it's strictly better than the status quo where *neither* resolved.
- The `stopID := stop.ID` fallback is unreachable — `stops` comes solely from `GetStopsByIDs(bareIDs)` where `bareIDs` are exactly the map's keys, so the `ok` branch always fires. Harmless, but it's dead.

Also: the PR description still describes the older assertion (checking for the presence of `_`) rather than the exact-set-equality one you actually landed in `f694603c`. Worth updating so the commit history reads accurately.

burma-shave (approved):

PR #1247, opened 2026-07-25
Remove dead RFC3339 time parse from stops-for-route

### LINKED ISSUE #1245: stops-for-route: remove dead RFC3339 time parse in parseStopsForRouteParams
**Description:**

`parseStopsForRouteParams` parses the `time` query param with`time.Parse(time.RFC3339, …)` into `stopsForRouteParams.Time`
([stops_for_route_handler.go:32-36](internal/restapi/stops_for_route_handler.go#L32-L36)),
but `params.Time` is never read. The handler re-reads the raw `time` param and does the real parsing through `utils.ParseTimeParameter`
([stops_for_route_handler.go:73-83](internal/restapi/stops_for_route_handler.go#L73-L83)),
which correctly accepts `yyyy-MM-dd` and epoch-ms per the spec.

Problems:
- `stopsForRouteParams.Time` and its RFC3339 parse are dead code — CONTRIBUTING
  explicitly calls out leftover dead code as a common review finding.
- It is misleading: RFC3339 is *not* an accepted input format for this
  parameter, so the block reads as if the endpoint supports a format it does
  not.
- The `time` param is read from the query twice.

Ahmedhossamdev:
Fixes: #1245 

`parseStopsForRouteParams` parsed the `time` query param as RFC3339 into a`Time` field that no handler code ever read. The live service-date filter reads `time` separately and parses it via `utils.ParseTimeParameter` (which accepts `yyyy-MM-dd` / epoch ms per the spec), so the RFC3339 block was dead code that also implied an input format the endpoint doesn't support.

With `Time` gone, the single-field `stopsForRouteParams` struct and its helper no longer earned their keep, so `includePolylines` is now derived inline at its point of use.

**No behavior change:** `includePolylines` still defaults to `true`, disabled
only by an explicit `false`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Route stop results now include route polylines by default.
  * Polylines can be disabled by explicitly setting the `includePolylines` query option to `false`.
  * Service-date filtering continues to work as expected.

### DIFF
--- a/internal/restapi/stops_for_route_handler.go
+++ b/internal/restapi/stops_for_route_handler.go
@@ -5,37 +5,12 @@ import (
 	"context"
 	"net/http"
 	"slices"
-	"time"
 
 	"maglev.onebusaway.org/gtfsdb"
 	"maglev.onebusaway.org/internal/models"
 	"maglev.onebusaway.org/internal/nulls"
 	"maglev.onebusaway.org/internal/utils"
 )
-
-type stopsForRouteParams struct {
-	IncludePolylines bool
-	Time             *time.Time
-}
-
-func (api *RestAPI) parseStopsForRouteParams(r *http.Request) stopsForRouteParams {
-	now := api.Clock.Now()
-	params := stopsForRouteParams{
-		IncludePolylines: true,
-		Time:             &now,
-	}
-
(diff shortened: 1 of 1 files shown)


Ahmedhossamdev:
@coderabbitai review

burma-shave (approved):

PR #1249, opened 2026-07-25
Return all stops for routes without `direction_id`

### LINKED ISSUE #1248: stops-for-route: NULL direction_id drops stops served by trips other than the first
## Description

Spec step 4 requires the flat `stopIds` list to be the union of every stop on
**all** qualifying trips for the route, and step 5 requires that when trips lack
`direction_id`, the server split them into similarity-based groups (`"0"`,
`"1"`, …).

When `direction_id` is `NULL`, `orderedStopIDsForGroup` falls back to
`GetOrderedStopIDsForTrip(group.Trips[0].ID)` — a **single** trip
([stops_for_route_handler.go:331-346](internal/restapi/stops_for_route_handler.go#L331-L346)).
Because both the flat `stopIds` and the group's `stopIds` are populated from
this ordered list
([stops_for_route_handler.go:280-282](internal/restapi/stops_for_route_handler.go#L280-L282)),
any stop that appears only on a later trip with a different stop pattern is
dropped entirely.

In addition, `groupTripsByDirection` lumps all `NULL`-direction trips into a
single group with ID `"0"`
([trips_helper.go:1200-1225](internal/restapi/trips_helper.go#L1200-L1225)),
rather than splitting them into similarity-based groups as required by the
specification.

`TestStopsForRouteNullDirectionID` currently exercises only a feed containing a
single trip, so this behavior is not covered by tests.

## Java Verification (CONFIRMED)

`RouteServiceImpl.getStopsForRouteCollectionForServiceInterval` (lines 67–95)
builds the flat stop set by iterating **every** trip on the route and adding
**every** stop time's stop to a `Set<AgencyAndId>`. This logic is completely
independent of direction grouping.

Maglev instead derives its flat `stopIds` from the union of the direction
groups' ordered stop lists. In the `NULL`-direction fallback, that ordered list
is generated from only `Trips[0]`, so stops that appear exclusively on later
trips are omitted.

`StopSequenceCollectionServiceImpl.getGroupsForStopSequences` (lines 159–230)
also confirms that trips without a `direction_id` are partitioned into
multiple similarity-based groups using:
- Union-find
- `STOP_SEQUENCE_MIN_COMMON_RATIO = 0.3`
- Integer-string group IDs (`"0"`, `"1"`, ...)

Maglev currently places all such trips into a single group `"0"`.

> **Note:** This issue only affects feeds where `direction_id` is `NULL`.
> For feeds where every trip has a `direction_id`, Maglev's per-direction query
> already considers all trips, so the flat `stopIds` are complete.

## Proposed Fix (Minimum)

For the `NULL`-direction fallback:

- Build the ordered stop list from the **union of stops across all trips in the
  group**, rather than only `Trips[0]`.
- This ensures both the flat `stopIds` list and the group's `stopIds` are
  complete.

Splitting `NULL`-direction trips into multiple similarity-based groups is a
larger behavioral change and can be handled as a follow-up or documented as an
Implementation Decision.

## Tests

Extend the `NULL`-direction fixture with a second trip that visits an additional
stop, and verify that:

- The flat `stopIds` contains stops from **both** trips.
- The group's `stopIds` also contains stops from **both** trips.

Ahmedhossamdev:
Fixes: #1248 

`stops-for-route` dropped stops when a feed's trips have no `direction_id`. The
per-direction ordered query returns nothing in that case (`direction_id = NULL`
→ UNKNOWN in SQL), so the handler fell back to ordering a single trip — and
since both the flat `stopIds` and the group's `stopIds` come from that list, any
stop served only by another trip was lost. The reference implementation builds
`stopIds` as the union across every qualifying trip.

**Fix:** add `GetOrderedStopIDsForTrips` (unions + orders stops across a set of
trip IDs, same as the per-direction query) and use it for the NULL-direction
fallback with all of the group's trips.

**Test:** extended the NULL-direction fixture with a second trip visiting a stop
the first doesn't; verified it fails on the old code and passes now.

Scope: minimal completeness fix. Splitting NULL-direction trips into
similarity-based groups remains a follow-up.

### DIFF
--- a/gtfsdb/db.go
+++ b/gtfsdb/db.go
@@ -216,6 +216,9 @@ func Prepare(ctx context.Context, db DBTX) (*Queries, error) {
 	if q.getOrderedStopIDsForTripStmt, err = db.PrepareContext(ctx, getOrderedStopIDsForTrip); err != nil {
 		return nil, fmt.Errorf("error preparing query GetOrderedStopIDsForTrip: %w", err)
 	}
+	if q.getOrderedStopIDsForTripsStmt, err = db.PrepareContext(ctx, getOrderedStopIDsForTrips); err != nil {
+		return nil, fmt.Errorf("error preparing query GetOrderedStopIDsForTrips: %w", err)
+	}
 	if q.getProblemReportsByStopStmt, err = db.PrepareContext(ctx, getProblemReportsByStop); err != nil {
 		return nil, fmt.Errorf("error preparing query GetProblemReportsByStop: %w", err)
 	}
@@ -712,6 +715,11 @@ func (q *Queries) Close() error {
 			err = fmt.Errorf("error closing getOrderedStopIDsForTripStmt: %w", cerr)
 		}
 	}
+	if q.getOrderedStopIDsForTripsStmt != nil {
+		if cerr := q.getOrderedStopIDsForTripsStmt.Close(); cerr != nil {
+			err = fmt.Errorf("error closing getOrderedStopIDsForTripsStmt: %w", cerr)
+		}
+	}
 	if q.getProblemReportsByStopStmt != nil {
 		if cerr := q.getProblemReportsByStopStmt.Close(); cerr != nil {
 			err = fmt.Errorf("error closing getProblemReportsByStopStmt: %w", cerr)
@@ -1100,6 +1108,7 @@ type Queries struct {
 	getNextStopInTripStmt                         *sql.Stmt
 	getOrderedStopIDsForRouteDirectionStmt        *sql.Stmt
 	getOrderedStopIDsForTripStmt                  *sql.Stmt
+	getOrderedStopIDsForTripsStmt                 *sql.Stmt
 	getProblemReportsByStopStmt                   *sql.Stmt
 	getProblemReportsByTripStmt                   *sql.Stmt
 	getRouteStmt                                  *sql.Stmt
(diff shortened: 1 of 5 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Nice catch, and a clean fix. The `direction_id = NULL` → UNKNOWN trap is genuinely easy to miss, and the old `group.Trips[0]` fallback silently returning one trip's stops instead of the group's is exactly the kind of bug that looks fine until you test a feed that omits `direction_id`.

Things I checked and liked:

- `GetOrderedStopIDsForTrips` mirrors `GetOrderedStopIDsForRouteDirection` exactly — same `GROUP BY stop_id ORDER BY MAX(stop_sequence)` semantics — so the fallback path and the normal path now order stops the same way instead of diverging.
- The `gtfsdb/` diff is straight sqlc output: statement registration, `Close`, the `Queries` struct, and `WithTx` all updated in the right alphabetical slots, matching the existing slice-query codegen verbatim. No hand-editing.
- The test actually fails against the old code. `tripB` visits `stopA3`, which `tripA` never does, and trips sort by ID so `Trips[0]` is `tripA` — so the union assertion is doing real work rather than restating current behavior. Swapping `assert.Len(..., 2)` for `ElementsMatch` on the full expected set is the right call too.
- The updated comment explains *why* the union matters, not just what changed.

I appreciate you calling out in the description that similarity-based grouping for NULL-direction feeds is still a follow-up rather than quietly leaving it. That's the right way to handle a known gap. For what it's worth, the reference implementation splits those trips into similarity-based groups with sequential integer ids, so a route with two real directions and no `direction_id` will still land in one group here with an approximate order — but that's strictly better than today, where half the stops were missing outright, and the flat `stopIds` list is sorted lexicographically by combined ID anyway, so it's unaffected.

Merging.

PR #1251, opened 2026-07-25
Disambiguate stops-for-route directions sharing a headsign

### LINKED ISSUE #1250: stops-for-route: direction groups sharing a headsign get identical names
Spec step 7 and the `stopGroups[].name.names` field state that when two or more
directions share the same most-common trip headsign, a direction disambiguator
must be appended to each name so the groups remain distinguishable.

`buildStopGroup` sets the name straight from `mostCommonHeadsign` with no
disambiguation ([stops_for_route_handler.go:301-311](internal/restapi/stops_for_route_handler.go#L301-L311),
[stops_for_route_handler.go:350-360](internal/restapi/stops_for_route_handler.go#L350-L360)),
so two directions whose most-common headsign is identical produce two groups
with the same `name.names`. Groups are then sorted by that name
([stops_for_route_handler.go:258-260](internal/restapi/stops_for_route_handler.go#L258-L260)),
making their order nondeterministic in the tie case.

**Fix:** after building the per-direction headsigns, detect collisions and
append a disambiguator (e.g. the direction id / terminal stop) as the Java
reference does.

**Tests:** fixture with two directions sharing a headsign; assert distinct
`name.names`.

Ahmedhossamdev:
Fixes: #1250 

When two direction groups resolved to the same most-common headsign,
both were emitted with identical name.names, and the name-based group
sort left their order nondeterministic. The reference implementation
appends a direction disambiguator whenever directions collide.

Append " - <direction id>" to every group's name on a collision, before
the sort so ordering is deterministic. Update the RABA end-to-end test
(both directions are "Shasta Lake") and add a unit test for the
collision and no-collision paths.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stop-group (destination) naming when a route has multiple directions with the same headsign.
  * Added deterministic disambiguation to prevent duplicate stop-group names and ensure consistent ordering.
  * Updated end-to-end behavior so outbound/inbound stop-group names and their display name lists reflect collision-free, direction-suffixed values.
* **Tests**
  * Added unit tests covering normal cases, duplicate collisions, and order-independent results for stop-group name disambiguation.

### DIFF
--- a/internal/restapi/stops_for_route_handler.go
+++ b/internal/restapi/stops_for_route_handler.go
@@ -255,6 +255,8 @@ func processTripGroups(
 		allStopGroups = append(allStopGroups, stopGroup)
 	}
 
+	disambiguateGroupNames(allStopGroups)
+
 	slices.SortFunc(allStopGroups, func(a, b models.StopGroup) int {
 		return cmp.Compare(a.Name.Name, b.Name.Name)
 	})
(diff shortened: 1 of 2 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is a tidy fix for a genuinely annoying bug — two groups both named "Shasta Lake" is useless to a rider, and it also left the name-based sort with a tie, so group order wasn't even stable. Nice that fixing the first problem fixes the second for free.

What I checked:

- The suffix format matches the reference implementation exactly: `name + " - " + direction`. (The Java side also capitalizes the first character, but that's a no-op for numeric direction ids, so the output is identical.)
- `name.type` stays `"destination"` and `name.names` is kept in sync with `name.name` — easy to forget one of those.
- The function only rewrites name fields; it never adds, removes, or reorders groups, so nothing downstream can see a duplicate or empty group appear.
- Mutating through `groups[i]` rather than a range copy is correct, and the result is a function of the group set alone — the `TestDisambiguateGroupNames` order-independence case is the right property to pin down, since it's what makes the subsequent sort deterministic.
- Re-checking against the *resulting* names rather than the original ones is a nice touch, and the bound is safe: direction ids are unique, so any two colliding groups separate on the next pass.

The one deliberate divergence — suffixing only the colliding groups instead of every group whenever any collision exists — is documented in the doc comment, which I appreciate. It's unreachable today regardless: `groupTripsByDirection` keys on `direction_id`, which GTFS restricts to 0 and 1 (NULL coerces to 0), so you can't get three groups without the similarity-based grouping we haven't implemented. If we ever add that, this is worth revisiting, and the comment will point whoever does it in the right direction.

Updating the RABA end-to-end expectations rather than working around them was the right call too.

Merging.

Issue #1235, opened 2026-07-24
Spec review: stops-for-location


burma-shave:
## What to review
Read the [wiki spec](https://github.com/OneBusAway/maglev/wiki/stops-for-location) and compare the full Maglev implementation of this endpoint against it. Look for:

- **Request parameters**: all params supported with correct types, defaults, and validation
- **Response structure**: all fields present with correct names, types, and nesting
- **Error handling**: error cases match the spec (404, 400, etc.)
- **Behaviour**: the endpoint behaves as the spec describes, including edge cases
- **Test coverage**: main scenarios and edge cases from the spec are covered by tests

## Workflow
1. Compare the implementation against the wiki spec. If no gaps found, move this card to **Done**.
2. For each gap or bug, file a GitHub issue labelled `spec-gap`, link it to this card, and assign it to yourself.
3. Move this card to **Issues Filed**.
4. Work on the issues you filed.
5. Once all linked issues are closed, move this card back to **In Review** and repeat from step 1.

## Deeper analysis
The skills at https://github.com/burma-shave/maglev-workspace can help when the spec or implementation behaviour is not clear. They analyse the Java OBA reference implementation by downloading, building, and running the server — enabling both static and runtime analysis.

Install them into your coding agent (e.g. Claude Code). Key capabilities:
- **Q&A mode**: ask arbitrary questions about this endpoint and get answers grounded in the Java source and live server behaviour
- **Spec generation**: regenerate or cross-check the spec directly from the Java source

**Requirements**: Docker and Git
**Note**: the skills are token-intensive and trigger many permission prompts — a VM or sandbox environment is recommended

Mister-Raggs:
Spec review of stops-for-location, compared against the wiki spec and verified on canonical legacy 2.7.1 (booted via maglev-workspace, KCM), with the maglev side confirmed on RABA.

**Gaps filed** 🔴 (both-sides evidence unless noted)
- #1238 — `maxCount > 250` returns HTTP 400 instead of legacy's silent cap at 250 (HTTP 200).
- #1239 — `query` mode is missing the closest-match fallback and uses a 500 m instead of 10 km default radius.
- #1244 — result list sorted by raw stop ID instead of the spec's combined `{agencyId}_{stopId}`. (Source-cited spec behaviour, not a Defect Workshop item; single-agency test data can't distinguish it at runtime, so not runtime-verified.)

[...]

PR #1253, opened 2026-07-25
fix: sort stops-for-location results by combined stop ID

### LINKED ISSUE #1244: stops-for-location: result list sorted by raw stop ID instead of combined {agencyId}_{stopId}
Part of #1235 (Spec review: stops-for-location).

### Spec Requirement

Main Success Scenario step 7: *"The result list is sorted lexicographically by combined stop ID"* — cited to [`StopsBeanServiceImpl.java#L274`](https://github.com/OneBusAway/onebusaway-application-modules/blob/095bf1ac3aeb7009b9f78e7f1cb4c65bd68b988a/onebusaway-transit-data-federation/src/main/java/org/onebusaway/transit_data_federation/impl/beans/StopsBeanServiceImpl.java#L274). "Combined" = `{agencyId}_{stopId}`.

### Current Maglev Behavior

`stopsForLocationHandler` sorts by the **raw** stop ID:

```go
slices.SortStableFunc(stops, func(a, b gtfsdb.Stop) int {
	return cmp.Compare(a.ID, b.ID)
})
```

Results are built with combined IDs afterward, so the returned order reflects raw stop IDs. For a single agency this is identical to combined-ID order (the `{agencyId}_` prefix is constant); it diverges only when stops from **multiple agencies** appear in one response, where combined-ID order groups by agency first.

Evidence note: not runtime-distinguishable on the single-agency test feeds (RABA/KCM); this rests on the spec's cited Java source plus the handler code above.

### Proposed Solution

Sort the result list by the combined `{agencyId}_{stopId}` ID rather than the raw stop ID. Add a multi-agency ordering test if a suitable fixture is available.

Mister-Raggs:
Fixes #1244 (sub-issue of the #1235 spec review).

Main Success Scenario step 7 orders the result list lexicographically by **combined** stop ID (`{agencyId}_{stopId}`), cited to [`StopsBeanServiceImpl.java#L274`](https://github.com/OneBusAway/onebusaway-application-modules/blob/095bf1ac3aeb7009b9f78e7f1cb4c65bd68b988a/onebusaway-transit-data-federation/src/main/java/org/onebusaway/transit_data_federation/impl/beans/StopsBeanServiceImpl.java#L274). The handler sorted the raw stop rows by their unprefixed IDs instead, which only agrees with the spec while every stop belongs to the same agency.

The sort now runs on the built result list, keyed on the combined ID, replacing the earlier raw-ID sort (one sort, correct key).

### Test

The single-agency fixtures can't distinguish the two orderings, so the test builds two agencies whose raw and combined orders disagree — raw `"2" < "9"`, combined `"SortA_9" < "SortB_2"`. Against the previous implementation it fails with:

```
expected: []string{"SortA_9", "SortB_2"}
actual  : []string{"SortB_2", "SortA_9"}
```

Verified: full suite in both `sqlite_fts5` and `purego` modes, `go vet` both, race detector on `restapi`/`gtfs`, `gofmt`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the ordering of stops returned by location search results.
  * Stops are now consistently sorted lexicographically by their combined stop ID, ensuring stable, predictable ordering across agencies.

### DIFF
--- a/internal/restapi/stops_for_location_handler.go
+++ b/internal/restapi/stops_for_location_handler.go
@@ -102,11 +102,6 @@ func (api *RestAPI) stopsForLocationHandler(w http.ResponseWriter, r *http.Reque
 
 	stops, limitExceeded := api.GtfsManager.GetStopsForLocation(ctx, loc, query, maxCount, routeTypes)
 
-	// Referenced Java code: "here we sort by distance for possible truncation, but later it will be re-sorted by stopId"
-	slices.SortStableFunc(stops, func(a, b gtfsdb.Stop) int {
-		return cmp.Compare(a.ID, b.ID)
-	})
-
 	results := []models.Stop{}
 	routeIDs := map[string]bool{}
 	agencyIDs := map[string]bool{}
@@ -239,6 +234,11 @@ func (api *RestAPI) stopsForLocationHandler(w http.ResponseWriter, r *http.Reque
 		api.clientCanceledResponse(w, r, ctx.Err())
 		return
 	}
+
+	// Spec: results are ordered lexicographically by combined stop ID.
+	slices.SortStableFunc(results, func(a, b models.Stop) int {
+		return cmp.Compare(a.ID, b.ID)
+	})
 
 	agencies := utils.FilterAgencies(allAgencies, agencyIDs)
 	routes := utils.FilterRoutes(api.GtfsManager.GtfsDB.Queries, ctx, routeIDs)
--- a/internal/restapi/stops_for_location_handler_test.go
+++ b/internal/restapi/stops_for_location_handler_test.go
@@ -1,8 +1,10 @@
 package restapi
 
 import (
+	"context"
 	"fmt"
 	"net/http"
+	"slices"
 	"strings"
 	"testing"
 	"time"
@@ -10,8 +12,10 @@ import (
 	"github.com/OneBusAway/go-gtfs"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"maglev.onebusaway.org/gtfsdb"
 	"maglev.onebusaway.org/internal/clock"
 	"maglev.onebusaway.org/internal/models"
+	"maglev.onebusaway.org/internal/nulls"
 )
 
 func TestStopsForLocationHandlerRequiresValidApiKey(t *testing.T) {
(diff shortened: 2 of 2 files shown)


Mister-Raggs:
@coderabbitai review

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Approved on the merits — this is correct, and the test is the part that convinced me.

The spec is unambiguous that this endpoint is sorted and by what key: Success Guarantees say "The list is sorted lexicographically by combined stop ID," step 7 repeats it, and the `data.list` field doc says the same. `main` was sorting the raw `gtfsdb.Stop` rows by their unprefixed feed ID, which coincides with the correct answer in a single-agency deployment and quietly diverges in a multi-agency one. Moving the sort onto the final `[]models.Stop` keyed on the combined ID is the fix.

The `SortB_2` / `SortA_9` fixture is what makes this a real test: raw ordering and combined ordering disagree, so it can actually distinguish the old behavior from the new one. A same-agency fixture would have passed either way.

I checked the truncation interaction too, since that's where this kind of change usually goes wrong. Both truncation paths live inside `Manager.GetStopsForLocation`, which runs well before the new sort, so this sorts *after* truncation — matching the spec, which shuffles and truncates at step 5 and only sorts at step 7. `limitExceeded` is untouched.

Good call deleting that comment, too. It claimed "here we sort by distance for possible truncation," but the code hadn't sorted by distance in some time — it sorted by ID. A comment that describes code that no longer exists is worse than none.

**Blocked only on the merge conflict**, which is purely textual: `main` added `TestStopsForLocationHandlerClampsMaxCountAboveCap` at the same insertion point in the test file, right after `TestStopsForLocationHandlerValidatesRadius`. The handler hunks don't conflict at all. Rebase and resolve and I'll merge it — no re-review needed.

One thing I want to note but explicitly not ask you to change: comparing the *concatenated* combined ID lexicographically differs from Java's `AgencyAndId` ordering when one agency ID is a prefix of another (agency `1` vs `12`: by concatenation `"12_y"` sorts before `"1_x"`, but agency-then-id puts `"1"` first). You implemented what the spec's wording says, which is the right call — if that wording is imprecise, that's a spec question for me to chase, not a defect in your PR.

Minor, non-blocking: the test seeds agencies, routes, stops, trips, and stop_times into the package-shared GTFS DB with no cleanup, so those rows persist for every later test in the package. That's an established pattern here (`stop_handler_test.go` and the servicedate regression test do the same), so I'm not going to single you out for it — but asserting the whole list with `assert.Equal` does mean any future test seeding a stop near 41.5/-123.5 will break this one. Scoping the assertion to "these two appear in this relative order" would be sturdier.

Mister-Raggs:
@aaronbrethorst 
Rebased and resolved, conflict was test-file only. Both
tests are kept; the handler hunks applied cleanly. Now mergeable.

I also took the assertion suggestion, since it costs nothing and prevents a real
future break: it now looks up the index of each of the two IDs and asserts
SortA_9 comes before SortB_2, rather than asserting the whole list. I re-ran the
mutation check against the old raw-ID sort and the looser assertion still fails
there, so it hasn't lost its teeth.

Noted on the AgencyAndId prefix case (agency 1 vs 12), leaving that alone.

Verified after the rebase: full suite in both sqlite_fts5 and purego modes, go vet
both, race detector, gofmt.

PR #1256, opened 2026-07-28
resolve interlining tripId mismatch in trips-for-route

### LINKED ISSUE #1254: Interlining contract not honored for tripId vs status.activeTripId in trips-for-route
### Description
There is a behavioral bug in the `trips-for-route` endpoint regarding how trips are represented for interlined blocks. The API fails to distinguish between the trip that caused a block to be selected (the queried route) and the trip the vehicle is currently executing, which violates the API specification.

### Current Behavior
In `internal/restapi/trips_for_route_handler.go`, the handler iterates over `fetchedTrips` and assigns `entry.TripId = utils.FormCombinedID(agencyID, tripID)` based directly on the fetched trip. If the active trip of a block belongs to a *different* route than the one queried by the client (due to interlining), the primary `tripId` incorrectly reflects the wrong route's trip.

### Expected Behavior (Spec Requirement)
According to the OBA API specifications:
- `data.list[].tripId`: MUST always reflect the **queried route's trip** that caused the block to be included in the response.
- `data.list[].status.activeTripId`: MUST reflect the trip the vehicle is actually running at the current moment (which may be on a completely different route due to interlining).

3rabiii:
### Description
This PR resolves a critical spec compliance issue in the `trips-for-route` endpoint regarding interlined blocks. Previously, the handler incorrectly assigned the currently executing trip to the outer `tripId` field, even if that trip belonged to a different route due to interlining. 

According to the OBA specification, the outer `tripId` must consistently reflect the **queried route's trip**, while `status.activeTripId` should reflect the trip the vehicle is currently executing on the ground.

### Changes Made
- **Batch Resolution for Interlining:** Introduced a mechanism to collect interlined block IDs and fetch their associated trips via a single batch query (`GetTripsByBlockIDs`), successfully avoiding an N+1 query problem.
- **State Separation:** Constructed an O(1) translation map (`blockTripForRoute`) to map block IDs to the correct queried-route trip ID. The outer `tripId` is now properly overridden using this map, while `BuildTripStatus` remains untouched to accurately populate `status.activeTripId`.
- **Robust Integration Testing:** Added `TestTripsForRouteHandler_InterlinedBlock`. This test constructs an in-memory GTFS zip fixture featuring a single block spanning two distinct routes. By manipulating a mock clock, it enforces strict assertions to guarantee that the outer `tripId` matches the queried route, while the inner `activeTripId` reflects the simulated active trip.

Closes: #1254

---
<a href="https://app.blacksmith.sh/OneBusAway/codesmith/maglev/pr/1256"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787861210&installation_model_id=429412&pr_number=1256&repository=OneBusAway%2Fmaglev&return_to=https%3A%2F%2Fgithub.com%2FOneBusAway%2Fmaglev%2Fpull%2F1256&signature=531241400f7d15b267ad483a11fb783c954eb91f3060a1a598fe8c1f2e3ebc13"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected trip identifiers and schedule details for interlined trips so results match the requested route.
  * Improved real-time status and situation information for interlined trips.
  * Enhanced overnight, cross-day, and looping-trip handling.
  * Improved selection across service days and schedule gaps.
  * Added graceful fallback when timing information or matching trips are unavailable.

* **Tests**
  * Added coverage for interlined, overnight, looping-route, cross-day, gap, service-day, and missing-time scenarios.


aaronbrethorst:
### Code review

Found 2 issues:

1. The interlining tripId selection silently falls back to the wrong trip whenever there is a layover gap. The overlap test `e.MinArrivalTime <= activeMax && e.MaxDepartureTime >= activeMin` compares a queried-route trip's window against the *active* trip's window — but trips within a single block never overlap in time by construction (one vehicle), so it can only be true when two trips are exactly contiguous (`e.MaxDepartureTime == activeMin`). The repo itself models the gaps that break this (`block_layover` in `gtfsdb/schema.sql`, plus `GetActiveLayoverBlockIDsForRoute` used earlier in this same handler). When no candidate overlaps, the loop never breaks and `bestIdx` stays `0`, so `entryTripID` becomes the *earliest* queried-route trip of the whole block/service day rather than the trip that caused the block to be selected. Example: block `A1(07:00–07:50) → B(08:05–08:35) → A2(08:40–09:40)`, query at 08:30 — the block is selected via `A2` (08:00–08:40 window), neither `A1` nor `A2` overlaps `B`, so the entry reports `A1`, hours off. The new `TestTripsForRouteHandler_LoopingRouteBlock` only passes because its fixture makes `tfr-loop-b` end and `tfr-loop-c` start at exactly 10:30; add any layover and it selects `tfr-loop-a`. The spec requires "the trip on the queried route that caused this block to be selected", so this should carry through the selecting trip from the block-index/layover window queries above rather than re-deriving it from an overlap heuristic — or at minimum use an explicit, documented nearest-trip fallback instead of index 0.

https://github.com/OneBusAway/maglev/blob/3d5748070c768972cb7415383d99505532c4c6b5/internal/restapi/trips_for_route_handler.go#L399-L420

2. `tripId` now identifies the queried-route trip, but `schedule` and `situationIds` in the same entry still describe the *active* trip, so a single entry mixes two trip identities. `buildScheduleForTrip(ctx, tripID, ...)` (line 381) and `GetSituationIDsForTrip(r.Context(), tripID)` (line 427) remain keyed on the active trip while `TripId` is built from `entryTripID`. Per the wiki spec, `data.list[].schedule.stopTimes` is "Ordered array of scheduled stop times for **this trip**" and `schedule.previousTripId` is "the preceding trip in this vehicle's block" — both relative to the entry's own trip. With the new `TestTripsForRouteHandler_InterlinedBlock` fixture the response says `tripId: tfr-agency_tfr-trip-a` while `schedule.stopTimes` are trip-b's 11:55/12:05 times and `schedule.previousTripId` resolves to `tfr-trip-a` — i.e. the entry claims its own tripId is the previous trip in its block. The test requests `includeSchedule=true` but asserts nothing about the schedule, so this is uncovered. Java OBA builds `tripId`, `schedule` and `situationIds` from the same `BlockTripInstance` and lets only `status.activeTripId` diverge; these three should move together.

https://github.com/OneBusAway/maglev/blob/3d5748070c768972cb7415383d99505532c4c6b5/internal/restapi/trips_for_route_handler.go#L379-L397

aaronbrethorst (dismissed):
You've identified the right problem and set up the right structure for it. The spec is explicit that these are two different things — `data.list[].tripId` is "the trip on the queried route that caused this block to be selected," while `status.activeTripId` is "the trip the vehicle is currently executing" — and before this PR we conflated them. Establishing that split, and keeping `BuildTripStatus` on the active trip, is correct.

The `(BlockID, ServiceID)` composite key is a genuine improvement too. Keying on block alone would collide across service days, and the overnight test shows you thought about that case rather than assuming it away.

Two things need fixing before I can merge it.

**1. The overlap test can't succeed for the case it's meant to handle, and the fallback is silently wrong.**

```go
bestIdx := 0
activeMin := fetchedTrip.MinArrivalTime.Int64
activeMax := fetchedTrip.MaxDepartureTime.Int64
for i, e := range entries {
    if e.MinArrivalTime <= activeMax && e.MaxDepartureTime >= activeMin {
        bestIdx = i
        break
    }
}
entryTripID = entries[bestIdx].ID
```

A block is one vehicle running trips in sequence, so two trips in the same block never overlap in time. The condition is a standard inclusive interval-overlap test, which means it can only be satisfied at exact contiguity — the queried-route trip ending at precisely the instant the active trip begins. Add any layover and it fails: a trip ending 10:25 against an active trip starting 10:30 gives `10:25 >= 10:30` → false.

We model layovers explicitly — there's a `block_layover` table in the schema, and this same handler calls `GetActiveLayoverBlockIDsForRoute` a couple hundred lines up — so this isn't a hypothetical.

When nothing matches, the loop never breaks and `bestIdx` stays at its initial `0`. Since the query orders by `t.block_id, t.min_arrival_time, t.id`, `entries[0]` is the *earliest* queried-route trip in that block and service day. So the common case doesn't fall back to something neutral; it emits a specific wrong trip that can be hours from the one that actually caused block selection. And it does so silently.

`TestTripsForRouteHandler_LoopingRouteBlock` passes only because the fixture has `tfr-loop-b` ending and `tfr-loop-c` starting at exactly 10:30:00. Nudge one of those by a minute and the test fails — worth doing as a way to see the bug.

What you want is "the queried-route trip in this block whose window contains or is nearest to the query window" — nearest-by-time with a deterministic tie-break — not overlap. And there shouldn't be a silent fallback: if no candidate can be chosen, that's worth a log line at minimum.

I should flag that CodeRabbit raised this same `bestIdx` fallback on the last review round, and the only commit after it was `Format trips for route handler`. Easy for a bot comment to get lost in the noise, but this one was right.

**2. `tripId` now disagrees with `schedule` and `situationIds` in the same entry.**

`schedule` is built from `tripID` (the active trip) at the `buildScheduleForTrip` call, `SituationIds` likewise from `tripID`, but `TripId` is now `entryTripID`. When those differ — precisely the interlining case this PR exists for — the entry claims to be trip A while carrying trip B's stop times and alerts.

The spec calls `schedule.stopTimes` "scheduled stop times for this trip," meaning the entry's trip. It gets worse with `schedule.previousTripId`, defined as "the preceding trip in this vehicle's block": built from the active trip, that resolves to the queried-route trip — so in your own `InterlinedBlock` fixture the entry would list itself as its own predecessor.

Java derives `tripId`, `schedule`, and `situationIds` from a single `BlockTripInstance` and lets only `status.activeTripId` diverge. Moving all three onto `entryTripID` would match.

The test requests `includeSchedule=true` but never asserts on the schedule, which is why this didn't surface. An assertion that `schedule.stopTimes` belong to the trip named in `tripId` would catch it and guard the fix.

Smaller things, none blocking:

- `allServiceIDs := serviceIDs` followed by `append(allServiceIDs, prevServiceIDs...)` aliases `serviceIDs`' backing array. Harmless today because `serviceIDs` isn't read afterward, but it's a trap for the next edit — an explicit copy costs nothing.
- The `GetTripsByBlockIDs` failure both logs a `Warn` and returns a 500. Beyond the double reporting, it's the only DB lookup in this handler that hard-fails; `layoverBlocks`, `prevServiceIDs`, `nullBlockTrips`, and `GetTripsInBlock` all degrade gracefully. Worth being consistent one way or the other.
- `entryAgencyID = agencyID` uses the agency parsed from the request path rather than the queried route's actual `agency_id`, bypassing the `routeAgencyMap` the handler already built. Same answer in practice, but the map is right there.
- The comments claiming the selected trip's window "overlaps with the active trip" describe something block trips can't do. Whatever selection rule you land on, please make the comments match it — these will mislead the next reader.
- The three fixture builders are near-identical ~50-line copies differing only in CSV payload and zip name; one parameterized helper would be easier to extend.

One pre-existing issue your new test happens to walk right past: `ServiceDate` is always `todayMidnight.UnixMilli()`, but the spec says trips extending past midnight take the *previous* calendar day. `TestTripsForRouteHandler_OvernightInterlinedBlock` exercises exactly that path and asserts nothing about `serviceDate`. Not this PR's job to fix, but since you've built the fixture, adding the assertion (even as a skipped/known-failing case) would be a gift to whoever picks it up.

The structure here is good and I'd like to land it. Fix the trip selection and the `schedule`/`situationIds` binding and I'll re-review quickly.

burma-shave (approved):

PR #1272, opened 2026-08-01
Cap route-type searches after the service filter

### LINKED ISSUE #1268: stops-for-location: pin routeType truncation behaviour with a regression test
Part of #1235. Direction confirmed by @burma-shave on the card.

### Spec Requirement

Suspected Defect #1 records that legacy sets `limitExceeded = true` when the post-`routeType`-filter count exceeds `maxCount` but never truncates the list. Defect Workshop **DC-6** gives the intended behaviour: *"the list capped at `maxCount` and `limitExceeded` indicating whether additional results were discarded."*

Measured on canonical legacy 2.7.1 — `routeType=3&maxCount=2` returned **599** stops with `limitExceeded: true`.

### Current Maglev Behavior

Maglev already implements the corrected behaviour. `GetStopsForLocation`'s route-type path truncates after filtering:

```go
if len(stops) > maxCount {
	limitExceeded = true
	stops = stops[:maxCount]
}
```

Same request against maglev returns 2 stops with `limitExceeded: true`.

So there is **no production change to make here** — the deviation is already in place and is the behaviour we want. What's missing is that nothing pins it: no test covers the route-type truncation path, so a future change could silently revert it to the legacy behaviour, and the deviation isn't recorded in the spec.

### Proposed Solution

Add a regression test asserting that `routeType` + `maxCount` truncates the list and reports `limitExceeded: true`.

The spec has no Implementation Decisions section for this endpoint; per CONTRIBUTING, recording the deviation there is a reviewer task, flagging it here rather than editing the wiki.

Mister-Raggs:
Fixes #1268 (sub-issue of the #1235 spec review).

Route-type searches capped the candidate list in `GetStopsForLocation`, but stops with no service on the queried date are dropped afterwards in the handler. The cap therefore ran against a set that was still about to shrink, and the response could fall far below `maxCount` while `limitExceeded` claimed results had been withheld.

Measured on the RABA fixture (clock 2025-12-26, `lat=40.583321&lon=-122.426966&radius=5000&routeType=3`):

| `maxCount` | before | after |
|---|---|---|
| 2 | `list=0`, `limitExceeded=true` | `list=2`, `limitExceeded=true` |
| 5 | `list=1`, `limitExceeded=true` | `list=5`, `limitExceeded=true` |
| 1000 | `list=122`, `limitExceeded=false` | unchanged |

Asking for two stops returned none, out of 122 matches.

This is the ordering issue recorded as Suspected Defect #4 (*"the limit check in the geospatial path runs before route filtering… a clean reimplementation would not have this ordering issue"*), and the result contradicts DC-6's intent that the list is capped at `maxCount` with `limitExceeded` signalling discarded results.

### Approach

The cap moves downstream of the service filter, into the handler, where the surviving stops are known. Truncation still happens before the combined-ID sort, so the nearest matches are the ones kept.

Bounds searches (no `routeType`) are untouched — they keep capping upstream, which is the documented legacy ordering from Main Success Scenario step 5, not a defect.

### Note

The candidate set handed to the batch queries is now the full route-type match within the search bounds rather than a pre-capped slice, since which stops survive the service filter isn't knowable before running it. That's the same shape legacy has.

Tests cover both cap sizes and the under-cap case; the capped cases were confirmed to fail against the previous ordering.

Verified: full suite in both `sqlite_fts5` and `purego` modes, `go vet` both, race detector on `restapi`/`gtfs`, `gofmt`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved location-based stop searches with route-type filters by removing inactive stops before applying result limits.
  * Corrected result counts and limit notifications so they accurately reflect the stops available for the selected date.
  * Ensured route and agency details correspond only to stops included in the results.

* **Tests**
  * Added coverage for capped results and cases where all matching stops fit within the configured limit.


aaronbrethorst:
### Code review

Found 1 issue:

1. `GetStopsForLocation`'s documentation still claims the ORDERED_BY_CLOSEST path truncates, but this PR removes that truncation. Both the doc comment (`// ORDERED_BY_CLOSEST mode (routeTypes present): sorts by distance, filters by route type, then truncates.`) and the inline marker inside the `else` branch (`// ORDERED_BY_CLOSEST mode: sort by distance, filter by route type, then truncate.`, line 352) now describe behavior the function no longer has — `maxCount` and the returned `limitExceeded` are dead for that mode. The mode-by-mode doc comment was added deliberately in 422b22df ("refactor: split GetStopsForLocation") precisely to record this truncation ordering, so leaving it inverted is worse than having no comment; a reader of the manager can't tell the cap moved to the handler. (CONTRIBUTING.md: "every exported function, type, and package still needs a doc comment starting with its name" and "Comment only where the code itself can't explain the why.")

https://github.com/OneBusAway/maglev/blob/3760aafe3e3ed114a724524326891dc172c60b00/internal/gtfs/gtfs_manager.go#L316-L320

Also checked and found clean: the cap runs after the service filter and before the combined-ID sort, so the nearest matches survive; `results` and `resultRawStopIDs` stay length-matched so the paired `[:maxCount]` slices are safe; `> maxCount` / `[:maxCount]` has no off-by-one; `isLimitExceeded` is now only set on a real overflow; references and `collectAlertsForStops` are correctly scoped to the capped set; `GetStopsForLocation` has exactly one caller, so removing the upstream cap regresses nothing else. The uncapped `stopIDs` now handed to `GetAgenciesForStops`/`GetActiveRouteIDsForStopsOnDate` is a subset of what `GetRoutesForStops` already receives on the same request today, so it adds no new SQLite parameter-count exposure.

aaronbrethorst (approved):
Good fix. Applying the cap after the service filter is the right ordering, and I verified the details hold up: distance ordering survives the route-type filter, the cap runs before the combined-ID sort so the nearest matches are the ones kept, results and resultRawStopIDs stay length-matched, and limitExceeded now only fires on a real overflow. Merging clean against main with go vet and the restapi/gtfs suites green.

One follow-up I'm filing separately rather than holding the merge on: GetStopsForLocation's doc comment and the ORDERED_BY_CLOSEST inline comment still say that path truncates, which this PR makes untrue.

PR #1274, opened 2026-08-01
Fix references.agencies null serialization in agencies-with-coverage

### LINKED ISSUE #1273: Fix leftover gaps in `agencies-with-coverage` spec review
**Description:**

### 1. `references.agencies` serialized as `null` instead of `[]`

**Current Behavior**
The spec's Minimal Guarantees require every response array to always be
present, even when empty. When the paginated agency list came back empty
(e.g. `offset` beyond the agency count), `references.agencies` serialized as
`null` instead of `[]`:

```
GET /api/where/agencies-with-coverage.json?key=TEST&offset=1
{"data":{"list":[],"references":{"agencies":null,"routes":[],...}}}
```

**Root Cause**
`buildAgencyReferences` (`internal/restapi/reference_utils.go`) declared a nil
slice and returned it unchanged when its input was empty, overwriting the
properly initialized `[]` from `models.NewEmptyReferences()`. The same helper
is shared by `routes_for_location_handler.go`, so the bug was latent there too.

**Fix Applied**
Pre-size the slice instead of declaring it nil:
`refs := make([]models.AgencyReference, 0, len(agencies))`. No handler changes
needed since both callers route through the shared helper.

### 2. No test coverage for `maxCount` and `limitExceeded`

**Current Behavior**
No handler test exercised the `maxCount` pagination parameter — all existing
cases used `limit` instead — and `limitExceeded=true` was never asserted for
this endpoint, since the RABA test fixture has exactly one agency and can
never overflow a default-sized page.

**Fix Applied**
Added a `maxCount=1` case, plus a second-agency fixture (inserted the same way
`TestAgenciesWithCoverageHandlerZeroStopTimes` does) to exercise the
truncation path and assert `limitExceeded == true`.

ARCoder181105:
## Description

  Spec review of `GET /api/where/agencies-with-coverage.json` (#1206) found
  `references.agencies` serialized as `null` instead of `[]` when the paginated
  agency list came back empty, violating the spec's Minimal Guarantees. Root
  cause: `buildAgencyReferences` returned a nil slice unchanged on empty input,
  overwriting `models.NewEmptyReferences()`. Same helper is shared by
  `routes-for-location`, so both fixed.

  Also added missing test coverage for `maxCount` and `limitExceeded=true`,
  neither previously exercised for this endpoint.

  ## Test plan

  - [x] `go vet` (both build tag variants)
  - [x] `make test`
  - [x] `go fmt`
  - [x] Raw-body regression test confirms `"agencies":[]`, not `null`

closes #1273 

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured agency reference lists are returned as empty arrays instead of `null` when no references are available.
  * Improved pagination responses to correctly report maximum counts and whether additional results remain.
* **Tests**
  * Added coverage for empty references, maximum-count handling, and pagination limit indicators.
  * Verified behavior when requested limits are below, equal to, or above the available agency count.


ARCoder181105:
@coderabbitai full reivew

Ahmedhossamdev:
Hi @ARCoder181105 
This fixes the `references.agencies` null-vs-empty-array bug by changing `buildAgencyReferences` to allocate a non-nil slice. The fix is correct and matches the spec, and the shared helper covers routes-for-location too. I verified the wire format against the live Java server, both now return `[]` on the empty path. Overall LGTM, just two optional test nits inline. Lint and the full test suite pass.

internal/restapi/agencies_with_coverage_handler_test.go
```diff
@@ -63,6 +63,49 @@ func TestAgenciesWithCoverageHandlerPagination(t *testing.T) {
...
+	defer api.Shutdown()
+
+	resp, model := callAPIHandler[CoverageResponse](t, api, "/api/where/agencies-with-coverage.json?key=TEST&offset=1")
+
+	assert.Equal(t, http.StatusOK, resp.StatusCode)
+	assert.NotNil(t, model.Data.References.Agencies)
+	assert.Empty(t, model.Data.References.Agencies)
+}
+
+func TestAgenciesWithCoverageHandlerMaxCount(t *testing.T) {
```
Ahmedhossamdev: Optional. With a 1-agency fixture, `maxCount=1` returning 1 result can't distinguish "maxCount honored" from "maxCount ignored." The `TestAgenciesWithCoverageHandlerLimitExceeded` test below (lines 93-109) is the one that actually proves the cap works, by inserting a second agency and asserting `limitExceeded:true`. This test is harmless coverage, but consider merging it as a sub-case of the `LimitExceeded` test to reduce redundancy. Low priority.

internal/restapi/agencies_with_coverage_handler_test.go
```diff
@@ -63,6 +63,49 @@ func TestAgenciesWithCoverageHandlerPagination(t *testing.T) {
...
 	assert.Len(t, model.Data.List, 0)
 	assert.False(t, model.Data.LimitExceeded)
+	assert.NotNil(t, model.Data.References.Agencies)
+	assert.Empty(t, model.Data.References.Agencies)
+}
+
+// TestAgenciesWithCoverageHandlerEmptyReferencesNotNull pins the wire format:
+// references.agencies must decode as an empty slice, never nil, so it
+// serializes as `[]` rather than `null`.
+func TestAgenciesWithCoverageHandlerEmptyReferencesNotNull(t *testing.T) {
```
Ahmedhossamdev, line 71: Optional. This test hits the same `offset=1` input already covered by the new assertions in `TestAgenciesWithCoverageHandlerPagination` (lines 66-67). The standalone version adds a status-code check and a good doc comment explaining the intent, which is valuable. If you want to keep both, that's fine. If you'd rather consolidate, the assertions in `Pagination` plus the doc comment here could merge into one. Low priority.

internal/restapi/agencies_with_coverage_handler_test.go
ARCoder181105: Done

internal/restapi/agencies_with_coverage_handler_test.go
ARCoder181105: Done

burma-shave (approved):

PR #1275, opened 2026-08-01
Honor latSpan and lonSpan in routes-for-location

### LINKED ISSUE #1226: routes-for-location: Bounding box spans are ignored when radius is omitted
### Description

**Current behavior:**
`loc.Radius` is unconditionally overwritten with a default value (500m or 10km) if it is 0 in `routes_for_location_handler.go`. This causes any provided `latSpan` and `lonSpan` parameters to be completely ignored by the downstream bounding box calculation in `BoundsFromParams`.

> Otherwise, if both `latSpan > 0` and `lonSpan > 0` are given, the box spans ±(`latSpan`/2) degrees of latitude and ±(`lonSpan`/2) degrees of longitude from the centre.

**What to change:**
1. In `internal/restapi/routes_for_location_handler.go`, check if `latSpan` and `lonSpan` are provided before overriding `loc.Radius`.
2. Only apply the default radius fallback if `loc.Radius == 0` AND both `loc.LatSpan == 0` and `loc.LonSpan == 0`.

ARCoder181105:
Fixes #1226

`routes-for-location` accepted `latSpan`/`lonSpan`, then ignored them , a 10km × 10km box request got a 600m box.

`BoundsFromParams` sizes the box from the spans only when no radius is set. The handler assigned the default radius before calling it, so the span branch was unreachable.

## Change

Apply the default radius only when neither a radius nor a usable pair of spans was supplied.

Guard is `Radius == 0 && !(LatSpan > 0 && LonSpan > 0)`, not the `LatSpan == 0 && LonSpan == 0` the issue specifies. With one span plus a `query`, the literal version leaves `Radius` at 0 and `BoundsFromParams` falls back to 600m instead of the specced 10km — its fallback isn't query-aware.

## Tests

Replaced `TestRoutesForLocationLatSpanAndLonSpan` with a table-driven `TestRoutesForLocationBoundingBoxSizing`. The old test used a 0.01° box, near-identical to the default-radius box at the fixture latitude, so it passed whether or not spans were honoured.

Four cases, each returning a different result than the default box: spans widening the box, spans narrowing it, radius taking precedence over spans, and a lone `latSpan` falling back to the 10km query radius.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved location-based route searches by sanitizing query input.
  * Corrected radius selection when latitude and longitude bounds are provided.
  * Improved fallback behavior when only one bounding-box span is specified.

* **Tests**
  * Expanded coverage for bounding-box sizing, radius precedence, and default-radius behavior.


Ahmedhossamdev (commented):
LGTM!

PR #1277, opened 2026-08-01
Fix: `includeTrip` / `includeStatus` interaction now matches the specification

### LINKED ISSUE #1276: trip-for-vehicle: add active trip to references from the status path
### Description

The `trips-for-vehicle` endpoint does not follow the specification when `includeTrip` and `includeStatus` are used together.

According to **Spec Step 4** and **Extension 4b**:

> When the `status` block is present, the active trip **must always** be added to `data.references`. The `includeTrip` flag only controls whether the trip is included in the response body, meaning it only changes behavior when `includeStatus=false`.

Currently, Maglev only populates `references.Trips` when `includeTrip=true`:

- `trip_for_vehicle_handler.go:169`
- `trip_details_handler.go:32`

To compensate, the endpoint currently defaults `includeTrip=true`, while the specification defines the default as `false`.

These two incorrect behaviors cancel each other out with the default request, but diverge as soon as either parameter is explicitly set.

### Current behavior

Request:

```text
includeTrip=false&includeStatus=true
```

| Implementation | Trips in `data.references` |
| -------------- | -------------------------- |
| Java           | ✅ 1 |
| Maglev         | ❌ 0 |

### Java verification

Verified in `TripStatusBeanServiceImpl`:

- Whenever a status bean is created, the active trip is added to `data.references`.
- This is independent of the `includeTripBean` flag.

### Expected behavior

- If `status != nil`, always add the active trip (and its route) to `data.references`.
- Change the default value of `includeTrip` to `false` to match the specification.

> **Note:** These two changes must land together. Changing the default alone would remove trip references from the default response.

### Tests

Update `TestTripForVehicleHandler_IncludeToggles`:

- Invert the existing `includeTrip=false omits trip references` test, since that behavior is incorrect.
- Add a new test for:

```text
includeTrip=false&includeStatus=false
```

In this case, `data.references` should correctly contain no trip or route references, matching Extension 4a.

Ahmedhossamdev:
Fixes: #1276 

### Description

The `trip-for-vehicle` endpoint did not follow the specification when `includeTrip` and `includeStatus` were used together.

According to **Spec Step 4** and **Extension 4b**:

> When the `status` block is present, the active trip **must always** be added to `data.references`. The `includeTrip` flag only controls whether the trip is included in the response body, meaning it only changes behavior when `includeStatus=false`.

Maglev only populated `references.Trips` when `includeTrip=true`:

- `trip_for_vehicle_handler.go:169`
- `trip_details_handler.go:32`

To compensate, the endpoint defaulted `includeTrip=true`, while the specification defines the default as `false`.

These two incorrect behaviors cancelled each other out with the default request, but diverged as soon as either parameter was explicitly set.

### Current behavior

Request:

```text
includeTrip=false&includeStatus=true
```

| Implementation | Trips in `data.references` |
| -------------- | -------------------------- |
| Java           | ✅ 1 |
| Maglev         | ❌ 0 |

### Java verification

Verified in `TripStatusBeanServiceImpl`:

- Whenever a status bean is created, the active trip is added to `data.references`.
- This is independent of the `includeTripBean` flag.

### Changes

- If `status != nil`, the active trip **and its route** are added to `data.references`.
- The default value of `includeTrip` is now `false`, matching the specification.

`parseTripParams` took its per-endpoint default as a positional bool. A second one would have been unreadable at the call site, so the defaults now travel in a `TripParamDefaults` struct:

```text
api.parseTripParams(r, TripParamDefaults{IncludeTrip: true, IncludeSchedule: true}, loc)   // trip-details
api.parseTripParams(r, TripParamDefaults{}, loc)                                           // trip-for-vehicle
```

The new `routeReferenceByID` helper lives in `reference_utils.go` and reuses the existing `buildRouteModels`.

> **Note:** These two changes landed together. Changing the default alone would have removed trip references from the default response.

### Resulting behavior

| Request | Trips in `data.references` |
| ------- | -------------------------- |
| default (`includeStatus=true`) | ✅ 1 |
| `includeTrip=false&includeStatus=true` | ✅ 1 |
| `includeTrip=true&includeStatus=false` | ✅ 1 |
| `includeTrip=false&includeStatus=false` | ✅ 0 |

### Tests

`TestTripForVehicleHandler_IncludeToggles`:

- Inverted the existing `includeTrip=false omits trip references` test, since that behavior was incorrect.

New `TestTripForVehicleHandler_TripReferences` covers the three-way interaction:

- `includeTrip=false` still yields trip refs via the status path.
- `includeTrip=false&includeStatus=false` yields no trip or route references, matching **Extension 4a**.
- `includeTrip=true&includeStatus=false` yields trip refs, which is the only case where the flag matters.

These live in a separate test function rather than as extra subtests: `createTestApi` sets `RateLimit: 5`, and six requests against one API instance returns `429`.

### Unrelated fix included

The first commit removes a `MockAddTrip` call from `setupTestApiWithMockVehicle`.

`MockAddTrip` issues an `INSERT OR REPLACE`, and the helper passed it the ID of a trip it had just read out of the fixture database — replacing a real trip row with a stub and wiping its `block_id`, `shape_id` and `service_id`. The test database is shared across the package via `testDbSetupOnce`, so the damage persisted for every test that ran afterwards, and it grew with each caller: the replace moves the row to a new rowid, so the next caller read and clobbered a different trip.

Two callers were enough to break `TestGetFirstStopOfNextTripInBlock_WithBlockContinuation`, which needs a trip whose block still has a second trip in it. The call was redundant to begin with — the trip already exists in the fixture database, which is where the helper read it from.

It is in this PR only because the new test function cannot pass without it.

### Handler consistency

`parseTripParams` is shared with `trip-details`, whose spec **does** default `includeTrip` to `true`. That default is preserved and only `trip-for-vehicle` changes, but the shared helper is worth a second look during review.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved trip-for-vehicle responses to include accurate trip and route references when requested.
  * Refined endpoint-specific defaults for trip and schedule details, providing more consistent results across trip-related requests.
  * Improved handling of missing or invalid route lookups without disrupting otherwise successful responses.
  * Corrected interpretation of Unix millisecond timestamps in trip parameters.
  * Improved validation and fallback handling for boolean query parameters.
* **Tests**
  * Expanded coverage for inclusion parameters, defaults, invalid values, timestamp parsing, and trip-reference combinations.

### DIFF
--- a/internal/restapi/reference_utils.go
+++ b/internal/restapi/reference_utils.go
@@ -83,6 +83,22 @@ func buildRouteModels(ctx context.Context, agencyID string, routes []gtfsdb.Rout
 	return modelRoutes, nil
 }
 
+// routeReferenceByID builds the Route reference for a single route, looked up by
+// its bare (agency-less) route ID.
+func (api *RestAPI) routeReferenceByID(ctx context.Context, agencyID string, routeID string) (models.Route, error) {
+	route, err := api.GtfsManager.GtfsDB.Queries.GetRoute(ctx, routeID)
+	if err != nil {
+		return models.Route{}, err
+	}
+
+	routeModels, err := buildRouteModels(ctx, agencyID, []gtfsdb.Route{route})
+	if err != nil {
+		return models.Route{}, err
+	}
+
+	return routeModels[0], nil
+}
+
 // routeReferencesForStops deduplicates the rows returned by GetRoutesForStops into Route
 // reference objects, keyed by each row's own agency ID so results spanning multiple
 // agencies are handled correctly.
--- a/internal/restapi/reference_utils.go
+++ b/internal/restapi/reference_utils.go
@@ -146,6 +146,25 @@ func (api *RestAPI) routeReferenceForTrip(ctx context.Context, routeID string, s
 	return api.routeReferenceByID(ctx, routeID)
 }
 
+// appendRouteAgencyReference adds a route's own agency to references when it is not
+// the agency the request was scoped to, so the agencyId a cross-agency route
+// reference carries resolves against references.agencies. A lookup that fails costs
+// the reference rather than the response, as with the route itself.
+func (api *RestAPI) appendRouteAgencyReference(ctx context.Context, references *models.ReferencesModel, routeAgencyID, requestAgencyID string) {
+	if routeAgencyID == requestAgencyID {
+		return
+	}
+
+	routeAgency, err := api.GtfsManager.GtfsDB.Queries.GetAgency(ctx, routeAgencyID)
+	if err != nil {
+		api.Logger.Warn("failed to fetch route agency for reference",
+			"agencyID", routeAgencyID, "error", err)
+		return
+	}
+
+	references.Agencies = append(references.Agencies, models.AgencyReferenceFromDatabase(&routeAgency))
+}
+
 // mergeParentRouteReferences appends any parent-station routes not already present in
 // routes, so parent stop references never point at an absent route.
 func mergeParentRouteReferences(routes []models.Route, parentRoutes map[string]gtfsdb.GetRoutesForStopsRow) []models.Route {
--- a/internal/restapi/reference_utils.go
+++ b/internal/restapi/reference_utils.go
@@ -133,6 +133,19 @@ func routeReferencesForStops(routesRows []gtfsdb.GetRoutesForStopsRow) []models.
 	return utils.MapValues(routesMap)
 }
 
+// routeReferenceForTrip returns the Route reference for the route a trip runs on.
+// A route already resolved for the trip's stop references is preferred over a fresh
+// lookup, since a trip's own route commonly also serves its closest and next stops.
+func (api *RestAPI) routeReferenceForTrip(ctx context.Context, routeID string, stopRoutes map[string]gtfsdb.GetRoutesForStopsRow) (models.Route, error) {
+	for _, row := range stopRoutes {
+		if row.ID == routeID {
+			return routeReferenceFromStopRow(row), nil
+		}
+	}
+
+	return api.routeReferenceByID(ctx, routeID)
+}
+
 // mergeParentRouteReferences appends any parent-station routes not already present in
 // routes, so parent stop references never point at an absent route.
 func mergeParentRouteReferences(routes []models.Route, parentRoutes map[string]gtfsdb.GetRoutesForStopsRow) []models.Route {
--- a/internal/restapi/reference_utils.go
+++ b/internal/restapi/reference_utils.go
@@ -84,19 +84,36 @@ func buildRouteModels(ctx context.Context, agencyID string, routes []gtfsdb.Rout
 }
 
 // routeReferenceByID builds the Route reference for a single route, looked up by
-// its bare (agency-less) route ID.
-func (api *RestAPI) routeReferenceByID(ctx context.Context, agencyID string, routeID string) (models.Route, error) {
+// its bare (agency-less) route ID. The reference is labelled with the route's own
+// agency, which need not be the agency of the entity that referenced it.
+func (api *RestAPI) routeReferenceByID(ctx context.Context, routeID string) (models.Route, error) {
 	route, err := api.GtfsManager.GtfsDB.Queries.GetRoute(ctx, routeID)
 	if err != nil {
 		return models.Route{}, err
 	}
 
-	routeModels, err := buildRouteModels(ctx, agencyID, []gtfsdb.Route{route})
+	routeModels, err := buildRouteModels(ctx, route.AgencyID, []gtfsdb.Route{route})
 	if err != nil {
 		return models.Route{}, err
 	}
 
 	return routeModels[0], nil
+}
+
+// routeReferenceFromStopRow builds a Route reference from a stops-derived route row,
+// labelled with the route's own agency rather than the agency that was queried, so
+// results spanning multiple agencies stay resolvable.
+func routeReferenceFromStopRow(row gtfsdb.GetRoutesForStopsRow) models.Route {
+	return models.NewRoute(
+		utils.FormCombinedID(row.AgencyID, row.ID),
+		row.AgencyID,
+		nulls.StringOrEmpty(row.ShortName),
+		nulls.StringOrEmpty(row.LongName),
+		nulls.StringOrEmpty(row.Desc),
+		models.RouteType(row.Type),
+		nulls.StringOrEmpty(row.Url),
+		nulls.StringOrEmpty(row.Color),
+		nulls.StringOrEmpty(row.TextColor))
 }
 
 // routeReferencesForStops deduplicates the rows returned by GetRoutesForStops into Route
@@ -110,16 +127,7 @@ func routeReferencesForStops(routesRows []gtfsdb.GetRoutesForStopsRow) []models.
 			continue
 		}
 
-		routesMap[combinedRouteID] = models.NewRoute(
-			combinedRouteID,
-			row.AgencyID,
-			nulls.StringOrEmpty(row.ShortName),
-			nulls.StringOrEmpty(row.LongName),
-			nulls.StringOrEmpty(row.Desc),
-			models.RouteType(row.Type),
-			nulls.StringOrEmpty(row.Url),
-			nulls.StringOrEmpty(row.Color),
-			nulls.StringOrEmpty(row.TextColor))
+		routesMap[combinedRouteID] = routeReferenceFromStopRow(row)
 	}
 
 	return utils.MapValues(routesMap)
--- a/internal/restapi/trip_details_handler.go
+++ b/internal/restapi/trip_details_handler.go
@@ -32,99 +32,116 @@ type TripParamDefaults struct {
 	IncludeSchedule bool
 }
 
+// Layouts accepted by the serviceDate and time params, alongside Unix millis.
+const (
+	serviceDateLayout = "2006-01-02"          // e.g. "2024-06-15"
+	tripTimeLayout    = "2006-01-02_15-04-05" // e.g. "2024-06-15_14-30-00"
+)
+
+// parseEpochOrLayoutTime parses a param accepting either a Unix timestamp in
+// milliseconds or a timestamp in the given layout. A missing param yields a nil
+// time and no error; ok is false only when a supplied value matches neither form.
+func parseEpochOrLayoutTime(value, layout string, loc *time.Location) (parsed *time.Time, ok bool) {
+	if value == "" {
+		return nil, true
+	}
+
+	if epochMillis, err := strconv.ParseInt(value, 10, 64); err == nil {
+		fromEpoch := time.Unix(epochMillis/1000, 0)
+		return &fromEpoch, true
+	}
+
+	fromLayout, err := time.ParseInLocation(layout, value, loc)
+	if err != nil {
+		return nil, false
+	}
+
+	return &fromLayout, true
+}
+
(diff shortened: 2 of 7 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (dismissed):
Thanks for this — the spec work here is solid. I checked the defaults against `testdata/openapi.yml` myself and they're right: `includeTrip` defaults to false for trip-for-vehicle (line 806) and to true for trip-details, and the `TripParamDefaults` struct expresses that cleanly. The `parseTripParams` refactor is behavior-preserving, and putting `routeReferenceByID` in `reference_utils.go` is the correct home for it per CONTRIBUTING's Code Reuse section. Flagging the shared-helper impact on `trip-details` in the PR description was the right call too.

One blocker before this can land.

**A missing route reference now 404s the entire request.** In `trip_for_vehicle_handler.go`, the new `routeReferenceByID` lookup turns `sql.ErrNoRows` into `sendNotFound`. That lookup is reference enrichment, not the primary resource — the vehicle and the trip were both found successfully by that point, and the trip reference has already been appended to `references.Trips` and is then thrown away. For a feed with a dangling `trips.route_id`, this endpoint stops serving trip details entirely rather than serving them with one route reference absent.

CONTRIBUTING's `ErrNoRows` → 404 rule is about the resource the request names, not about entities you're dereferencing into the references block. The rest of the codebase already treats those differently: `BuildRouteReferences` goes through the batch `GetRoutesByIDs`, where an absent row simply doesn't appear in the result.

The fix is small — on `ErrNoRows`, log and skip the route reference (keeping the trip reference you already built) and let the response go out as a 200. Any other error should still be a 500, which you have right.

Also worth knowing for sequencing: #1279 has landed and #1281 is ahead of you in the queue, both touching this same references block. You'll want to rebase onto `main` once #1281 goes in.

Happy to re-review as soon as the 404 path is sorted.

internal/restapi/reference_utils.go
```diff
@@ -83,6 +83,22 @@ func buildRouteModels(ctx context.Context, agencyID string, routes []gtfsdb.Rout
...
 
+// routeReferenceByID builds the Route reference for a single route, looked up by
+// its bare (agency-less) route ID.
+func (api *RestAPI) routeReferenceByID(ctx context.Context, agencyID string, routeID string) (models.Route, error) {
+	route, err := api.GtfsManager.GtfsDB.Queries.GetRoute(ctx, routeID)
+	if err != nil {
+		return models.Route{}, err
+	}
+
+	routeModels, err := buildRouteModels(ctx, agencyID, []gtfsdb.Route{route})
```
burma-shave: `routeReferenceByID` builds the route reference using the caller's `agencyID` parameter instead of the row's own `route.AgencyID` (via `buildRouteModels`, which always stamps whatever `agencyID` it's passed). This reintroduces the same class of bug fixed in e8de6a5 ("Fix route agency mis-prefixing in trip-for-vehicle references") for the stops-derived route path just above (`routeReferencesForStops`, which correctly keys/labels by `row.AgencyID`).

On a feed where the trip's route belongs to a different agency than the vehicle's own `agencyID` (multi-agency GTFS), this would re-prefix and mislabel the route with the wrong agency ID, producing a `references.routes` entry a client can't actually resolve.

Suggest building the reference from `route.AgencyID` rather than the passed-in `agencyID`, matching `routeReferencesForStops`.

internal/restapi/trip_for_vehicle_handler.go
```diff
@@ -173,8 +174,25 @@ func (api *RestAPI) tripForVehicleHandler(w http.ResponseWriter, r *http.Request
 			utils.FormCombinedID(agencyID, trip.ShapeID.String),
 		)
 		references.Trips = append(references.Trips, *tripRef)
+
+		routeRef, err := api.routeReferenceByID(ctx, agencyID, trip.RouteID)
```
burma-shave: This always issues a fresh `GetRoute` query for the trip's active route, even though that same route is often already present in `routeRefs` from `BuildStopReferencesAndRouteIDsForStops` a few lines above (the trip's own route commonly also serves its closest/next stop). Worth checking `routeRefs` for the combined route ID before calling `routeReferenceByID`, to skip the redundant DB round trip in the common case.

internal/restapi/trip_details_handler.go
```diff
@@ -24,100 +24,124 @@ type TripParams struct {
...
-			params.IncludeTrip = val
-		} else {
-			fieldErrors["includeTrip"] = []string{"must be a boolean value (true/false)"}
-		}
+	return &fromLayout, true
+}
+
+// parseIncludeParam reads a boolean include* param, falling back to fallback when
+// the request omits it. ok is false when a supplied value is not a boolean.
+func parseIncludeParam(value string, fallback bool) (parsed bool, ok bool) {
```
burma-shave: `parseIncludeParam` is a generic boolean query-param parser with fallback — per CONTRIBUTING.md's parameter-parsing convention, this shape belongs in `internal/utils/api.go` alongside the other `Parse*` helpers (`ParseFloatParam`, `ParseTimeParameter`, etc.) rather than living locally in this handler file.

burma-shave (changes requested):
Reviewed against legacy Java (`TripStatusBeanServiceImpl` / `BeanFactoryV2`, traced directly rather than taken from the wiki) and against `/code-review medium`, CodeRabbit, and SonarQubeCloud findings on this PR.

The core fix is spec-correct and well-tested: the active trip (and now its route) is added to `data.references` whenever a `status` block is built, independent of `includeTrip`, and `trip-for-vehicle`'s `includeTrip` default now correctly matches the spec's documented `false`. Confirmed directly against `TripStatusBeanServiceImpl.getBlockLocationAsStatusBean` and `BeanFactoryV2.getTripStatus`, both of which add the active trip to references with no dependency on `includeTrip`.

Found one correctness bug and two minor cleanup opportunities in the new code — see inline comments.

This PR's own description invokes `trip-for-vehicle.md`'s Extension 4b directly. Tracing that same `includeTrip`/`status` interaction through the Java source (`TripStatusBeanServiceImpl.getBlockLocationAsStatusBean` L294-295, `BeanFactoryV2.getTripStatus` L528-537) shows the sibling `trip-details` endpoint has the identical exemption — the active trip is added to references whenever a `status` block is built, regardless of `includeTrip` — and `trip_details_handler.go`'s existing (unchanged) code already implements it correctly. `trip-details.md`'s Extension 4c didn't document this exemption at all, describing `includeTrip=false` as suppressing the trip record with no carve-out.

No code change is needed here, since `trip_details_handler.go` already behaves correctly — I've corrected the wiki text on our end to match `trip-for-vehicle.md`'s Extension 4b language, so the spec stops contradicting itself across two endpoints backed by the same Java code path. Flagging it here for visibility since it surfaced directly from this PR's own change.

internal/restapi/trip_for_vehicle_handler.go
```diff
@@ -160,25 +161,38 @@ func (api *RestAPI) buildTripForVehicleReferences(ctx context.Context, agencyID
...
-			route.Desc.String,
-			models.RouteType(route.Type),
-			route.Url.String,
-			route.Color.String,
-			route.TextColor.String)
+		routeRefs[combinedID] = routeReferenceFromStopRow(route)
 	}
-	references.Routes = utils.MapValues(routeRefs)
-
 	references.Agencies = append(references.Agencies, models.AgencyReferenceFromDatabase(&agency))
```
burma-shave, line 165: `buildTripForVehicleReferences` only ever appends the vehicle's own agency to `references.Agencies`, even though `routeReferenceForTrip`/`routeReferenceByID` (added in this PR) can now resolve the active trip's route from a *different* agency than the vehicle's own.

`TestTripForVehicleHandler_RouteFromAnotherAgency` (added in this PR) already exercises this exact scenario — a route from `OTHERAGENCY` — and confirms the route reference correctly carries the `OTHERAGENCY_` prefix. But the test never asserts on `references.Agencies`, so it doesn't catch that `OTHERAGENCY` itself is missing from that list. A client resolving `references.routes[0].agencyId` against `references.agencies` won't find a matching entry.

`arrival_and_departure_for_stop_handler.go:434-443` already has the right pattern for this — fetch and append the route's own agency when it differs from the primary one:
```go
if route.AgencyID != stopAgency.ID {
    routeAgency, err := api.GtfsManager.GtfsDB.Queries.GetAgency(ctx, route.AgencyID)
    if err == nil {
        references.Agencies = append(references.Agencies, models.AgencyReferenceFromDatabase(&routeAgency))
    } else {
        api.Logger.Warn("failed to fetch route agency for reference", "agencyID", route.AgencyID, "error", err)
    }
}
```
Worth applying the same here, and extending `TestTripForVehicleHandler_RouteFromAnotherAgency` to assert `references.Agencies` contains the route's agency.

internal/restapi/trip_details_handler.go
```diff
@@ -24,100 +24,104 @@ type TripParams struct {
...
+	}
+
+	if timeParam, ok := parseEpochOrLayoutTime(query.Get("time"), tripTimeLayout, parseLoc); ok {
+		params.Time = timeParam
+	} else {
+		fieldErrors["time"] = []string{"must be a valid Unix timestamp in milliseconds or a datetime in yyyy-MM-dd_HH-mm-ss format"}
 	}
 
-	params.VehicleID = r.URL.Query().Get("vehicleId")
+	includeParams := map[string]*bool{
```
burma-shave: Minor readability nit: this map of `*bool` pointers exists to loop over three fixed, compile-time-known fields. Three explicit calls read top-to-bottom without the indirection of dereferencing a pointer and cross-referencing a map key back to a struct field:
```go
params.IncludeTrip, fieldErrors = utils.ParseBoolParam(query, "includeTrip", params.IncludeTrip, fieldErrors)
params.IncludeSchedule, fieldErrors = utils.ParseBoolParam(query, "includeSchedule", params.IncludeSchedule, fieldErrors)
params.IncludeStatus, fieldErrors = utils.ParseBoolParam(query, "includeStatus", params.IncludeStatus, fieldErrors)
```
Not a bug, just a step down in directness for a fixed, small set of fields — per CONTRIBUTING's readability guidance on avoiding unnecessary indirection.

burma-shave (changes requested):
Reviewed the includeTrip/includeStatus fix for trip-for-vehicle. The core fix (active trip + route always added to references when status is present, includeTrip default corrected to false) is spec-consistent and well tested — verified against the legacy Java implementation and confirmed no client impact across Wayfinder, JS SDK, iOS, and Android.

Two things worth addressing before merge, both scoped to this PR's own diff:

1. A correctness gap in the new cross-agency route handling (blocking).
2. A readability nit in the new include-param parsing loop (non-blocking, but worth cleaning up).

See inline comments.

---

Process note: this branch was force-pushed after earlier review comments (CodeRabbit and a human reviewer) had already landed on it. All of those prior comments are now anchored to commit `d1ce9c0`, which isn't an ancestor of the current head (`f8c2762`) — the history was rewritten out from under them. Everything they flagged does still appear addressed in the current code, but confirming that required re-checking each one against live source rather than trusting the "addressed in commit X" annotations, since those referenced commits are dangling too. Per CONTRIBUTING.md: "Add new commits when addressing requested changes — don't rewrite already-reviewed history. Rebasing or force-pushing after review has started means the whole PR has to be re-reviewed." Please avoid force-pushing this branch again once review is underway — use merge commits or additional commits to resolve conflicts instead, so reviewers don't have to redo work that's already been done.

burma-shave (approved):

PR #1279, opened 2026-08-01
## Fix: schedule stops are now present in `data.references.stops`

### LINKED ISSUE #1278: trip-for-vehicle: add schedule stops to references when includeSchedule=true
### Description

Spec **Extension 4c** states:

> Each stop in the schedule is added to `data.references.stops`.

Currently, Maglev only passes `closestStop` and `nextStop` into `BuildStopReferencesAndRouteIDsForStops`:

- `trip_for_vehicle_handler.go:130-152`

As a result, the remaining `stopId`s in `schedule.stopTimes` are missing from `data.references.stops`, making them impossible to resolve from the response.

### Current behavior

Request:

```text
includeSchedule=true
```

| Implementation | Stops in `data.references` | Routes in `data.references` |
| -------------- | -------------------------: | --------------------------: |
| Java           | 42                         | 14                          |
| Maglev         | 2                          | 5                           |

The schedule itself contains **43 stop times**, but only the `closestStop` and `nextStop` are included in the stop references.

### Java verification

Verified against the live reference server.

### Expected behavior

When the schedule block is included:

- Collect every `stopId` from `schedule.stopTimes`.
- Merge those stop IDs with `closestStop` and `nextStop`.
- Populate `data.references.stops` with all referenced stops.

### Tests

Add a test verifying that when:

```text
includeSchedule=true
```

every `schedule.stopTimes[].stopId` can be resolved from `data.references.stops`.

Ahmedhossamdev:
Fixes: #1278

### Description

The `trip-for-vehicle` endpoint did not follow the specification when `includeSchedule=true` was used.

According to **Extension 4c**:

> The `schedule` block is populated with the active trip's stop times, its timezone, and (when applicable) IDs for the preceding and following trips in the same block. Each stop in the schedule is added to `data.references.stops`.

Maglev collected only the two stops named by the `status` block — `closestStop` and `nextStop` — and passed those to the reference builder:

- `trip_for_vehicle_handler.go:133-152`

Every `stopId` in `schedule.stopTimes` was therefore undereferenceable: the client received a stop ID with no matching entry in `data.references.stops`.
Extension 4c requires every stop in the schedule block to appear in data.references.stops. Only the status block's closest and next stops were collected, so with includeSchedule=true each stopId in schedule.stopTimes was undereferenceable: the reference server returns 42 stops and 14 routes for the same request where Maglev returned 2 and 5.

Collect the referenced stop IDs in one place instead of extracting them inline per status field, and include the schedule's stops. The existing reference builder already dedupes and batches the lookups, so the extra IDs cost one widened query rather than a query per stop.
### Current behavior

Request:

```text
includeSchedule=true
```

The schedule block returns 43 stop times.

| Implementation | Stops in `data.references` | Routes in `data.references` |
| -------------- | -------------------------- | --------------------------- |
| Java           | ✅ 42 | ✅ 14 |
| Maglev         | ❌ 2  | ❌ 5  |

### Java verification

Verified against the live reference server, which runs the same static GTFS bundle and the same GTFS-RT feed as the local config.

### Changes

- The stops named in `schedule.stopTimes` are now collected alongside the status block's `closestStop` and `nextStop`.
- The two inline extractions are replaced by a single `referencedStopIDs` helper, which converts the entry's combined `{agencyID}_{stopID}` identifiers into the bare IDs the reference builders expect.

`BuildStopReferencesAndRouteIDsForStops` already deduplicates its input and batches the underlying lookups via `GetStopsByIDs` / `GetRoutesForStops`, so the additional IDs widen one query rather than adding a query per stop.

Route references improve as a side effect: they are derived from the stops, so the routes serving the schedule's stops now appear too.

### Resulting behavior

| Request | Schedule stops dereferenceable |
| ------- | ------------------------------ |
| `includeSchedule=true` | ✅ all |
| `includeSchedule=false` (default) | ✅ n/a, no schedule block |

## Summary by CodeRabbit

* **Bug Fixes**
  * Trip responses now include references for all stops listed in the schedule, including stops beyond the current trip status.
  * Stop references are normalized consistently, and invalid stop identifiers are handled correctly.

* **Tests**
  * Added coverage confirming scheduled stops include corresponding stop-time and reference data.

### DIFF
--- a/internal/restapi/trip_for_vehicle_handler.go
+++ b/internal/restapi/trip_for_vehicle_handler.go
@@ -130,26 +130,12 @@ func (api *RestAPI) tripForVehicleHandler(w http.ResponseWriter, r *http.Request
 
 	agencyModel := models.AgencyReferenceFromDatabase(&agency)
 
-	stopIDs := []string{}
-
-	if status != nil {
-		if status.ClosestStop != "" {
-			_, closestStopID, err := utils.ExtractAgencyIDAndCodeID(status.ClosestStop)
-			if err != nil {
-				api.serverErrorResponse(w, r, err)
-				return
-			}
-			stopIDs = append(stopIDs, closestStopID)
-		}
-		if status.NextStop != "" {
-			_, nextStopID, err := utils.ExtractAgencyIDAndCodeID(status.NextStop)
-			if err != nil {
-				api.serverErrorResponse(w, r, err)
-				return
-			}
-			stopIDs = append(stopIDs, nextStopID)
-		}
+	stopIDs, err := referencedStopIDs(status, schedule)
+	if err != nil {
+		api.serverErrorResponse(w, r, err)
+		return
 	}
+
 	stops, uniqueRouteMap, err := BuildStopReferencesAndRouteIDsForStops(api, ctx, agencyID, stopIDs)
 	if err != nil {
 		api.serverErrorResponse(w, r, err)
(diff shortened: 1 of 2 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Nice, focused fix. The helper reuses the batching/deduping reference builder rather than hand-rolling lookups, and the table-driven unit test covers every branch including the malformed-ID error path. Verified locally: merges clean against main, go vet passes, and the restapi/gtfs suites are green.

PR #1281, opened 2026-08-02
Fix/trip for vehicle include references

### LINKED ISSUE #1280: trip-for-vehicle: honour includeReferences=false
Spec: "When `false`, the `data.references` object is omitted from the response." Maglev returns it fully populated. The shared helper
[`ShouldIncludeReferences`](internal/restapi/reference_utils.go#L308) already exists and is used by `stop_handler.go` — `tripForVehicleHandler` never calls it.

Live, `includeReferences=false`: reference server returns all reference arrays empty; Maglev returns `agencies: 1, routes: 5, stops: 2, trips: 1`.

**Java verification:** confirmed against the live reference server.

**Fix:** call `ShouldIncludeReferences(r)` and skip reference building when
false. Match the existing convention (`routes_for_agency_handler.go` notes the
block stays present but empty) rather than omitting the key outright.

Ahmedhossamdev:
Fixes: #1280 

## Summary by CodeRabbit

- **Bug Fixes**
  - Trip-for-vehicle responses now preserve the trip entry when reference inclusion is disabled.
  - Reference collections, including agencies, routes, trips, stops, situations, and stop times, are returned as present but empty when references are not requested.
  - Responses now complete reliably after reference processing succeeds.

### DIFF
--- a/internal/restapi/trip_for_vehicle_handler.go
+++ b/internal/restapi/trip_for_vehicle_handler.go
@@ -1,12 +1,14 @@
 package restapi
 
 import (
+	"context"
 	"database/sql"
 	"errors"
 	"net/http"
 	"strconv"
 	"time"
 
+	"maglev.onebusaway.org/gtfsdb"
 	"maglev.onebusaway.org/internal/models"
 	"maglev.onebusaway.org/internal/utils"
 )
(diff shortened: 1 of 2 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Approved on the merits — this is the right shape. It reuses `ShouldIncludeReferences` rather than re-deriving the flag, mirrors the empty-but-present convention already established in `trip_handler.go`, and moves the reference-building block verbatim so the `includeReferences=true` path is provably untouched. The test earns its place too: asserting every collection serializes as `[]` rather than `null` is exactly the part that would regress on the wire and wouldn't be caught by a shape-only check.

Heads up on ordering: #1279 just landed on `main` and restructured the same references block in `trip_for_vehicle_handler.go`, so this branch now has conflicts. Please merge `main` in and resolve.

One thing to watch in the resolution: #1279 added a `referencedStopIDs(status, schedule)` helper so that schedule stops end up dereferenceable in `references.stops`. Your `buildTripForVehicleReferences` currently only takes `status`, so the merge needs to thread `schedule` through as well — otherwise resolving the conflict will silently drop #1279's fix.

Once it's mergeable it's good to go; no re-review needed unless the resolution changes behavior.

PR #1283, opened 2026-08-02
Fix scheduledDistanceAlongTrip when scheduleDeviation is zero

### LINKED ISSUE #1282: trip-status: fix scheduledDistanceAlongTrip when scheduleDeviation is 0
<html><head></head><body><h2>Issue</h2><p><code>scheduledDistanceAlongTrip</code> is only computed when <code>scheduleDeviation != 0</code>:</p><ul><li><p><code>internal/restapi/trips_helper.go#L201</code></p></li></ul><p>As a result, on-time vehicles (<code>scheduleDeviation == 0</code>) incorrectly return:</p><pre><code class="language-json">"scheduledDistanceAlongTrip": 0
</code></pre><p>instead of a value equal to <code>distanceAlongTrip</code>.</p><p>According to the API specification:</p><blockquote><p><strong>scheduledDistanceAlongTrip</strong>: Metres the vehicle is scheduled to have progressed along the active trip at the requested time.</p></blockquote><h2>Example</h2><p>For vehicle <code>LMU_4021</code>:</p>
Server | scheduleDeviation | scheduledDistanceAlongTrip
-- | -- | --
Maglev | 0 | 0.0
Reference (Java) | 0 | 4306.7

<p>The Java reference server reports:</p><pre><code class="language-text">scheduledDistanceAlongTrip == distanceAlongTrip
</code></pre><p>whenever <code>scheduleDeviation == 0</code>.</p><h2>Proposed Fix</h2><p>Remove the <code>scheduleDeviation != 0</code> guard.</p><p>When the schedule deviation is zero, the scheduled distance should simply be equal to the actual <code>distanceAlongTrip</code>.</p><h2>Tests</h2><p>Add a test case covering a vehicle with <code>scheduleDeviation == 0</code> to ensure <code>scheduledDistanceAlongTrip</code> matches <code>distanceAlongTrip</code>.</p></body></html>

Ahmedhossamdev:
Fixes: #1282 

## Summary

This PR fixes the computation of `scheduledDistanceAlongTrip` for on-time vehicles.

Previously, the value was only calculated when `scheduleDeviation != 0`, causing vehicles with no schedule deviation to incorrectly report:

```json id="g1m8qa"
"scheduledDistanceAlongTrip": 0
```

instead of a value equal to `distanceAlongTrip`.

This behavior differs from the Java reference implementation, which reports:

```text id="k4n2pw"
scheduledDistanceAlongTrip == distanceAlongTrip
```

whenever `scheduleDeviation == 0`.

## Changes

* Remove the `scheduleDeviation != 0` guard when computing `scheduledDistanceAlongTrip`.
* Ensure on-time vehicles report `scheduledDistanceAlongTrip` equal to `distanceAlongTrip`.
* Add a test covering the zero schedule deviation case.

## Result

`scheduledDistanceAlongTrip` now matches the Java implementation and the API specification for both delayed/early and on-time vehicles.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved trip status accuracy for vehicles with valid GPS and route data.
  * Scheduled distance now reflects actual distance when a vehicle is on time, even when stop-time or schedule-deviation data is unavailable.

### DIFF
--- a/internal/restapi/trips_helper.go
+++ b/internal/restapi/trips_helper.go
@@ -194,13 +194,14 @@ func (api *RestAPI) BuildTripStatus(
 				}
 			}
 
-			if scheduleDeviation != 0 && len(stopTimes) > 0 {
-				scheduledDistance := api.calculateEffectiveDistanceAlongTrip(
-					ctx, actualDistance, scheduleDeviation, currentTime, serviceDate,
-					stopTimes, shapePoints, cumulativeDistances,
-				)
-				status.ScheduledDistanceAlongTrip = scheduledDistance
-			}
+			// Called unconditionally: an on-time vehicle's scheduled distance is its
+			// actual distance, not zero, and calculateEffectiveDistanceAlongTrip
+			// already falls back to actualDistance when there is no deviation to
+			// project away from and when the trip has no stop times.
+			status.ScheduledDistanceAlongTrip = api.calculateEffectiveDistanceAlongTrip(
+				ctx, actualDistance, scheduleDeviation, currentTime, serviceDate,
+				stopTimes, shapePoints, cumulativeDistances,
+			)
 		}
 	}
 
--- a/internal/restapi/trips_helper_test.go
+++ b/internal/restapi/trips_helper_test.go
@@ -560,6 +560,13 @@ func TestBuildTripStatus_ShapeData_ComputesDistanceAlongTrip(t *testing.T) {
 	assert.Greater(t, status.TotalDistanceAlongTrip, float64(0), "TotalDistanceAlongTrip should be > 0 with shape data")
 	assert.Greater(t, status.DistanceAlongTrip, float64(0), "DistanceAlongTrip should be > 0 for a vehicle mid-route")
 	assert.Less(t, status.DistanceAlongTrip, status.TotalDistanceAlongTrip, "DistanceAlongTrip should be less than total for a mid-route vehicle")
+
+	// No trip update is mocked, so the vehicle is on time. There is no deviation
+	// to project away from, which makes the scheduled distance the actual one —
+	// not zero, which is what a guard on a non-zero deviation used to produce.
+	require.Zero(t, status.ScheduleDeviation, "an unmocked trip update leaves the vehicle on time")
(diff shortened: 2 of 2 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (changes requested):
The reasoning in this one is sound — `calculateEffectiveDistanceAlongTrip` already early-returns `actualDistance` when the deviation is zero, so the outer guard was redundant-with-inversion, and removing it is behavior-preserving except for the on-time case you're targeting. The test isn't vacuous either: `require.Zero(status.ScheduleDeviation)` pins the precondition and the equality assertion would fail under the old guarded code.

Unfortunately `main` has moved out from under it.

**The function this PR modifies no longer exists.** `calculateEffectiveDistanceAlongTrip` was removed in `f264d7ad` ("Java parity for arrivals/departures trip status calculations"), which moved `ScheduledDistanceAlongTrip` onto the block-snapshot path — it's now set from `snap.ActiveTripScheduledDistance` in `trips_helper.go`. A repo-wide grep on `main` finds no remaining references. That's also why the branch is showing conflicts.

So rather than rebasing this as-is, the useful next step is to check whether #1282 still reproduces against the new snapshot code. It may already be fixed there, in which case the issue can just be closed. If it does still reproduce, the fix will look quite different and is worth a fresh PR against the current code path.

Sorry for the churn — this sat in the queue while that refactor landed, which isn't on you. Happy to look at a follow-up if the bug is still live.

Ahmedhossamdev:
You're right, it's fixed by f264d7ad.

Ran the same fixture against main: on-time vehicle, scheduleDeviation=0, and scheduledDistanceAlongTrip=2785.41 instead of the 0 the old guarded code gave. So the bug's gone with the refactor.

Closing this PR and #1282. No worries about the churn.

One thing I spotted while checking: Java reports scheduledDistanceAlongTrip == distanceAlongTrip for on-time vehicles, but the snapshot code has them diverging (7730 vs 2785). Might be intentional given the two are meant to be distinct now — my live capture predates the refactor, so I'll re-check against current main before filing anything.

PR #1286, opened 2026-08-03
fix: correct serviceDate for past-midnight trips

### LINKED ISSUE #1285: Incorrect `serviceDate` for past-midnight trips in TripsForRoute API
### Description
In the `trips-for-route` handler, the `serviceDate` for an active trip is currently being computed directly from the `currentTime` (using `todayMidnight`). However, this causes a logical bug for trips that run past midnight and belong to the previous service day. 

When a trip is resolved using the `prevServiceIDs` path (meaning the trip started on the previous day but is still active early the next morning), assigning the current day's midnight as its `serviceDate` violates the API/GTFS specifications.

### Expected Behavior
According to the specification: 
> `serviceDate` is the "Unix millisecond timestamp of midnight at the start of the service day for this trip."

If a trip departs at `25:30:00` (1:30 AM the next calendar day) on a Monday service schedule, and the API is queried at 1:00 AM on Tuesday, the `serviceDate` returned must be **Monday's midnight**, not Tuesday's.

### Actual Behavior
The API currently assigns `todayMidnight` to all trips uniformly, regardless of whether they were matched via current `serviceIDs` or `prevServiceIDs`. This results in past-midnight trips returning a `serviceDate` that is 24 hours in the future relative to their actual schedule.

3rabiii:
### Description
This PR fixes a logical bug in the `trips-for-route` handler where the `serviceDate` for active trips was being unconditionally set to the current calendar day's midnight (`todayMidnight`). 

According to the GTFS API specification, `serviceDate` should represent the midnight at the start of the service day for the trip. For trips running past midnight (e.g., departing at 23:00 on Thursday and still active at 00:30 on Friday), the previous implementation incorrectly returned Friday's midnight instead of Thursday's.

### Changes Made
- **Dynamic Service Date Tracking:** Introduced a `tripServiceDay` map to track the correct midnight timestamp for each active trip at the time it is resolved.
- **Previous-Day Resolution:** Trips resolved via the `prevServiceIDs` path (previous-day null-block trips) and block-resolved active trips now correctly carry the midnight of their respective service day (`prevDayMidnight`).
- **Response Builder Update:** The entry builder now fetches the exact `serviceDate` for each trip from the map instead of falling back to the current day's midnight.

### Testing
- Added a new custom fixture `createTestApiWithOvernightFixture` containing a Thursday-only service with a trip running from `23:00` to `24:45`.
- Added `TestTripsForRouteHandler_PastMidnightServiceDate`: Uses a mock clock set to Friday `00:30 UTC` to verify that the returned `serviceDate` correctly reflects Thursday's midnight. All assertions pass cleanly.

Closes: #1285

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected service dates for trips operating across midnight, including previous-day and duplicated trips.
  * Improved date accuracy for trips selected through scheduled, block-based, null-block, and cross-agency linked-block paths.
  * Ensured trip statuses and adjacent-trip references use the applicable service date and midnight timestamp.

* **Tests**
  * Added regression coverage for overnight, duplicated, null-block, and linked-block trip scenarios.
  * Added fixtures validating trip IDs, service dates, status dates, and adjacent-trip references.


aaronbrethorst:
### Code review

Found 1 issue:

1. The per-trip service day is threaded into the entry's `serviceDate` but not into the status built four lines above, which still receives `todayMidnight`. For exactly the past-midnight trips this PR targets, the same entry now reports two different service dates — `serviceDate` = yesterday's midnight, `status.serviceDate` = today's midnight (`BuildTripStatus` sets `status.ServiceDate` from that parameter in `internal/restapi/trips_helper.go`). Worse, `BuildTripStatus` also uses that value as the epoch for `GetScheduleDeviationForBlock`, `calculateOffsetForStop` and the distance-along-trip math, so for a trip whose stop times are 24:00–25:00 the seconds-since-service-date lands before its first arrival — `position`, `closestStop`, and `scheduleDeviation` stay 24 hours off. Hoisting the `serviceDate` computation above the `includeStatus` block and passing it to `BuildTripStatus` would make the whole entry self-consistent (bug due to `status, statusErr = api.BuildTripStatus(ctx, agencyID, tripID, nil, todayMidnight, currentTime)`)

https://github.com/OneBusAway/maglev/blob/60cc46cbfd7977c68634ecd5533c8d4e49b52f39/internal/restapi/trips_for_route_handler.go#L370-L395

aaronbrethorst (dismissed):
The core approach is right: a per-trip service-day map is the correct shape for this, and using `AddDate(0, 0, -1)` rather than subtracting 24h is the DST-safe way to get the previous midnight. The two overnight fixtures with a mock clock at 00:30 are genuine tests — they'd fail against the old unconditional today's-midnight.

The fix stops one line short, though.

**`BuildTripStatus` still receives `todayMidnight`.** The resolved per-trip service date is used only for `entry.ServiceDate`; the status built immediately above it still gets today's midnight, and `BuildTripStatus` sets `status.ServiceDate` from that parameter. So a past-midnight trip now carries two contradicting service dates in the same entry — and, more importantly, the status's schedule-deviation, closest-stop, and distance-along-trip math stays 24h off for exactly the trips this PR is about. Before the change both fields were consistently wrong; now they disagree, which is harder to debug.

Threading the resolved `serviceDate` into `BuildTripStatus` should be all it takes. For reference, #1316 does exactly this and gets both fields agreeing — worth a look at how it's structured there.

Also: the branch now conflicts with `main`. `trips_for_route_handler.go` has gained interlined-block resolution since you cut it, and the entry loop distinguishes `tripID` from a resolved `entryTripID`, so the `tripServiceDay` lookup will need to be keyed deliberately against one or the other. The test file also gained a generic `createTestApiWithGTFSFixture(t, clock, zipName, files)` helper, which should let you drop both of your near-identical fixture builders in favor of two file maps.

One smaller thing: in the null-block path, `tripServiceDay[id] = prevDayMidnight` overwrites an earlier `todayMidnight` for the same trip, whereas the block path iterates today-first and breaks, so today wins there. Only reachable for a >23h trip active on both days, but the two paths should agree.

Happy to re-review once it's rebased and the status date is threaded through.

internal/restapi/trips_for_route_handler.go
```diff
@@ -449,10 +479,15 @@ func (api *RestAPI) tripsForRouteHandler(w http.ResponseWriter, r *http.Request)
 			collectStopIDsFromSchedule(schedule, stopIDsMap)
 		}
 
+		serviceDate := todayMidnight
+		if midnight, ok := tripServiceDay[baseTripID]; ok {
+			serviceDate = midnight
+		}
```
burma-shave, line 516: **Bug: DUPLICATED trips can still report the wrong `serviceDate` past midnight.**

`tripServiceDay` is only populated from the null-block and block-matching queries above, both constrained to the ±30min/+10min window around `currentTime` (`timeRangeStart`/`timeRangeEnd`). But `GetDuplicatedVehiclesForRoute` (`internal/gtfs/gtfs_manager.go:610`) pulls DUPLICATED vehicles straight from the real-time feed, indexed only by route ID — entirely independent of that time window. A DUPLICATED trip genuinely active past midnight whose base trip falls outside the window (plausible, since real-time DUPLICATED vehicles aren't bound by it at all) will miss `tripServiceDay[baseTripID]` and fall back to `todayMidnight` here — the same 24h-off bug this PR fixes for regular trips, just missed for DUPLICATED ones.

Worth resolving the service day independently for DUPLICATED trips too (e.g. by checking whether `baseTripID`'s scheduled stop times place `currentSinceMidnight` in the previous day's window, the same way the null-block/block paths do), and adding a regression test with a DUPLICATED trip crossing midnight.

internal/restapi/trips_for_route_handler.go
```diff
@@ -383,12 +404,21 @@ func (api *RestAPI) tripsForRouteHandler(w http.ResponseWriter, r *http.Request)
...
 		}
 
+		// Per spec, serviceDate is "Unix millisecond timestamp of midnight at
+		// the start of the service day for this trip." A trip running past
+		// midnight was matched via the previous service day, so its service day
+		// starts at yesterday's midnight, not today's.
+		serviceDate := todayMidnight
+		if midnight, ok := tripServiceDay[tripID]; ok {
+			serviceDate = midnight
+		}
```
burma-shave: **Nit: duplicated lookup pattern.** This `tripServiceDay` lookup-with-fallback (`serviceDate := todayMidnight; if midnight, ok := tripServiceDay[id]; ok { serviceDate = midnight }`) is duplicated verbatim here and at lines 482-485 for the DUPLICATED-trip branch. Consider extracting a small helper, e.g. `serviceDateFor(tripServiceDay, id, todayMidnight)`, so both call sites stay in sync if the fallback logic ever changes.

internal/restapi/trips_for_route_handler.go
```diff
@@ -188,6 +202,11 @@ func (api *RestAPI) tripsForRouteHandler(w http.ResponseWriter, r *http.Request)
 			api.Logger.Warn("trips-for-route: failed to fetch previous-day null-block trips", "error", err)
 		} else {
 			nullBlockTrips = append(nullBlockTrips, prevNullBlockTrips...)
+			for _, id := range prevNullBlockTrips {
+				if _, ok := tripServiceDay[id]; !ok {
```
burma-shave, line 206: **Missing test coverage.** This guard prioritizes a current-day null-block match over a same-trip-ID previous-day match, protecting >23h-active trips from having their service day overwritten by the previous-day loop. No fixture currently constructs a trip ID appearing in both `nullBlockTrips` (today) and `prevNullBlockTrips` to exercise it — worth a regression test per CONTRIBUTING.md's expectation that new branches get coverage.

burma-shave (changes requested):
## Review summary

**Spec conformance** (`/oba-api-review`, cross-checked against legacy Java and the maglev.wiki spec): no deviations found. This brings `trips-for-route` in line with already-documented spec (mirroring Java's `BlockStatusServiceImpl`/`BlockCalendarServiceImpl` resolving candidate service dates against the block's own scheduled time range). Client-impact check found no adverse impact on Wayfinder, the JS SDK, or iOS (iOS doesn't call this endpoint), and a net-positive fix for Android's schedule-overlay extrapolation, which was silently 24h misaligned for overnight trips before this fix.

**Code quality**: one correctness bug in the DUPLICATED-trip path below should be fixed before merge; two test-coverage gaps and one small duplication nit are also flagged.

(For context: CodeRabbit's earlier comment about missing block-path test coverage is already resolved by the `test: add past-midnight block trip test coverage` commit, and a previous review comment about `BuildTripStatus` receiving the wrong `serviceDate` for regular entries is already resolved by the `fix: synchronize serviceDate across trip entries and statuses` commit — no further action needed on either.)

burma-shave (changes requested):
Just needs merge conflicts resolved

burma-shave:

burma-shave:
@coderabbit review

burma-shave:
@3rabiii checking in on this older PR. It is still marked dirty after the earlier conflict request, and CodeRabbit now has an open follow-up as well. Could you please merge `main`, resolve the conflicts/review comments, and let us know when it is ready for re-review?

3rabiii:
@burma-shave Done!

burma-shave (approved):

PR #1289, opened 2026-08-04
Search route short and long names in routes-for-location query

### LINKED ISSUE #1228: routes-for-location: Query parameter only searches route short names
### Description

**Current behavior:**
The `GetActiveRoutesWithinBounds` SQL query in `gtfsdb/stops_rtree.go` only filters routes using `routes.short_name = ?7 COLLATE NOCASE`. It completely ignores `long_name` and uses an exact match rather than a text index for partial or text-based matching (e.g., stop-words, partial queries).

> With query: the server searches a text index of route short names and long names for routes matching the query string, considering at most `maxCount + 1` candidates by relevance score.

**What to change:**
1. Update `GetActiveRoutesWithinBounds` in `gtfsdb/stops_rtree.go` (and `stops_rtree.sql` if using sqlc) to match against both `short_name` and `long_name`.
2. Replace the exact equality check with a `LIKE` match or implement full-text search behavior to support partial matching and relevance scoring.

ARCoder181105:
## Summary

`query` only matched `routes.short_name` exactly, so long-name-only
matches (e.g. "Shasta View/Shasta College") never worked. Now searches
both short and long names via the existing `routes_fts` index, ranked
by relevance, then filtered to routes with a stop in bounds.

- Reuse `routes_fts` / `Manager.SearchRoutes` for the candidate lookup
- Fix punctuation-only terms (e.g. `%`) crashing FTS5 instead of matching nothing
- Drop unused `queryTime` param from `GetRoutesForLocation`
- Multi-term queries OR instead of AND, matching legacy Java's Lucene default
  (`query=Shasta Airport` now returns routes matching either term)

Fixes https://github.com/OneBusAway/maglev/issues/1228

## Note for reviewer

Verified against `api.pugetsound.onebusaway.org`: the real server does
exact-token matching (`query=Rap` → 0 results, needs `Rap*`). This PR
always does prefix matching (`query=Ro` matches "Roosevelt" routes, no
`*` needed) — kept intentionally, better for live search. Flagging, not
filing as a defect. (This and three other deviations — description-field
indexing, no stop-word filtering, no 400 on parse error — are now
documented as accepted on the [routes-for-location wiki page](https://github.com/OneBusAway/maglev/wiki/routes-for-location#implementation-decisions).)

`buildRouteSearchQuery`'s AND→OR switch also broadens `/api/where/search/route.json`,
since both endpoints share it. Intentional — that endpoint should match Java too —
and safe: no existing test there depends on AND-only matching (all covered `input`
values are single-term).

Separately, spotted during this pass but out of scope here: `maxCount`'s default for
this endpoint is `50` in code vs `10` in both the wiki spec and prod's actual behavior,
and the overflow-above-max behavior is a 400 here vs a silent clamp in the spec. Since
the FTS candidate cap is `maxCount + 1`, that default also affects how many text-search
candidates get considered. Flagging for a possible follow-up, not fixing here.

## Test plan

- [x] `go vet` (both build tags), `make test`, `go fmt`
- [x] New tests: long-name match, prefix match, multi-word OR, no-match,
      candidate cap, out-of-bounds text match, empty-RouteIDs-means-no-filter
- [x] Manually verified multi-term OR against live King County Metro feed vs prod


gtfsdb/stops_rtree.go
```diff
@@ -152,29 +159,40 @@ SELECT
...
+	Lon      float64
+	MinLat   float64
+	MaxLat   float64
+	MinLon   float64
+	MaxLon   float64
+	RouteIDs []string
+	MaxCount int
 }
 
 func (q *Queries) GetActiveRoutesWithinBounds(ctx context.Context, arg GetActiveRoutesWithinBoundsParams) ([]Route, error) {
```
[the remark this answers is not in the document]
ARCoder181105: @coderabbitai i added the commend on both the places as u requested

Ahmedhossamdev (commented):
Hey @ARCoder181105, Great work on this.

I verified the long-name search against the live Unitrans Java server and it checks out. The punctuation-only FTS5 fix and the `ContainsLetterOrDigit` extraction are clean. LGTM, one optional note on the dynamic SQL inline. Lint and all tests pass.

gtfsdb/stops_rtree.go
```diff
@@ -152,29 +159,44 @@ SELECT
...
+	MaxLat   float64
+	MinLon   float64
+	MaxLon   float64
+	RouteIDs []string
+	MaxCount int
 }
 
+// GetActiveRoutesWithinBounds returns routes serving stops within the given bounding box,
+// ordered by ascending distance from (Lat, Lon), optionally restricted to RouteIDs.
 func (q *Queries) GetActiveRoutesWithinBounds(ctx context.Context, arg GetActiveRoutesWithinBoundsParams) ([]Route, error) {
```
Ahmedhossamdev, line 183: Optional. The dynamic `WHERE routes.id IN (?7, ?8, ...)` construction is safe from injection since only positional placeholders (`?N`) are inserted, not user values. A one-line comment noting that the `?N` strings are generated, not user input, would help future readers who might flag this as a injection risk. Low priority.

ARCoder181105: Done :)

gtfsdb/stops_rtree.go
```diff
@@ -152,29 +159,47 @@ SELECT
...
-WHERE ?7 == "" OR routes.short_name = ?7 COLLATE NOCASE
+%[1]s
 GROUP BY routes.id
 ORDER BY min_distance ASC
-LIMIT ?8
+LIMIT ?%[2]d
 `
 
+// GetActiveRoutesWithinBoundsParams holds the bind parameters for GetActiveRoutesWithinBounds.
+// RouteIDs is optional: when empty, no route-ID filter is applied.
```
burma-shave, line 169: Optional/non-blocking: `RouteIDs` empty-slice is overloaded to mean two different things depending on caller - "no filter" here, but `GetRoutesForLocation` (`gtfs_manager.go`) has to special-case an empty candidate list as "return nothing" with an early return *before* ever calling this, specifically because passing it through unchanged would mean "no filter" and wrongly return every in-bounds route. That invariant currently only lives in comments on both sides. Since this is the only call site today, it might be worth enforcing it at the type level (e.g. distinguishing "unset" from "empty" with a pointer or small wrapper type) or at least adding a test directly on `GetActiveRoutesWithinBounds` pinning that an empty `RouteIDs` returns all in-bounds routes, so a future second caller doesn't get bitten by this silently.

gtfsdb/stops_rtree.go
```diff
@@ -152,29 +159,47 @@ SELECT
...
+	MaxLat   float64
+	MinLon   float64
+	MaxLon   float64
+	RouteIDs []string
+	MaxCount int
 }
 
+// GetActiveRoutesWithinBounds returns routes serving stops within the given bounding box,
+// ordered by ascending distance from (Lat, Lon), optionally restricted to RouteIDs.
 func (q *Queries) GetActiveRoutesWithinBounds(ctx context.Context, arg GetActiveRoutesWithinBoundsParams) ([]Route, error) {
```
burma-shave, line 183: Optional/non-blocking: the `%[2]d` / `len(args)+1` bookkeeping for the dynamic `IN (...)` clause is correct today, but it's fragile - adding a third optional filter ahead of `MaxCount` means re-deriving both the template's `%[2]d` index and the Go-side arg-count arithmetic by hand, and an off-by-one here fails as a wrong bind value or a SQLite bind-index error rather than a compile-time one.

(One thing this *isn't* a candidate for: sqlc's `sqlc.slice()`, used elsewhere in `query.sql` for exactly this IN-clause pattern. It can't apply here because this whole file is hand-written specifically because sqlc can't parse the `stops_rtree` virtual-table join - per CLAUDE.md that's meant to stay a narrow, sanctioned exception, not a template to extend.) A comment pinning the invariant, or pulling the arg-building into a small named helper, would help the next person who touches this.

burma-shave (changes requested):
Reviewed this against legacy Java (`RouteCollectionSearchServiceImpl`) for spec conformance and did a code-quality pass. Four semantic gaps between Maglev's FTS5-based search and Java's Lucene-based search were found and discussed; three (prefix-vs-exact-token matching, description-field inclusion, no stop-word filtering) plus the resulting no-400-on-parse-error behavior are now documented as deliberate, accepted deviations on the [routes-for-location wiki page](https://github.com/OneBusAway/maglev/wiki/routes-for-location#implementation-decisions), so no action needed on those here.

Two things do need changes before merge. Both touch code this PR's diff doesn't actually modify, so I couldn't leave these as inline comments - GitHub only allows anchoring to lines within the diff:

**1. Multi-term queries should OR, not AND, to match Java's default.**

[`buildRouteSearchQuery`](https://github.com/OneBusAway/maglev/blob/fca1a842c1fef08a84945369abe15e9a6743dfb3/internal/gtfs/route_search.go#L33) joins terms with `" AND "`, but legacy Java's `RouteCollectionSearchServiceImpl` defaults to Lucene's OR operator between terms - `QueryParserBase` initializes `operator = OR_OPERATOR` ([QueryParserBase.java#L64-L66,L74](https://github.com/apache/lucene-solr/blob/releases/lucene-solr/7.1.0/lucene/queryparser/src/java/org/apache/lucene/queryparser/classic/QueryParserBase.java#L74)), and this service never calls `setDefaultOperator`.

Concretely, on this PR's own RABA fixture, `query=Shasta Airport` should return both Route 17 (long name contains "Shasta") and Route 15 (long name contains "Airport") under Java's OR semantics - but this PR's own new test ([`TestRoutesForLocationQueryMatchesRouteText`](https://github.com/OneBusAway/maglev/blob/fca1a842c1fef08a84945369abe15e9a6743dfb3/internal/restapi/routes_for_location_handler_test.go#L181-L185), subtest "multiple terms with one absent from any route match nothing") asserts an *empty* result instead, which is AND behavior, not Java's.

Could you switch `route_search.go:33` to `strings.Join(safeTerms, " OR ")` and update that test's expectation? One thing to flag while you're in there: `buildRouteSearchQuery` is shared with `/api/where/search/route.json` (`route_search_handler.go`), so this changes that endpoint's multi-word matching too. No existing test there depends on AND-only semantics (`route_search_handler_test.go` only exercises single-term `input` values), so it looks safe, but worth double-checking that broadening `search/route.json`'s results is intentional since it's outside this PR's stated scope.

Full Java citation and reasoning: https://github.com/OneBusAway/maglev/wiki/routes-for-location#main-success-scenario (step 5).

**2. CI is red - a pre-existing, unrelated test needs its expectation updated.**

[`TestRoutesForLocationBoundingBoxSizing`](https://github.com/OneBusAway/maglev/blob/fca1a842c1fef08a84945369abe15e9a6743dfb3/internal/restapi/routes_for_location_handler_test.go#L49), subtest "one span alone falls back to the query default radius" (line 81), is currently failing on `Test (ubuntu-latest)` and `Test (windows-latest)`. Confirmed real, not flaky - reproduced locally on this branch and confirmed passing on `origin/main`.

Root cause: on `origin/main`, `query=3` did an exact `short_name = '3'` match, matching only Route 3 (`25_153`). This PR's switch to FTS5 search across short *and* long names means `query=3` now also legitimately matches Route 99X (`25_24`, long name "Route 99X/Amtrak Thruway Route 3" - trailing token "3"). That's correct behavior for this PR's own goal (search long names too, not just short names), so the fix is updating that test's expectation to `[]string{"25_153", "25_24"}` at line 81, not changing production code.

(If you also switch AND→OR per the finding above, this specific case is unaffected - it's a single-term query, and AND vs OR only matters for 2+ terms.)

Plus two optional code-quality suggestions on the dynamic SQL in `stops_rtree.go`, left as inline comments below.

ARCoder181105:
Pushed fixes for both blocking items:

**Multi-term OR** (`f481824`): switched `buildRouteSearchQuery` to join terms with OR, matching Lucene's default. Verified against prod (`api.pugetsound.onebusaway.org`) before changing: `query=Tacoma Zzzznomatch` still returns the Tacoma route, and `query=Tacoma District` returns a second route matching only "District" — confirms OR, not AND. Updated `TestBuildRouteSearchQuery`'s five multi-term cases and the two `TestRoutesForLocationQueryMatchesRouteText` subtests that encoded AND semantics. Also confirmed this broadens `/api/where/search/route.json` too (shared `buildRouteSearchQuery`) — intentional, noted in the PR description, no existing test there assumed AND.

**CI red** (`d7a8fe5`): already fixed before this review landed — same root cause and same expectation you diagnosed (`["25_153", "25_24"]`).

Replied inline on the two optional items separately.

gtfsdb/stops_rtree.go
ARCoder181105: Added `TestGetActiveRoutesWithinBounds_EmptyRouteIDsMeansNoFilter` (`d311bd4`) pinning both halves of the contract at the query layer — empty `RouteIDs` returns everything, populated filters down. Left the type-level unset-vs-empty distinction (pointer/wrapper) out for now since there's still only the one caller; happy to revisit if a second caller shows up.

gtfsdb/stops_rtree.go
ARCoder181105: Agreed this is fragile, deferring the helper extraction for now to keep this PR's diff focused — it's real but not blocking, and no second filter is being added here. Worth a follow-up if/when a third optional param gets added to this query.

burma-shave (approved):

PR #1291, opened 2026-08-04
routes-for-location: silently clamp maxCount to 50

### LINKED ISSUE #1231: routes-for-location: maxCount is not silently clamped to 50
### Description

**Current behavior:**
The implementation uses `utils.ParseMaxCount` which allows `maxCount` up to 250 (and returns a `fieldErrors` validation error if exceeded). It does not silently clamp large requests.

> `maxCount` exceeds the absolute maximum: The server silently clamps `maxCount` to 50. No error is returned.

**What to change:**
1. In `internal/restapi/routes_for_location_handler.go`, check if `maxCount > 50` after parsing.
2. If it is, reassign `maxCount` to `50` (or clear the error and clamp it) before passing it to the database/manager layer, ensuring it silently clamps without returning an HTTP 400.

ARCoder181105:
## Summary
- Per the [wiki spec](https://github.com/OneBusAway/maglev/wiki/routes-for-location), `maxCount` values above the endpoint's cap must be silently clamped to 50, not rejected.
- The handler used `utils.ParseMaxCount` (the rejecting variant), so requests with `maxCount` between 51-250 were passed through uncapped, and requests above 250 returned HTTP 400 instead of being clamped.
- Switched to `utils.ParseMaxCountClamped` and applied a new `models.MaxCountForRoutesForLocation = 50` cap on top of the global 250 ceiling.

Closes #1231

## Test plan
- [x] Added `TestRoutesForLocationHandlerClampsMaxCountAboveCap`
- [x] `go test ./internal/restapi -run TestRoutesForLocation -tags "sqlite_fts5 sqlite_math_functions" -v`
- [x] `go vet -tags "sqlite_fts5 sqlite_math_functions" ./...` and `go vet -tags "purego" ./...`
- [x] `make test`

## Note for reviewer
Spec says `maxCount` default is 10; code defaults to `models.DefaultMaxCountForRoutes = 50`. Separate, unfiled gap under #1225 — left out of scope here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Location-based route requests now enforce a maximum of 50 routes.
  * Requests with higher route-count values are safely clamped to the supported limit.
  * Responses accurately indicate when the requested count exceeds the endpoint’s maximum while still returning available route results.
  * Improved coverage verifies behavior for counts below, at, and above the limit.


ARCoder181105:
@coderabbitai full review

internal/restapi/routes_for_location_handler.go
```diff
@@ -19,12 +19,14 @@ func (api *RestAPI) routesForLocationHandler(w http.ResponseWriter, r *http.Requ
 
 	var fieldErrors map[string][]string
 	loc, fieldErrors := api.parseLocationParams(r, fieldErrors)
-	maxCount, fieldErrors := utils.ParseMaxCount(queryParams, models.DefaultMaxCountForRoutes, fieldErrors)
+	maxCount, fieldErrors := utils.ParseMaxCountClamped(queryParams, models.DefaultMaxCountForRoutes, fieldErrors)
```
Ahmedhossamdev, line 21: The core fix is correct. Switching to `ParseMaxCountClamped` plus the `min(maxCount, 50)` clamp matches Java's `MaxCountSupport(10, 50)` behavior from `RoutesForLocationAction.java:62`. I ran a live parity diff against the Unitrans Java server (`api.unitrans.onebusawaycloud.com`) for `maxCount` values 10, 50, 51, 250, 300, 1000, 0, -1, and a non-numeric input. All match: the previously-rejected values (300, 1000) now return 200 with `limitExceeded:false`, matching Java. Lint and tests pass.

Two optional inline notes, nothing blocking.

One pre-existing gap that's out of scope: omitting `maxCount` defaults to 50 in maglev vs 10 in Java. This is #1225, called out in the PR description. Flagging only for visibility.

Ahmedhossamdev, line 21: Optional cleanup. `ParseMaxCountClamped` already silently clamps to 250, then line 29 clamps the result down to 50. Since 50 < 250, the 250 clamp does nothing useful for any input above 250. Java does this in a single `Math.min`. Harmless as-is, but if you want to mirror Java more directly, the rejecting `ParseMaxCount` + `min(maxCount, 50)` would reject nothing new (any positive value lands at 50 or below) and removes the redundant clamp. Low priority.

ARCoder181105: `ParseMaxCount` doesn't just skip clamping above 250  it sets a `maxCount` field error and resets to default (`api.go:248-254`). Handler returns 400 whenever `fieldErrors` is non-empty, so `maxCount=300`/`1000` would go back to being rejected the exact case your Unitrans diff shows should be 200. Keeping `ParseMaxCountClamped`.

Redundancy point is fair, but fixing it needs a third parse variant with no 250 ceiling not worth a new helper for one inert `min`. Leaving as-is.

burma-shave (changes requested):
Merge in main and fix conflicts

ARCoder181105:
Done resolved the merge conflicts

burma-shave (approved):

PR #1293, opened 2026-08-04
test: add coverage for includeStatus parameter

### LINKED ISSUE #1292: Missing test coverage for `includeStatus` in `trips-for-route` handler
### Description
The `trips-for-route` API handler currently lacks test coverage for the `includeStatus` query parameter. While the `includeSchedule` parameter is tested, there are no assertions verifying that `includeStatus=false` successfully omits the `status` object from the response entries (as per spec extension 5c).

### Expected Behavior
Add a table-driven test to `internal/restapi/trips_for_route_handler_test.go` that asserts:
- `includeStatus=false` -> the `status` object is `nil` / omitted.
- `includeStatus=true` (or omitted) -> the `status` object is present and populated correctly.

3rabiii:
### Description
This PR addresses the missing test coverage for the `includeStatus` query parameter in the `trips-for-route` handler.

### Changes Made
- Added a new table-driven test `TestTripsForRouteHandler_StatusInclusion` in `internal/restapi/trips_for_route_handler_test.go`.
- The test explicitly verifies that:
  - `includeStatus=false` correctly omits the `status` object from the response entries.
  - `includeStatus=true` (or omitted) correctly includes the populated `status` object.

Closes: #1292 

## Summary by CodeRabbit

* **Tests**
  * Added coverage to verify trip responses correctly include or omit status information based on the `includeStatus` setting.
  * Confirmed default and explicitly enabled behavior returns valid status details for active trips.

### DIFF
--- a/internal/restapi/trips_for_route_handler_test.go
+++ b/internal/restapi/trips_for_route_handler_test.go
@@ -343,6 +343,50 @@ func TestTripsForRouteHandlerWithMalformedID(t *testing.T) {
 	assert.Equal(t, http.StatusBadRequest, model.Code)
 }
 
+func TestTripsForRouteHandler_StatusInclusion(t *testing.T) {
+	api := createTestApiWithTripsForRouteFixture(t, clock.NewMockClock(tripsForRouteTestClock))
+	combinedRouteID := utils.FormCombinedID(tripsForRouteAgencyID, tripsForRouteRouteID)
+
+	tests := []struct {
+		name          string
+		includeStatus string
+		wantStatus    bool
+	}{
+		{name: "Include Status (default)", includeStatus: "", wantStatus: true},
+		{name: "Include Status Explicit", includeStatus: "true", wantStatus: true},
+		{name: "Exclude Status", includeStatus: "false", wantStatus: false},
+	}
+
+	timeMs := tripsForRouteTestClock.UnixMilli()
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			url := fmt.Sprintf("/api/where/trips-for-route/%s.json?key=TEST&includeSchedule=true&time=%d",
+				combinedRouteID, timeMs)
+			if tt.includeStatus != "" {
+				url += "&includeStatus=" + tt.includeStatus
+			}
+
+			resp, model := callAPIHandler[TripsForRouteResponse](t, api, url)
+
+			assert.Equal(t, http.StatusOK, resp.StatusCode)
+			require.NotEmpty(t, model.Data.List,
+				"fixture guarantees a trip at the pinned clock")
+			for i, entry := range model.Data.List {
+				if tt.wantStatus {
+					require.NotNil(t, entry.Status, "list[%d].status should be present when includeStatus=true", i)
+					assert.NotEmpty(t, entry.Status.ActiveTripID, "list[%d].status.activeTripId should be set", i)
+					assert.Contains(t, []string{"scheduled", "in_progress", "completed"}, entry.Status.Phase,
+						"list[%d].status.phase should be a known value", i)
+				} else {
+					assert.Nil(t, entry.Status,
+						"list[%d].status should be omitted when includeStatus=false", i)
+				}
+			}
+		})
+	}
+}
+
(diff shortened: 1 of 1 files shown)


aaronbrethorst:
### Code review

Found 2 issues:

1. The new test calls `createTestApiWithTripsForRouteFixture`, which no longer exists on `main` — it was removed in `2af6f7f0` when the per-scenario fixture builders were collapsed into `createTestApiWithGTFSFixture(t, clock, zipName, files)`. This PR is based on `3b73fd9d` (Aug 4) and its green checks were computed against that stale base, so CI never saw the conflict; GitHub still reports the merge as "clean" because there is no *textual* conflict. Merging as-is would leave `package restapi` unable to compile (`undefined: createTestApiWithTripsForRouteFixture`), breaking every test in the package. Needs a rebase onto `main` plus a switch to `createTestApiWithGTFSFixture(t, clock.NewMockClock(tripsForRouteTestClock), "trips-for-route.zip", basicTripsForRouteFiles())`, matching every other test in this file.

https://github.com/OneBusAway/maglev/blob/9268bda98296adec0106f84622165a0d7517d4be/internal/restapi/trips_for_route_handler_test.go#L346-L348

2. The coverage this test adds already landed on `main` in `9fdf56b7` ("Standardize boolean parameter parsing across all flags") as `TestTripsForRouteHandler_BoolParamParsing`, which exercises `includeStatus` for omitted / `=true` / `=false` / `=` (empty) / `=abc` (junk) and asserts `entry.Status` non-nil vs. nil in exactly the same way — a strict superset of the three cases here. CONTRIBUTING.md says to "check whether it already exists" before writing new code and to "match the codebase's existing test-coverage conventions rather than maximizing coverage for its own sake." The only genuinely new assertions are `status.activeTripId` non-empty and `status.phase` being one of three values; if those are worth keeping, folding them into the existing `assertFlag` helper's `includeStatus` case would be the smaller change. Worth confirming whether #1292 is already satisfied on `main` before landing anything here.

https://github.com/OneBusAway/maglev/blob/9268bda98296adec0106f84622165a0d7517d4be/internal/restapi/trips_for_route_handler_test.go#L345-L358

aaronbrethorst (changes requested):
The test itself is written the way I'd want — table-driven, uses `callAPIHandler`, asserts nil vs non-nil `entry.Status` rather than something vaguer. Two problems, neither of them about the code you wrote.

**1. This won't compile once merged.** It calls `createTestApiWithTripsForRouteFixture`, which was removed from `main` and replaced by the generic `createTestApiWithGTFSFixture(t, clock, zipName, files)`. I merged your branch into current `main` locally and `go vet` fails:

```
internal/restapi/trips_for_route_handler_test.go:549:9: undefined: createTestApiWithTripsForRouteFixture
```

Worth knowing why CI didn't catch this: GitHub reports the PR as MERGEABLE and every check is green, because the checks ran against a merge ref computed before that helper was deleted. Green CI on a stale base isn't proof the merge compiles — that's on our tooling, not on you.

**2. The coverage already exists on `main`.** `TestTripsForRouteHandler_BoolParamParsing` covers `includeStatus` for omitted / true / false / empty / junk with the same nil vs non-nil assertions — a strict superset of the three cases here. The only genuinely new content is the `activeTripId` non-empty check and the `phase` membership check.

My suggestion: rather than rebasing this, check whether #1292 can be closed against what's already on `main`, and if you want to keep the two extra assertions, fold them into the existing `assertFlag` helper as a small follow-up. That's a much smaller change than resurrecting this one.

Separately, and not something to fix in this PR — you've surfaced a real spec question. `testdata/openapi.yml` says `includeStatus` and `includeSchedule` on trips-for-route default to **false**, but `parseBoolQueryParam` defaults omitted flags to **true**, and `trips_for_location_handler.go` has a comment deliberately going the other way. CLAUDE.md says the spec wins, so that's a discrepancy I need to resolve on our side before more tests pin the current behavior. I'll take that one.

3rabiii:
@aaronbrethorst Thanks for the review and the clear explanation! 
Since the core test coverage for `includeStatus` already landed on `main` and this branch is stale, I will close this PR. I am working on a small follow-up PR to fold the two extra assertions (`activeTripId` and `phase`) into the new `assertFlag` helper, as you suggested.

PR #1296, opened 2026-08-04
fix: thread api.Clock into ParseTimeParameter and add coverage for time edges

### LINKED ISSUE #1295: Missing test coverage for omitted and invalid `time` parameter in `trips-for-route` handler
### Description
The `trips-for-route` API handler currently lacks test coverage for edge cases regarding the `time` query parameter. While explicit valid time parameters are tested, we need to ensure the handler correctly processes missing or malformed time inputs. 

### Expected Behavior
Add tests in `internal/restapi/trips_for_route_handler_test.go` to assert that:
- **Omitted `time=` parameter:** The handler correctly defaults to the current clock time and returns a `200 OK` response.
- **Invalid `time=` parameter:** Passing a malformed value triggers `ParseTimeParameter` and correctly returns a `400 Bad Request` validation error.

3rabiii:
### Description
This PR addresses the missing test coverage for edge cases regarding the `time` query parameter in the `trips-for-route` handler. 

### Changes Made
- **Added `TestTripsForRouteHandler_TimeOmitted`:** Verifies that when the `time` parameter is not provided, the handler correctly defaults to the current clock time and returns a `200 OK` response.
- **Added `TestTripsForRouteHandler_InvalidTimeParameter`:** Asserts that passing malformed time values (e.g., garbage strings, overflowing epochs, or invalid calendar dates) correctly triggers a validation error, returning a `400 Bad Request`.

Closes: #1295 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of optional time parameters across trip, stop, and vehicle endpoints.
  * Requests without a specified time now consistently use the API’s configured current time.
  * Malformed, overflowing, or invalid-date time values now return HTTP 400 responses with the expected error details.

* **Tests**
  * Added coverage for default-time behavior and invalid time parameter responses.

### DIFF
--- a/internal/restapi/stops_for_route_handler.go
+++ b/internal/restapi/stops_for_route_handler.go
@@ -48,7 +48,7 @@ func (api *RestAPI) stopsForRouteHandler(w http.ResponseWriter, r *http.Request)
 
 	var serviceIDs []string
 	if filterByDate {
-		formattedDate, _, fieldErrors, success := utils.ParseTimeParameter(timeParam, currentLocation)
+		formattedDate, _, fieldErrors, success := utils.ParseTimeParameter(timeParam, currentLocation, api.Clock)
 		if !success {
 			api.validationErrorResponse(w, r, fieldErrors)
 			return
--- a/internal/restapi/trips_for_location_handler.go
+++ b/internal/restapi/trips_for_location_handler.go
@@ -202,7 +202,7 @@ func (api *RestAPI) resolveCurrentTime(timeParam string, currentLocation *time.L
 	if timeParam == "" {
 		return api.Clock.Now().In(currentLocation), nil
 	}
-	_, currentTime, timeFieldErrors, _ := utils.ParseTimeParameter(timeParam, currentLocation)
+	_, currentTime, timeFieldErrors, _ := utils.ParseTimeParameter(timeParam, currentLocation, api.Clock)
 	return currentTime, timeFieldErrors
 }
 
--- a/internal/restapi/trips_for_route_handler.go
+++ b/internal/restapi/trips_for_route_handler.go
@@ -53,7 +53,7 @@ func (api *RestAPI) tripsForRouteHandler(w http.ResponseWriter, r *http.Request)
 	}
 
 	timeParam := r.URL.Query().Get("time")
-	formattedDate, currentTime, fieldErrors, success := utils.ParseTimeParameter(timeParam, currentLocation)
+	formattedDate, currentTime, fieldErrors, success := utils.ParseTimeParameter(timeParam, currentLocation, api.Clock)
 	if !success {
 		api.validationErrorResponse(w, r, fieldErrors)
 		return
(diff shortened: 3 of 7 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. The `fix:` commit makes an existing comment in this same file factually wrong. `TestTripsForRouteHandler_RouteVariations` still says "ParseTimeParameter ignores api.Clock when no time= is given" — after this PR, `utils.ParseTimeParameter` takes the clock and uses it for the omitted-`time` branch, which is precisely what `TestTripsForRouteHandler_TimeOmitted` now relies on. Left as-is it will mislead the next contributor about the helper's contract (CONTRIBUTING.md: "Add an inline comment when there's a non-obvious constraint, workaround, or invariant a reader couldn't infer from the code itself"). Passing an explicit `time=` there is still worth doing, but the stated reason is no longer true.

https://github.com/OneBusAway/maglev/blob/a6b13490967861740faf4502ccc30e0ee6ca5e1a/internal/restapi/trips_for_route_handler_test.go#L151-L155

aaronbrethorst (changes requested):
The substance here is good, and better than the `test:` title suggests — threading `api.Clock` into `ParseTimeParameter` is a real fix, and you updated all four production call sites and all five test call sites consistently. The omitted-`time` test is genuinely load-bearing too: it fails without the clock change, because the fixture trip only falls inside the window at the pinned clock. That's the right way to write a regression test.

One blocker, and it isn't about your code.

**This won't compile once merged.** It calls `createTestApiWithTripsForRouteFixture`, which was removed from `main` and replaced by the generic `createTestApiWithGTFSFixture(t, clock, zipName, files)`. Merging your branch into current `main` locally and running `go vet`:

```
internal/restapi/trips_for_route_handler_test.go:422:9: undefined: createTestApiWithTripsForRouteFixture
```

The PR shows MERGEABLE with every check green because CI ran against a merge ref computed before that helper was deleted — so nothing on our side flagged it. Rebasing onto current `main` and porting the two new tests to `createTestApiWithGTFSFixture` should be all it takes.

While you're in there, one small thing this PR invalidates: the comment at `trips_for_route_handler_test.go:152` still says "ParseTimeParameter ignores api.Clock when no time= is given", which your own fix makes untrue. It's on a line you didn't touch, but this is the change that breaks it.

Worth splitting the title/description too — a signature change to a shared util under a `test:` heading is easy to miss in the log. Your commits do delineate the two concerns properly, which is the part that actually matters.

Re-review as soon as it's rebased; the fix itself I'm happy with.

aaronbrethorst:
### Code review

aaronbrethorst (approved):

The clock threading is complete and consistent. I grepped for `ParseTimeParameter` and all four production call sites are updated with no stragglers, and since `createClock` returns `clock.RealClock{}` for everything but the test environment, this is a genuine no-op in production — the only reachable behavior change is under a mock clock, which is the whole point.

What I appreciated is that the new tests actually depend on the change. `TestTripsForRouteHandler_TimeOmitted` only passes because the fixture trip falls inside the window of the pinned clock; under the old `time.Now()` behavior it would fail outside a narrow wall-clock window. That is the difference between a test that documents behavior and one that just happens to be green.

The stale comment I flagged last time ("ParseTimeParameter ignores api.Clock when no time= is given") is gone and replaced with an accurate one.

PR #1298, opened 2026-08-05
test: add E2E coverage for DUPLICATED real-time trips

### LINKED ISSUE #1297: Missing end-to-end test coverage for DUPLICATED real-time trips in `trips-for-route`
### Description
The `trips-for-route` API handler currently lacks end-to-end test coverage for the `DUPLICATED` real-time trip path. While the `stripNumericSuffix` utility has unit tests, there is no handler-level test verifying that real-time trips with `schedule_relationship=DUPLICATED` and a suffixed trip ID (e.g., `_1` or `.00060`) are processed correctly.

### Expected Behavior
Add an end-to-end test in `internal/restapi/trips_for_route_handler_test.go` using a real-time fixture that asserts:
- A `DUPLICATED` vehicle injected into the feed appears in the response with its correct combined `tripId`.
- The base-trip references (route, headsign, block) and schedule are successfully resolved and populated end-to-end.
- The numeric suffix fallback successfully resolves the original base trip.

3rabiii:
### Description

This PR addresses end-to-end test coverage for the `DUPLICATED` real-time trips path in the `trips-for-route` handler. It ensures that the handler correctly resolves base trips for duplicated vehicles using the `stripNumericSuffix` fallback and successfully populates the schedule and reference data.

### Changes Made

* **End-to-End Assertion:** Added `TestTripsForRouteHandler_DuplicatedRealtimeTrip` to verify that:
  * The base trip falls outside the handler's -30/+10 minute window, ensuring it is reached *only* through the `DUPLICATED` fallback block.
  * The duplicated trip appears in the response list with its preserved combined suffixed `tripId`.
  * The schedule is successfully populated from the resolved base trip (proving `stripNumericSuffix` works correctly).
  * `ActiveTripID` is correctly set from the vehicle's own suffixed ID by `BuildVehicleStatus`, not the base ID.
  * The resolved base trip correctly appears in `references.trips` with all associated static data.
* **Cleanup:** Removed the previous custom fixture zip builder and mock GTFS-RT injection in favor of a cleaner, more direct testing approach using the updated `MockAddDuplicatedVehicle`.

Closes: #1297

## Tests

- Added integration coverage for duplicated real-time vehicle trips.
- Verified that suffixed real-time trips appear correctly and resolve their associated schedules and stop times.
- Confirmed duplicated trips report the appropriate in-progress status and reference the original trip.
- Improved synthetic transit fixtures and real-time feed setup to support reliable validation of these scenarios.


3rabiii:
@coderabbitai full review

aaronbrethorst:
### Code review

Found 1 issue:

1. The `references.trips` block asserts nothing about the DUPLICATED path — it passes regardless of that behavior. In this fixture `tfr-trip` is an active scheduled trip, so it is already in `fetchedTrips`/`filteredRouteTrips` from the static loop and is emitted into `references.trips` by `buildTripReferences(..., preFetchedTrips, ...)` with route, headsign, and block populated. The DUPLICATED branch that would add it (`if !filteredRouteTrips[baseTripID] { ... }` in `trips_for_route_handler.go`) is therefore dead in this test — delete that block and these three assertions still pass. The comment claiming "stripNumericSuffix resolved the base trip: references.trips must contain..." misattributes what is being verified (the suffix fallback is genuinely covered, but by the `entry.Schedule` stop-ID assertions above, not by this block). To actually exercise it, the duplicated vehicle needs a base trip that is *not* independently active at the query time.

https://github.com/OneBusAway/maglev/blob/00c17c02f0070c35e95abbe1ee4b7eaaa40ac574/internal/restapi/trips_for_route_handler_test.go#L238-L252

aaronbrethorst (changes requested):
The schedule assertions in here are the good part — they genuinely cover the `stripNumericSuffix` fallback and confirm the suffixed trip keeps its own entry ID while carrying the base trip's stop times. The RT feed setup is correct too; `InitGTFSManager` does a synchronous first fetch, so there's no race to worry about.

One assertion doesn't do what its comment claims.

**The `references.trips` block is tautological for the DUPLICATED path.** In this fixture `tfr-trip` is an active scheduled trip, so it reaches `references.trips` through the ordinary static path — `fetchedTrips` → `filteredRouteTrips` → `buildTripReferences(..., preFetchedTrips, ...)` — fully populated. The handler's DUPLICATED-side resolution (`if !filteredRouteTrips[baseTripID] { ... }`, `trips_for_route_handler.go:472-479`) never executes, and deleting it entirely would leave this test green. The comment says the block verifies that `stripNumericSuffix` resolved the base trip, but that's covered by the schedule assertions instead.

Since base-trip reference resolution is one of the things #1297 specifically asked for, it's worth making real: make the base trip inactive at query time, or query at a moment when only the duplicated run is live, so the fallback branch is the only way the reference can appear.

The branch also conflicts with `main`, and more substantially than the conflict markers suggest: `createTestApiWithTripsForRouteFixture` no longer exists — it was replaced by a generic `createTestApiWithGTFSFixture(t, clock, zipName, files map[string]string)` plus per-scenario file maps. So your `buildTripsForRouteFixtureZip` extraction is largely redundant with what `main` now does more generally, and `createTestApiWithTripsForRouteAndRealtime` will want rebuilding on top of the new builder. Sorry — that landed while this was in the queue.

Two smaller notes for the rework: the URL only passes `includeSchedule=true`, so the `DUPLICATED → ("DUPLICATED", "in_progress")` status mapping is never exercised, which seems worth having in a PR with this title; and there's no assertion on list length, so a spurious extra entry would slip through.

Re-review whenever it's rebased.

aaronbrethorst:
### Code review

burma-shave (changes requested):
merge in main and fix conflicts

aaronbrethorst (approved):
Approving on the merits — this looks good, but it can't merge as-is. See below.

All four points from my last review are addressed, and the important one is genuinely fixed. I checked it rather than assuming: at the 11:30 clock the base trip falls outside the handler's -30/+10 minute window, so it can only reach `references.trips` through the DUPLICATED fallback block — delete that block and the test fails. That's the difference from last time, when the assertion held no matter what the handler did.

The rest holds up too. `dupEntry.Schedule` would come back with empty stop IDs if `stripNumericSuffix` were removed, so the fallback is really exercised. `ActiveTripID` is set from the vehicle's own suffixed ID by `BuildVehicleStatus`, not the base ID, so that assertion has teeth. The `DUPLICATED`/`in_progress` pair matches `GetVehicleStatusAndPhase`, and `require.Len(..., 2)` guards against extra entries.

Two things stand between this and merging, neither of them the code:

1. The branch conflicts with `main` — it's one commit ahead and thirty-seven behind, and both files it touches have changed since. It needs another rebase. Sorry for the treadmill; that file is busy right now.
2. burma-shave has a change request still open on this PR. My approval doesn't clear that one, so please square it with them before merging.

While you're rebasing: the PR description still describes `buildTripsForRouteFixtureZip` and a mock GTFS-RT feed, neither of which exists in the current diff. Worth a refresh so the next reader isn't misled.

burma-shave (changes requested):
needs merge conflicts resolved again

burma-shave:

burma-shave (approved):

Issue #1299, opened 2026-08-05
Missing test coverage for `status` sub-object fields in `trips-for-route`


3rabiii:
### Description
While some `status` fields like `activeTripId` and `phase` are covered by existing tests, the API lacks strict assertions for other critical fields in the `trips-for-route` response. To prevent potential frontend crashes, we need to ensure these fields strictly follow the API specification.

### Expected Behavior
Add a real-time fixture test in `internal/restapi/trips_for_route_handler_test.go` to assert the following on the `status` object:
- `blockTripSequence` is a non-negative integer.
- `position` is present and contains `lat` and `lon`.
- `occupancyCount` returns the `-1` sentinel value when unavailable.
- `situationIds` and `vehicleFeatures` are never `null` (must default to empty arrays `[]`).

PR #1300, opened 2026-08-05
test: add E2E coverage for status sub-object fields

3rabiii:
### Description
This PR addresses Issue #1299 by adding strict end-to-end test coverage for the `status` sub-object fields in the `trips-for-route` API response.

### Changes Made
* **Strict Assertions:** Added `TestTripsForRouteHandler_StatusFields` to verify:
  * `blockTripSequence` is a non-negative integer.
  * `position.lat` and `position.lon` are accurately populated (handling the GTFS-RT `float32` to Go `float64` conversion).
  * `occupancyCount` correctly returns the `-1` sentinel constructor default value.
  * `situationIds` and `vehicleFeatures` are strictly non-nil empty slices (`[]`), differentiating them from `null`.
* **Testing Pattern Alignment:** Uses the unified real-time mock injection approach to cleanly test the `SCHEDULED` vehicle tracking without race conditions.

## Tests

- Expanded coverage for trip status information returned by the API.
- Verifies vehicle positions, block sequencing, occupancy values, service status, and trip phase details.
- Confirms consistent handling of empty vehicle-feature and service-alert data.
- Adds scheduled-vehicle scenarios with GPS coordinates and timestamps to improve confidence in trip and vehicle-status results.


aaronbrethorst:
### Code review

Found 1 issue:

1. `assert.GreaterOrEqual(t, entry.Status.BlockTripSequence, 0)` cannot fail, so it provides no regression coverage for the field the PR is meant to pin down. `BuildTripStatus` computes the sequence via `calculateBlockTripSequence`, which returns `0` on any failure ("Returns 0 when the sequence is unavailable"), and then only assigns it with `if blockTripSequence > 0` (`internal/restapi/trips_helper.go:62-65`) — the field is non-negative by construction on this path. The synthetic fixture also makes the value fully deterministic: `tfr-trip` is the only trip in block `tfr-block`, and `GetBlockTripSequence` is zero-based (`ROW_NUMBER() ... - 1`), so the expected value is exactly `0`. If `BuildTripStatus` stopped computing `blockTripSequence` entirely, this assertion would still pass. `assert.Equal(t, 0, entry.Status.BlockTripSequence)` would actually pin the behavior. (The neighbouring `position` / `occupancyCount` / `situationIds` / `vehicleFeatures` assertions are load-bearing — only this one is vacuous.)

https://github.com/OneBusAway/maglev/blob/9805ac48a7c5c68adb6bde59ef139c25bd7b3d8d/internal/restapi/trips_for_route_handler_test.go#L228-L234

aaronbrethorst (changes requested):
Most of this test is load-bearing. Injecting a real GTFS-RT vehicle with a GPS fix and asserting on `status.position.lat/lon`, the `-1` `occupancyCount` sentinel, and non-null-but-empty `situationIds`/`vehicleFeatures` would all catch real regressions — and the `[]` vs `null` checks in particular guard something easy to break and hard to notice. I confirmed none of it contradicts `openapi.yml`'s `TripStatus` schema.

The one field the PR headlines first, though, asserts nothing.

**`assert.GreaterOrEqual(t, entry.Status.BlockTripSequence, 0)` can never fail.** `calculateBlockTripSequence` returns 0 on any failure, and `BuildTripStatus` only assigns the field under `if blockTripSequence > 0` (`trips_helper.go:62-65`) — so it's non-negative by construction, whatever happens. The fixture is fully deterministic here (a single trip in `tfr-block`, and `GetBlockTripSequence` is zero-based), so the real assertion is `assert.Equal(t, 0, ...)`. One-line change.

I know `GreaterOrEqual(seq, 0)` appears elsewhere in the suite, but those run over RABA data where the value genuinely isn't pinned. Over a controlled fixture — and in a PR whose stated goal is strict assertions — it's worth being exact.

The branch also conflicts with `main`, and more deeply than it looks: `createTestApiWithTripsForRouteFixture`, the helper this refactors, was removed in `c905b6c8` and replaced by a generic `createTestApiWithGTFSFixture(t, clock, zipName, files)` plus per-scenario file maps. So the `buildTripsForRouteFixtureZip` extraction is now redundant, and `createTestApiWithScheduledRealtimePosition` needs rebuilding on top of the new builder. That landed while this sat in the queue — not your doing. `TestTripsForRouteHandler_StatusFields` itself should transplant cleanly once the fixture dependency is sorted.

One last thing: the helper doc and test comment both emphasize "a SCHEDULED vehicle", but the feed never sets `ScheduleRelationship` (it relies on the proto default) and the test never asserts `Status`/`Phase`. Either assert it or soften the comment.

Happy to re-review once it's rebased.

aaronbrethorst:
### Code review

burma-shave (changes requested):
merge main and fix conflicts

aaronbrethorst (approved):
Approving on the merits — this looks good, but it can't merge as-is. See below.

All three points from my last review are addressed: `GreaterOrEqual(seq, 0)` is now a real equality assertion, the redundant fixture-zip extraction is gone in favor of `createTestApiWithGTFSFixture`, and the "SCHEDULED vehicle" claim is asserted rather than narrated.

I checked the parts that could have made this test quietly meaningless and they're sound. `InitGTFSManager` does a synchronous first RT fetch before `MarkReady()`, so there's no race. The fixture has no `shapes.txt`, which means `BuildTripStatus` skips shape projection and `status.Position` stays the raw GPS fix — so the `float64(float32(...))` expectation is correct rather than accidentally correct. And the `situationIds`/`vehicleFeatures` assertions are load-bearing, since decoding `null` into `[]string` yields nil and `require.NotNil` distinguishes `[]` from `null`.

Two things blocking the merge, neither of them the code:

1. The branch conflicts with `main` — that file gained ~198 lines after your 08-11 rebase, touching the import block and the fixture region you also edit. Mechanical to resolve, but it needs doing.
2. burma-shave has a change request still open here. My approval doesn't dismiss it, so please square that before merging.

Two notes for the rebase:

- `createTestApiWithScheduledRealtimePosition` takes a `clock.Clock` parameter but hardcodes `tripsForRouteTestClock` for the feed header and vehicle timestamps. A future caller passing a different clock more than 15 minutes away would silently trip `StaleDetector` and get `default`/`scheduled` instead of a clear failure. Deriving the timestamp from `c.Now()` removes the trap.
- The `occupancyCount` assertion's message says "when the feed omits occupancy data," but `NewTripStatus()` sets -1 unconditionally and nothing on this path overwrites it — the assertion pins a constructor default, not feed-driven behavior.

One coordination note: #1298 adds a second real-time fixture builder to this same file using a different injection mechanism (`SetRealTimeVehiclesForTest` rather than an httptest feed). If both land we end up with two overlapping patterns — worth settling which one wins before the second merges.

burma-shave (changes requested):
conflicts need to be resolved again

burma-shave:

burma-shave:

burma-shave (approved):

PR #1305, opened 2026-08-06
Keep operator words in stop search input

### LINKED ISSUE #1303: Stop dropping FTS5 operator words from user input
### Description

**Spec Requirement:**
Prefix matching applies to any word the user types — "a value of `pine` matches any stop
whose name contains a token starting with `pine`" (*Request Parameters*). The wiki does
not carve out an exception for particular words. Confirmed against the live OBA server:
`GET /api/where/search/stop.json?input=or` returns HTTP 200 with 20 stops
("Orion Dr at Polaris Ln", …).

**Current Maglev Behavior:**
`fts5OperatorsRegex` (`internal/restapi/search_stops_handler.go:22`) strips `AND`, `OR`,
`NOT` and `NEAR` as whole words during sanitization. `input=or` produces zero terms and
returns HTTP 200 with an empty list; `input=or lake` silently searches only `lake`.

RABA fixtures already contain affected stops — `1100` "Browning St at Mission De Oro Dr",
`1062` "Shasta Dam Blvd at 5001 Oregon", `9030` "Balls Ferry Blvd at Anderson Square (MB)",
`4404` "Shingletown Park and Ride Lot" — all currently unreachable by `input=or` or
`input=and`.

The stripping is unnecessary: `sanitizeFTS5Query` runs before each term is wrapped in
double quotes, and FTS5 only treats these as operators when they appear unquoted.
`"or"* AND "lake"*` is already a valid query that matches the literal token.

**Proposed Solution:**
- Remove `fts5OperatorsRegex` and its use in `sanitizeFTS5Query`. Special-character
  removal stays as-is.
- Update `TestSanitizeFTS5Query` — the `AND`/`OR`/`NEAR` cases currently assert the
  stripping behaviour.
- Add handler tests asserting `input=or` and `input=and` return the RABA stops above.

ARCoder181105:
Closes #1303.

## Problem

`sanitizeFTS5Query` stripped `AND`, `OR`, `NOT`, `NEAR` as whole words. Query `or` produced
zero terms and returned HTTP 200 with an empty list, making stops with those words
unreachable.

The stripping was never needed: every term is wrapped in double quotes before the query is
assembled, so FTS5 reads them as literal tokens. The sibling `search/route.json` path
(`buildRouteSearchQuery`) does no operator filtering for the same reason.

## Changes

- Removed `fts5OperatorsRegex` and its call from `sanitizeFTS5Query`. Special-character
  removal unchanged.
- Documented why operator words are safe to keep.

## Verification against deployed OBA

Same King County Metro feed, local instance vs `api.pugetsound.onebusaway.org`:

| Query | Deployed | Before | After |
|---|---|---|---|
| `or` | 200, n=20 | 200, n=0 | 200, n=20 |
| `and` | 200, n=20 | 200, n=0 | 200, n=20, same first four names |
| `broadway and` | 200, n=1 | 200, n=0 | 200, n=1, same stop |
| `1st ave & spring` | 200, n=1 `1_100` | 200, n=1 `1_100` | unchanged |

`not` and `1st ave and spring` return 404 upstream, 200 with empty list here — pre-existing
no-match deviation, not touched by this PR.

## Behavior change

`1st ave and spring` no longer matches "1st Ave & Spring St" — the query now requires an
`and` token. Deployed server returns 404 for it, so this moves toward parity.

## Tests

- Added `TestSearchStopsHandlerOperatorWordInput` over `or`, `and`, `not`, `near` using
  existing RABA fixtures. `or` reaches `25_1062`, `and` reaches `25_4404`.
- Eight `TestSanitizeFTS5Query` cases now assert operator words are preserved, renamed
  accordingly. Special-character cases untouched.
- `TestSearchStopsHandlerFTSInjectionAttempt` passes unmodified.

## Test plan

- [x] `go vet -tags "sqlite_fts5 sqlite_math_functions" ./...`
- [x] `go vet -tags "purego" ./...`
- [x] `make test`
- [x] `go fmt ./...` (no changes)

## Summary by CodeRabbit

* **Bug Fixes**
  * Search queries now preserve operator words such as “AND,” “OR,” and “NEAR.”
  * Special characters continue to be sanitized and whitespace normalized.
  * Searches using these terms now return successful, relevant results.


ARCoder181105:
@coderabbitai full review

Ahmedhossamdev (commented):
LGTM!

PR #1313, opened 2026-08-07
Make situation references resolve on every endpoint

### LINKED ISSUE #1307: trips-for-location: references.situations is never populated
Part of #1051.

## Problem

`trips-for-location` emits `situationIds` on list entries, but the `references.situations` array is always empty. Clients get IDs that resolve to nothing.

## Current behaviour

`internal/restapi/trips_for_location_handler.go`

- Each entry sets `SituationIds: api.GetSituationIDsForTrip(ctx, tripID)` (line 415).
- `referenceBuilder.toReferencesModel()` (line 759) assigns only `Agencies`, `Routes`, `Stops`, `Trips`. `Situations` stays as the empty slice from `NewEmptyReferences()`.

## Spec requirement

Wiki spec, `data.references`: `situations` — service alert beans for the situation IDs appearing in the response.

## Proposed fix

Reuse the existing helper `api.BuildSituationReferences(alerts)` (`internal/restapi/reference_utils.go:164`), as nine peer handlers already do (`routes_for_location_handler.go:77`, `stops_for_location_handler.go:252`, `route_handler.go:67`, `trip_details_handler.go:341`).

- Collect the alerts matched while building entries, dedupe by alert ID.
- Assign `references.Situations` in `toReferencesModel`.

No new DB query needed.

## Test

Assert `references.situations` is non-empty whenever any entry has a non-empty `situationIds`, and that every emitted ID resolves to a situation. May require adding a service-alerts fixture to `createTestApiWithRealTimeData`.

ARCoder181105:
Closes #1307.

`situationIds` resolved to nothing on every endpoint that emits them. Three never populated `references.situations` at all; three populated it with raw alert IDs while emitting combined ones. The IDs were also double-prefixed — the feed ships `1_92239`, maglev published `1_1_92239`.

## Changes

- Add `TripSituations` to `reference_utils.go`: a trip's situation IDs and their references from one lookup, so the two cannot disagree.
- Populate `references.situations` on `trips-for-location`, `trips-for-route` and `trip-for-vehicle`, which never did; point the other three endpoints at the shared helper.
- Make agency prefixing idempotent: an ID already carrying this agency's prefix is left alone. maglev now emits `1_92239`, matching upstream.
- Scope a situation ID to the alert's own informed entity rather than to whichever lookup found it, so an alert reachable through both a route and a stop of different agencies gets one ID instead of two.
- Return the situations `BuildTripStatus` already resolved, so `trip-details`, `trip-for-vehicle` and `arrival-and-departure-for-stop` reuse them instead of resolving the same trip twice. `GetSituationIDsForTrip` goes with its last caller.
- `trips-for-route`: index a DUPLICATED trip before its situation lookup, and resolve the agency when a trip's route is absent from `routeAgencyMap` rather than emitting an unprefixed ID.

## Handler consistency

`trip-for-vehicle` gained `includeReferences` handling on main while this branch was open. Its situation references are assigned inside that gate, so `includeReferences=false` now returns an empty `situations` array along with the rest of the block — matching `trip-details`. The entry's `situationIds` are unaffected.

## Deviation from Java

Java prefixes unconditionally in `GtfsRealtimeSource.createId`, and every other ID in maglev goes through `FormCombinedID`. The idempotent check here is deliberate — it is what makes maglev's output match upstream on the feeds that exist, which export alert IDs already prefixed. It has a known edge: an alert whose raw ID happens to begin with the agency ID plus an underscore is left unprefixed, a plausible shape for a numeric agency.

## Verification

Live against the King County feed with a Puget Sound alerts feed: every emitted `situationId` resolves on all six endpoints, and the IDs match upstream exactly (`1_92239`).

`TestSituationIDsResolveToReferences` asserts resolution across all six endpoints, including with `includeStatus=false`; `TestSituationRefsFromAlertsAgencyScope` covers the cross-agency ID; `TestTripSituationRefsAgencyFallback` covers the unmapped-route fallback. `go vet` (both tag sets) and `make test` pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured all situation IDs returned in trip, stop, vehicle, route, and location responses resolve to matching situation references.
  * Improved agency-aware situation ID generation, including cross-agency and unknown-agency alerts.
  * Deduplicated situation references while preserving associated list-entry identifiers.
* **Tests**
  * Added coverage across trip, stop, vehicle, route, and location responses to validate alert and situation reference consistency.


ARCoder181105:
@coderabbitai full review

ARCoder181105:
Found the issue to be a global , I tried to make the changes complete for other endpoint as well so that a final PR fix be made , the reason the PR went this large...

Ahmedhossamdev:
Verified the premise of the prefix fix against real data rather than taking it on trust: pulled the live Puget Sound alerts export (/api/gtfs_realtime/alerts-for-agency/1.pb) and protoc --decode_raw confirms FeedEntity.id already carries the agency prefix (1_86736, 1_89529), while the informed entities inside carry raw IDs. So 1_1_86736 was real and 1_86736 is the right output. The one-source approach in TripSituations also matches BeanFactoryV2.getTripDetails, which builds situationIds and references.situations from the same bean list.

One thing to fix: situationCollector dedupes on the agency-scoped ID rather than the alert ID, so at a stop served by a route from another agency the same alert can be emitted twice under two IDs. Details inline. The singular arrival-and-departure-for-stop also changed without a test, and it is one more row in the table you already wrote.

internal/restapi/arrivals_and_departures_for_stop_handler.go
```diff
@@ -444,17 +443,7 @@ func (api *RestAPI) arrivalsAndDeparturesForStopHandler(w http.ResponseWriter, r
...
-			if alert.ID == "" {
-				continue
-			}
-
-			situationIDs = append(situationIDs, utils.FormCombinedID(route.AgencyID, alert.ID))
-			if _, seen := collectedAlerts[alert.ID]; !seen {
-				collectedAlerts[alert.ID] = alert
-			}
-		}
+		situationIDs := situations.add(tripAlerts, route.AgencyID)
```
Ahmedhossamdev, line 446: > Must fix. `situationCollector` dedupes on `ref.ID`, which is agency-scoped, but
> trip alerts come in under `route.AgencyID` here and stop alerts under
> `alertAgencyID` at line 621. At a stop served by a route from another agency,
> one alert arrives on both paths with two different IDs and lands in
> `references.situations` twice. The old `collectedAlerts` map deduped on the raw
> alert ID so this could not happen. KCM alert `1_86736` has informed entities of
> `{agency: 1, route: 100264, stop: 10190}`, so it hits both `alertIdx.byRoute`
> and `alertIdx.byStop`.
>
> Java avoids this by fixing the alert ID at ingest from the feed's agency
> (`GtfsRealtimeSource.createId`), never from whichever route matched. Resolving
> the agency once per alert, from its informed entity or the feed's
> `agency-ids[0]`, and passing that into `situationRefsFromAlerts` would give an
> alert one ID on every path. Deduping on `alert.ID` alone would fix the
> duplicate but leave entries pointing at whichever prefix lost.

internal/restapi/situation_references_test.go
```diff
@@ -0,0 +1,192 @@
...
+	rawAgencyID := "25"
+	api.GtfsManager.AddAlertForTest(gogtfs.Alert{
+		ID:               "situation-resolution-alert",
+		InformedEntities: []gogtfs.AlertInformedEntity{{AgencyID: &rawAgencyID}},
+		Header:           []gogtfs.AlertText{{Text: "Test Agency Alert", Language: "en"}},
+	})
+
+	tripID, stopID := anyTripAndStop(t, api)
+	vehicleID := anyRealTimeVehicleID(t, api)
+
```
Ahmedhossamdev, line 87: Add a case for arrival-and-departure-for-stop. The singular handler changed at lines 401 and 598 and is the only one of the six endpoints with no new coverage. It needs a tripId and serviceDate alongside the stop, so it does not drop into the table as cleanly as the others, but the assertion body is the same.

internal/restapi/reference_utils.go
```diff
@@ -448,3 +451,129 @@ func (api *RestAPI) buildStopModel(ctx context.Context, agencyID string, stop gt
...
+	return refs
+}
+
+// situationID returns the combined-form ID for an alert.
+//
+// Prefixing is idempotent, not unconditional: a GTFS-RT feed exported by an OBA
+// instance already carries the agency prefix on its alert IDs — Puget Sound
+// ships "1_92239" — and prefixing again would publish "1_1_92239", which
+// resolves against nothing a client has seen. Upstream reports "1_92239".
+func situationID(alertID, agencyID string) string {
```
Ahmedhossamdev, line 504: Low priority, no change needed. Worth calling out in the PR description that this diverges from Java, which prefixes unconditionally in GtfsRealtimeSource.createId, and from every other ID in maglev, which goes through FormCombinedID. The prefix check also skips an alert whose raw ID starts with the agency ID plus an underscore, which is a plausible shape for a numeric agency. Right call for the feeds that exist, just not self-evident from the code.

internal/restapi/arrivals_and_departures_for_stop_handler.go
ARCoder181105: Fixed in 9185418. Took the agency from the alert's own informed entity instead of deduping on alert.ID, so the route path and the stop path derive the same ID rather than one prefix winning. Applied in situationRefsFromAlerts, so it covers every call site, not just this handler.

Skipped the feed agency-ids[0] half of the suggestion: alerts are merged across feeds before any handler sees them, so plumbing feed identity through is a much bigger change than the informed-entity path it would only back up. The remaining gap is an alert that names no agency entity at all it still falls back to the caller's agency and can diverge. Doesn't apply to the alerts you decoded, which all carry one.

`TestSituationRefsFromAlertsAgencyScope` (14cdf96) runs both caller agencies through one collector and asserts the collected IDs: one entry when the alert names an agency, two when it doesn't, so the limitation is pinned rather than implied.

internal/restapi/situation_references_test.go
ARCoder181105: Added in 181a084  dropped into the table as you said, taking the trip anyTripAndStop already returns for the stop plus the pinned fixture date as serviceDate. The assertions were also tightened in 14cdf96 to require the seeded alert's own ID rather than just any situationId.

internal/restapi/reference_utils.go
ARCoder181105: Written up in the PR description under "Deviation from Java", including the edge you flagged where an alert ID beginning with the agency ID plus an underscore is left unprefixed.

internal/restapi/trip_details_handler.go
```diff
@@ -254,12 +254,9 @@ func (api *RestAPI) tripDetailsHandler(w http.ResponseWriter, r *http.Request) {
...
 
-	var situationsIDs []string
-	if status != nil && len(status.SituationIDs) > 0 {
-		situationsIDs = status.SituationIDs
-	} else {
-		situationsIDs = api.GetSituationIDsForTrip(r.Context(), tripID)
-	}
+	// trip is looked up by tripID, and BuildTripStatus was given trip.ID, so the
+	// status carries the same situations this resolves.
+	situationsIDs, situationRefs := api.TripSituations(ctx, tripID)
```
burma-shave: `TripSituations` here re-resolves the same trip's alerts that `BuildTripStatus` (line 232) already resolved via `GetSituationIDsForTrip` → `situationRefsForTrip`, which does its own `GetTrip`+`GetRoute` lookup. `trip.ID` and `tripID` are the same value — the comment right above this line already notes that — so this doubles the DB round-trip for every trip-details request, the same class of duplicate lookup this PR's own description says it removed elsewhere. Consider resolving `situationRefsForTrip` once and threading the result into both `status.SituationIDs` and the entry/reference building, the way `trips_for_route_handler.go`'s `tripSituationRefs` already avoids re-querying for trips it has already loaded.

burma-shave (commented):
Stage 2 code-quality pass on top of the situation-references fix. Three follow-ups on the redundant-lookup theme this PR already tackles, plus one narrow edge case to flag for awareness.

Stage 1 (spec-conformance) is settled separately: the idempotent agency-prefixing, the informed-entity agency-scoping, and the pre-existing situationIds absent-vs-empty-array behavior were all reviewed and are being kept as-is, documented in maglev.wiki's Implementation Decisions sections rather than changed here.

internal/restapi/trip_for_vehicle_handler.go
```diff
@@ -109,12 +109,9 @@ func (api *RestAPI) tripForVehicleHandler(w http.ResponseWriter, r *http.Request
...
 
-	var situationIDs []string
-	if status != nil && len(status.SituationIDs) > 0 {
-		situationIDs = status.SituationIDs
-	} else {
-		situationIDs = api.GetSituationIDsForTrip(r.Context(), tripID)
-	}
+	// BuildTripStatus above was given this same tripID, so the status carries
+	// the same situations this resolves.
+	situationIDs, situationRefs := api.TripSituations(ctx, tripID)
```
burma-shave: Same redundant lookup as in `trip_details_handler.go`: `TripSituations(ctx, tripID)` re-resolves what `BuildTripStatus` (line 73) already resolved for the identical `tripID`, doubling `GetTrip`+`GetRoute` per request. The comment above ("BuildTripStatus above was given this same tripID...") documents the coincidence without acting on it — worth the same fix as the sibling handlers.

internal/restapi/arrival_and_departure_for_stop_handler.go
```diff
@@ -398,7 +398,7 @@ func (api *RestAPI) arrivalAndDepartureForStopHandler(w http.ResponseWriter, r *
 	blockTripSequence := api.calculateBlockTripSequence(ctx, tripID, serviceMidnight)
 
 	lastUpdateTime := api.GtfsManager.GetVehicleLastUpdateTime(vehicle)
-	situationIDs := api.GetSituationIDsForTrip(r.Context(), tripID)
+	situationIDs, situationRefs := api.TripSituations(ctx, tripID)
```
burma-shave: Same pattern again: `BuildTripStatus` (line 360) already resolves this trip's situations internally, and `TripSituations(ctx, tripID)` here re-resolves the identical trip right after. The PR description says the duplicate trip lookup was dropped for this handler — this is a *different* duplicate lookup than the one that was fixed, and it's still here.

internal/restapi/trips_helper.go
```diff
@@ -819,6 +819,17 @@ func distanceToLineSegment(px, py, x1, y1, x2, y2 float64) (distance, ratio floa
 }
 
 func (api *RestAPI) GetSituationIDsForTrip(ctx context.Context, tripID string) []string {
```
burma-shave: Now that `TripSituations` (`reference_utils.go`) runs the same `situationRefsForTrip` call and extracts IDs the same way, this could delegate — `ids, _ := api.TripSituations(ctx, tripID); return ids` — instead of re-implementing the extraction loop. Minor, but two copies of the same loop is one more place a future change to ID derivation has to be applied twice.

internal/restapi/trips_for_route_handler.go
```diff
@@ -464,7 +481,7 @@ func (api *RestAPI) tripsForRouteHandler(w http.ResponseWriter, r *http.Request)
 			Schedule:     schedule,
 			Status:       status,
 			ServiceDate:  todayMidnight.UnixMilli(),
-			SituationIds: api.GetSituationIDsForTrip(r.Context(), baseTripID),
+			SituationIds: situations.addRefs(api.tripSituationRefs(ctx, baseTripID, tripsByID, routeAgencyMap)),
```
burma-shave, line 497: This `tripSituationRefs` call always misses the `tripsByID` cache for a DUPLICATED trip, because `baseTrip` is only fetched and added to `fetchedTrips`/`filteredRouteTrips` a few lines below (around line 490) — and even there it's never added to `tripsByID`. Compare the interlined-trip path above (line ~375), which explicitly caches `resolution.SelectedTrip` into `tripsByID` with a comment explaining exactly why: an unindexed trip sends situation lookups back to the database. Every DUPLICATED trip therefore falls back to a full `situationRefsForTrip` DB round-trip here, even though `baseTrip` gets fetched moments later for that same ID by the `GetTrip` call below. Resolving `baseTrip` before this call (instead of after) and adding it to `tripsByID` would let this reuse the same cache the interlined path already relies on, and would also remove the separate `GetTrip` call below for the same trip.

internal/restapi/trips_for_route_handler.go
```diff
@@ -691,156 +715,182 @@ func collectStopIDsFromSchedule(schedule *models.TripsSchedule, stopIDsMap map[s
...
+	tripID string,
+	tripsByID map[string]gtfsdb.Trip,
+	routeAgencyMap map[string]string,
+) []situationRef {
+	trip, ok := tripsByID[tripID]
+	if !ok {
+		return api.situationRefsForTrip(ctx, tripID)
+	}
+
+	agencyID := routeAgencyMap[trip.RouteID]
```
burma-shave: If a trip is present in `tripsByID` but its route isn't in `routeAgencyMap`, `agencyID` silently resolves to `""` here, and `situationID()` returns the bare, unprefixed alert ID instead of the combined form the rest of the response uses — no error, just an ID shaped differently, with a theoretical collision risk against a same-named bare ID reached via a different agency's alert on a different path. The `tripsByID`-miss branch just above already has a correct fallback (`situationRefsForTrip`, which resolves the route/agency itself); this branch could use the same fallback when `routeAgencyMap` lacks an entry instead of silently falling through with an empty agency. Reachability depends on whether `GetRoutesByIDs` can ever return fewer routes than `routeIDSet` for a route id a trip actually references — flagging for a second look, not necessarily blocking.

burma-shave (changes requested):

Ahmedhossamdev:
Hey @ARCoder181105 
Could you fix the merge conflicts?

internal/restapi/trip_details_handler.go
ARCoder181105: Addressed in 6033fa4. BuildTripStatus now returns the situations it
resolved, and this line reuses them instead of re-querying. Falls back to
its own lookup only when includeStatus was false and no status was built.

internal/restapi/trip_for_vehicle_handler.go
ARCoder181105: Addressed in 6033fa4, same fix as the sibling handlers — the situations
come back from BuildTripStatus now, so the second lookup is gone.

internal/restapi/arrival_and_departure_for_stop_handler.go
ARCoder181105: Addressed in 6033fa4. This handler calls BuildTripStatus unconditionally,
so it always reuses the resolved situations — no fallback path here.

internal/restapi/trips_helper.go
ARCoder181105: Resolved in 6033fa4. GetSituationIDsForTrip had only one caller, the
status line inside BuildTripStatus; with the references now reaching
callers intact it was deleted, and the shared situationIDsFromRefs is the
single extraction point.

internal/restapi/trips_for_route_handler.go
ARCoder181105: Addressed in 8264b72. baseTrip is resolved before the situation lookup and
indexed into tripsByID, so this reuses the same cache the interlined path
does. The separate GetTrip below is removed.

internal/restapi/trips_for_route_handler.go
ARCoder181105: Addressed in 45f26fd — this branch now takes the same situationRefsForTrip
fallback as the tripsByID miss above.

Worth noting the impact is larger than an ID-shape difference: with an
empty agency the alert isn't matched at all. Reverting the fix against a
constructed unmapped-route case returns no refs rather than a bare ID.
That's from a unit test, not proof GetRoutesByIDs can drop a referenced
route in production — but the fallback is cheap and the failure mode is
silent, so it seemed worth closing either way.

ARCoder181105:
@Ahmedhossamdev done resolved the merge conflict also addressed the requested changes ready for review

burma-shave (approved):

PR #1315, opened 2026-08-07
Fix schedule timezone and outOfRange bounds in trips-for-location

### LINKED ISSUE #1310: trips-for-location: schedule.timeZone and outOfRange use wrong inputs
Part of #1051. Two small coupled cleanups in the same handler path.

## 1. `schedule.timeZone` always uses the first agency

**Current**: `parseAndValidateRequest` resolves the timezone once from `agencies[0]` (`trips_for_location_handler.go:161-165`) and stamps it on every trip's schedule (line 818). Wrong for multi-agency feeds spanning timezones.

**Spec**: `schedule.timeZone` is the IANA identifier for the trip's own agency.

**Fix**: `tripAgencyMap` is already built at lines 88-98 — resolve the timezone per trip's agency instead of using the global one.

## 2. `outOfRange` computed from unclamped bounds

**Current**: `CheckIfOutOfBounds` calls `BoundsFromParams(loc)` with no clamp flag (`internal/gtfs/location_params.go:62`), while the search path passes `clamp=true`. So `radius=5000000` reports `outOfRange: false` from anywhere on earth while actually searching only 20 km.

**Spec**: search area is clamped to 20 km; `outOfRange` reports whether that area falls outside the service area.

**Fix**: pass the clamp flag through in `CheckIfOutOfBounds` so both paths use the same bounds.

## Test

- `BoundsFromParams` clamp parity in `internal/gtfs/location_params_test.go`.
- Table-driven unit test on the schedule builder for the per-agency timezone (cheaper than adding a second agency to the RABA fixture).

ARCoder181105:
Closes #1310.

Depends on #1314 (and transitively #1313) and includes their commits; merge those first.

Two correctness bugs in the same handler path:

- `schedule.timeZone` came from `agencies[0]` and was stamped onto every trip, so on a feed spanning timezones most trips reported the wrong one.
- `outOfRange` was computed from unclamped bounds while the search clamped to 20 km, so `radius=5000000` reported `outOfRange: false` from anywhere on earth.

## Changes

- Resolve every agency's timezone once per request and look up the trip's own agency when building its schedule. The query `time` is still interpreted in the first agency's zone.
- Give `CheckIfOutOfBounds` the same variadic `clamp` flag `BoundsFromParams` has, and pass `true` from trips-for-location. `routes-for-location` and `stops-for-location` do not clamp their searches, so they keep unclamped behaviour and stay self-consistent.
- Pass the parsed request struct into `buildTripsForLocationEntries` instead of eleven positional parameters, five of which were unpacked from it at the single call site. No behaviour change; it is what keeps the timezone map from becoming a twelfth.

## Verification

Live: LA with `radius=2000000` now reports `outOfRange: true`, previously `false`.

The timezone fix is covered by unit tests rather than the fixture — King County is single-agency, so it cannot exercise a multi-timezone feed. `TestCheckIfOutOfBoundsClamping`, `TestTripsForLocationRequest_ScheduleLocation`, `TestAgencyLocations`.

`go vet` (both tag sets) and `make test` pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved location-based trip searches by consistently applying search boundary limits.
  * Corrected out-of-range indicators for large-radius searches, preventing misleading results when searches overlap distant areas.
  * Added coverage to verify behavior with both clamped and unclamped search bounds.

### DIFF
--- a/internal/gtfs/location_params.go
+++ b/internal/gtfs/location_params.go
@@ -53,13 +53,17 @@ func BoundsFromParams(loc *LocationParams, clamp ...bool) utils.CoordinateBounds
 
 // CheckIfOutOfBounds returns true if the user's search area is completely
 // outside every agency's region bounds.
-func (manager *Manager) CheckIfOutOfBounds(loc *LocationParams) bool {
+//
+// clamp must match what the caller passed to BoundsFromParams when it ran the
+// search. Reporting on unclamped bounds while searching clamped ones lets an
+// oversized radius overlap a region it never actually searched.
+func (manager *Manager) CheckIfOutOfBounds(loc *LocationParams, clamp ...bool) bool {
 	boundsMap := manager.GetRegionBounds()
 	if len(boundsMap) == 0 {
 		return false
 	}
 
-	innerBounds := BoundsFromParams(loc)
+	innerBounds := BoundsFromParams(loc, clamp...)
 
 	for _, region := range boundsMap {
 		outerBounds := utils.CalculateBoundsFromSpan(region.Lat, region.Lon, region.LatSpan/2, region.LonSpan/2)
--- a/internal/gtfs/location_params_test.go
+++ b/internal/gtfs/location_params_test.go
@@ -154,3 +154,19 @@ func TestCheckIfOutOfBounds(t *testing.T) {
 		assert.False(t, manager.CheckIfOutOfBounds(params))
 	})
 }
+
+func TestCheckIfOutOfBoundsClamping(t *testing.T) {
+	manager := &Manager{
+		regionBounds: map[string]*RegionBounds{
+			"agency1": {Lat: 40.0, Lon: -70.0, LatSpan: 1.0, LonSpan: 1.0},
+		},
+	}
+	// Seattle, with a radius wide enough to reach the east-coast region only if
+	// the radius is left unclamped.
+	params := &LocationParams{Lat: 47.6, Lon: -122.3, Radius: 5_000_000}
+
+	assert.False(t, manager.CheckIfOutOfBounds(params),
+		"unclamped bounds still overlap the distant region")
+	assert.True(t, manager.CheckIfOutOfBounds(params, true),
+		"clamped to 20km the search never reaches the region, so it is out of range")
+}
--- a/internal/gtfs/location_params_test.go
+++ b/internal/gtfs/location_params_test.go
@@ -165,8 +165,25 @@ func TestCheckIfOutOfBoundsClamping(t *testing.T) {
 	// the radius is left unclamped.
 	params := &LocationParams{Lat: 47.6, Lon: -122.3, Radius: 5_000_000}
 
-	assert.False(t, manager.CheckIfOutOfBounds(params),
-		"unclamped bounds still overlap the distant region")
-	assert.True(t, manager.CheckIfOutOfBounds(params, true),
-		"clamped to 20km the search never reaches the region, so it is out of range")
+	tests := []struct {
+		name  string
+		clamp []bool
+		want  bool
+	}{
+		{
+			name: "Unclamped bounds still overlap the distant region",
+			want: false,
+		},
+		{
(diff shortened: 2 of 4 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Approved on the merits — both fixes are correct and narrowly scoped.

The clamp flag is right: the handler searches with `BoundsFromParams(..., true)` and `GetStopsInBounds(..., true)`, so computing `outOfRange` against the unclamped bounds was genuinely inconsistent with what was searched. I checked the other two `CheckIfOutOfBounds` callers — routes-for-location and stops-for-location — and neither clamps its search, so leaving the variadic default alone keeps them self-consistent. The per-agency timezone resolution is only consumed for the `TimeZone` string in `buildScheduleFromMemory`, so no arrival/departure math moves and there's no DST exposure.

Two things for a follow-up rather than this PR:

- `agencyLocations` calls `time.LoadLocation` for every agency on every request, and Go doesn't memoize it. On a multi-agency feed that's N tzdata parses per hit on a load-tested endpoint, and it runs even when `includeSchedule=false`, where the map is never read. Worth resolving lazily or caching on the manager.
- `serviceDate` is still the first agency's midnight while `timeZone` is now per-agency, so a multi-timezone feed reports e.g. `America/Chicago` alongside an LA-midnight `serviceDate`. Not a regression — it was equally wrong before — but the timezone fix is only half the multi-timezone story. I'll open an issue.

Ordering note: this can't land before #1313 and #1314. It's approved and ready whenever the stack clears; no re-review needed unless a rebase changes behavior.

ARCoder181105:
Thanks — nothing changed here on rebase. The three commits are untouched and still touch only the same four files; the endpoint moved underneath from #1314's query narrowing and stop-reference IDs, reviewed there.

Both follow-ups noted:
- `agencyLocations` eager `LoadLocation` — agreed, and it running under `includeSchedule=false` where the map is dead is the worse half. Happy to take it once the stack clears.
- `serviceDate` vs per-agency `timeZone` — agreed it's half the multi-timezone story. Will pick up your issue.

soumajitgh:
I opened #1329 to cover the remaining multi-timezone service-date problem tracked in #1325. It keeps schedule.timeZone and serviceDate on the same trip agency timezone, and also uses that local service date for block-schedule lookups. If #1315 is preferred as the integration point, these changes can be incorporated there before merge.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Re-approving on the merits — this looks good, but it can't merge until the stack below it clears.

I re-reviewed after the force-push rather than assuming nothing moved, and your claim checks out: `location_params.go` is byte-identical to the SHA I approved, and the only genuinely new commit converts `TestCheckIfOutOfBoundsClamping` to table-driven form, which is what CONTRIBUTING asks for anyway.

Both fixes still read correctly. The per-trip `schedule.timeZone` only feeds `TimeZone: currentLocation.String()` — stop times come through as durations since midnight, so there's no epoch or DST exposure. And computing `outOfRange` from the same 20 km clamped bounds the search actually used matches the precedent already set in `stops_for_location_handler.go`; leaving the other two callers on the variadic default keeps them self-consistent. The clamping test genuinely discriminates: at radius 5,000,000 m the unclamped bounds overlap the seeded region and the clamped box can't.

What's blocking:

The branch is conflicting with `main`, and notably none of your own four files are the problem — the conflicts are all in files inherited from #1313. That stack (#1313 → #1314 → #1315) has to land in order, and #1313 and #1314 both still have changes requested. Resolving those two should clear this one without further work from you here.

The two follow-ups we agreed to defer are still deferred and still fine as such: the eager `time.LoadLocation` for every agency, and the `serviceDate`-versus-per-agency-`timeZone` mismatch. On that second one — I've just merged #1329, which fixes exactly that by resolving midnight in each trip's own agency timezone. It touches `buildTripsForLocationEntries`, so expect to reconcile with it when you rebase.

burma-shave (approved):

PR #1328, opened 2026-08-09
Correct route-type truncation comments

### LINKED ISSUE #1322: Stale truncation comments in GetStopsForLocation after #1272
#1272 removed the truncation from the ORDERED_BY_CLOSEST path in `Manager.GetStopsForLocation` — route-type searches are now capped downstream in `stopsForLocationHandler`, after stops with no service on the queried date are dropped.

Two comments still describe the old behavior:

- The doc comment on `GetStopsForLocation`: "ORDERED_BY_CLOSEST mode (routeTypes present): sorts by distance, filters by route type, then truncates."
- The inline comment on the same branch: "ORDERED_BY_CLOSEST mode: sort by distance, filter by route type, then truncate."

Neither truncates any more, and `maxCount` plus the returned `limitExceeded` are now dead for that mode. The doc comment was added deliberately in `422b22df` to record the ordering, so leaving it inverted actively misleads.

Fix: update both comments to say the cap is applied by the caller, and note that `limitExceeded` is only meaningful for BOUNDS mode.

soumajitgh:
Correct the stale truncation comments for route-type stop searches.

#1272 moved route-type capping to the handler, after inactive stops are
removed. The manager comments still described the old behavior.

Closes #1322

## Summary by CodeRabbit

* **Documentation**
  * Clarified how location-based stop results are sorted and filtered.
  * Documented that result limits apply differently depending on the search mode.

### DIFF
--- a/internal/gtfs/gtfs_manager.go
+++ b/internal/gtfs/gtfs_manager.go
@@ -315,7 +315,9 @@ func (manager *Manager) RoutesForAgencyID(ctx context.Context, agencyID string) 
 //
 // GetStopsForLocation is used by the stops-for-location endpoint.
 // BOUNDS mode (no routeTypes): shuffles stops then truncates before route-type filtering.
-// ORDERED_BY_CLOSEST mode (routeTypes present): sorts by distance, filters by route type, then truncates.
+// ORDERED_BY_CLOSEST mode (routeTypes present): sorts by distance and filters by route type.
+// The caller caps results after dropping stops without service on the queried date;
+// limitExceeded is therefore only meaningful for BOUNDS mode.
 func (manager *Manager) GetStopsForLocation(
 	ctx context.Context,
 	loc *LocationParams,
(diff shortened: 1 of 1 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. The commit body is a single 133-character line rather than wrapped prose (CONTRIBUTING.md says "Wrap the body at 72 characters"). The comment change itself checks out: `RoutesForAgencyID`/`GetStopsForLocation`'s ORDERED_BY_CLOSEST branch no longer truncates — #1272 moved that cap into `stopsForLocationHandler`, after stops with no service on the queried date are dropped — so the old "then truncates" wording was stale and the replacement matches the code.

https://github.com/OneBusAway/maglev/blob/262b7ca6c0d12884573f20aa55a2a92e24600127/CONTRIBUTING.md#L21-L23

aaronbrethorst (approved):

I verified the claim rather than just reading it: PR #1272 removed the `stops = stops[:maxCount]` from the ORDERED_BY_CLOSEST branch and moved the cap into `stopsForLocationHandler`, where it fires only when `len(routeTypes) > 0` and only after stops with no service on the queried date are dropped. So the old "then truncates" comment was genuinely stale, and the replacement describes what the code actually does.

Correcting a comment that has drifted from the code is real work, and this is exactly the kind of thing that saves the next person a confusing half hour.

One small thing for next time, no need to respin: the commit body is a single 133-character line, and CONTRIBUTING asks for it wrapped at 72.

PR #1331, opened 2026-08-12
Memoize repeated block and calendar lookups per request

dev-aditya-hub:
## Why

`BuildTripStatus` resolves a trip's block to compute position, distance and schedule deviation. For a handler that calls it once per response row, that resolution repeats for every row, and each one reloads the stop times and shape points of every trip in the block. The work ends up scaling with (rows x block size) rather than with the number of distinct blocks.

There is a second, smaller repeat in the same path: `GetActiveServiceIDsForDate` is called once per row for a date that cannot change within a request.

This is the same problem `WithSnapshotCache` already solves for whole block snapshots on the plural arrivals handler. These entries sit one level below it, and are still reached on the schedule-deviation path, which resolves a block without going through a snapshot.

## What this changes

A request-scoped memo in `internal/restapi/request_cache.go`, carried on the context, holding three things:

- `loadBlockTripData`, keyed on the trip-ID set. This deduplicates across vehicles sharing a block, and also within a single `BuildTripStatus` call, which resolves the same block twice (once via `computeScheduledBlockSnapshot`, once via `blockShiftTripIDsSortedByStartTime`).
- `GetActiveServiceIDsForDate`, keyed on the date. I only swapped the three hot call sites (`blockTripIDsForServiceDate`, `blockTripSequence`, `activeTripInBlockAt`); the other nine are untouched.
- Stop times, seeded by `prefetchStopTimes`, so a handler can resolve every row's stop times in one query instead of one per row.

The memo is request-scoped rather than a field on `RestAPI` on purpose. Static GTFS data reloads in the background, and anything longer-lived would need invalidation tied to that reload. A handler that does not opt in gets the existing behaviour unchanged, since every accessor falls through to the original query when the context carries no memo.

`loadBlockTripData` hands out `slices.Clone` of the cached entry because `loadShiftTrips` sorts it in place. The elements themselves are shared and must be treated as read-only; I traced every consumer and none of them mutate, but `BuildTripStatus` does take `&stopTimes[i]` pointers into the backing array and pass them to the offset helpers, so I documented the constraint on the accessor rather than leave it implicit.

## Deliberately not prefetched: shape points

I had shapes in the prefetch initially and took them back out. `GetShapePointsByTripIDs` joins shapes to trips, so a shape shared by many trips comes back once per trip with no deduplication. At roughly 1,500 points per shape that is about 96 KB per vehicle held for the life of the request, which for a large agency is hundreds of megabytes per concurrent request. It also pushes the `IN (?,?,...)` expansion towards `SQLITE_MAX_VARIABLE_NUMBER`. Measured, it made things worse: 26.7 MB and 343,680 allocations against 21.4 MB and 328,843 with stop times alone. Stop times are around 4 KB per trip and carry no duplication, so that half stays.

## Numbers

Measured with `BenchmarkVehiclesForAgency_50Vehicles` (added in the follow-up PR below, since this PR has no handler calling `BuildTripStatus` per row yet), 50 vehicles on distinct active trips, pure-Go SQLite:

| | bytes/op | allocs/op |
|---|---|---|
| without the memo | 210 MB | 2,751,307 |
| with the memo | 21 MB | 332,215 |
| with the memo and the stop-times prefetch | 21.4 MB | 328,843 |

Allocation counts are deterministic and are the figures I trust. Wall-clock on my machine varied between 235 ms and 810 ms across repeated runs of the same benchmark, so I have not quoted a speedup; the ratio would not be reproducible.

The structural change matters more than the ratio. Per-vehicle allocations go from growing with vehicle count (12,888 at one vehicle, 55,026 per vehicle at fifty) to flat at roughly 6,600 per vehicle. The quadratic term is gone.

## Callers

`arrivals-and-departures-for-stop` installs the memo. That handler builds a trip
status per arrival row, and each one previously resolved its own stop times; it was
also already batch-loading the same rows for `totalStopsInTrip` and discarding them,
so the data was fetched twice. `prefetchStopTimes` now returns the rows grouped by
trip, the stop counts come from that same result, and the per-row lookups become map
reads — one existing query doing two jobs rather than a new one.

Over a 10-hour window on the fixture's busiest stop that is ~53ms → ~45ms per
request and 237k → 229k allocations (`BenchmarkArrivalsAndDeparturesWideWindow`).

`trips-for-route` looks like the worst offender of all, around four per-trip queries
per row with no cache installed at all, and `trips-for-location` is in the same
shape. Both are bigger changes and I have left them alone.

## Testing

`internal/restapi/request_cache_test.go` covers each memo: fall-through when the handler has not opted in, seeded entries being preferred over the database, first resolution populating the memo, and callers getting independent slices. The prefetch tests cover a real trip, a trip with no stop times (which must still be seeded so it is not retried per row), and the no-memo no-op path.

`go fmt` clean and the full suite passes on the pure-Go path (`CGO_ENABLED=0 go test -tags purego ./...`).

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved response times for schedule and trip-status requests by reusing repeated GTFS lookups within each request.
  * Added efficient prefetching for stop-time data shared across multiple trips.
  * Reduced redundant service-date and block-trip database queries.

* **Reliability**
  * Preserved fallback behavior when lookup queries fail.
  * Avoided retaining failed or incomplete lookups in the request cache.
  * Ensured cached results remain safe when callers modify returned data.

### DIFF
--- a/internal/restapi/arrivals_and_departures_for_stop_handler.go
+++ b/internal/restapi/arrivals_and_departures_for_stop_handler.go
@@ -98,7 +98,11 @@ func (api *RestAPI) arrivalsAndDeparturesForStopHandler(w http.ResponseWriter, r
 	// in the same shift. Without this each row pays for the full compute
 	// chain (blockTripIDsForServiceDate + loadBlockTripData + emitBlock
 	// Stops), amplifying to thousands of round-trips on wide time windows.
-	ctx := WithSnapshotCache(r.Context(), newSnapshotCache())
+	//
+	// The request cache sits one level below that, covering the block trip
+	// data and service calendar that rows outside a shared shift still
+	// resolve individually, plus the stop times prefetched further down.
+	ctx := withRequestCache(WithSnapshotCache(r.Context(), newSnapshotCache()))
 
 	// Capture parsing errors
 	params, fieldErrors := api.parseArrivalsAndDeparturesParams(r)
@@ -275,17 +279,13 @@ func (api *RestAPI) arrivalsAndDeparturesForStopHandler(w http.ResponseWriter, r
 		tripsLookup[trip.ID] = trip
 	}
 
-	// Batch-fetch stop counts per trip to avoid per-arrival N+1 queries for totalStopsInTrip.
-	tripStopCountMap := make(map[string]int, len(uniqueTripIDs))
-	if len(uniqueTripIDs) > 0 {
-		allStopTimesForTrips, err := api.GtfsManager.GtfsDB.Queries.GetStopTimesForTripIDs(ctx, uniqueTripIDs)
-		if err != nil {
-			api.Logger.Warn("failed to batch fetch stop times for trips", slog.Any("error", err))
-		} else {
-			for _, st := range allStopTimesForTrips {
-				tripStopCountMap[st.TripID]++
-			}
-		}
+	// One batch load serves two purposes: totalStopsInTrip below, and seeding the
+	// request memo so the per-arrival BuildTripStatus calls read each trip's stop
+	// times from it instead of issuing a query per row.
+	stopTimesByTrip := api.prefetchStopTimes(ctx, uniqueTripIDs)
+	tripStopCountMap := make(map[string]int, len(stopTimesByTrip))
+	for tripID, stopTimes := range stopTimesByTrip {
+		tripStopCountMap[tripID] = len(stopTimes)
 	}
 
 	for _, ast := range allActiveStopTimes {
--- a/internal/restapi/request_cache.go
+++ b/internal/restapi/request_cache.go
@@ -1,0 +1,183 @@
+package restapi
+
+import (
+	"context"
+	"log/slog"
+	"strings"
+	"sync"
+
+	"maglev.onebusaway.org/gtfsdb"
+)
+
+// requestCacheKey is the context key under which a request carries its memo.
+type requestCacheKey struct{}
+
+// requestCache memoizes GTFS lookups that repeat across the rows of a single
+// response.
+//
+// Endpoints that build a trip status per row — vehicles-for-agency across every
+// vehicle in an agency, for instance — resolve the same block and the same
+// service calendar over and over. Without a memo the work scales with
+// (rows × block size) rather than with the number of distinct blocks, and the
+// service-ID lookup repeats once per row for a date that never changes within
+// the request.
+//
+// The memo is deliberately request-scoped rather than a field on RestAPI: static
+// GTFS data is reloaded in the background, and a longer-lived cache would need
+// invalidation wired to that reload. It is mutex-guarded because nothing
+// prevents a handler from building statuses concurrently.
+//
+// This complements the existing snapshotCache (see WithSnapshotCache), which
+// memoizes whole block snapshots. These entries sit a level below that and are
+// still reached on the schedule-deviation path, which resolves a block without
+// going through a snapshot.
+type requestCache struct {
+	mu               sync.Mutex
+	blockTripData    map[string][]blockTripData
+	activeServiceIDs map[string][]string
+	stopTimes        map[string][]gtfsdb.StopTime
+}
+
+// withRequestCache returns a context carrying a request-scoped memo. Handlers
+// that build many trip statuses in one request should wrap their context with
+// it; handlers that build a single trip status gain nothing and can leave it off.
+func withRequestCache(ctx context.Context) context.Context {
+	return context.WithValue(ctx, requestCacheKey{}, &requestCache{
+		blockTripData:    make(map[string][]blockTripData),
+		activeServiceIDs: make(map[string][]string),
+		stopTimes:        make(map[string][]gtfsdb.StopTime),
+	})
+}
+
+// requestCacheFrom returns the request's memo, or nil when the handler did not
+// opt in.
+func requestCacheFrom(ctx context.Context) *requestCache {
+	cache, _ := ctx.Value(requestCacheKey{}).(*requestCache)
+	return cache
+}
+
+// blockTripDataCacheKeyFor joins trip IDs into a single map key. GTFS IDs cannot
+// contain a NUL byte, so the separator cannot collide across different splits of
+// the same concatenated text.
+func blockTripDataCacheKeyFor(tripIDs []string) string {
+	return strings.Join(tripIDs, "\x00")
+}
+
+func (c *requestCache) getBlockTripData(key string) ([]blockTripData, bool) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	trips, ok := c.blockTripData[key]
+	return trips, ok
+}
+
+func (c *requestCache) putBlockTripData(key string, trips []blockTripData) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	c.blockTripData[key] = trips
+}
+
+func (c *requestCache) getActiveServiceIDs(date string) ([]string, bool) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	ids, ok := c.activeServiceIDs[date]
+	return ids, ok
+}
+
+func (c *requestCache) putActiveServiceIDs(date string, ids []string) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	c.activeServiceIDs[date] = ids
+}
+
+func (c *requestCache) getStopTimes(tripID string) ([]gtfsdb.StopTime, bool) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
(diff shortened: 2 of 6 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (changes requested):

The cache itself is well built, and I went looking specifically for the ways this kind of change goes wrong. It's genuinely request-scoped with no package-level state; the maps are mutex-guarded with immediate `defer Unlock()` and no nested acquisition; failures aren't cached as successes on any of the three paths; and only static GTFS data is memoized, so real-time staleness isn't in play. The `slices.Clone` in `loadBlockTripData` is necessary and correct given `loadShiftTrips` sorts in place, and I traced every consumer of the aliased inner slices to confirm the read-only contract your doc comment asserts actually holds. `GetStopTimesForTripIDs` grouped by trip really is row-for-row equivalent to the per-trip query. That's careful work.

What's holding it up:

**Nothing calls it.** `withRequestCache` and `prefetchStopTimes` are reachable only from `request_cache_test.go` — no handler installs the memo, so this lands as +382 lines that are inert on `main`. You offered in the description to wire up `arrivals-and-departures-for-stop` in this PR; please do that, or split so the caller lands with the cache. CONTRIBUTING is explicit that leftover unreachable code is a review finding, and I'd rather not have a caching layer sitting in the tree with no way to tell whether it works in production.

Two smaller things while you're in there:

- A `GetShapePointsByTripIDs` failure inside `loadBlockTripDataFromDB` is non-fatal and yields a haversine-degraded but non-nil result, which *is* cached. One transient blip pins reduced distance precision for that block for the rest of the request, where before each call would retry. Worth either not caching the degraded result or noting why it's fine.
- `TestActiveServiceIDsForDateWithoutCache` hardcodes `20241104` and asserts equality against a direct query, so it would pass trivially if RABA has no service that day. The companion sentinel test does carry real signal.

Also note the CLA check is still pending — that will block the merge independently of the code.

dev-aditya-hub:
Hie Aaron i signed the CLA but it is not showing in the Cheaks can you help me with this ClA issue 


image

dev-aditya-hub:
Thanks for the careful read  the shape-failure caching in particular was a real hole and I'd missed it.

**Wired it up.** `arrivals-and-departures-for-stop` now installs the memo. It turned out better than just adding a caller: the handler was already batch-loading stop times for `totalStopsInTrip` and then throwing the rows away, while `BuildTripStatus` re-queried the same trip once per arrival row. So the data was being fetched twice over. `prefetchStopTimes` now returns the rows grouped by trip, the stop counts are derived from that same result, and the per-row lookups become map reads. No new query  one existing query doing two jobs.

On a 10-hour window over the fixture's busiest stop, ~53ms → ~45ms per request and 237k → 229k allocations. The ranges across four runs each don't overlap, so the wall-clock difference is real, though I'd still trust the allocation count more than the millisecond figure on my hardware.

That also exposed something worth knowing: the existing `BenchmarkArrivalsAndDeparturesForStop` runs the real clock against a fixture whose calendar expired in 2024, so it returns an empty list and never enters the per-arrival loop at all. It has been measuring the empty path. I added `BenchmarkArrivalsAndDeparturesWideWindow`, which holds the clock inside the service window and uses the same stop the handler tests use. Happy to fix the original in a separate PR rather than widen this one.

**Shape-failure caching.** Fixed properly rather than documented away. `loadBlockTripDataFromDB` now returns a completeness flag, and only a fully successful load is memoized. A shape failure still degrades that one caller to haversine, but the next row retries instead of inheriting it. Separate commit (`109cf64`) since it's a different concern from the wiring.

**The trivially-passing test.** You were right that it proved nothing. It now asserts the fixture actually has service on that date before comparing, so an empty-equals-empty pass is impossible. I also added a case for a date with genuinely no service, to pin that an empty answer is memoized as a legitimate result rather than re-queried on every row.

**On CodeRabbit's singleflight suggestion**  I don't think it's warranted and I've left it out. Nothing in `internal/restapi` builds trip statuses concurrently, so there are no concurrent misses to collapse. The bot was reacting to my own doc comment, which claimed handlers might build statuses concurrently; that was overstated, so I've rewritten it to say the mutex is defensive rather than load-bearing. Worth noting for whenever someone does parallelize a loop here: the sibling `snapshotCache` is an unguarded map and would fault first.

Still outstanding on the CLA that's on me, not the code, and I'm sorting it out.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
You addressed all three of my August 14 points properly rather than papering
over them, and the hardest part of this change holds up under scrutiny.

Taking them in order:

- **"Nothing calls it."** Fixed the right way. `arrivalsAndDeparturesForStopHandler`
  now installs the memo, and `prefetchStopTimes` replaces the
  `GetStopTimesForTripIDs` batch whose result used to be discarded — so one query
  now does two jobs (seeding the memo and deriving `totalStopsInTrip`) where there
  were previously two. I checked the row-count derivation is equivalent to the old
  `tripStopCountMap[st.TripID]++` loop, failure path included: a nil map yields
  zero counts, same as before.
- **Caching a haversine-degraded result.** `loadBlockTripDataFromDB` returning
  `([]blockTripData, bool)` with `complete == (shapeErr == nil)`, and only
  memoizing on `complete`, is exactly right — and
  `TestLoadBlockTripDataDoesNotCacheFailedLoad` pins it.
- **The trivially-passing service-IDs test.** The `require.NotEmpty(t, expected,
  "fixture must have service on %s for this test to prove anything")` guard is the
  fix I wanted, and adding
  `TestActiveServiceIDsForDateDoesNotConfuseEmptyWithMissing` to pin that an
  empty-but-valid answer still memoizes is a good instinct — that's the bug this
  class of cache usually ships with.

The thing I looked hardest at is the aliasing, since memoized slices are now
shared where they used to be freshly allocated per call. I traced the consumers:
`stopTimesForTrip` into `BuildTripStatus` (which takes `&stopTimes[i]` pointers
across half a dozen helpers) and `blockTripData.stopTimes/shapePoints/cumDistances`
into `emitBlockStops`, `projectStopsInSequence` and the snapshot fields. All
read-only. The one in-place mutation, the `slices.SortFunc` in `loadShiftTrips`,
is precisely what the `slices.Clone` guards. That's the correct place to have put
the clone, and I'm satisfied it's the only one needed.

Cache keys check out too: `activeServiceIDs` keys on the formatted date string,
which is the query's only input, with timezone resolution happening before
formatting; `blockTripData` keys on the ordered trip-ID set; and
`blockTripIDsForServiceDate` is deliberately left unmemoized. Declining the
singleflight suggestion was also the right call — there are no goroutines
anywhere in `internal/restapi`, and rewriting the overstated doc comment instead
of adding machinery for a concurrency that doesn't exist is the better outcome.

**Approving on the merits — but I can't merge it yet, and it isn't about the
code.** `license/cla` is still pending, and the reason is a mismatch rather than
an oversight: your commits are authored as `Aditya Kuchekar
<adityakuchekar0077@gmail.com>`, which GitHub links to the `aditya-systems-hub`
account, not to `dev-aditya-hub` which opened this PR. The CLA bot can't match
the two. The test suite hasn't run on this PR either, which I expect is
downstream of the same thing.

To unblock: either add `adityakuchekar0077@gmail.com` to the email addresses on
the `dev-aditya-hub` account, or re-author the commits with an email that account
already owns, then re-run the CLA check. Once it's green and the test suite has
run clean, this goes in — no re-review needed.

Two follow-ups you can fold in whenever, neither worth another round on its own:

- `arrivals_and_departures_for_stop_handler.go:170` still calls
  `GetActiveServiceIDsForDate` directly instead of `api.activeServiceIDsForDate`.
  It's nearly free either way since those three calls use three distinct dates,
  but it's inconsistent with the rest.
- `prefetchStopTimes` logs via package-level `slog.Warn` where the code it
  replaced used `api.Logger.Warn`, which moves that failure off the app logger.

burma-shave:
This PR has identified a scalability issue which is worth addressing.

However, I don’t think this improvement justifies adding `requestCache` alongside the existing `snapshotCache`. This creates two request-context caching layers with overlapping responsibilities. The mutex protects against concurrent trip-status construction, which no current handler performs. The largest reported memory improvement applies to `vehicles-for-agency`, but that endpoint is not used in production, and the PR that would have used it (#1332) is being closed separately.

The remaining benefit is limited to repeated work within certain requests. It does not help single-row requests or persist across requests, and it does not address `trips-for-route` or `trips-for-location`, which may be the most affected endpoints.

The Java OBA server handles this by computing trip status once per GTFS-RT update and caching the result. That approach fits Maglev’s real-time refresh path better than a general request-scoped cache. I've documented the investigation and benchmark results in #1379: https://github.com/OneBusAway/maglev/issues/1379.

Thanks for the thorough benchmarking and failure-mode tests. They provide useful groundwork for an ingest-side implementation.

PR #1332, opened 2026-08-12
Build vehicles-for-agency tripStatus from the shared builder

### LINKED ISSUE #1195: vehicles-for-agency: tripStatus fields left unpopulated
### Description
Per maglev.wiki/vehicles-for-agency.md, data.list[].tripStatus defines many fields that the vehicles-for-agency handler never populates. The handler builds tripStatus inline and only sets a subset (activeTripId, blockTripSequence, phase, status, position, orientation, lastUpdateTime, lastLocationUpdateTime, serviceDate, occupancyStatus), leaving the rest at their zero/default values.
Unpopulated spec'd fields:
- closestStop, closestStopTimeOffset
- nextStop, nextStopTimeOffset
- scheduleDeviation
- scheduledDistanceAlongTrip, totalDistanceAlongTrip, distanceAlongTrip
- predicted (always false)
- situationIds (always empty)
These are already computed for the other real-time endpoints (trip-details, trip-for-vehicle, trips-for-route, arrival(s)-and-departures-for-stop) by the shared api.BuildTripStatus in internal/restapi/trips_helper.go. The vehicles-for-agency handler duplicates a minimal version of this logic instead of reusing it, which is why these fields are missing — a handler-consistency deviation.
### Expected behavior
tripStatus is populated with the spec'd fields when the underlying data is available, consistent with the other real-time endpoints:
- predicted reflects whether real-time (vehicle or trip-update) data was used.
- situationIds lists active alerts affecting the trip.
- Stop-relative fields (closestStop, nextStop, and their offsets) are resolved from schedule/real-time data, with the referenced stops appearing in references.stops.
- Distance fields (scheduledDistanceAlongTrip, totalDistanceAlongTrip, distanceAlongTrip) are computed from shape data.
### Actual behavior
The listed fields are emitted at their zero values ("", 0, false, []) regardless of available data, so the tripStatus is structurally complete but semantically empty for them.

dev-aditya-hub:
Fixes #1195.

> **Depends on #1331.** This branch is stacked on that one, so its diff includes that commit as well. Please review/merge that first; GitHub will drop it from this diff once it lands.

## Why

`vehicles-for-agency` builds its `tripStatus` inline, and only fills a subset of the fields the spec defines. Missing today: `closestStop` and `closestStopTimeOffset`, `nextStop` and `nextStopTimeOffset`, `scheduleDeviation`, `scheduledDistanceAlongTrip`, `totalDistanceAlongTrip`, `distanceAlongTrip`, `predicted` (always false) and `situationIds` (always empty).

All of it is already computed by `api.BuildTripStatus`, which every other real-time endpoint uses: trip-details, trip-for-vehicle, trips-for-route, and both arrivals endpoints. This handler was the odd one out, carrying a reduced copy of the same logic.

## What this changes

The inline block is replaced by a call to the shared builder, wrapped in `buildVehicleTripStatus`. Roughly fifty lines of duplicated field assignment come out.

Three behaviours specific to this endpoint are re-applied on top, because changing them is outside the scope of this issue:

1. `activeTripId` still comes from `resolveActiveTripID`, resolved against the request's `time` parameter, and `blockTripSequence` still reports -1 when it cannot be resolved rather than the builder's 0. Both are pinned by existing tests.
2. `status`, `phase` and the two update timestamps are mirrored from the enclosing `vehicleStatus`, so the nested object cannot contradict the outer one within a single entry.
3. Raw GPS position and orientation are restored when the builder leaves them unset. More on that below.

## The stale-vehicle case

Worth calling out, because I nearly shipped it as a regression. `BuildVehicleStatus` returns early for a vehicle it considers stale, leaving position at (0, 0) and orientation at 0. The old inline code set both unconditionally. `StaleDetector` compares an absolute difference, so any `time` parameter more than fifteen minutes from the feed timestamp makes every vehicle stale, not just genuinely old ones.

The result was an entry carrying a real `location` at the top level next to a `tripStatus.position` of (0, 0), which is Null Island in the middle of the Atlantic. `applyRawVehiclePosition` restores the reported position and bearing when the builder has not set them. There is a regression test for it.

While in there I pulled the GTFS-bearing-to-OBA-orientation conversion out into `OrientationFromGTFSBearing` rather than have a second copy of it.

## Discrepancies worth flagging

Two things I did not change but that a reviewer should probably rule on:

**`lastLocationUpdateTime` semantics differ between handlers.** `BuildVehicleStatus` only sets it when the vehicle has a position; this endpoint has always set it from the vehicle timestamp regardless. Adopting the shared behaviour breaks `TestVehiclesForAgencyHandler_UpdateTimesPresentWhenSet`. I preserved the existing behaviour rather than flip it inside this PR. If the shared semantics are the correct ones, I will file it separately.

**`activeTripId` can now describe a different trip than the distance fields.** The override in point 1 above picks the active trip by time window, while `scheduledDistanceAlongTrip`, `totalDistanceAlongTrip` and `distanceAlongTrip` were computed by the builder against its own distance-along-block choice. On an interlined block at a trip transition those two disagree, and the response reports one trip's ID with another trip's distances. This is a pre-existing tension between the endpoint's pinned `activeTripId` behaviour and the shared builder, not something this PR introduces, but it becomes visible here.

**`tripStatus.position` is now shape-projected** for fresh vehicles rather than raw GPS, since that is what the builder produces. Consistent with the other endpoints, but it is a change in the emitted number.

## Cost

Reusing the builder is more expensive per vehicle: it loads that vehicle's stop times and shape and does the projection maths, which is what computing `closestStop` and `distanceAlongTrip` actually requires. The memoization in the prerequisite PR removes the part of that cost that was quadratic in vehicle count. What remains is linear and, as far as I can tell, irreducible without changing what the fields mean.

Against main at 50 vehicles it is still roughly 15x the allocations, all of it per-vehicle work that main was simply not doing. If that trade is not acceptable for large agencies I would rather hear it now than after more of the response is built on it.

## Testing

`TestVehiclesForAgencyHandler_TripStatusSpecFields` parks a vehicle at a real stop on an active trip and asserts the previously-missing fields. I confirmed it fails on main on exactly those assertions and passes with the change, so it is a real regression test rather than one written to fit.

`TestVehiclesForAgencyHandler_StaleVehicleKeepsPosition` covers the Null Island case above.

Also adds `BenchmarkVehiclesForAgency_1Vehicle` and `BenchmarkVehiclesForAgency_50Vehicles`. The existing `BenchmarkVehiclesForAgency` runs against the single-vehicle RABA fixture whose position goes stale during a long run, so it ends up measuring the empty path; these hold the clock still and take a vehicle count.

`go fmt` clean and the full suite passes on the pure-Go path (`CGO_ENABLED=0 go test -tags purego ./...`).

### DIFF
--- a/internal/restapi/request_cache.go
+++ b/internal/restapi/request_cache.go
@@ -1,0 +1,183 @@
+package restapi
+
+import (
+	"context"
+	"log/slog"
+	"strings"
+	"sync"
+
+	"maglev.onebusaway.org/gtfsdb"
+)
+
+// requestCacheKey is the context key under which a request carries its memo.
+type requestCacheKey struct{}
+
+// requestCache memoizes GTFS lookups that repeat across the rows of a single
+// response.
+//
+// Endpoints that build a trip status per row — vehicles-for-agency across every
+// vehicle in an agency, for instance — resolve the same block and the same
+// service calendar over and over. Without a memo the work scales with
+// (rows × block size) rather than with the number of distinct blocks, and the
+// service-ID lookup repeats once per row for a date that never changes within
+// the request.
+//
+// The memo is deliberately request-scoped rather than a field on RestAPI: static
+// GTFS data is reloaded in the background, and a longer-lived cache would need
+// invalidation wired to that reload. It is mutex-guarded because nothing
+// prevents a handler from building statuses concurrently.
+//
+// This complements the existing snapshotCache (see WithSnapshotCache), which
+// memoizes whole block snapshots. These entries sit a level below that and are
+// still reached on the schedule-deviation path, which resolves a block without
+// going through a snapshot.
+type requestCache struct {
+	mu               sync.Mutex
+	blockTripData    map[string][]blockTripData
+	activeServiceIDs map[string][]string
+	stopTimes        map[string][]gtfsdb.StopTime
+}
+
+// withRequestCache returns a context carrying a request-scoped memo. Handlers
+// that build many trip statuses in one request should wrap their context with
+// it; handlers that build a single trip status gain nothing and can leave it off.
+func withRequestCache(ctx context.Context) context.Context {
+	return context.WithValue(ctx, requestCacheKey{}, &requestCache{
+		blockTripData:    make(map[string][]blockTripData),
+		activeServiceIDs: make(map[string][]string),
+		stopTimes:        make(map[string][]gtfsdb.StopTime),
+	})
+}
+
+// requestCacheFrom returns the request's memo, or nil when the handler did not
+// opt in.
+func requestCacheFrom(ctx context.Context) *requestCache {
+	cache, _ := ctx.Value(requestCacheKey{}).(*requestCache)
+	return cache
+}
+
+// blockTripDataCacheKeyFor joins trip IDs into a single map key. GTFS IDs cannot
+// contain a NUL byte, so the separator cannot collide across different splits of
+// the same concatenated text.
+func blockTripDataCacheKeyFor(tripIDs []string) string {
+	return strings.Join(tripIDs, "\x00")
+}
+
+func (c *requestCache) getBlockTripData(key string) ([]blockTripData, bool) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	trips, ok := c.blockTripData[key]
+	return trips, ok
+}
+
+func (c *requestCache) putBlockTripData(key string, trips []blockTripData) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	c.blockTripData[key] = trips
+}
+
+func (c *requestCache) getActiveServiceIDs(date string) ([]string, bool) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	ids, ok := c.activeServiceIDs[date]
+	return ids, ok
+}
+
+func (c *requestCache) putActiveServiceIDs(date string, ids []string) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	c.activeServiceIDs[date] = ids
+}
+
+func (c *requestCache) getStopTimes(tripID string) ([]gtfsdb.StopTime, bool) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	stopTimes, ok := c.stopTimes[tripID]
+	return stopTimes, ok
+}
+
+// prefetchStopTimes resolves the stop times of every trip in tripIDs with one
+// query, seeding the request memo so the per-trip lookup inside BuildTripStatus
+// becomes a map read instead of a query per response row.
+//
+// Trips that return no rows are recorded as empty entries, so a genuinely
+// stop-time-less trip is not retried once per row.
+//
+// Shape points are deliberately NOT prefetched. GetShapePointsByTripIDs joins
+// shapes to trips, so a shape shared by many trips is returned once per trip; a
+// large agency's worth of vehicles would materialize hundreds of megabytes of
+// duplicated geometry for the life of the request. Stop times are ~two orders of
+// magnitude smaller per trip and carry no such duplication.
+//
+// A failed prefetch is logged and swallowed: an absent entry falls back to the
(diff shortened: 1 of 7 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. `tripStatus.closestStop` / `nextStop` are now populated, but `references.stops` is still never filled, so those IDs dangle. The handler only sets `references.Agencies`, `Routes`, `Trips` and `Situations`; `NewEmptyReferences()` leaves `Stops` as `[]`. This was not a problem on main because both fields were always `""`. Issue #1195 (which this PR says it fixes) calls it out explicitly: "Stop-relative fields (closestStop, nextStop, and their offsets) are resolved from schedule/real-time data, **with the referenced stops appearing in references.stops**." The three sibling handlers that emit a `tripStatus` all populate it (`trip_for_vehicle_handler.go:159`, `trip_details_handler.go:360`, `trips_for_route_handler.go:914`).

https://github.com/OneBusAway/maglev/blob/db72902ca292c6ca2167f1fd03110aa29d5e9933/internal/restapi/vehicles_for_agency_handler.go#L230-L244

Two smaller things, not blocking, for the maintainer to rule on:

- `situationIds` has the same shape of gap. `GetSituationIDsForTrip` matches alerts by trip, route *and* agency, but the reference block only collects route-scoped and agency-scoped alerts (`collectAlertsForRoutes(routeIDs)` + `GetAlertsByIDs("", "", id)`). An alert whose informed entity names only a `trip_id` will appear in `situationIds` with no matching entry in `references.situations`.
- The doc comment on `buildVehicleTripStatus` says it "re-applies the two behaviours that are specific to this endpoint" and then lists activeTripId/blockTripSequence, but the function also mirrors status/phase/update times and restores the raw GPS position. The PR description says three; the comment says two.

aaronbrethorst (changes requested):

I did the field-by-field comparison this kind of refactor demands, and it came out clean: `serviceDate` still normalizes to service-day midnight, `occupancyStatus` is still set unconditionally from the vehicle, `occupancyCapacity`/`occupancyCount` still default to -1 through `NewTripStatus()`, and the status/phase mapping is unchanged. The raw-GPS fallback for stale vehicles correctly preserves the endpoint's long-standing behavior rather than quietly regressing it. Replacing fifty lines of inline construction with the shared builder while re-applying only the genuinely endpoint-specific bits is exactly the right shape for this change, and the description is unusually candid about its own trade-offs — that made reviewing it much faster.

One thing to fix:

**`references.stops` is never populated.** The handler sets Agencies, Routes, Trips, and Situations, and `NewEmptyReferences()` leaves `Stops` as `[]`. On `main` that was fine because `closestStop` and `nextStop` were always empty strings — but the shared builder emits them now, so those IDs come back as dangling references. #1195, which this PR says it fixes, names that requirement directly, and all three sibling `tripStatus` handlers populate it.

Related and worth handling at the same time: `situationIds` can carry trip-scoped alert IDs that aren't in `references.situations` either. Rarer, since it needs an alert whose informed entity names only a trip, but it's the same class of gap.

Two notes, not blocking:

- For a stale vehicle, `tripStatus.position` is the old GPS fix while `distanceAlongTrip` and `scheduledDistanceAlongTrip` come from the schedule snapshot at `referenceTime` — so the two describe different places. That's inherent to the fallback you chose and it beats the previous behavior of reporting 0, but it's worth a comment so the next reader doesn't think it's a bug.
- The `buildVehicleTripStatus` doc comment says "the two behaviours specific to this endpoint," but four are re-applied.

The allocation increase you flagged — per-vehicle shape query, per-vehicle `GetTrip`+`GetRoute` inside `GetSituationIDsForTrip`, per-vehicle block snapshot — is real and I'd like to see numbers on a large agency before this lands, since `vehicles-for-agency` is on the hot path.

Finally, sequencing: this is stacked on #1331, which I've asked for changes on (its cache has no production caller yet), so this can't land before that resolves. The CLA check is also still pending on both.

burma-shave:
Thanks for your work on this. The PR for the caching mechanism that this PR depends on has been closed. The linked issue is currently blocked until the underlying scalability issue has been addressed in a different way.

PR #1335, opened 2026-08-14
Batch RT agency lookups; skip stale success

### LINKED ISSUE #1334: Performance: N+1 query in RT poller and inaccurate circuit breaker reset for stale vehicles
### Describe the bug / performance issue
While reviewing the real-time polling logic, I noticed two related issues that cause unnecessary SQLite contention and mask data staleness:

1. **N+1 Query in RT Filter:** In `realtime.go` and the poller, we are currently executing a `GetRoute` database call for every single entity during each poll. This N+1 pattern creates significant SQLite contention and increases polling latency at scale.
2. **False Circuit Breaker Resets:** When a vehicle feed is skipped because it is stale, the system still sets `hasNewData = true`. This resets the circuit breaker, meaning frozen vehicle feeds are falsely reported as healthy/successful.

### Expected behavior
1. The poller should batch route-to-agency lookups in memory before looping through entities to eliminate the N+1 query.
2. The `hasNewData` flag should only evaluate to `true` (and thus reset the circuit breaker) if the real-time data was actually applied, ignoring skipped stale feeds.

### Impact
* **Latency:** High DB roundtrips during every poll cycle.
* **Reliability:** The system incorrectly assumes stale data is actively updating, masking potential upstream feed failures.

### Proposed Solution
I propose refactoring the polling logic to:
- Extract all unique route IDs upfront and build an in-memory map of route-to-agency mappings using existing bulk fetch patterns.
- Tie the `hasNewData` toggle strictly to the successful application of the entity data.

I am already exploring a fix for this locally and would love to assign this to myself and open a PR for review shortly!

tejasva-vardhan:
### Description
Closes #1334

Agency RT filtering issued one `GetRoute` per entity every poll, and skipped stale vehicle feeds still set `hasNewData`, which reset the circuit breaker as if the feed were healthy.

- Batch unique route IDs with `GetRoutesByIDs` and resolve route→agency in memory (same pattern as trips-for-route / vehicles-for-agency).
- Count a vehicle sub-feed as success only when its data is actually applied (`applyVehicleUpdate`). A fetched-but-stale vehicle feed no longer sets `hasNewData`.

**Testing**
- Agency filter tests still pass with the batched lookup.
- Stale vehicle-only feeds return `hasNewData == false`.
- Successful trip updates still succeed when vehicles are stale.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented stale vehicle data from being treated as a successful update or refreshing feed timestamps.
  - Continued removing expired vehicles during feed processing.
  - Preserved successful trip updates when vehicle data is stale.
  - Improved agency filtering for trips, vehicles, and alerts, including proper handling of unresolved route references.
  - Improved reliability when processing large sets of route, stop, and trip references.

- **Tests**
  - Added coverage for initial updates, repeated stale vehicle feeds, mixed stale and successful updates, and large reference sets.


aaronbrethorst:
### Code review

Found 1 issue:

1. `buildRouteAgencyMap` sends every unique route ID in one unbatched `IN (...)` list and, on any query error, silently returns an empty map — which drops *every* trip/vehicle/alert for that feed rather than just the unresolvable ones. The doc comment's claim that this matches "the previous per-entity `GetRoute` skip-on-error behavior" only holds for per-row misses; a statement-level failure now nukes the whole feed. The route ID set is bounded only by feed size, and this repo already treats unbounded ID lists as a correctness hazard — `idsPerBatchedQuery` in `internal/restapi/reference_utils.go` documents that "SQLite rejects a statement carrying more bind variables than it allows rather than truncating it," and commit d3b2d350 ("Batch every unbounded stop ID lookup") went through and batched all of them. On a large aggregated feed — exactly the multi-agency case `agency-ids` filtering exists for — this fails deterministically every poll, and the failure is invisible: nothing is logged, `applyVehicleUpdate` still advances `feedVehicleTimestamp`, and `hasNewData` still returns true, so the feed reports healthy while serving zero entities. That directly undercuts this PR's own goal of not masking upstream feed problems. Suggest batching the lookup and propagating/logging the error instead of collapsing it into an empty map.

https://github.com/OneBusAway/maglev/blob/3d5407569749e19d6b3c2fc36c5badfbf7dc8105/internal/gtfs/realtime.go#L529-L536

aaronbrethorst (changes requested):
Two changes in one PR, and they're in quite different shape. Let me take them
separately.

**The `hasNewData` half is right and I'd merge it on its own.** Hoisting
`applyVehicleUpdate` so `vehiclesUpdated` only goes true when vehicle data was
actually stored is exactly the fix #1334 wanted — a fetch that parses fine but
carries a stale header timestamp shouldn't reset the poller's circuit breaker.
The concurrency picture is also clean, which I checked carefully since this file
is where it would hurt: the filters run *before* `realTimeMutex.Lock()`, there's
no nested locking, no new unguarded read of shared state, and the `defer Unlock`
pattern is preserved.

**The batched-lookup half needs work before it lands.**
`buildRouteAgencyMap` (internal/gtfs/realtime.go:521-541):

```go
routes, err := manager.GtfsDB.Queries.GetRoutesByIDs(ctx, routeIDs)
if err != nil {
    return map[string]string{}
}
```

Two problems, and the second is the one that worries me.

*The query isn't batched.* `routeIDs` is bounded only by feed size. We have an
explicit guard against this — `idsPerBatchedQuery` / `queryInBatches` in
internal/restapi/reference_utils.go:511, whose doc comment spells out why:
"SQLite rejects a statement carrying more bind variables than it allows rather
than truncating it." Commit d3b2d350 was a dedicated pass to batch every
unbounded ID lookup, and this reintroduces one. I realize `queryInBatches` lives
in `internal/restapi` and this code is in `internal/gtfs`, so reuse means moving
it somewhere shared rather than importing across — `internal/utils` seems like
the right home, and that's a fine thing to do in this PR.

*Swallowing the error turns a query failure into a silently empty feed.* On any
error — SQLITE_BUSY under exactly the contention #1334 describes, or a bind-limit
overflow from the above — this returns an empty map. Then every trip, vehicle and
alert filters to zero, the empty result gets **stored**
(`manager.feedVehicles[feedID] = validVehicles`), `feedVehicleTimestamp`
advances, and `hasNewData` returns `true`. The feed reports healthy while
serving nothing at all.

That's the same class of bug the other half of this PR sets out to fix, which is
why I want to be careful about it rather than wave it through. The doc comment
says this "match[es] the previous per-entity `GetRoute` skip-on-error behavior",
but that's only true for per-row misses — the old code lost one entity when one
route didn't resolve; this loses the entire feed when the statement fails.

Please log the error and either propagate it so the update is treated as failed,
or fall back to the per-entity path. Silently returning an empty map is the one
option that shouldn't survive.

Smaller things, none blocking:

- `agency_filter_test.go:217` hardcodes `routeIDSet` as `{"R1","R2","R999"}`
  rather than deriving it from the table cases, so a future case with a new route
  ID would get an unpopulated map and could pass for the wrong reason.
- `TestUpdateFeedRealtime_StaleVehiclesWithSuccessfulTripsStillSucceed` serves a
  vehicle-positions payload from the trip-updates URL. It has zero trip entities,
  so `tripsUpdated` is true purely because parsing succeeded — the test would
  pass even if trip handling were broken. It does verify the boolean semantics it
  claims to, so it isn't worthless, just weaker than it looks.
- Pre-existing, not yours: `logger.Error("realtime feed update failed", ...)`
  now fires on a healthy-but-stale skip with every `*_error` field `false`. Worth
  a follow-up so the new skip path doesn't read as an outage in the logs.

Fix the error handling and the batching and I'll get this straight back to you —
the underlying idea is right and the N+1 is worth removing.

tejasva-vardhan:
@aaronbrethorst pushed the lookup/batching changes as a new commit.

- `GetRoutesByIDs` is chunked with `idsPerBatchedQuery` via `utils.QueryInBatches` (moved out of restapi so gtfs can reuse it).
- One route→agency map is built per poll and passed to all three filters.
- A statement-level lookup error is logged and the update returns false, so previous entities stay and the feed is not reported healthy while empty.
- The mixed stale/success test now serves a trip-update payload and asserts the trip is stored. Alert-match tests derive the route ID set from the cases.

Left the stale-skip `logger.Error` follow-up as you suggested — not in this PR.

burma-shave:
Thanks for the PR and pointing out the N+1 route lookup concern, but we’re going to close this for now.

The realtime `agency-ids` filtering path is closely tied to Maglev’s current single-static-feed/raw-ID model. Planned multi-agency static feed support will likely rewrite this area, especially the way realtime IDs are resolved and scoped against static data. Given that, we’d rather avoid optimizing/refactoring this interim path before the multi-agency design lands.

This may still be worth revisiting after that work, particularly if mixed-agency GTFS-RT feeds continue to need route→agency filtering. Thanks again for digging into this.

PR #1336, opened 2026-08-14
Align trips bool defaults with OpenAPI

### LINKED ISSUE #1323: trips-for-route includeStatus/includeSchedule defaults diverge from OpenAPI spec
`testdata/openapi.yml` documents `includeStatus` and `includeSchedule` on `/api/where/trips-for-route/{routeID}.json` as "Defaults to false". The implementation defaults omitted bool params to **true** via `parseBoolQueryParam`, and `TestTripsForRouteHandler_BoolParamParsing` on `main` pins that behavior ("omitted defaults to true").

Meanwhile `trips_for_location_handler.go` carries the comment "Intentionally defaulting includeStatus to false to align with includeSchedule" — so the two endpoints disagree with each other as well as with the spec.

CLAUDE.md is explicit that the spec is the single source of truth and "if the codebase diverges from the spec, the spec wins", so this needs a deliberate resolution rather than drift.

Surfaced while reviewing #1293, which was adding tests that would have further pinned the current behavior. Worth settling before more coverage codifies it.

Decide: change the handlers to match the spec, or change the spec upstream in [sdk-config](https://github.com/OneBusAway/sdk-config/blob/main/openapi.yml) and re-sync. Either way both endpoints should end up consistent.

tejasva-vardhan:
Closes #1323

OpenAPI documents `includeStatus` and `includeSchedule` on trips-for-route as defaulting to false. Maglev defaulted omitted bools to true via `parseBoolQueryParam`, and tests pinned that. trips-for-location already defaulted both to false, so the two endpoints disagreed with each other and with the spec.

- Default omitted `includeStatus` / `includeSchedule` to false on both endpoints.
- Share `parseBoolQueryParam` so they stay consistent.
- Leave `includeTrip` and `includeReferences` defaulting to true.
- Update `TestTripsForRouteHandler_BoolParamParsing` so omitted no longer expects true.

trip-details and trip-for-vehicle are unchanged; their spec still defaults those flags to true.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected boolean query parameter handling for trip searches.
  * Omitted schedule and status options now default to disabled, while trip and reference options retain their existing defaults.
  * Invalid boolean values are handled consistently instead of being silently misinterpreted.


ARCoder181105:
@tejasva-vardhan follow contributing.md , no co-authored comments ; amend the commit ; make sure u read the code before pushing any changes ; also fix the sonar cloud issue as well ;

aaronbrethorst:
### Code review

The default changes themselves check out against `testdata/openapi.yml`: trips-for-route documents `includeStatus` and `includeSchedule` as "Defaults to false", trips-for-location documents `includeSchedule` with no default in its schema and "Defaults to false" in its description (and does not document `includeStatus` at all), while `includeTrip` there is explicitly `default: true`. Leaving `includeTrip`/`includeReferences` on the existing `parseIncludeTrip`/`ShouldIncludeReferences` helpers is the right call, and the updated `TestTripsForRouteHandler_BoolParamParsing` genuinely fails without the production change.

Found 2 issues:

1. The single commit carries a `Co-authored-by: Cursor <cursoragent@cursor.com>` trailer. CONTRIBUTING.md says "Do not attribute commits to a coding agent (e.g. no `Co-Authored-By` lines for Claude or similar tools). The human contributor authored and is responsible for the code, regardless of what tooling assisted in writing it." Please amend the commit to drop the trailer.

https://github.com/OneBusAway/maglev/commit/828513e5199a40b220352c1642a03d00455e1f76

2. The SonarCloud quality gate is red ("D Maintainability Rating on New Code", required ≥ A), which is the one failing check on this PR. The single open issue is `go:S107` — "This function has 11 parameters, which is greater than the 7 authorized" — on `buildTripsForLocationEntries`. That is *not* code this PR touches, and on current `main` the same function is already down to 7 parameters (it now takes `request *tripsForLocationRequest`). The branch is based on an older `main` and is still carrying the 11-parameter version, so Sonar is scoring a smell that no longer exists upstream. Merging/rebasing on current `main` should clear the gate — no refactor of that function is needed here.

https://github.com/OneBusAway/maglev/blob/828513e5199a40b220352c1642a03d00455e1f76/internal/restapi/trips_for_location_handler.go#L242-L257

Two smaller, non-blocking notes:

- In the `values` table, the `omitted` row's `want: true` is now dead — the loop unconditionally overwrites it from `omittedDefault` whenever `tt.query == ""`. Dropping the field for that row (or making the override read less like a special case) would keep the table honest about what it pins.
- `parseBoolQueryParam` now serves two handlers but still lives in `trips_for_route_handler.go`. CONTRIBUTING.md points parameter parsing at `internal/utils/api.go`. Not worth churning this PR over, given `parseIncludeTrip` is already shared out of `trips_for_location_handler.go`, but the pair is drifting toward a home in `utils`.

aaronbrethorst (changes requested):
The functional change is right and I checked it against the spec directly.
`testdata/openapi.yml` says "Defaults to false" for both `includeStatus` and
`includeSchedule` on trips-for-route, so flipping `parseBoolQueryParam`'s
fallback is correct, and leaving `includeTrip` (`parseIncludeTrip`) and
`includeReferences` (`ShouldIncludeReferences`) alone is right — those genuinely
do default to true and shouldn't move. Switching trips-for-location off its
inline `strconv.ParseBool` calls onto the shared helper is a clean bit of
consolidation on the way past. The updated `TestTripsForRouteHandler_BoolParamParsing`
would fail without the production change, which is what I want to see.

Three things before this lands.

**1. Drop the `Co-authored-by: Cursor` trailer.** Commit `828513e5` carries it,
and CONTRIBUTING.md is explicit: "Do not attribute commits to a coding agent
(e.g. no `Co-Authored-By` lines for Claude or similar tools)." We merge with
merge commits rather than squashing, so it lands in `main`'s history for good.
`git commit --amend` and `git push --force-with-lease`. (Same note as on your
#1346 — worth checking your Cursor config so it stops adding these.)

**2. Merge `main` in to clear the SonarCloud gate.** The red check is the
Quality Gate on a `go:S107` MAJOR — `buildTripsForLocationEntries` having 11
parameters. That's not yours: on current `main` that function takes 7 parameters
(it was refactored to accept `request *tripsForLocationRequest`), and your
branch is based on an older main that still has the 11-parameter version. Merge
main in and the gate should go green with no refactor on your part. Every
Actions check already passes.

**3. The one that actually matters — this makes a latent dangling-reference bug
the default.** In `trips_for_route_handler.go`, `stopIDsMap` is only populated
inside the `if includeSchedule` branches (~lines 400 and 479). So with
`includeSchedule` now defaulting to false, a plain `?includeStatus=true` returns
entries whose `status.closestStop`/`nextStop` name stop IDs that aren't in
`references.stops`.

That combination was reachable before via an explicit
`?includeSchedule=false&includeStatus=true`, so the bug isn't new — but your
change promotes it from "reachable if you ask for it" to "what you get by
default," which is a meaningfully different exposure.

Good news: #1353 fixes exactly this, by routing trips-for-route's stop
references through the same `stopsReferencedByEntries` path trips-for-location
already uses. I've asked for some cleanup there, but it's the right fix. So
please note the interaction in your description and let's land #1353 first — I
don't want to ship a default that emits dangling references even briefly. I'll
keep the two moving together.

Two small ones, neither blocking:

- `parseBoolQueryParam` now serves two handlers but still lives in
  `trips_for_route_handler.go`. CONTRIBUTING.md points parameter parsing at
  `internal/utils/api.go`. I'll accept it here since `parseIncludeTrip` already
  sits in the mirror-image spot, but the pair of them are a good follow-up.
- In the test table, the `omitted` row's `want: true` is dead — the loop always
  overwrites it from `omittedDefault` when `tt.query == ""`. Worth deleting so
  it doesn't read as a real expectation.

burma-shave:
The specs explicitly state that this param defaults to true. This is the legacy behaviour. Several PRs have already been merged defaulting this value to true. Any client that does not make use of the param will loose reference data in responses, having said that all known clients set this param explicitly to true which demonstrates, that references are nearly always needed and so true is a safe default.

PR #1337, opened 2026-08-14
Fix future service checks for removed dates

### LINKED ISSUE #1301: Exclude removed regular-calendar dates in both SQL representations.
> [!CAUTION]
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> 
> 
> <details>
> <summary>⚠️ Outside diff range comments (1)</summary><blockquote>
> 
> <details>
> <summary>gtfsdb/query.sql (1)</summary><blockquote>
> 
> `1317-1343`: _🎯 Functional Correctness_ | _🟠 Major_ | _🏗️ Heavy lift_
> 
> **Exclude removed regular-calendar dates in both SQL representations.**
> 
> `RouteHasFutureService` claims to report an effective future service date, but its regular-calendar branch checks only calendar overlap and enabled weekdays. It can return `1` when every date in the interval has a removal exception. This creates an incorrect service-date classification.
> 
> - `gtfsdb/query.sql#L1317-L1343`: require at least one enabled, non-removed date and add a regression test.
> - `gtfsdb/query.sql.go#L5295-L5323`: regenerate this derived query from the corrected SQL source with `make models`.
> 
> As per coding guidelines, modify the SQL source and regenerate `gtfsdb/query.sql.go`; do not hand-edit generated files.
> 
> <details>
> <summary>🤖 Prompt for AI Agents</summary>
> 
> ```
> Verify each finding against current code. Fix only still-valid issues, skip the
> rest with a brief reason, keep changes minimal, and validate.
> 
> In `@gtfsdb/query.sql` around lines 1317 - 1343, The regular-calendar EXISTS
> branch of RouteHasFutureService must exclude dates covered by removal
> exceptions, requiring at least one enabled, non-removed date and adding a
> regression test. Update gtfsdb/query.sql at lines 1317-1343; regenerate
> gtfsdb/query.sql.go at lines 5295-5323 with make models rather than editing it
> manually.
> ```
> 
> </details>
> 
> 
> 
> _Source: Coding guidelines_
> 
> </blockquote></details>
> 
> </blockquote></details>

<details>
<summary>🤖 Prompt for all review comments with AI agents</summary>

```
Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

Outside diff comments:
In `@gtfsdb/query.sql`:
- Around line 1317-1343: The regular-calendar EXISTS branch of
RouteHasFutureService must exclude dates covered by removal exceptions,
requiring at least one enabled, non-removed date and adding a regression test.
Update gtfsdb/query.sql at lines 1317-1343; regenerate gtfsdb/query.sql.go at
lines 5295-5323 with make models rather than editing it manually.
```

</details>

---

<details>
<summary>ℹ️ Review info</summary>

<details>
<summary>⚙️ Run configuration</summary>

**Configuration used**: Repository UI

**Review profile**: ASSERTIVE

**Plan**: Pro Plus

**Run ID**: `aa762f97-87a6-452d-b989-32d7fd795503`

</details>

<details>
<summary>📥 Commits</summary>

Reviewing files that changed from the base of the PR and between ac03ef628a9bca87192249b0d37e191f3a75eafb and e6101b55bc3b0b41231f719486c9c42b5bf247c6.

</details>

<details>
<summary>📒 Files selected for processing (6)</summary>

* `gtfsdb/query.sql`
* `gtfsdb/query.sql.go`
* `internal/restapi/reference_utils.go`
* `internal/restapi/search_stops_handler.go`
* `internal/restapi/search_stops_handler_test.go`
* `internal/restapi/trip_for_vehicle_handler.go`

</details>

</details>

_Originally posted by @coderabbitai[bot] in https://github.com/OneBusAway/maglev/pull/1221#pullrequestreview-4871818453_

soumajitgh:
## Summary

- account for calendar_dates removal exceptions when checking regular future service
- add regression coverage for a fully removed future weekday calendar
- regenerate the sqlc query output

## Root cause

The regular-calendar branch treated any overlapping enabled calendar as future service, even when every candidate date was explicitly removed.

Closes #1301

## Validation

- `CGO_ENABLED=1 go test -tags 'sqlite_fts5 sqlite_math_functions' ./gtfsdb -run 'TestRouteHasFutureService' -count=1`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved future service availability checks to evaluate each date accurately over the upcoming two years.
  * Correctly excludes regularly scheduled service dates canceled by calendar exceptions.
  * Continues to recognize service dates added through calendar exceptions.
  * Resolves cases where routes could incorrectly appear to have future service after all scheduled dates were canceled.

* **Tests**
  * Added regression coverage for fully canceled future regular service.


aaronbrethorst:
### Code review

Found 1 issue:

1. Part 2's per-day enumeration reverts commit `ee6aa59b` ("perf(gtfsdb): drop day-by-day scan in RouteHasFutureService") and reintroduces the O(days × trips) scan it deliberately removed — but without the julianday arithmetic the pre-`ee6aa59b` version used, and joined against `trips` rather than the route's *distinct* `service_id`s. Because the `dates` CTE has no index, SQLite rescans all ~730 materialized rows once per trip. Measured on SQLite 3.54 with a 400-trip route on two expired calendars, `ref_date` = today (exactly the case this query exists to answer — `scheduleForRouteHandler`'s `noTripsResponse` on a feed whose service has ended): **237 ms** for this PR vs **0.22 ms** for the query on `main`, ~1000x. It also gets worse without bound as `ref_date` moves into the past, which is caller-controlled: the recursion anchors at `ref_date + 1 day`, not at `MAX(ref_date, today)`, and `?date=` accepts arbitrary epoch-ms, so `date=-62135596800000` builds a 740,577-row CTE (0.55 s just to materialize) before any join. Hoisting the route's services into a `SELECT DISTINCT service_id FROM trips WHERE route_id = ?` CTE and driving Part 2 off that brought the 400-trip case back to 1.4 ms in the same harness, so the correctness fix does not have to cost this. The correctness fix itself and its regression test look right — this is about how Part 2 is executed, not whether the removals should be honored.

https://github.com/OneBusAway/maglev/blob/69b27f301c21e2a08da718441068be74431696a3/gtfsdb/query.sql#L1336-L1352

aaronbrethorst (changes requested):
The craft here is good — the SQL change went into `query.sql` with
`query.sql.go` properly regenerated, and
`TestRouteHasFutureService_AllFutureRegularDatesRemoved` genuinely fails against
`main`, which is exactly the kind of regression test I want. So this isn't a
"you did it wrong" review. It's that the thing being fixed was a deliberate
decision, and the fix costs more than the problem.

**This reverts commit ee6aa59b, and that commit's trade-off was intentional.**

`ee6aa59b` ("perf(gtfsdb): drop day-by-day scan in RouteHasFutureService")
replaced exactly this recursive day-by-day CTE with two indexed `EXISTS` checks.
Its message calls out the very case this PR fixes:

> Part 2 is a mild over-approximation for calendar rows whose only overlap-days
> are disabled or all cancelled by exception_type=2, but those cases only flip
> the caller's error text between ServiceDateOutOfRange and NoServiceThatDay —
> both surface an empty schedule, so the miscategorization is benign and is
> called out in the query's doc comment.

That's the accepted cost, documented on purpose. This PR deletes the doc comment
that recorded it, which tells me the history wasn't visible when you picked this
up — completely understandable, and not something I'd expect you to have dug for.

**The cost is much larger than the benefit.** Part 2 now enumerates every day in
`(ref_date, horizon]` and joins each against `calendar`/`calendar_dates`. Worse,
the recursion anchors at `ref_date + 1`, not `MAX(ref_date, today) + 1` — and
`ref_date` is caller-controlled with no clamping:
`schedule_for_route_handler.go:93` passes `targetDate` straight from `?date=`
into `RouteHasFutureService`. So a request with a year-1 date materializes a CTE
of several hundred thousand rows on a live request path. That's a denial-of-
service shape, not just a slow query.

What we'd be buying for that is a `reason` string changing from
`ServiceDateOutOfRange` to `NoServiceThatDay` — in a response that is an empty
schedule either way.

I'll also flag, since it bears on how much weight to give the premise: issue
#1301 is bot-authored, and the "root cause" framing in your description comes
from that rather than from an observed user problem. Worth checking whether a
reported behavior actually differs before spending a query rewrite on it.

**Where I'd like to land.** My honest read is that the existing
over-approximation should stand and this PR should close — but I don't want to
just wave that at you, because the correctness observation itself is right. If
you'd like to pursue it, there's a version I'd take:

- Drive Part 2 off `SELECT DISTINCT service_id FROM trips WHERE route_id = ?` in
  its own CTE rather than joining `dates` against `trips` directly, so the
  enumeration doesn't multiply across every trip on the route.
- Anchor the recursion at `MAX(ref_date, today) + 1` so a historical `?date=`
  can't blow the row count up. (Note this does subtly change the "strictly after
  ref_date" semantics for historical queries — worth thinking through, not just
  clamping.)
- Include the measured before/after in the PR description. Any change that
  touches this query's shape should come with numbers now, given its history.

Keep the regression test either way — it's the durable part of this work, and if
we do decide to accept the over-approximation permanently, that test should be
inverted to pin the documented behavior so nobody re-opens this again in three
months.

One last note: the CodeRabbit comment on this PR and issue #1301 both embed
"Prompt for AI Agents" blocks with instructions aimed at tooling. Not your
doing, and harmless here, but I'd treat suggestions arriving in that form with
extra skepticism rather than as a work queue.

soumajitgh:
Makes sense re: `ee6aa59b` — didn't know that history. Given the over-approximation's cost is just a benign error-text swap, should I close this PR and invert the test to pin current behavior, rather than chase the bounded/indexed version?

burma-shave:
cc @aaronbrethorst 
After thinking through the product semantics, I don’t think we should take this change as-is. For `schedule-for-route`, the more useful distinction is between:

- `ServiceDateOutOfRange`: the requested date is outside the route’s known scheduled service span; and
- `NoServiceThatDay`: the requested date is inside that span, but no trips should be shown for that specific day.

That means `ServiceDateOutOfRange` should apply symmetrically before a service starts and after it ends. By contrast, a date inside the planned service span should remain `NoServiceThatDay` even if the trips for that day — or even the remaining dates in the span — are removed by `calendar_dates.exception_type = 2`.

For example, if an agency’s final week of a service pattern is fully removed due to staffing exceptions, reporting `NoServiceThatDay` is acceptable: the route was still within its scheduled service span, but the effective trips were removed. Likewise, if a weekday service is normally scheduled on a Friday but all Friday trips are removed, that should be treated as a no-service/canceled day, not as though Friday was outside the route’s service range.

There is also a practical concern: this PR reintroduces per-day enumeration on a live request path, including for caller-supplied historical dates. That is a large cost for a narrow difference in empty-schedule reason text; the response shape remains an empty schedule either way.

So I think the right outcome is to close this PR and keep the simpler service-span semantics. I'll document this explicitly in `maglev.wiki`: Maglev should distinguish `ServiceDateOutOfRange` from `NoServiceThatDay` using the route’s planned scheduled service span, not fully materialized effective operating dates after removal exceptions. That is also a cleaner rule than the legacy “any later effective service exists” behavior, because it treats dates before service starts and after service ends consistently.

PR #1333, opened 2026-08-13
Split Frequency model into clear types

3rabiii:
### Description
This PR resolves the architectural conflation in the `models.Frequency` struct, which previously mixed two distinct legacy Java beans (`FrequencyV2Bean` and `ScheduleFrequencyInstanceV2Bean`). 

By properly scoping these types, we fix the missing `exactTimes` field in trip-centric endpoints and remove redundant schedule identifiers from contexts where they are not needed.

### Changes Included
* **Model Layer:** 
  * Extracted common fields into a new `FrequencyWindow` base struct.
  * Updated the generic `Frequency` struct to embed `FrequencyWindow`, removed schedule-specific fields (`serviceDate`, `serviceId`, `tripId`), and removed the `json:"-"` tag from `exactTimes`.
  * Introduced `ScheduleFrequency` for `schedule-for-stop` containing the required routing identifiers and boolean flags.
* **Handler Layer:** 
  * Updated `trip_details_handler.go` to remove redundant assignments of `ServiceDate`, `ServiceID`, and `TripID` to the frequency object.
* **Testing:** 
  * Updated all existing model and JSON tests to use the new `FrequencyWindow` embedding.
  * Added comprehensive serialization tests for `ScheduleFrequency`, including `omitempty` edge cases for `stopHeadsign`.

**Resolves:** #1147
**Fixes:** #1056 (Exposes `exactTimes` in JSON output)
**Fixes:** #618 (Prepares exact schema for `schedule-for-stop`)

## Summary by CodeRabbit

* **API Improvements**
  * Refined frequency data structures to separate reusable timing information from schedule-specific details.
  * Frequency responses now include exact-time information when available.
  * Schedule frequency data now supports service dates, trips, stop headsigns, and arrival/departure details.
  * Empty schedule frequency lists are returned consistently instead of as null values.

* **Bug Fixes**
  * Improved JSON serialization and omission of empty optional fields for frequency and schedule data.

PR #1343, opened 2026-08-14
Add frequency-based schedule-for-stop service

### LINKED ISSUE #1031: schedule-for-stop: Frequency-based service is completely missing
### Description
The `schedule-for-stop` endpoint does not return frequency-based trips.

### Spec Requirement
Frequency-based service (from GTFS `frequencies.txt`) must be queried and represented separately from fixed-schedule stop times inside the `scheduleFrequencies` array.

### Implementation Gap
The handler lacks any logic to fetch frequency-based trips. When creating the `StopRouteDirectionSchedule`, it explicitly passes `nil` for the frequencies slice, resulting in an empty array in the JSON response.

### Acceptance Criteria
- Frequency-based trips are successfully fetched from the database.
- Frequencies populate the `scheduleFrequencies` array within the correct `StopRouteDirectionSchedule` direction grouping.
- Unit tests verify that frequency objects exist and are correctly structured.

soumajitgh:
## Summary

- join active stop schedules with GTFS frequency rows in the existing query
- return `exact_times=0` service through `scheduleFrequencies`
- expand `exact_times=1` service into first-stop-relative virtual stop times
- preserve route/direction grouping, block-boundary flags, headsign weighting, and deterministic ordering
- validate malformed frequency data and make large expansions cancellation-aware
- add focused unit and HTTP integration coverage for both frequency modes

## Root cause

The schedule-for-stop query did not read the `frequencies` table, so the endpoint always returned an empty `scheduleFrequencies` array and omitted frequency-defined service.

## Spec notes

- The wiki does not define a distinct response matrix for `exact_times=1`. This implementation follows the explicit clarification in #1031 and expands deterministic frequency service into virtual `scheduleStopTimes`.
- The corrected `ScheduleFrequency` model includes `stopHeadsign`, `arrivalEnabled`, and `departureEnabled`, while the current OpenAPI schema documents only the other six fields. These legacy fields are additive and existing OpenAPI conformance checks pass.

## Scope note

This exceeds the preferred 200-line PR size because the query, generated sqlc row, handler classification, exact-time expansion, and coverage are tightly coupled parts of the same issue. Most of the diff is generated query code and tests; splitting it would leave an intermediate implementation without complete behavior or validation.

## Validation

- `go fmt ./...`
- `make test`
- `make test-pure`
- `make check-openapi`
- `git diff --check`

Closes #1031
Refs #993

### DIFF
--- a/gtfsdb/query.sql
+++ b/gtfsdb/query.sql
@@ -551,12 +551,25 @@ SELECT
     t.min_arrival_time,
     t.max_departure_time,
     t.direction_id,
+    first_st.departure_time AS first_departure_time,
+    f.start_time AS frequency_start_time,
+    f.end_time AS frequency_end_time,
+    f.headway_secs AS frequency_headway_secs,
+    f.exact_times AS frequency_exact_times,
     r.id as route_id,
     r.agency_id
 FROM
     stop_times st
     JOIN trips t ON st.trip_id = t.id
+    JOIN stop_times first_st
+      ON first_st.trip_id = t.id
+     AND first_st.stop_sequence = (
+        SELECT MIN(first_st_lookup.stop_sequence)
+        FROM stop_times first_st_lookup
+        WHERE first_st_lookup.trip_id = t.id
+     )
     JOIN routes r ON t.route_id = r.id
+    LEFT JOIN frequencies f ON f.trip_id = t.id
     LEFT JOIN (
         SELECT c.id AS service_id
         FROM calendar c
@@ -1463,5 +1476,4 @@ FROM
     JOIN stops s ON s.id = st.stop_id
 GROUP BY
     r.agency_id;
-
 
--- a/gtfsdb/query.sql.go
+++ b/gtfsdb/query.sql.go
@@ -2998,12 +2998,25 @@ SELECT
     t.min_arrival_time,
     t.max_departure_time,
     t.direction_id,
+    first_st.departure_time AS first_departure_time,
+    f.start_time AS frequency_start_time,
+    f.end_time AS frequency_end_time,
+    f.headway_secs AS frequency_headway_secs,
+    f.exact_times AS frequency_exact_times,
     r.id as route_id,
     r.agency_id
 FROM
     stop_times st
     JOIN trips t ON st.trip_id = t.id
+    JOIN stop_times first_st
+      ON first_st.trip_id = t.id
+     AND first_st.stop_sequence = (
+        SELECT MIN(first_st_lookup.stop_sequence)
+        FROM stop_times first_st_lookup
+        WHERE first_st_lookup.trip_id = t.id
+     )
     JOIN routes r ON t.route_id = r.id
+    LEFT JOIN frequencies f ON f.trip_id = t.id
     LEFT JOIN (
         SELECT c.id AS service_id
         FROM calendar c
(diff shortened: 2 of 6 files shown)


soumajitgh:
@coderabbitai review

aaronbrethorst:
### Code review

Found 1 issue:

1. This duplicates work already in flight in open PR #1348 ("GTFS Frequencies Phase 3"). That PR touches the same two files and implements the same schedule-for-stop behavior: `exact_times=0` routed to `scheduleFrequencies` with run-count-weighted headsign voting and start-time sorting, and `exact_times=1` expanded in memory into `scheduleStopTimes` relative to the trip's first stop departure. The two approaches are incompatible and will conflict: this PR adds the frequency columns to the hot `GetScheduleForStopOnDate` query (plus a `stop_times` self-join with a correlated `MIN(stop_sequence)` subquery to get the trip's first departure), while #1348 leaves that query alone and batch-fetches via the existing `GetFrequenciesForTrips`. Only one of these should land for `schedule_for_stop_handler.go`; the rest of #1348 (trip-details, trips-for-route, arrivals-and-departures) is genuinely additional scope. Worth deciding which implementation wins before either merges, rather than resolving it as a merge conflict.

https://github.com/OneBusAway/maglev/blob/276ab1f15585902228536a4f10b9a88019c70ed1/internal/restapi/schedule_for_stop_handler.go#L391-L415

aaronbrethorst (changes requested):
This is careful work and I want to be clear up front that I'm not bouncing it
for a defect — I went looking and didn't find a correctness bug worth blocking
on. The `frequencyInstanceCount` arithmetic and the expansion loop's break
condition are mutually consistent in nanosecond units throughout, the
`first_st` self-join is exactly-one-row given `stop_times`' `(trip_id,
stop_sequence)` primary key so it can't drop or duplicate rows, the generated
`query.sql.go` is consistent with a real `make models` run, and leaving
`startTime`/`endTime` unshifted for mid-route `exact_times=0` stops matches
legacy OBA. You also correctly left `internal/models/frequency.go` alone rather
than fixing a pre-existing deviation in passing.

**The problem is collision, and that's a coordination failure on my side, not
yours.**

PR #1348 ("GTFS Frequencies Phase 3", @3rabiii) rewrites the same file —
`internal/restapi/schedule_for_stop_handler.go`, +250/-65 against your +259/-73 —
implementing the same feature: `exact_times=0` → `scheduleFrequencies` entries,
`exact_times=1` → in-memory expansion relative to the trip's first stop. Two
different mechanisms for one outcome, and only one can land.

I'm going to give that file to #1348, for two reasons:

1. It's the next step in a sequence that's already partly on `main` — #1333
   ("Split Frequency model into clear types") merged, and
   `GetFrequenciesForTrips` already exists in `query.sql` specifically to serve
   this. #1348 uses that batch query; this PR instead extends
   `GetScheduleForStopOnDate` with a `LEFT JOIN frequencies` plus a `stop_times`
   self-join. Given we already have a purpose-built batch query, growing the hot
   schedule query is the larger blast radius for the same result.
2. #1348 covers the whole surface consistently — arrivals-and-departures,
   trip-details, trip-for-vehicle, trips-for-location, trips-for-route — where
   this is scoped to schedule-for-stop. Frequencies showing up on one endpoint
   and not its neighbours is exactly the kind of inconsistency that's painful to
   unpick later.

Neither of those is a knock on your implementation. You had no way to see #1348
coming, and I should have flagged the overlap when this opened rather than
letting you build it out. Sorry about that.

**What I'd suggest.** Two pieces here are independently useful and don't collide
with #1348 at all:

- `nulls.ValidInt64Count` and its test — small, general, and belongs in the
  package regardless of who implements the handler.
- The test suite. Your `schedule_for_stop_handler_test.go` coverage is more
  thorough than #1348's on this endpoint (~491 added lines vs ~131). Once #1348
  lands, porting your cases onto its implementation would be a genuinely valuable
  PR, and I'd review it quickly.

If you'd rather make the case that this approach should win over #1348's, I'll
hear it — I've asked for changes there too, so it isn't a done deal. But I'd want
that argument made before either of you writes more code, not after.

For the record, a few smaller things I noticed, useful whichever way this goes:

- `buildScheduleFrequency` takes a `row` parameter it never reads.
- Exceeding the 10,000-instance `maxExpandedScheduleStopTimes` budget returns a
  500 for the entire stop schedule rather than truncating. A valid feed with
  short `exact_times=1` headways would get an error instead of a partial
  schedule, and the message always cites 10000 even when the remaining budget was
  smaller.
- The new unexported helpers (`getScheduleDirectionGroup`,
  `parseScheduleFrequencyRow`, `expandExactScheduleStopTimes`, `bestHeadsign`,
  and the two new types) have no doc comments, where every pre-existing helper in
  that file does.
- At +817/-101 this is 4x CONTRIBUTING's ~200-line target. I saw your scope note
  and the coupling argument is reasonable, but the SQL change, the `nulls`
  helper, and the handler rework were three separable commits' worth of PR.

soumajitgh:
Closing per the discussion above — this collides with #1348, which is taking over `schedule_for_stop_handler.go`. Extracted the reusable piece into #1364 (`nulls.ValidInt64Count` + test). Will look at porting this PR's test coverage onto #1348's implementation once that lands.

PR #1347, opened 2026-08-15
Reuse resolved situations for arrivals

### LINKED ISSUE #1341: arrivals-and-departures re-resolves each trip's alerts per arrival row
Depends on #1313, which added the `tripStatusExtras` plumbing.

**Current**: the per-arrival loop calls `BuildTripStatus`, which resolves the trip's situations into `statusExtras.situations` (`trips_helper.go:63`). The loop then resolves alerts a second time with `GetAlertsForTrip` (`arrivals_and_departures_for_stop_handler.go:445`). `BuildTripStatus`'s own doc comment tells callers not to do this, naming this endpoint as the reason.

Not a pure duplicate: `situationRefsForTrip` matches on trip + route + agency, while `GetAlertsForTrip` matches on trip alone. So the two produce different sets, and switching to the resolved one changes what an arrival reports.

**Question to settle first**: which set is right for `arrival.situationIds`? Java attaches trip, route, and agency alerts, which points at the resolved set — worth confirming against a deployed server before changing it.

**Fix**: if the wider set is correct, reuse `statusExtras.situations` via `situations.addRefs` and drop the second lookup. That also removes one alert resolution per arrival row, which matters here — `minutesBefore`/`minutesAfter` can each be 24h.

**Test**: assert an arrival reports its route-level alert, and that `situationIds` resolve in `references.situations`.

soumajitgh:
## Summary

- Reuse situation references resolved by `BuildTripStatus` for each plural arrival.
- Add regression coverage for route-level alerts and situation-reference resolution.

## Why

The arrivals-and-departures handler resolved each trip’s alerts twice per arrival row. Reusing the existing references removes that repeated work and keeps entry IDs aligned with `references.situations`.

## Validation

- `make test`

Closes #1341

## Summary by CodeRabbit

* **Bug Fixes**
  * Route-level service alerts now appear correctly on affected arrival and departure results.
  * Alert situation references are now consistently included with the associated stop information.
  * Resolved situations are handled consistently when displaying alerts for individual arrivals and departures.

### DIFF
--- a/internal/restapi/arrivals_and_departures_for_stop_handler_test.go
+++ b/internal/restapi/arrivals_and_departures_for_stop_handler_test.go
@@ -94,6 +94,61 @@ func TestArrivalsAndDeparturesForStopHandlerEndToEnd(t *testing.T) {
 	require.NotEmpty(t, model.Data.References.Stops)
 }
 
+// TestArrivalsAndDeparturesForStopHandler_RouteAlertReferences verifies that a
+// route-level alert appears on the affected arrival and that its situation ID
+// resolves in references.situations.
+func TestArrivalsAndDeparturesForStopHandler_RouteAlertReferences(t *testing.T) {
+	api, cleanup := createTestApiWithRealTimeData(t, clock.NewMockClock(arrivalsTestClock))
+	defer cleanup()
+	t.Cleanup(api.GtfsManager.MockResetRealTimeData)
+
+	requestURL := arrivalsAndDeparturesURL(arrivalsTestStopID,
+		url.Values{"minutesBefore": {"60"}, "minutesAfter": {"240"}})
+	_, initial := callAPIHandler[ArrivalsAndDeparturesResponse](t, api, requestURL)
+	require.NotEmpty(t, initial.Data.Entry.ArrivalsAndDepartures,
+		"fixture stop should have an arrival in the test window")
+
+	target := initial.Data.Entry.ArrivalsAndDepartures[0]
+	agencyID, routeID, err := utils.ExtractAgencyIDAndCodeID(target.RouteID)
+	require.NoError(t, err)
+
+	const alertID = "plural-arrivals-route-alert"
+	api.GtfsManager.MockAddAlert("feed-0", gtfs.Alert{
+		ID:               alertID,
+		InformedEntities: []gtfs.AlertInformedEntity{{RouteID: &routeID}},
+		Header:           []gtfs.AlertText{{Text: "Plural arrivals route alert", Language: "en"}},
+	})
+
+	resp, model := callAPIHandler[ArrivalsAndDeparturesResponse](t, api, requestURL)
+	require.Equal(t, http.StatusOK, resp.StatusCode)
+
+	wantSituationID := utils.FormCombinedID(agencyID, alertID)
+	var affectedArrival *models.ArrivalAndDeparture
+	for i := range model.Data.Entry.ArrivalsAndDepartures {
+		arrival := &model.Data.Entry.ArrivalsAndDepartures[i]
+		if arrival.RouteID == target.RouteID {
+			affectedArrival = arrival
+			break
+		}
+	}
+	require.NotNil(t, affectedArrival, "expected an arrival on the alerted route")
+	require.NotNil(t, affectedArrival.TripStatus)
+	assert.Equal(t, affectedArrival.TripStatus.SituationIDs, affectedArrival.SituationIDs,
+		"arrival situationIds must be the ones BuildTripStatus resolved, not a second lookup")
+	assert.Contains(t, affectedArrival.SituationIDs, wantSituationID,
+		"an arrival must include its route-level alert")
+
+	var referenced bool
+	for _, situation := range model.Data.References.Situations {
+		if situation.ID == wantSituationID {
+			referenced = true
+			break
+		}
+	}
+	assert.True(t, referenced,
+		"arrival situationId %q must resolve in references.situations", wantSituationID)
+}
+
 func TestArrivalsAndDeparturesForStopHandlerTimeParams(t *testing.T) {
 	api, cleanup := createTestApiWithRealTimeData(t, clock.NewMockClock(arrivalsTestClock))
 	defer cleanup()
--- a/internal/restapi/arrivals_and_departures_for_stop_handler.go
+++ b/internal/restapi/arrivals_and_departures_for_stop_handler.go
@@ -442,8 +442,10 @@ func (api *RestAPI) arrivalsAndDeparturesForStopHandler(w http.ResponseWriter, r
 
 		lastUpdateTime := api.GtfsManager.GetVehicleLastUpdateTime(vehicle)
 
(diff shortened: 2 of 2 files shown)


ARCoder181105 (changes requested):
Traced both paths — the refactor is output-identical, and the perf win is real.

`GetAlertsForTrip` and `situationRefsForTrip` both do `GetTrip` → `GetRoute` → `GetAlertsByIDs(tripID, routeID, agencyID)` and both end at `situationRefsFromAlerts(alerts, agencyID)`. Same alerts, same IDs — so this drops one duplicated `GetTrip`+`GetRoute`+alert-index scan per arrival row, which matters on wide `minutesBefore`/`minutesAfter` windows. It also makes the plural handler match the singular one (`arrival_and_departure_for_stop_handler.go:401`).

One request, two notes.

**1. `arrivals_and_departures_for_stop_handler_test.go:135` — the test passes on the pre-change code too.**

Revert the one-line change and it still goes green, because the old path resolved the same alert with the same ID. Good coverage of the situationIds↔references invariant, but it doesn't pin what this PR changed. Asserting the IDs came from the status would, replacing line 135:

```go
	require.NotNil(t, affectedArrival.TripStatus)
	assert.Equal(t, affectedArrival.TripStatus.SituationIDs, affectedArrival.SituationIDs,
		"arrival situationIds must be the ones BuildTripStatus resolved, not a second lookup")
	assert.Contains(t, affectedArrival.SituationIDs, wantSituationID,
```

That's exactly the property the commit claims, and it breaks if someone reintroduces a separate lookup.

**2. `arrivals_and_departures_for_stop_handler.go:448` — consistency nit only, fine to leave.**

`statusExtras.snapshot` is read inside `if status != nil` (line 393) while `statusExtras.situations` is read outside it, so the two reads of the same struct look like they have different safety requirements. Both guards are in fact unreachable — every return in `BuildTripStatus` (`trips_helper.go:103`, `108`, `366`) passes a non-nil `status` and a non-nil `extras`. No change needed.

*Correcting my earlier version of this point:* I suggested `api.tripSituationsFor(ctx, st.TripID, statusExtras)` here. That was wrong — it returns `([]string, []models.Situation)` and bypasses the `situationCollector`, which this handler needs so dedupe stays in one place and `references.Situations` is built once at line 624. Using it would mean hand-appending and hand-deduping per arrival. A nil guard would also be dead code, and risks leaving `situationIDs` nil, which marshals as `"situationIds": null` instead of `[]`; the current `addRefs` avoids that via `make([]string, 0, len(refs))`. The code as written is correct.

**3. Commit message body has literal `\n` instead of newlines** — renders as one unwrapped line. Worth an amend while it's a single commit.

Verified against the deployed Puget Sound server: stop `1_10190` (Route 70 reroute) returns `1_86736` on each affected arrival with a matching `references.situations` entry — the shape this PR preserves. Two notes from that, both out of scope here:

- Alert IDs arrive already prefixed (`1_86736`), so the idempotency guard in `situationID()` is load-bearing.
- Prod returns `entry.situationIds: []` at the top level on that stop, while we emit the union of every collected ref. Pre-existing, but this PR now feeds that union from a new source, so it may be worth a separate look.

internal/restapi/arrivals_and_departures_for_stop_handler_test.go
```diff
@@ -94,6 +94,58 @@ func TestArrivalsAndDeparturesForStopHandlerEndToEnd(t *testing.T) {
...
+	var affectedArrival *models.ArrivalAndDeparture
+	for i := range model.Data.Entry.ArrivalsAndDepartures {
+		arrival := &model.Data.Entry.ArrivalsAndDepartures[i]
+		if arrival.RouteID == target.RouteID {
+			affectedArrival = arrival
+			break
+		}
+	}
+	require.NotNil(t, affectedArrival, "expected an arrival on the alerted route")
+	assert.Contains(t, affectedArrival.SituationIDs, wantSituationID,
```
ARCoder181105, line 138: This test passes on the pre-change code too — revert the one-line change and it still goes green, because `GetAlertsForTrip` resolved the same alert with the same agency and the same ID. Good coverage of the situationIds↔references invariant, but it doesn't pin what this PR changed.

Asserting the IDs came from the status would:

```suggestion
	require.NotNil(t, affectedArrival.TripStatus)
	assert.Equal(t, affectedArrival.TripStatus.SituationIDs, affectedArrival.SituationIDs,
		"arrival situationIds must be the ones BuildTripStatus resolved, not a second lookup")
	assert.Contains(t, affectedArrival.SituationIDs, wantSituationID,
```

That's exactly the property the commit message claims, and it breaks if someone reintroduces a separate lookup.

soumajitgh: Addressed in a73efdf. The test now asserts that the arrival’s `SituationIDs` exactly match `TripStatus.SituationIDs`, while retaining the expected alert-ID assertion. I also amended the commit message body to use real line breaks.

ARCoder181105:
@soumajitgh can you rebase and squash the PR into single atomic commit , as the commit history still consist the anomaly of `\n` we have to remove that..

soumajitgh:
@ARCoder181105 done

ARCoder181105 (approved):
LGTM

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Four lines of production code for a real win on an endpoint where
`minutesBefore`/`minutesAfter` can each be 24 hours. Good trade.

I didn't take the equivalence on faith — I traced both sides.
`GetAlertsForTrip` (internal/gtfs/realtime.go:154) and `situationRefsForTrip`
(internal/restapi/trips_helper.go:833) do the same resolution: `GetTrip` →
`GetRoute` → `GetAlertsByIDs(tripID, routeID, agencyID)`, both feeding
`situationRefsFromAlerts` with the same agency. So the reused set is
trip+route+agency scoped exactly as before, `addRefs` allocates a fresh slice
per call and dedupes on the same key, and `references.situations` composition is
unchanged. Worth noting for the record: issue #1341's premise that
`GetAlertsForTrip` "matches on trip alone" is stale — the two sets were already
identical, so this is a pure performance refactor rather than a behavior fix.
That's a better outcome, not a worse one, but it's worth being precise about.

Also a nice side effect: the line you removed was the handler's only `r.Context()`
use, which bypassed the snapshot-cache-wrapped `ctx` established earlier in the
function. Everything now flows through the right context.

Two notes, neither blocking:

- The new test is a good invariant guard (situationIds ↔ references.situations),
  but it doesn't actually pin *this* change — since both paths resolve the same
  set, it passes on pre-change code too. @ARCoder181105 made this point on the
  earlier revision and it still stands; the test is worth keeping, just not for
  the reason the description gives.
- This PR removes the last production caller of
  `Manager.GetAlertsForTrip` — after it lands, only `realtime_test.go` exercises
  it. CONTRIBUTING.md calls out leftover dead code specifically. I'd rather not
  grow a 4-line perf PR to cover it, but a quick follow-up deleting that method
  and its test would be welcome if you want to take it.

Merging.

PR #1348, opened 2026-08-15
Feat: GTFS Frequencies Phase 3 - API Integration & Handlers

### LINKED ISSUE #666: Feature: GTFS Frequencies Phase 3 (API Integration & Handlers)

3rabiii:
### Description
This PR implements Phase 3 of the GTFS Frequencies integration, successfully wiring frequency data from the database into the API handlers and response models while maintaining strict parity with the legacy Java OBA system.

### Implementation Details

**1. Core Helpers & Models**
* Added `NewScheduleFrequencyFromDB` constructor for safe timezone and over-midnight frequency calculations.
* Updated `BuildTripStatus` and `BuildTripSchedule` to populate `Frequency` fields, automatically propagating data to all downstream consumers.
* Reduced cognitive complexity and eliminated literal duplication in trip parameters parsing by extracting focused helper functions (e.g., `ParseBoolParam`).

**2. Single & Batch Handlers**
* **Single-Trip:** Wired frequency population into `trip_details`, `trip_for_vehicle`, and the singular `arrival_and_departure` handlers.
* **Batch Optimization:** Implemented highly optimized batch-fetching (`freqMap`) in `arrivals_and_departures`, `trips_for_route`, and `trips_for_location` handlers to prevent N+1 queries. 
* Added `frequencyForEntry` helper utilizing robust negative caching (seeding empty trips with `nil`) and fallback querying.
* Enforced strict error propagation (HTTP 500) for database failures during frequency lookups.

**3. Schedule For Stop**
* **`exact_times=0` (Headway-based):** Routed to the `scheduleFrequencies` array with accurate frequency-weighted headsign voting (weight = window duration / headway) and applied start-time sorting. Added guards against division-by-zero panics.
* **`exact_times=1` (Schedule-based):** Implemented dynamic in-memory expansion, converting a single frequency window into discrete `ScheduleStopTime` entries calculated via arrival/departure offsets relative to the trip's first stop.

### Scope Adjustments for Legacy Parity
To guarantee exact behavioral parity with the legacy Java system, the following scope decisions were applied:
* **Dropped:** Including frequency data in the `schedule_for_route` handler was skipped, as the legacy system did not support this.
* **Maintained:** Time-window filtering in `trips_for_location` remains unchanged; frequency trips only appear if they have an active GTFS-RT vehicle, matching legacy behavior.

### Testing
* Created `frequency_fixture_test.go` providing a fully mocked, in-memory GTFS dataset with mixed trip types (`exact_times=0`, `exact_times=1`, and normal scheduled trips) and block boundary testing.
* Added comprehensive handler unit tests and HTTP-level End-to-End tests ensuring correct frequency population, verifying that `exact_times=1` trips expand correctly into the `scheduleStopTimes` array, and ensuring non-frequency trips cleanly return `null`. Full regression suite is green.

**Resolves #666** 

## New Features
- Added frequency details to trip, stop arrival/departure, schedule, route, location, and vehicle responses.
- Included headways, exact-times indicators, service dates, and frequency start/end times.
- Added frequency-based schedules and expanded exact-time frequency trips into stop times.
- Improved frequency support across single and batch API responses.

## Bug Fixes
- Frequency data is omitted for non-frequency trips.
- Added support for schedules extending past midnight and across agencies with different time zones.
- Improved timezone handling, frequency lookup, and error reporting.


internal/restapi/trips_helper.go
```diff
@@ -366,6 +378,28 @@ func (api *RestAPI) BuildTripStatus(
...
+		}
+		freqs = rows
+		if freqMap != nil {
+			freqMap[tripID] = rows
+		}
+	}
+	if len(freqs) == 0 {
+		return nil, nil
+	}
+	converted := models.NewFrequencyFromDB(freqs[0], serviceDate)
```
aaronbrethorst: The first frequency row is used regardless of which frequency window the request time actually falls in. A trip with a peak/off-peak split in `frequencies.txt` — e.g. two rows for the same `trip_id`, `06:00-09:00 @600s` and `09:00-15:00 @1200s`, which is ordinary GTFS — will report the 06:00-09:00 window and a 10-minute headway for a 10:30 arrival. `startTime`, `endTime`, and `headway` are all wrong for any request outside the earliest window.

`serviceDate` (and the caller's effective time) is already in hand here, so the row whose `[start_time, end_time)` contains the requested time should be selected, falling back to the first row only when none matches. The same `freqs[0]` / `freqRows[0]` shortcut is repeated in `trip_for_vehicle_handler.go:120`, `arrival_and_departure_for_stop_handler.go:437`, and `arrivals_and_departures_for_stop_handler.go:502`, so a shared selector here would fix all of them at once.

internal/restapi/schedule_for_stop_handler.go
```diff
@@ -174,32 +198,33 @@ func (api *RestAPI) scheduleForStopHandler(w http.ResponseWriter, r *http.Reques
...
+			tripHeadsign := pluralityHeadsign(routeDirectionHeadsignCounts, combinedRouteID, dirID)
 
-			directionSchedule := models.NewStopRouteDirectionSchedule(tripHeadsign, stopTimes, nil)
+			frequencies := frequencyMap[dirID]
+			// Java OBA orders scheduleFrequencies by start time (FrequencyBeanComparator).
+			slices.SortStableFunc(frequencies, func(a, b models.ScheduleFrequency) int {
+				return cmp.Compare(a.StartTime.UnixMilli(), b.StartTime.UnixMilli())
+			})
+
+			directionSchedule := models.NewStopRouteDirectionSchedule(tripHeadsign, directionMap[dirID], frequencies)
```
aaronbrethorst: When a route/direction has only `exact_times=0` frequency trips at this stop, `routeDirectionScheduleMap` has no entry for the route, so `directionMap` is a nil map and `directionMap[dirID]` yields a nil slice. `NewStopRouteDirectionSchedule` special-cases nil `frequencies` (`[]ScheduleFrequency{}`) but passes `stopTimes` straight through, so the response emits `"scheduleStopTimes": null`.

`testdata/openapi.yml` lists `scheduleStopTimes` in `StopRouteDirectionSchedule.required`, and Java OBA emits an empty list — a `null` here will break clients that iterate the array unguarded. The fixture in `frequency_fixture_test.go` puts a plain scheduled trip in the same route+direction as the frequency trips, so this path is never exercised by the new tests.

Simplest fix is to mirror the frequencies handling in `NewStopRouteDirectionSchedule`:

```suggestion
			stopTimes := directionMap[dirID]
			if stopTimes == nil {
				stopTimes = []models.ScheduleStopTime{}
			}
			directionSchedule := models.NewStopRouteDirectionSchedule(tripHeadsign, stopTimes, frequencies)
```

internal/restapi/arrivals_and_departures_for_stop_handler.go
```diff
@@ -287,6 +287,24 @@ func (api *RestAPI) arrivalsAndDeparturesForStopHandler(w http.ResponseWriter, r
 		}
 	}
 
+	// Batch-fetch frequencies; seeding nil prevents BuildTripStatus from
+	// falling back to a per-arrival query. First row (earliest start_time)
+	// becomes the arrival's frequency field.
+	freqMap := make(map[string][]gtfsdb.Frequency, len(uniqueTripIDs))
+	if len(uniqueTripIDs) > 0 {
+		for _, tripID := range uniqueTripIDs {
+			freqMap[tripID] = nil
```
aaronbrethorst: Seeding every trip ID with `nil` *before* the batch query means a failed batch query is indistinguishable from "these trips have no frequencies". If `GetFrequenciesForTrips` returns an error, the code logs a warning and falls through with the map fully seeded, so `frequencyForEntry` finds `ok == true` for every trip, skips its fallback query, and returns `nil` — the whole response goes out as a 200 with `frequency: null` on every arrival and `status.frequency: null` throughout.

That contradicts the PR description's "strict error propagation (HTTP 500) for database failures during frequency lookups", and per CONTRIBUTING a non-`sql.ErrNoRows` DB error should go through `serverErrorResponse`. Moving the seeding loop into the `else` branch (so it only runs after a successful batch) would at least let the per-trip fallback take over; propagating the error is better still.

The same pattern is in `trips_for_location_handler.go:437-444` and `trips_for_route_handler.go:335-343`.

internal/restapi/trips_helper.go
aaronbrethorst: The first frequency row is used regardless of which frequency window the request time actually falls in. A trip with a peak/off-peak split in `frequencies.txt` — e.g. two rows for the same `trip_id`, `06:00-09:00 @600s` and `09:00-15:00 @1200s`, which is ordinary GTFS — will report the 06:00-09:00 window and a 10-minute headway for a 10:30 arrival. `startTime`, `endTime`, and `headway` are all wrong for any request outside the earliest window.

`serviceDate` (and the caller's effective time) is already in hand here, so the row whose `[start_time, end_time)` contains the requested time should be selected, falling back to the first row only when none matches. The same `freqs[0]` / `freqRows[0]` shortcut is repeated in `trip_for_vehicle_handler.go:120`, `arrival_and_departure_for_stop_handler.go:437`, and `arrivals_and_departures_for_stop_handler.go:502`, so a shared selector here would fix all of them at once.

internal/restapi/schedule_for_stop_handler.go
aaronbrethorst: When a route/direction has only `exact_times=0` frequency trips at this stop, `routeDirectionScheduleMap` has no entry for the route, so `directionMap` is a nil map and `directionMap[dirID]` yields a nil slice. `NewStopRouteDirectionSchedule` special-cases nil `frequencies` (`[]ScheduleFrequency{}`) but passes `stopTimes` straight through, so the response emits `"scheduleStopTimes": null`.

`testdata/openapi.yml` lists `scheduleStopTimes` in `StopRouteDirectionSchedule.required`, and Java OBA emits an empty list — a `null` here will break clients that iterate the array unguarded. The fixture in `frequency_fixture_test.go` puts a plain scheduled trip in the same route+direction as the frequency trips, so this path is never exercised by the new tests.

Simplest fix is to mirror the frequencies handling in `NewStopRouteDirectionSchedule`:

```suggestion
			stopTimes := directionMap[dirID]
			if stopTimes == nil {
				stopTimes = []models.ScheduleStopTime{}
			}
			directionSchedule := models.NewStopRouteDirectionSchedule(tripHeadsign, stopTimes, frequencies)
```

internal/restapi/arrivals_and_departures_for_stop_handler.go
aaronbrethorst: Seeding every trip ID with `nil` *before* the batch query means a failed batch query is indistinguishable from "these trips have no frequencies". If `GetFrequenciesForTrips` returns an error, the code logs a warning and falls through with the map fully seeded, so `frequencyForEntry` finds `ok == true` for every trip, skips its fallback query, and returns `nil` — the whole response goes out as a 200 with `frequency: null` on every arrival and `status.frequency: null` throughout.

That contradicts the PR description's "strict error propagation (HTTP 500) for database failures during frequency lookups", and per CONTRIBUTING a non-`sql.ErrNoRows` DB error should go through `serverErrorResponse`. Moving the seeding loop into the `else` branch (so it only runs after a successful batch) would at least let the per-trip fallback take over; propagating the error is better still.

The same pattern is in `trips_for_location_handler.go:437-444` and `trips_for_route_handler.go:335-343`.

internal/restapi/arrival_and_departure_for_stop_handler.go
```diff
@@ -430,6 +430,14 @@ func (api *RestAPI) arrivalAndDepartureForStopHandler(w http.ResponseWriter, r *
 		situationIDs,                                   // situationIds
 	)
 
+	// Populate frequency for the arrival's trip. The singular handler serves
+	// one trip per request, so a per-trip query is appropriate here.
+	freqRows, freqErr := api.GtfsManager.GtfsDB.Queries.GetFrequenciesForTrip(ctx, tripID)
+	if freqErr == nil && len(freqRows) > 0 {
```
aaronbrethorst: `freqErr` is swallowed here — a DB failure is silently indistinguishable from "no frequencies for this trip", and the arrival goes out with `frequency: null` and a 200. Every other frequency lookup added in this PR either propagates (`trip_for_vehicle_handler.go:115`, `schedule_for_stop_handler.go:172`) or at minimum logs. Per CONTRIBUTING, a non-`sql.ErrNoRows` error should reach `serverErrorResponse`:

```suggestion
	freqRows, freqErr := api.GtfsManager.GtfsDB.Queries.GetFrequenciesForTrip(ctx, tripID)
	if freqErr != nil {
		api.serverErrorResponse(w, r, freqErr)
		return
	}
	if len(freqRows) > 0 {
```

internal/restapi/trips_helper.go
```diff
@@ -152,6 +155,15 @@ func (api *RestAPI) BuildTripStatus(
...
 			slog.String("error", err.Error()))
 	}
+
+	// Reuse the batch-fetched frequencies when supplied; single-entry handlers
+	// fall back to a per-trip query.
+	frequency, freqErr := api.frequencyForEntry(ctx, freqMap, dbTripID, serviceDate)
+	if freqErr != nil {
+		return status, extras, freqErr
+	}
+	status.Frequency = frequency
```
aaronbrethorst: This assignment sits below the `CANCELED` early return at line 103, so a canceled frequency-based trip gets `status.frequency: null` while the enclosing entry (`TripDetails.Frequency`, `ArrivalAndDeparture.Frequency`, `TripsForRouteListEntry.Frequency`) still carries the frequency — the same trip reports two different answers in one response. Frequency is static schedule data and is just as valid for a canceled trip as `serviceDate` and `blockTripSequence`, both of which the code deliberately computes before that early return. Consider moving the lookup up next to `blockTripSequence` for the same reason.

internal/restapi/schedule_for_stop_handler.go
```diff
@@ -155,13 +156,36 @@ func (api *RestAPI) scheduleForStopHandler(w http.ResponseWriter, r *http.Reques
...
+	if len(scheduleRows) > 0 {
+		tripIDSet := make(map[string]bool, len(scheduleRows))
+		for _, row := range scheduleRows {
+			tripIDSet[row.TripID] = true
+		}
+		tripIDs := make([]string, 0, len(tripIDSet))
+		for tripID := range tripIDSet {
+			tripIDs = append(tripIDs, tripID)
+		}
+		allFreqs, freqErr := api.GtfsManager.GtfsDB.Queries.GetFrequenciesForTrips(ctx, tripIDs)
```
aaronbrethorst: `tripIDs` here is unbounded — it is every distinct trip serving this stop on the target date, which at a busy downtown stop can be well over a thousand. This is the only query in the handler whose bind-variable count scales with trip count (`GetScheduleForStopOnDate` binds route IDs, `GetTripsByBlockIDs` binds block IDs), and it now hard-fails the whole request with a 500 ("too many SQL variables") on exactly the stops most likely to be requested.

`queryInBatches` in `reference_utils.go:520` exists for this and is documented as staying under the oldest 999-variable limit so behavior doesn't depend on which SQLite the build links against:

```suggestion
		allFreqs, freqErr := queryInBatches(ctx, tripIDs, api.GtfsManager.GtfsDB.Queries.GetFrequenciesForTrips)
```

internal/restapi/trip_for_vehicle_handler.go
```diff
@@ -105,21 +102,34 @@ func (api *RestAPI) tripForVehicleHandler(w http.ResponseWriter, r *http.Request
...
+			api.serverErrorResponse(w, r, scheduleErr)
+			return
 		}
 	}
 
 	// BuildTripStatus above was given this same tripID, so when the status was
 	// built its situations are this trip's and are reused here.
 	situationIDs, situationRefs := api.tripSituationsFor(ctx, tripID, statusExtras)
 
+	freqRows, err := api.GtfsManager.GtfsDB.Queries.GetFrequenciesForTrip(ctx, tripID)
```
aaronbrethorst: When `includeStatus` is true (the default), `BuildTripStatus` was already called at line 76 with a nil `freqMap`, which runs `GetFrequenciesForTrip(ctx, tripID)` for this same trip inside `frequencyForEntry`. This is a second identical round-trip on every request. Passing a one-entry `freqMap` seeded here (or reading `status.Frequency` back) removes the duplicate — minor, but this PR's stated goal is eliminating redundant frequency queries. The same duplication exists in `arrival_and_departure_for_stop_handler.go:436`.

aaronbrethorst (dismissed):
please fix the items noted in the inline review and we'll get this merged!

internal/restapi/trips_helper.go
burma-shave: `frequencyForEntry` always uses `freqs[0]` — the first frequency row for the trip — regardless of which window the request's `serviceDate`/effective time actually falls into. A trip with a peak/off-peak split in `frequencies.txt` (ordinary GTFS: e.g. `06:00-09:00@600s` and `09:00-15:00@1200s` for the same `trip_id`) will always report the `06:00-09:00` window and headway, even for a 10:30 request.

Legacy Java doesn't do this — `ArrivalAndDepartureServiceImpl` resolves the frequency entry whose `[start_time, end_time)` actually contains the query time (`isFrequencyBasedArrivalInRange`), never "just the first row." The wiki spec already says "the applicable headway window" for this field, so this isn't just undocumented — it contradicts what's already written.

Confirmed client impact: Wayfinder (`ArrivalDeparture.svelte`) and Android (`ArrivalInfo.kt`) both render this field directly as "every N min, from/until HH:MM" — riders would see the wrong window/headway for any multi-window frequency trip. Android's `ArrivalAdapters.kt` also uses `frequency.startTime` in a sentinel-detection heuristic that would compare against the wrong boundary.

`serviceDate` (and the caller's effective time) is already available here, so please select the row whose `[start_time, end_time)` contains it, falling back to the first row only when none matches. The identical `freqs[0]`/`freqRows[0]` shortcut is repeated in `trip_for_vehicle_handler.go:125`, `arrival_and_departure_for_stop_handler.go:437`, `arrivals_and_departures_for_stop_handler.go:504`, and reimplemented again in `BuildTripSchedule` (`trips_helper.go:441`) instead of reusing this function — a single shared selector here would fix all five call sites at once.

internal/restapi/trips_helper.go
burma-shave: This assignment sits after the `CANCELED` early return at line 103, so a canceled frequency-based trip gets `status.frequency: null` while the enclosing entry's own `Frequency` field (populated separately in `trip_details_handler.go`, `trip_for_vehicle_handler.go`, `arrival_and_departure_for_stop_handler.go`, etc.) still carries the frequency data — the same trip reports two different answers for the same field concept in one response. Frequency is static schedule data, just as valid for a canceled trip as `serviceDate`/`blockTripSequence`, both of which are computed before this early return. Consider moving the lookup up alongside those.

internal/restapi/schedule_for_stop_handler.go
burma-shave: `tripIDs` here is unbounded — every distinct trip serving this stop on the target date, which at a busy stop can be well over a thousand. This is the only query in the handler whose bind-variable count scales with trip count, and it can hard-fail the whole request with a "too many SQL variables" error on exactly the stops most likely to be requested. `queryInBatches` (`reference_utils.go:520`) exists for this and keeps behavior independent of which SQLite build is linked:

```suggestion
		allFreqs, freqErr := queryInBatches(ctx, tripIDs, api.GtfsManager.GtfsDB.Queries.GetFrequenciesForTrips)
```

The same unbounded-list issue applies to `trips_for_route_handler.go:338`, `trips_for_location_handler.go:440`, and `arrivals_and_departures_for_stop_handler.go:298`.

Separately: this near-identical batch-fetch-and-seed block (build trip IDs, seed `freqMap[id] = nil`, call `GetFrequenciesForTrips`, append by `TripID`) is duplicated across all four of those files, with inconsistent failure handling — this file returns a 500 on query failure, the other three log a warning and continue with an empty/partial map (see the next comment for why that's also a correctness bug in at least one of them). Worth consolidating into one shared helper so there's a single place to fix both the batching and the error policy.

internal/restapi/arrivals_and_departures_for_stop_handler.go
burma-shave: Seeding every trip ID with `nil` *before* the batch query means a failed batch query is indistinguishable from "these trips have no frequencies." If `GetFrequenciesForTrips` errors, the code logs a warning and falls through with the map fully seeded, so `frequencyForEntry` finds `ok == true` for every trip, skips its fallback query, and returns `nil` — the response goes out as a 200 with `frequency: null` on every arrival and `status.frequency: null` throughout.

That contradicts this PR's own stated goal of "strict error propagation (HTTP 500) for database failures during frequency lookups," and per CONTRIBUTING a non-`sql.ErrNoRows` DB error should reach `serverErrorResponse`. Moving the seeding loop into the `else` branch (so it only runs after a successful batch) would at least let the per-trip fallback take over; propagating the error, as `schedule_for_stop_handler.go` already does, is more consistent with the rest of this PR. The same pattern is in `trips_for_location_handler.go:438` and `trips_for_route_handler.go:336`.

internal/restapi/schedule_for_stop_handler.go
burma-shave: When a route/direction has only `exact_times=0` frequency trips at this stop, `routeDirectionScheduleMap` has no entry for the route, so `directionMap` is a nil map and `directionMap[dirID]` yields a nil slice. `NewStopRouteDirectionSchedule` special-cases nil `frequencies` (`[]ScheduleFrequency{}`) but passes `stopTimes` straight through, so the response emits `"scheduleStopTimes": null`.

`testdata/openapi.yml` lists `scheduleStopTimes` in `StopRouteDirectionSchedule.required`, and Java always emits an empty list here — `null` will break clients that iterate the array unguarded. The fixture in `frequency_fixture_test.go` puts a plain scheduled trip in the same route+direction as the frequency trips, so this path isn't exercised by the new tests.

```suggestion
			stopTimes := directionMap[dirID]
			if stopTimes == nil {
				stopTimes = []models.ScheduleStopTime{}
			}
			directionSchedule := models.NewStopRouteDirectionSchedule(tripHeadsign, stopTimes, frequencies)
```

internal/restapi/schedule_for_stop_handler.go
```diff
@@ -469,34 +558,130 @@ func blockBoundaries(
...
+		firstStopTime = row.ArrivalTime
+	}
+
+	arrivalOffset := row.ArrivalTime - firstStopTime
+	departureOffset := row.DepartureTime - firstStopTime
+
+	isFirstInBlock, isLastInBlock := blockBoundaries(row, rowCtx.activeServiceBlockTripsMap)
+
+	expanded := make([]models.ScheduleStopTime, 0)
+	for offset := int64(0); freq.StartTime+offset < freq.EndTime; offset += headwayNs {
```
burma-shave, line 562: This headway loop has no cap on iteration count or generated entries. An `exact_times=1` frequency block with a small headway over a long window (e.g. 30s headway across a 20-hour span) generates on the order of 2,400 synthetic `ScheduleStopTime` entries per trip at that stop, with no upper bound the way `GetStopsForLocation`-style endpoints cap their result size elsewhere. A few such trips serving one busy stop could make a single request allocate and sort tens of thousands of entries. Worth adding a sane cap (with a warning log when hit), especially since this expansion is being kept as intentional behavior rather than reverted to match Java.

internal/restapi/arrival_and_departure_for_stop_handler.go
```diff
@@ -430,6 +430,14 @@ func (api *RestAPI) arrivalAndDepartureForStopHandler(w http.ResponseWriter, r *
 		situationIDs,                                   // situationIds
 	)
 
+	// Populate frequency for the arrival's trip. The singular handler serves
+	// one trip per request, so a per-trip query is appropriate here.
+	freqRows, freqErr := api.GtfsManager.GtfsDB.Queries.GetFrequenciesForTrip(ctx, tripID)
```
burma-shave: `freqErr` is swallowed here — a DB failure is silently indistinguishable from "no frequencies for this trip," and the arrival goes out with `frequency: null` and a 200. Every other frequency lookup added in this PR either propagates (`trip_for_vehicle_handler.go:115`, `schedule_for_stop_handler.go:172`) or at minimum logs. Per CONTRIBUTING, a non-`sql.ErrNoRows` error should reach `serverErrorResponse`:

```suggestion
	freqRows, freqErr := api.GtfsManager.GtfsDB.Queries.GetFrequenciesForTrip(ctx, tripID)
	if freqErr != nil {
		api.serverErrorResponse(w, r, freqErr)
		return
	}
	if len(freqRows) > 0 {
```

internal/restapi/trip_for_vehicle_handler.go
```diff
@@ -73,13 +73,10 @@ func (api *RestAPI) tripForVehicleHandler(w http.ResponseWriter, r *http.Request
...
 		var statusErr error
-		status, statusExtras, statusErr = api.BuildTripStatus(ctx, agencyID, tripID, nil, serviceDate, currentTime)
+		status, statusExtras, statusErr = api.BuildTripStatus(ctx, agencyID, tripID, nil, serviceDate, currentTime, nil)
 		if statusErr != nil {
-			api.Logger.Warn("failed to build trip status",
-				"tripID", tripID,
-				"agencyID", agencyID,
-				"error", statusErr)
-			status = nil
+			api.serverErrorResponse(w, r, statusErr)
```
burma-shave: `BuildTripStatus`/`BuildTripSchedule` failures here now call `api.serverErrorResponse` and abort the whole request for *any* error the function can return — not just the new frequency-lookup error this PR adds. Previously, any `BuildTripStatus` error (e.g. a transient DB hiccup elsewhere in the block-snapshot logic) degraded gracefully to `status: null`/`schedule: null` in a 200. That's a wider behavior change than the PR's stated goal of propagating frequency errors, and it's inconsistent with sibling handlers touched by this same PR — `trips_for_route_handler.go` and `arrival_and_departure_for_stop_handler.go` still log-and-continue (`status = nil`) on the identical `BuildTripStatus` error path. Worth confirming this scope expansion is intentional, and if so, applying it consistently rather than per-handler.

internal/restapi/trip_for_vehicle_handler.go
burma-shave: When `includeStatus` is true (the default), `BuildTripStatus` was already called above with a nil `freqMap`, which runs `GetFrequenciesForTrip(ctx, tripID)` for this same trip inside `frequencyForEntry`. This is a second, identical round-trip for the same trip on every request. Passing a one-entry `freqMap` seeded here (or reading `status.Frequency` back) removes the duplicate — minor, but this PR's stated goal is eliminating redundant frequency queries. The same duplication exists in `arrival_and_departure_for_stop_handler.go:436`.

burma-shave (changes requested):
Reviewed against legacy Java (pinned at the SHA the wiki already cites) and the codebase's own conventions, alongside @aaronbrethorst's review and an independent `/code-review` pass. Most of CodeRabbit's earlier actionable comments on this PR (N+1 caching, division-by-zero guard, error propagation in several handlers, the routeDirection key-type refactor, missing doc comments) are already fixed in the current HEAD and don't need further action — I re-verified each against the code rather than trusting the stale comment text.

Two spec-conformance items came out of comparing against legacy Java source:

1. **Frequency window selection always picks the first `frequencies.txt` row for a trip, not the one matching the request time** (see inline comment on `frequencyForEntry`). Legacy Java resolves the time-matching `FrequencyEntry` per request; this PR's `frequencyForEntry`/`freqs[0]` shortcuts don't. This needs fixing — see inline comment.
2. **`schedule-for-stop` expanding `exact_times=1` trips into discrete `scheduleStopTimes`** deviates from legacy Java, which lumps all frequency trips (`exact_times` 0 or 1) into `scheduleFrequencies` undifferentiated. We've decided to keep this behavior — it's a defensible improvement (GTFS's `exact_times=1` semantics call for exact timetable-equivalent output, and schedule-for-stop's underlying Java bean has no `exactTimes` field to fall back on, unlike the `frequency` object elsewhere). This has been documented in the `maglev.wiki` `schedule-for-stop` page's Implementation Decisions section — no action needed here, just flagging so the intentional divergence is visible in-PR too.

The rest are correctness/consistency findings from combining @aaronbrethorst's review with an independent `/code-review medium` pass — each verified against current HEAD and confirmed still applicable.

burma-shave (changes requested):
Looks like we have merge conflicts

burma-shave (approved):

Issue #1187, opened 2026-07-17
routes-for-agency: honor includeReferences=false (references not suppressed)


Mister-Raggs:
Part of #1179 (Spec review: routes-for-agency).

### Spec Requirement

The [wiki spec](https://github.com/OneBusAway/maglev/wiki/routes-for-agency) documents `includeReferences`:

> **`includeReferences`** — Controls whether the `references` block is populated. Defaults to `true`. Set to `false` to suppress reference data and reduce response size.

Confirmed against the canonical legacy binary (OBA 2.7.1, booted locally via maglev-workspace, KCM agency `1`):

```
includeReferences=false -> references.agencies: 0   (suppressed)
includeReferences=true  -> references.agencies: 1   (populated)
```

### Current Maglev Behavior

`routesForAgencyHandler` (`internal/restapi/routes_for_agency_handler.go`) never reads `includeReferences`; it always populates `references.agencies`:

```
# RABA test feed, agency 25
includeReferences=false -> references.agencies: 1   (not suppressed)
includeReferences=true  -> references.agencies: 1
```

### Proposed Solution

Parse `includeReferences` (default `true`); when `false`, return `*models.NewEmptyReferences()` instead of building the agency reference. Follows the same fix applied to `search-stop` (#1160) and `trip` (#1063). Add a test asserting the `references` block is empty when `includeReferences=false`.

### Note

This was initially (incorrectly) assessed as parity on #1179 based on production Puget Sound, which is version/feed-drifted and currently returns empty references even for `includeReferences=true`. The canonical 2.7.1 binary above is the authoritative reference.

Issue #1160, opened 2026-07-10
Support `includeReferences=false` query parameter


ARCoder181105:
### Description
**Spec Requirement:**
When `includeReferences=false` is passed in the query string, the endpoint must return the `list` of matching stops without populating the `references` block.

**Current Maglev Behavior:**
`searchStopsHandler` completely ignores `r.URL.Query().Get("includeReferences")`. It always executes reference batch queries and returns full `references` (`routes`, `agencies`, `situations`) regardless of what the client requested.

**Proposed Solution:**
- Parse `includeReferences` (`true` by default; check if `r.URL.Query().Get("includeReferences") == "false"`).
- If `includeReferences == false`, skip building `references.Routes`, `references.Agencies`, `references.Stops`, and `references.Situations`, returning `References: *models.NewEmptyReferences()`.

PR #1351, opened 2026-08-15
Honor includeReferences in stops-for-location

### LINKED ISSUE #1349: stops-for-location: includeReferences=false is ignored
Part of #1235.

### Spec Requirement

Extension 8a: *"includeReferences=false: The references block is present in the response but all its sub-arrays (`routes`, `agencies`, `stops`, `trips`, `situations`) are empty."*

Confirmed on canonical legacy 2.7.1 (maglev-workspace, KCM, `lat=47.6062&lon=-122.3321&radius=1000`):

```
includeReferences=true  -> references: agencies=3  routes=43
includeReferences=false -> references: agencies=0  routes=0
```

### Current Maglev Behavior

`stopsForLocationHandler` never reads the parameter — `ShouldIncludeReferences` is not called anywhere in the handler, so references are always populated:

```
# RABA, lat=40.583321&lon=-122.426966&radius=2000
includeReferences=true  -> references: agencies=1  routes=2
includeReferences=false -> references: agencies=1  routes=2
```

### Proposed Solution

Guard the reference block with `ShouldIncludeReferences(r)` (`internal/restapi/reference_utils.go`), returning `*models.NewEmptyReferences()` when false, as `routes-for-agency` (#1187), `search-stop` (#1160) and `trip` (#1063) already do. Add a test asserting the sub-arrays are empty when `includeReferences=false`.

Mister-Raggs:
Fixes #1349. Part of the second pass on the #1235 spec-review card.

## Problem

`stopsForLocationHandler` never read `includeReferences` — `ShouldIncludeReferences` was not called anywhere in the file, so the references block was populated identically whether the parameter was `true`, `false`, or absent.

Spec extension 8a: *"includeReferences=false: The references block is present in the response but all its sub-arrays are empty."*

Confirmed against canonical legacy 2.7.1 (maglev-workspace, KCM, `lat=47.6062&lon=-122.3321&radius=1000`):

```
includeReferences=true  -> references: agencies=3  routes=43
includeReferences=false -> references: agencies=0  routes=0
```

Maglev on RABA returned `agencies=1 routes=2` either way.

## Fix

Guard the reference block with `ShouldIncludeReferences(r)`, the same way `routes-for-agency` (#1187), `search-stop` (#1160) and `trip` (#1063) already do. The block is still emitted — `models.NewEmptyReferences()` gives empty (not null) sub-arrays — only its contents are skipped.

The guard also skips the work that fed it: the agency/route ID collection loop, both `Filter*` calls (one of which runs a full `ListRoutes`), and the per-stop alert lookup.

The early "no stops found" return already produced empty sub-arrays — `FilterAgencies`/`FilterRoutes` filter against empty maps there — so it needed no change.

## Testing

`TestStopsForLocationHonorsIncludeReferences` is table-driven over omitted / `true` / `false`, and asserts all three sub-arrays including `situations` (using the existing stop-2042 alert fixture, so the situations arm is non-trivially populated in the `true` cases).

Mutation-verified: forcing the guard true fails all three `false`-case assertions.

Full gauntlet green in both tag modes — gofmt, `go vet`, full suite, race detector, `purego`, `check-openapi`.

## Summary by CodeRabbit

* **New Features**
  * Added support for controlling whether agency, route, and situation references are included in stop-location responses.
  * Stops continue to be returned regardless of the reference setting.
  * References are included by default and can be omitted with `includeReferences=false`.

* **Bug Fixes**
  * Ensured references are collected consistently after result limits are applied.

* **Tests**
  * Added coverage for omitted, enabled, and disabled reference options.

### DIFF
--- a/internal/restapi/stops_for_location_handler.go
+++ b/internal/restapi/stops_for_location_handler.go
@@ -25,6 +25,7 @@ func (api *RestAPI) stopsForLocationHandler(w http.ResponseWriter, r *http.Reque
 	loc, fieldErrors := api.parseLocationParams(r, fieldErrors)
 	maxCount, fieldErrors := utils.ParseMaxCountClamped(queryParams, models.DefaultMaxCountForStops, fieldErrors)
 	query := queryParams.Get("query")
+	includeReferences := ShouldIncludeReferences(r)
 
 	var routeTypes []int
 	if routeTypeStr := queryParams.Get("routeType"); routeTypeStr != "" {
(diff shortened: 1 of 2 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is the version of this change I'd want other handlers copied from.

- Reuses `ShouldIncludeReferences` and `models.NewEmptyReferences()`, so the
  suppressed case is present-but-empty rather than null, matching
  `routes_for_agency_handler.go` and `stops_for_route_handler.go`. Consistency
  across handlers on this semantic matters more than the individual handler,
  so I'm glad you went and looked at what the others do.
- Moving the ID-collection loop inside the guard is the right call — that work
  only exists to feed references, so skipping it when references are suppressed
  is real savings, not just a shape change. And hoisting the sort above it is
  behavior-neutral since the loop iterates `resultRawStopIDs`, not `results`.
- The test seeding the stop-2042 alert is what makes it worth having: the
  `true` arm asserts agencies, routes *and* situations are non-empty, so all
  three assignments inside the guard are genuinely covered rather than passing
  vacuously. That's better than the sibling routes-for-location PR does it.
- Thanks for flagging the `agencies=1 routes=2` vs legacy `3/43` discrepancy on
  KCM in the description rather than quietly fixing it here. That's the right
  handling — it's a separate question and I'll take it up against the wiki spec.

Merging.

PR #1352, opened 2026-08-15
Accept the documented time formats in stops-for-location

### LINKED ISSUE #1350: stops-for-location: time in yyyy-MM-dd_HH-mm-ss form is parsed as epoch millis
Part of #1235.

### Spec Requirement

The `time` parameter *"Accepts Unix milliseconds as a digit string, or the format `yyyy-MM-dd_HH-mm-ss`"*.

Confirmed on canonical legacy 2.7.1 (maglev-workspace, KCM, `radius=1000&maxCount=250`; `maxCount` raised above the candidate count so legacy's shuffle/truncate can't add noise — the same request repeated three times returned an identical count):

```
today       epoch=157  string(2026-08-15_12-00-00)=157
far_future  epoch=0    string(2030-09-23_12-00-00)=0
far_past    epoch=0    string(2022-07-07_12-00-00)=0
no time param = 157
```

The far-date rows are the discriminating ones: the string form tracks the epoch form to 0 rather than falling back to the 157 returned when `time` is absent, so legacy parses the format.

### Current Maglev Behavior

The handler parses `time` with `fmt.Sscanf(timeStr, "%d", &timeMs)`. On `2025-06-13_14-00-00` that reads `2025` off the front and **succeeds**, so the value is treated as 2025 *milliseconds* — 1 Jan 1970 — and the active-service lookup finds nothing:

```
# RABA
time=1749848400000        -> 7 stops
time=2025-06-13_14-00-00  -> 0 stops   (same instant, documented format)
time=garbage              -> 0 stops
```

Note this is worse than the malformed case: `garbage` fails the parse and correctly falls back to the current time, while a *valid* documented value silently produces a wrong result with HTTP 200.

### Proposed Solution

Use `utils.ParseTimeParameter` (`internal/utils/api.go`), which already handles epoch millis and `yyyy-MM-dd_HH-mm-ss`, instead of the local `Sscanf`. Add tests for the string form and for a malformed value falling back to the current time.

Mister-Raggs:
Fixes #1350. Part of the second pass on the #1235 spec-review card.

## Problem

The `time` parameter *"accepts Unix milliseconds as a digit string, or the format `yyyy-MM-dd_HH-mm-ss`"*. The handler parsed it with `fmt.Sscanf(timeStr, "%d", &timeMs)`, which reads `2025` off the front of `2025-06-13_14-00-00` and **succeeds**. The value is then treated as 2025 *milliseconds* — 1 Jan 1970 — so no service is active and the response is empty, with HTTP 200 and no indication anything went wrong.

```
# RABA
time=1749848400000        -> 7 stops
time=2025-06-13_14-00-00  -> 0 stops   (same instant, documented format)
```

This is worse than the malformed case: `time=garbage` fails the parse and correctly falls back to the current time, while a *valid* documented value silently produces a wrong answer.

Confirmed against canonical legacy 2.7.1 (maglev-workspace, KCM, `radius=1000&maxCount=250`, chosen so the candidate count stays under the cap and legacy's shuffle/truncate can't add noise — the same request repeated three times returned an identical count):

```
today       epoch=157  string(2026-08-15_12-00-00)=157
far_future  epoch=0    string(2030-09-23_12-00-00)=0
far_past    epoch=0    string(2022-07-07_12-00-00)=0
no time param = 157
```

The far-date rows are the discriminating ones: the string form tracks the epoch form down to 0 rather than falling back to the 157 returned when `time` is absent, so legacy does parse the format.

## Fix

Use the existing `utils.ParseTimeParameter`, which already handles epoch millis, `yyyy-MM-dd_HH-mm-ss` and the date-only form. Two details:

- **Timezone.** The value is read against the agency's timezone, the same way `trips-for-location` resolves its own `time` for a location-scoped (non-agency-scoped) request. Previously `time.UnixMilli` gave the server's local zone.
- **Malformed values still fall back to the current time.** `ParseTimeParameter` returns field errors, but propagating them would turn `time=garbage` into a 400. Legacy serves that request (HTTP 200, same count as no `time` param at all), so the field errors are deliberately discarded and the fallback preserved. This is a preserve, not a fix.

The 15-minute binning is dropped rather than reimplemented. It cannot affect this handler: the value is only ever used at date granularity (`Format("20060102")`), 900000 ms divides a day evenly, and every IANA zone offset is itself a multiple of 15 minutes — so rounding down to a 15-minute UTC boundary can never cross a local midnight.

Resolving the time after the empty-result early return also shortens the variable's travel — it is now declared immediately above its only use.

## Testing

`TestStopsForLocationAcceptsDocumentedTimeFormats` is table-driven over epoch millis / `yyyy-MM-dd_HH-mm-ss` / date-only / malformed / absent. The clock is pinned to 2031, outside the fixture's service dates, so a time that parses shows up as a *non-empty* result rather than as a count that merely shifts — and the three accepted forms are asserted to return the identical stop list, not just a non-empty one.

Mutation-verified: restoring the `Sscanf` parse empties both the `yyyy-MM-dd_HH-mm-ss` and date-only cases.

Full gauntlet green in both tag modes — gofmt, `go vet`, full suite, race detector, `purego`, `check-openapi`.

## Also included: the current-time fallback timezone

Raised by review. The helper this PR adds read an explicit `time` in the agency's timezone but formatted its own fallback — used when `time` is absent or unparseable — from the clock's location. A service date is a local calendar date, so on a UTC-hosted instance serving a Pacific agency that named the *next* day's service for roughly seven hours every evening.

Reproduced on the RABA fixture. Both instants are Friday 2025-06-13 in `America/Los_Angeles`; only the second has rolled over in UTC:

```
13:00 local (2025-06-13T20:00Z) -> 15 stops
22:00 local (2025-06-14T05:00Z) -> 12 stops
```

`20250613` has 5 active service IDs against `20250614`'s 4, so the two dates really do select different service.

Loading the agency location before resolving the parameter fixes both routes to the fallback, and delegating the empty parameter to `ParseTimeParameter` removes a branch. Covered by `TestStopsForLocationUsesAgencyDateForCurrentTime`, which exercises the absent and malformed cases separately because they reach the fallback by different paths.

**This half is a pre-existing defect on the no-`time` path, not part of #1350**, and it changes the service date every request without a `time` param resolves to. It is in its own commit and can be split out if preferred — see the [review reply](https://github.com/OneBusAway/maglev/pull/1352#issuecomment-5304929259).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stop results for location searches by consistently interpreting supported time formats.
  * Corrected service-date selection to use the agency’s local calendar date, including around UTC midnight.
  * Invalid or missing time values now safely return no stops without rejecting the request.

### DIFF
--- a/internal/restapi/stops_for_location_handler.go
+++ b/internal/restapi/stops_for_location_handler.go
@@ -62,17 +62,6 @@ func (api *RestAPI) stopsForLocationHandler(w http.ResponseWriter, r *http.Reque
 		}
 	}
 
-	queryTime := api.Clock.Now()
-
-	if timeStr := queryParams.Get("time"); timeStr != "" {
-		var timeMs int64
-		if _, err := fmt.Sscanf(timeStr, "%d", &timeMs); err == nil {
-			// Bin to 15 minutes
-			binnedMs := timeMs - (timeMs % 900000)
-			queryTime = time.UnixMilli(binnedMs)
-		}
-	}
-
 	if len(fieldErrors) > 0 {
 		api.validationErrorResponse(w, r, fieldErrors)
 		return
@@ -142,8 +131,7 @@ func (api *RestAPI) stopsForLocationHandler(w http.ResponseWriter, r *http.Reque
 		return
 	}
 
-	// Get active service IDs for the requested queryTime
-	currentDate := queryTime.Format("20060102")
+	currentDate := api.queryTimeForStops(queryParams.Get("time"), allAgencies).Format("20060102")
 	activeServiceIDs, err := api.GtfsManager.GtfsDB.Queries.GetActiveServiceIDsForDate(ctx, currentDate)
 	if err != nil {
 		api.serverErrorResponse(w, r, err)
@@ -265,6 +253,30 @@ func (api *RestAPI) stopsForLocationHandler(w http.ResponseWriter, r *http.Reque
 
 	response := models.NewListResponseWithRange(results, *references, outOfRange, api.Clock, isLimitExceeded)
 	api.sendResponse(w, r, response)
+}
+
+// queryTimeForStops resolves the time parameter that selects which service date's
+// routes are considered active. The value is read in the agency's timezone, as in
+// trips-for-location. A value that does not parse falls back to the current time
+// rather than failing the request, which is what legacy does with one.
+func (api *RestAPI) queryTimeForStops(timeParam string, agencies []gtfsdb.Agency) time.Time {
+	now := api.Clock.Now()
+	if timeParam == "" || len(agencies) == 0 {
+		return now
+	}
+
+	location, err := loadAgencyLocation(agencies[0].ID, agencies[0].Timezone)
+	if err != nil {
+		api.Logger.Warn("failed to load agency timezone for time parameter",
+			"agencyID", agencies[0].ID, "error", err)
+		return now
+	}
+
+	_, parsedTime, _, ok := utils.ParseTimeParameter(timeParam, location, api.Clock)
+	if !ok {
+		return now
+	}
+	return parsedTime
 }
 
 // routeIDsByStop maps each stop to the routes serving it. Stop-code queries report every
--- a/internal/restapi/stops_for_location_handler.go
+++ b/internal/restapi/stops_for_location_handler.go
@@ -255,26 +255,26 @@ func (api *RestAPI) stopsForLocationHandler(w http.ResponseWriter, r *http.Reque
 	api.sendResponse(w, r, response)
 }
 
-// queryTimeForStops resolves the time parameter that selects which service date's
-// routes are considered active. The value is read in the agency's timezone, as in
-// trips-for-location. A value that does not parse falls back to the current time
-// rather than failing the request, which is what legacy does with one.
+// queryTimeForStops resolves the time that selects which service date's routes are
+// considered active. A service date is a local calendar date, so both an explicit
+// time parameter and the current-time fallback are read in the agency's timezone,
+// as in trips-for-location. A value that does not parse falls back to the current
+// time rather than failing the request, which is what legacy does with one.
 func (api *RestAPI) queryTimeForStops(timeParam string, agencies []gtfsdb.Agency) time.Time {
-	now := api.Clock.Now()
-	if timeParam == "" || len(agencies) == 0 {
-		return now
+	if len(agencies) == 0 {
+		return api.Clock.Now()
 	}
 
 	location, err := loadAgencyLocation(agencies[0].ID, agencies[0].Timezone)
 	if err != nil {
 		api.Logger.Warn("failed to load agency timezone for time parameter",
 			"agencyID", agencies[0].ID, "error", err)
-		return now
+		return api.Clock.Now()
 	}
 
 	_, parsedTime, _, ok := utils.ParseTimeParameter(timeParam, location, api.Clock)
 	if !ok {
-		return now
+		return api.Clock.Now().In(location)
 	}
 	return parsedTime
 }
--- a/internal/restapi/stops_for_location_handler_test.go
+++ b/internal/restapi/stops_for_location_handler_test.go
@@ -690,3 +690,46 @@ func TestStopsForLocationAcceptsDocumentedTimeFormats(t *testing.T) {
 		})
 	}
 }
+
+// A service date is a local calendar date, so the date is taken in the agency's
+// timezone even when the server clock's own date has already rolled over.
+func TestStopsForLocationUsesAgencyDateForCurrentTime(t *testing.T) {
+	// Both instants are Friday 2025-06-13 in RABA's America/Los_Angeles, but the
+	// second has already rolled over to the 14th in UTC.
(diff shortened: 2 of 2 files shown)


Mister-Raggs:
Good catch — confirmed and fixed in f69b66f.

It reproduces on the RABA fixture. Both of these instants are Friday 2025-06-13 in `America/Los_Angeles`, and only the second has rolled over to the 14th in UTC:

```
13:00 local (2025-06-13T20:00Z) -> 15 stops
22:00 local (2025-06-14T05:00Z) -> 12 stops
```

The two dates have genuinely different service: `20250613` has 5 active service IDs, `20250614` has 4. So a UTC-hosted instance serving a Pacific agency answered against the wrong service date for roughly seven hours every evening.

Loading the agency location before resolving the parameter fixes both routes to the fallback. Delegating the empty parameter to `ParseTimeParameter` — which already returns the current time in the location it is given — also drops a branch.

`TestStopsForLocationUsesAgencyDateForCurrentTime` covers the absent and malformed cases separately, since they reach the fallback by different paths; each was mutation-verified against its own path.

@aaronbrethorst worth noting this half is a pre-existing defect on the no-`time` path rather than part of #1350, and it changes the service date every request without a `time` param resolves to. I fixed it here because the helper this PR introduces would otherwise read the parameter in one timezone and its own fallback in another, which the doc comment would have been lying about. Happy to split it into its own issue and PR if you'd rather keep #1350 to the format parsing — and either way it may deserve a card entry, since the same fallback pattern likely appears on other endpoints.

Mister-Raggs:
@coderabbitai review

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Nice catch on the underlying bug — `fmt.Sscanf(timeStr, "%d", &timeMs)` happily
reading `2025` off the front of `2025-06-13_14-00-00` and resolving to 1970 is
exactly the kind of silent wrong-answer that's hard to notice from the outside.

Things I checked and that hold up:

- It routes through `utils.ParseTimeParameter` rather than growing another
  parser, which is what CONTRIBUTING.md's Code Reuse section asks for and what
  trips-for-route, stops-for-route, vehicles-for-agency and trips-for-location
  already do. Epoch-vs-string disambiguation is handled inside that helper, so
  a digit-only string can't be mis-read as a date.
- Dropping the 15-minute binning is safe — `queryTime` fed exactly one consumer,
  `.Format("20060102")`, so binning to a quarter hour never affected anything.
- The second commit is the one I'd have asked for if it weren't here: a service
  date is a local calendar date, so taking the fallback in the agency's zone
  rather than the server clock's is correct, and it brings this in line with
  trips-for-location.
- `TestStopsForLocationUsesAgencyDateForCurrentTime` is the good test — two
  instants that are the same local Friday but straddle UTC midnight will fail
  if the fallback ever drifts back to the server zone. That's a real regression
  guard, not a shape check.
- Serving a malformed `time` as 200 rather than 400 preserves legacy behavior,
  and thanks for calling that out in the description rather than leaving me to
  spot it — that's the discrepancy handling CONTRIBUTING.md asks for.

Non-blocking: the comment at the `currentDate` line still reads "for the
requested queryTime", but `queryTime` is the variable this PR deleted. Worth a
one-word fix next time you're in the file.

Merging.

aaronbrethorst:
Approved on the merits — see my review above. Heads up: I just merged your #1351
into `main`, and since both PRs restructure the same tail end of
`stopsForLocationHandler`, this branch has picked up merge conflicts.

Please merge `main` in and resolve. The two changes are complementary — #1351
wrapped the reference block in the `includeReferences` guard, and this one
replaces the time parsing further up — so the resolution should be mechanical.
Once it's mergeable it's good to go; no re-review needed unless the resolution
changes behavior.

Mister-Raggs:
@aaronbrethorst rebased.

burma-shave:
FYI (non-blocking, no code change needed): did a full spec-conformance pass on this while reviewing and updated the wiki with what we found — no action needed here.

- Documented the malformed-`time` fallback (falls back to current time rather than erroring), which matched legacy but wasn't written down: [stops-for-location § Extensions, 4a](https://github.com/OneBusAway/maglev/wiki/stops-for-location#extensions).
- Traced legacy's actual timezone handling for the current-time fallback: Java resolves it from the server JVM's own default timezone (a single process-wide setting), never per-agency. Your fix — resolving it against the queried agency's timezone instead — is a deliberate improvement over that, not a literal port of Java's mechanism, so it's now written up as its own decision rather than left implied: [stops-for-location § Implementation Decisions](https://github.com/OneBusAway/maglev/wiki/stops-for-location#implementation-decisions).
- Flagged the `agencies[0]` stand-in this (and `trips-for-location`) uses for "the queried agency" as a single-feed-limitation deviation, tracked in the [Multi-Agency-Static-Feeds-Stop-Consolidation §10 table](https://github.com/OneBusAway/maglev/wiki/Multi-Agency-Static-Feeds-Stop-Consolidation#10-deviations-from-legacy-due-to-the-single-feed-limitation) to revisit once multi-feed support lands.

Nice catch on the pre-existing UTC-fallback bug, and thanks for the reproduction detail in your commit message — made this easy to verify.

PR #1353, opened 2026-08-15
Fix/status stops missing from ref stops when include schedule is false

### LINKED ISSUE #1338: trips-for-route: status stops are missing from references.stops
Depends on #1314 — it touches the same reference-building path.

**Current**: `stopIDsMap` is written only by `collectStopIDsFromSchedule` (`trips_for_route_handler.go:400`, `:479`), and both calls sit inside `if includeSchedule`. Nothing collects `status.closestStop` / `status.nextStop`.

So `includeSchedule=false&includeStatus=true` returns entries whose statuses name stop IDs against an empty `references.stops`, and every one of those IDs dangles.

**Spec**: every ID an entry emits must resolve in the references block.

**Fix**: collect the status stop IDs into `stopIDsMap` before the reference build, for normal and DUPLICATED entries alike. `trips-for-location` does exactly this in `stopsReferencedByEntries` (`trips_for_location_handler.go:236`) — same shape, already reviewed.

**Test**: drive the handler with `includeSchedule=false&includeStatus=true` and assert each non-empty `closestStop`/`nextStop` resolves. `TestTripsForLocationHandler_StatusStopsAreReferenced` is the sibling case.

JohnAkindipe:
# Fix dangling status stops in references.stops in trips-for-route

Closes #1338

## Problem

`trips-for-route` builds `references.stops` exclusively from the stop IDs collected out of each entry's schedule. That collection runs only inside the `includeSchedule` branch — in both the normal entry loop and the loop that appends DUPLICATED (extra-run) entries from the real-time feed — so when a request asks for `includeSchedule=false&includeStatus=true`, no stop IDs are collected at all, while every entry's status still names a `closestStop` and `nextStop`. Those IDs end up having no corresponding value in `references.stops`.

## Fix

After both entry loops finish and before the reference is built, the handler now makes a single pass over the finished entries and records each non-nil status's `closestStop` and `nextStop` into the same stop-ID accumulator the schedule collection feeds. Additionally, for cross-agency interlined entries, where the same physical stop could possibly be referred to by different agency-qualified combined stop IDs, the handler resolves all agency-qualified stopIDs in `references.stops`.

## Testing

`TestTripsForRouteHandler_StatusStopsAreReferenced` drives the handler with `includeSchedule=false&includeStatus=true` and asserts that every non-empty `status.closestStop`/`nextStop` across all returned entries resolves in `references.stops`, while also asserting that no schedule information is returned. `TestTripsForRouteHandler_CrossAgencyInterlinedBlock` now includes a sub-test to verify that all agency-qualified stopIDs in cross-agency interlined blocks (which may reference the same physical stop as the queried-route agency) all resolve in `references.stops`. Multiple agencies referencing the same physical stop use the same underlying `gtfsdb.Stop` row. These tests fail on the pre-fix code and pass post-fix.

## Note
Previously, if `GetStopsByIDs` returned an error, the handler logged a warning and degraded `references.Stops` to an empty slice; the response then continued and returned `200 OK` with `references.Stops` totally empty. This behavior contradicts CONTRIBUTING.md, on handling query failures. Now, if `GetStopsByIDs` returns an error the handler sends a `500` server error response instead.

## Summary by CodeRabbit

* **Bug Fixes**
  * Trip results now include stops referenced by real-time status information, even when schedules are excluded.
  * Closest and next status stops are handled independently from scheduled stops.
  * Missing status data and malformed stop identifiers continue to be handled safely.
  * Requests now report an error when referenced stop information cannot be resolved.

* **Tests**
  * Added coverage confirming that status stops appear in trip results without schedules.


ARCoder181105:
@JohnAkindipe Fix sonarcloud issues , follow what contributing.md says ; as i can there is duplications of code fix that

aaronbrethorst:
### Code review

Found 1 issue:

1. `stopsReferencedByEntries` is now shared by two handlers but still lives in `trips_for_location_handler.go`, and its doc comment no longer describes what it does. CONTRIBUTING.md says reference-building helpers belong in `internal/restapi/reference_utils.go` ("Reference building (`internal/restapi/reference_utils.go`) — building the Agency/Route/Situation reference blocks used in list/entry responses"), and c00a305a ("Build every stop reference through one helper") already moved the sibling `stopReferences` there for exactly this reason. The comment still says "those on each entry's schedule, plus the closest and next stops on its status" and "the in-bounds stop set is deliberately not included" — the function no longer takes entries, and trips-for-route has no in-bounds stop set. The signature change is also why both call sites now repeat the same 7-line entry → `(schedules, statuses)` split; a shared helper in `reference_utils.go` taking the two entry types (or a small shared accessor) would keep that in one place.

https://github.com/OneBusAway/maglev/blob/27b3ff41018607e9ebdce5aa387b2fd1a7fd7f55/internal/restapi/trips_for_location_handler.go#L240-L247

aaronbrethorst (dismissed):
The diagnosis is right and the fix works — I confirmed the new test fails
against `main` and passes with the change. With `includeSchedule=false` there
are no schedule stop IDs at all, so `status.closestStop`/`nextStop` are the only
thing `references.stops` has to resolve from, and building the stops block
purely from schedule IDs left them dangling. Reusing trips-for-location's
existing helper instead of writing a second one is the right instinct.

What I'd like changed before this lands:

**1. The now-shared helper is still in a handler-specific file, and its doc
comment no longer describes it.**
`stopsReferencedByEntries` is called from both `trips_for_route_handler.go` and
`trips_for_location_handler.go`, but lives in the latter. CONTRIBUTING.md puts
reference-building helpers in `internal/restapi/reference_utils.go`, and commit
c00a305a moved the sibling `stopReferences` there for exactly this reason.

The comment is the bigger problem, because it now misleads:

> stopsReferencedByEntries fetches the stops the response actually refers to:
> those on each entry's schedule, plus the closest and next stops on its status.
> The in-bounds stop set is deliberately not included — it is a candidate-trip
> selection detail...

There are no entries in the signature any more, and "the in-bounds stop set" is
a trips-for-location concept that means nothing to a trips-for-route reader.
Please move the function to `reference_utils.go`, rewrite the comment for the
shared contract, and rename it — it takes schedules and statuses now, so
something like `stopsReferencedBySchedulesAndStatuses` says what it does.

**2. The parallel-slice signature pushes duplication into both call sites.**
Both handlers now repeat the same seven lines splitting entries into
`schedules`/`statuses` (`trips_for_route_handler.go` ~505, and
`trips_for_location_handler.go` ~118). That's the duplication @soumajitgh
flagged, and it's a consequence of the signature choice rather than an oversight
— two parallel slices that must stay index-aligned is a shape worth avoiding on
its own. Consider having the helper take the pairs directly (a small
`{Schedule, Status}` struct, or a variadic of them) so neither caller has to
build two lists in lockstep.

**3. Please note the error-handling change in the description.**
A `GetStopsByIDs` failure used to log a warning and degrade to an empty stops
block; it now returns a 500. I think that's the *right* change — CONTRIBUTING.md
is explicit that a failed query shouldn't be collapsed into a
looks-like-success response — but it's a behavior change beyond the stated scope
and a reviewer shouldn't have to discover it from the diff.

Small one: typo in the new test's comment — "the whole of whatreferences.stops".

One logistics note: I just merged #1360, which restructures `buildTripReferences`
in the same handler, so you'll need to merge `main` in and resolve. The two
changes are complementary — #1360 seeds routes from the stop references, this
one changes where the stops themselves come from — but the resolution isn't
purely mechanical, so give it a careful read. Happy to re-review promptly.

JohnAkindipe:
Thanks — the requested changes have been addressed:

- Renamed `stopsReferencedByEntries` to `prestopsReferencedBySchedulesAndStatuses` and moved it to `internal/restapi/reference_utils.go`. The doc comment has been updated to accurately describe its purpose.
- Refactored the helper to accept a slice of `{Schedule, Status}` structs, avoiding the need to build and maintain index-aligned slices in lockstep.
- Also noted the error-handling change in the PR description.

burma-shave:
@JohnAkindipe checking in on this PR. It is still marked dirty/blocked after earlier requested changes. Could you please merge `main`, resolve the merge conflicts, and let us know when it is ready for re-review?

burma-shave:
@JohnAkindipe checking in on this PR. It is still marked dirty after earlier requested changes. Could you please merge `main`, resolve the merge conflicts, and let us know when it is ready for re-review?

JohnAkindipe:
@burma-shave per the comment noted by coderabbit, it describes a situation where for an interlined block entry spanning multiple agencies, the queried-route trip's `schedule.stopTimes.stopID` prefixed by its `agencyID`, and then the same stop is referenced by the `status.closestStop.StopID` by a different `agencyID` (both agencies refer to the same stop, only the `agencyID` prefix differs), the current PR's implementation would mean that the queried-route trip's combined stop id is the only one which would resolve in references. The combined `stopID` in `status.closestStop.StopID` will not resolve in references because the current implementation assumes that stops are prefixed by the same one agency. 

However, after reflecting more on this, I realized that this scenario is not scoped to just a single entry alone. It is guaranteed that the `schedule.stopTimes.stopID` across all entries are prefixed by the same `agencyID`. However, each entry could possibly have a `status.closestStop.StopID` with a different `agencyID` prefix referencing the same stop in another entirely different entry. This means that we can have an arbitrary number of combined `stopIDs` with different `agency_ID` prefix which all reference the same bare `stopID`. 

A possible fix is to resolve every combined `stopID` in references, regardless of if multiple combined `stopIDs` reference the same bare `stopID`.

I decided to confirm if this is a suitable direction to go with this and to gain some more clarification on this.

burma-shave:
Thanks for digging into the CodeRabbit note. Yes, your proposed direction sounds right to me and is aligned with the OneBusAway response model: `references.stops` should include the exact combined stop IDs that the response entries emit.

The important contract is that IDs such as `schedule.stopTimes[*].stopId`, `status.closestStop`, and `status.nextStop` should resolve by exact string match in `references.stops`. So if an interlined/cross-agency response can emit both `agencyA_stop1` and `agencyB_stop1` for the same underlying bare GTFS stop ID, the reference builder should retain and publish both combined IDs rather than collapsing them to one entry keyed only by bare `stop1`.

Concretely, I’d suggest changing the collection from “one combined ID per bare stop ID” to “all combined IDs per bare stop ID”, then emitting one stop reference for each combined ID using the same underlying stop row.

Also, please merge/rebase against current `main`; the PR is currently marked dirty and conflicts in `internal/restapi/trips_for_route_handler.go`. Once that’s resolved and this reference preservation case is handled, we can re-review.

JohnAkindipe:
@aaronbrethorst  @burma-shave , this is ready for re-review. The CodeRabbit comments have been fully addressed. For a `stop` referenced by multiple agencies, `trips-for-route-handler` now resolves all the agency-qualified stop IDs for bare stop IDs in `references.stops`. This PR also includes regression tests which ensure the intended behaviour. The PR description has also been updated to describe the latest changes.

PR #1357, opened 2026-08-17
Default routes-for-location maxCount to 10

### LINKED ISSUE #1355: routes-for-location: maxCount defaults to 50 instead of 10
### Current behavior

`routesForLocationHandler` passes `models.DefaultMaxCountForRoutes` (= 50,
`internal/models/constants.go:27`) as the default to `ParseMaxCountClamped`
(`internal/restapi/routes_for_location_handler.go:21`). A request with no `maxCount`
therefore returns up to 50 routes.

This constant predates the spec review (added in #172); it was not introduced by the
`maxCount` clamping work in #1231, which added the separate ceiling constant
`MaxCountForRoutesForLocation = 50`. Default and ceiling are currently conflated.

### Expected behavior (per spec)

> "maxCount": { "type": "integer", "default": 10, "description": "Maximum number of routes to return; silently clamped to 50" }

Java matches: `new MaxCountSupport(10, 50)` in `RoutesForLocationAction` — 10 is the default,
50 the ceiling.

Observable difference: at a centre with 13 in-bounds routes and no `maxCount`, the spec
returns 10 routes with `limitExceeded: true`; maglev returns 13 with `limitExceeded: false`.

### What to change

1. Replace `models.DefaultMaxCountForRoutes = 50` with an endpoint-specific
   `DefaultMaxCountForRoutesForLocation = 10`, placed beside the existing
   `MaxCountForRoutesForLocation = 50` so default and ceiling read together. The old constant
   has exactly one call site.
2. Use it at `routes_for_location_handler.go:21`. The 50 ceiling and the `maxCount <= 0`
   validation error are unchanged.
3. Test: a request with no `maxCount` at a location with more than 10 in-bounds routes returns
   exactly 10 entries and `limitExceeded: true`.
4. `TestRoutesForLocationBoundingBoxSizing` case "spans widen the box beyond the default radius"
   expects all 13 RABA routes and relies on the old default — pin it with an explicit `maxCount`.

ARCoder181105:
Closes #1355

`maxCount` defaulted to 50 on `routes-for-location`; the spec and Java's
`MaxCountSupport(10, 50)` both use 10 as the default and 50 as the ceiling. The 50 default
predates the spec review and was unrelated to the clamping fix in #1231.

- `DefaultMaxCountForRoutes` (50) replaced by `DefaultMaxCountForRoutesForLocation` (10),
  placed beside the existing ceiling constant.
- New test: no `maxCount` with more than 10 in-bounds routes returns 10 and `limitExceeded: true`.
- The span-sizing case now passes an explicit `maxCount` so it stays about box sizing.

### DIFF
--- a/internal/models/constants.go
+++ b/internal/models/constants.go
@@ -24,9 +24,11 @@ const (
 )
 
 const (
-	DefaultMaxCountForRoutes = 50
-	DefaultMaxCountForStops  = 100
-	MaxAllowedCount          = 250
+	DefaultMaxCountForStops = 100
+	MaxAllowedCount         = 250
+	// DefaultMaxCountForRoutesForLocation is the routes-for-location default when the
+	// caller omits maxCount; MaxCountForRoutesForLocation is the ceiling it is clamped to.
+	DefaultMaxCountForRoutesForLocation = 10
 	// MaxCountForRoutesForLocation is the routes-for-location ceiling; the endpoint
 	// silently clamps larger requests instead of rejecting them.
 	MaxCountForRoutesForLocation = 50
--- a/internal/restapi/routes_for_location_handler.go
+++ b/internal/restapi/routes_for_location_handler.go
@@ -18,7 +18,7 @@ func (api *RestAPI) routesForLocationHandler(w http.ResponseWriter, r *http.Requ
 
 	var fieldErrors map[string][]string
 	loc, fieldErrors := api.parseLocationParams(r, fieldErrors)
-	maxCount, fieldErrors := utils.ParseMaxCountClamped(queryParams, models.DefaultMaxCountForRoutes, fieldErrors)
+	maxCount, fieldErrors := utils.ParseMaxCountClamped(queryParams, models.DefaultMaxCountForRoutesForLocation, fieldErrors)
 
 	if len(fieldErrors) > 0 {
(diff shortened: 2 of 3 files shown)


Ahmedhossamdev (commented):
LGTM!

aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is exactly the right shape for the fix. A few things I checked and they
hold up:

- `DefaultMaxCountForRoutes` had a single call site, so retiring the shared
  constant in favour of an endpoint-specific one doesn't leave anything
  dangling — and pairing `DefaultMaxCountForRoutesForLocation` right next to
  `MaxCountForRoutesForLocation` makes the default/ceiling relationship legible
  at the definition instead of at the call site.
- The clamp path is untouched: `min(10, 50)` is still 10, and explicit large
  values still clamp the way #1291 set up. That PR deliberately deferred this
  default, so this is the clean follow-on rather than a re-litigation.
- The new test seeds 15 in-bounds routes, which is the detail that makes it
  worth having — with only the 13 RABA fixture routes the assertion couldn't
  tell 10 from 50. Good instinct.
- Pinning `maxCount=50` on the span-sizing case with a comment explaining why
  is the right call; that case is about box geometry and shouldn't quietly
  become a truncation test.

Merging.

PR #1358, opened 2026-08-17
Honor includeReferences on routes-for-location

### LINKED ISSUE #1356: routes-for-location: includeReferences=false is ignored
### Current behavior

`routesForLocationHandler` always builds `references.agencies` (via `GetAgenciesByIDs`) and
`references.situations` (via `collectAlertsForRoutes` / `BuildSituationReferences`),
`internal/restapi/routes_for_location_handler.go:71-83`. The `includeReferences` query
parameter is never read, so `includeReferences=false` still returns a fully populated
references block.

### Expected behavior (per spec)

Java honours `includeReferences` globally for every V2 endpoint:
`ApiActionSupport` holds `_includeReferences` and passes it to `new BeanFactoryV2(...)`.
When false, the references block is present but empty.

Same defect class as #1187 (routes-for-agency), #1211 (agencies-with-coverage),
#1280 (trip-for-vehicle) and #1349 (stops-for-location).

### What to change

1. Read `ShouldIncludeReferences(r)` (`internal/restapi/reference_utils.go:395`) in the handler
   and skip both the agency lookup and the situation build when false, leaving the empty
   references from `models.NewEmptyReferences()`. Follow `routes_for_agency_handler.go:50-56`.
2. `data.list` is unchanged either way — only `data.references` empties.
3. Tests: `includeReferences=false` empties `references.agencies` and `references.situations`
   while the list is unchanged; `includeReferences=true`, absent, and unparseable values all
   populate references.

ARCoder181105:
Closes #1356

`routes-for-location` ignored `includeReferences`, always returning populated
`references.agencies` and `references.situations`. Java applies the flag to every V2 endpoint via
`BeanFactoryV2`.

- Reference building gated on `ShouldIncludeReferences(r)`, following `routes_for_agency_handler.go`.
- `data.list` is unchanged in every case; only `data.references` empties.
- Table-driven test covering `false`, `true`, absent and unparseable values.

Stacked on #1357 (maxCount default) — this branch is built on top of it, so the diff will
shrink to just this change once #1357 merges.

### DIFF
--- a/internal/models/constants.go
+++ b/internal/models/constants.go
@@ -24,9 +24,11 @@ const (
 )
 
 const (
-	DefaultMaxCountForRoutes = 50
-	DefaultMaxCountForStops  = 100
-	MaxAllowedCount          = 250
+	DefaultMaxCountForStops = 100
+	MaxAllowedCount         = 250
+	// DefaultMaxCountForRoutesForLocation is the routes-for-location default when the
+	// caller omits maxCount; MaxCountForRoutesForLocation is the ceiling it is clamped to.
+	DefaultMaxCountForRoutesForLocation = 10
 	// MaxCountForRoutesForLocation is the routes-for-location ceiling; the endpoint
 	// silently clamps larger requests instead of rejecting them.
 	MaxCountForRoutesForLocation = 50
--- a/internal/restapi/routes_for_location_handler.go
+++ b/internal/restapi/routes_for_location_handler.go
@@ -69,18 +69,20 @@ func (api *RestAPI) routesForLocationHandler(w http.ResponseWriter, r *http.Requ
 	}
 
 	references := models.NewEmptyReferences()
+	// When includeReferences=false the references block is present but empty.
+	if ShouldIncludeReferences(r) {
+		agencyIDList := slices.Collect(maps.Keys(agencyIDs))
+		agencies, err := api.GtfsManager.GtfsDB.Queries.GetAgenciesByIDs(ctx, agencyIDList)
+		if err != nil {
+			api.serverErrorResponse(w, r, err)
+			return
+		}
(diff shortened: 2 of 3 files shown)


Ahmedhossamdev (approved):
Verified the change against a hosted OBA on the same feed: response list is byte-identical for false/true/absent/out-of-range; 
only references empties, matching Java's BeanFactoryV2. 

Lint, tests, gofmt, check-openapi all pass. No blockers. 

One note: includeReferences=notabool populates refs in maglev but not Java (shared helper quirk, pre-existing). LGTM.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Good change, and it lands in the right place. What I checked:

- It reuses `ShouldIncludeReferences` rather than re-parsing the param, and
  `models.NewEmptyReferences()` for the suppressed case, so the block stays
  present-but-empty instead of going null. That matches what
  `routes_for_agency_handler.go` already does, which is what I want — this
  semantic should look identical across handlers.
- Asserting `data.list` is unchanged in every table row is the assertion that
  actually earns its keep here; it's the thing that would catch a future
  refactor accidentally coupling the gate to result construction.
- The `len(routes) == 0` early return already emitted empty references, so
  leaving it alone is correct.

One non-blocking note for whenever you're next in this file:
`assert.Empty(t, model.Data.References.Situations)` is vacuous as written —
`createTestApi(t)` loads no GTFS-RT alerts, so situations are empty whether or
not the guard exists, and the `wantReferences == true` rows don't assert on
situations at all. So the situations half of the guard isn't really covered.
#1351 does this properly by seeding an alert fixture. Not worth another round
trip on this PR, but worth copying that pattern next time.

Merging after #1357, which this is stacked on.

PR #1360, opened 2026-08-17
Fix/stop routes missing from reference routes in trips-for-route-handler

### LINKED ISSUE #1339: trips-for-route: routes named by stop references are not in references.routes
Depends on #1314 — the fix reads the route IDs that PR's `stopReferences` already returns.

**Current**: `references.Routes` comes from `sets.routes`, built from the returned trips. Each stop reference separately carries `routeIds`/`staticRouteIds` from `GetRouteIDsForStops` — every route serving that stop, which can include routes no returned trip runs on (`trips_for_route_handler.go:748`).

Those route IDs resolve to nothing, and their agencies can be absent from `references.agencies` too.

**Spec**: every ID an entry or reference emits must resolve in the references block.

**Fix**: `stopReferences` already returns the route IDs by stop as its second value, discarded at line 748. Register them in `sets.routes` before `fillRoutesAndAgencies` runs, so routes and agencies are both filled in. `trips-for-location`'s `buildStopList` follows this pattern.

**Test**: assert every `routeId` on a stop reference resolves in `references.routes`, and that its agency resolves in `references.agencies`.

JohnAkindipe:
# Fix dangling stop route references in trips-for-route
Closes #1339

## Problem
In the `trips-for-route` response, `references.routes` was built exclusively from
the routes of the returned trips: `buildTripReferences` seeded `sets.routes` from
`trip.RouteID` (`collectPreFetchedTrips`) and then resolved routes and agencies in
`fillRoutesAndAgencies`. A stop served by a route that **no returned trip runs on**
 emits a `routeId` that resolves to nothing in `references.routes`.

## Fix

`internal/restapi/trips_for_route_handler.go` — `buildTripReferences`:

1. Call `api.stopReferences` **before** `fillRoutesAndAgencies` and keep the returned
   route-IDs-by-stop map.
2. Register those route IDs into `sets.routes` before route/agency resolution runs.
3. `fillRoutesAndAgencies` then resolves both the full route objects **and** their
   agencies (`addAgencyReference`), so routes and agencies are filled in by the one
   existing mechanism.

## Testing

Added a regression test `TestTripsForRouteHandler_StopRoutesResolveInReferences` with a 
synthetic fixture, `orphanStopRouteFiles()`. The test asserts that routes serving stops which
are not associated with any returned trips are correctly resolved in `references.routes`
alongside the route agencies in `references.agencies`. It fails in the pre-fix code and passes
post-fix

## Summary by CodeRabbit

* **Bug Fixes**
  * Route and agency references are now included when they are associated with returned stops, even if no active trip is returned for those routes.
  * Results continue to include only active trips for the specifically queried route.

### DIFF
--- a/internal/restapi/trips_for_route_handler.go
+++ b/internal/restapi/trips_for_route_handler.go
@@ -737,15 +737,31 @@ type tripReferenceParams struct {
 
 func (api *RestAPI) buildTripReferences(ctx context.Context, params tripReferenceParams) models.ReferencesModel {
 	sets := newTripReferenceSets()
+
 	sets.collectPreFetchedTrips(params.PreFetchedTrips)
 	sets.collectTripIDsFromEntries(params.Trips)
 	api.fillMissingTrips(ctx, sets)
-	api.fillRoutesAndAgencies(ctx, sets)
 
 	references := models.NewEmptyReferences()
+	var routeIDsByStopID map[string][]string
+	references.Stops, routeIDsByStopID = api.stopReferences(ctx, params.Stops, params.StopIDMap)
+
+	for _, combinedRouteIDs := range routeIDsByStopID {
+		for _, combinedID := range combinedRouteIDs {
+			rawID, err := utils.ExtractCodeID(combinedID)
+			if err != nil {
+				continue
+			}
+			if _, exists := sets.routes[rawID]; !exists {
+				sets.routes[rawID] = models.Route{}
+			}
+		}
+	}
+
+	api.fillRoutesAndAgencies(ctx, sets)
+
 	references.Agencies = utils.MapValues(sets.agencies)
 	references.Routes = sets.routeList()
-	references.Stops, _ = api.stopReferences(ctx, params.Stops, params.StopIDMap)
 	references.Trips = sets.tripReferenceList(params.IncludeTrip)
 	references.Situations = api.situationReferences(params.Situations)
 	return *references
@@ -780,7 +796,7 @@ func (api *RestAPI) tripSituationRefs(
 // to, keyed by bare ID so each is emitted once.
 type tripReferenceSets struct {
 	trips    map[string]models.Trip
-	routes   map[string]models.Route
+	routes   map[string]models.Route // maps raw route ids to models.Route
 	agencies map[string]models.AgencyReference
 	// missing holds the trips the response refers to — entry tripIds,
 	// schedule.nextTripId/previousTripId, status.activeTripId — whose full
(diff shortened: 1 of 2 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. The stop-route registration loop is inlined into `buildTripReferences` and is a third copy of logic that already exists as `referenceBuilder.buildStopList` in `trips_for_location_handler.go:665-679` (same `stopReferences` → `ExtractCodeID` → seed the route set shape). `buildTripReferences` otherwise reads as a sequence of named stages, and commit c00a305a ("Build every stop reference through one helper") folded the previous trips-for-location/trips-for-route divergence of this exact pair of functions into `reference_utils.go` precisely because the copies drifted. Worth extracting a shared helper (e.g. in `reference_utils.go`), or at minimum a named `sets.collectRoutesFromStops(routeIDsByStopID)` method alongside the existing `collectPreFetchedTrips`/`collectTripIDsFromEntries`. CONTRIBUTING.md says "Before writing new logic, check whether it already exists — this is the single most common category of review comment" and "Wrap related code in functions. If a block performs a distinct sub-task, extract it into its own well-named function rather than inlining it." Note the inline copy also drops the explanatory comment the trips-for-location copy carries about why the raw rather than combined ID is registered.

https://github.com/OneBusAway/maglev/blob/e026efb95dfb1b0bb934e391b18774c471b710e8/internal/restapi/trips_for_route_handler.go#L748-L760

aaronbrethorst (approved):
Good fix, and the test fixture is the part I want to call out.

`orphanStopRouteFiles` builds a stop served by a route from a *second* agency
that no returned trip runs on, on a different block, at a different time — which
is precisely the case where the old code emitted a `routeIds` entry with nothing
behind it. Constructing that scenario deliberately is more work than asserting
against the RABA fixture and worth much more, because it fails pre-fix for the
right reason rather than by coincidence. The `assert.Contains(stop.RouteIDs,
combinedOrphanRouteID)` self-guard on the fixture is a nice touch too — it means
the test can't silently stop testing anything if the fixture drifts.

Correctness checks that hold up:

- Moving `stopReferences` above `fillRoutesAndAgencies` and seeding `sets.routes`
  from the stops' route IDs is the right ordering — the routes have to be known
  before the batch fetch runs, not after.
- No N+1 introduced: the new loop is pure in-memory over the map
  `stopReferences` already returns, and route resolution still goes through the
  single batched `GetRoutesByIDs`.
- Agencies come along via `addAgencyReference`, so the second agency resolves too
  — which is the half of #1339 that's easy to miss.

One follow-up, not blocking: the stop-route registration loop is close to a
verbatim copy of `buildStopList` in `trips_for_location_handler.go`, and per
CONTRIBUTING.md's Code Reuse section that belongs in `reference_utils.go`
alongside `stopReferences` — commit c00a305a consolidated the sibling function
there for the same reason. I'd rather land the correctness fix now and do that
consolidation once, covering this and #1353 together, than block a 19-line bug
fix on it. If you want to take that follow-up PR, I'd welcome it.

Merging.

PR #1362, opened 2026-08-19
Add GET /api/where/metrics.json endpoint

### LINKED ISSUE #1363: Add `GET /api/where/metrics.json` to Maglev, matching OneBusAway Java's undocumented endpoint
## Summary

Add `GET /api/where/metrics.json` to Maglev, matching OneBusAway Java's undocumented `metrics.json` endpoint field-for-field.

This gives existing watchdog and monitoring tooling the same response shape it expects from Java.

Full technical spec: **[[Metrics Endpoint — Feature Spec](https://github.com/OneBusAway/maglev/wiki/Metrics%E2%80%90Endpoint)]** 

## Motivation

- Provide visibility into GTFS-RT feed health and schedule matching and other upcoming GTFS_STATIC data metrics
- Allow existing Java-based watchdog/alerting tooling to work with Maglev without changes.

## Response Shape

```json
{
  "agenciesWithCoverageCount": 2,
  "agencyIDs": ["1", "2"],
  "scheduledTripsCount": { "1": 69 },
  "realtimeRecordsTotal": { "1": 69 },
  "realtimeTripCountsMatched": { "1": 57 },
  "realtimeTripCountsUnmatched": { "1": 0 },
  "realtimeTripIDsUnmatched": { "1": [] },
  "stopIDsMatchedCount": { "1": 1701 },
  "stopIDsUnmatchedCount": { "1": 0 },
  "stopIDsUnmatched": { "1": [] },
  "timeSinceLastRealtimeUpdate": { "1": 1 }
}

### PR DESCRIPTION
**Pull request by Ahmedhossamdev**
Fixes: #1363

Adds `GET /api/where/metrics.json` to Maglev, matching OneBusAway Java's undocumented but widely used monitoring endpoint field-for-field. This ensures existing watchdog and alerting tooling built around the Java response shape works unmodified against a Maglev instance.

Per agency, the endpoint reports:
- Currently active block count
- GTFS-RT record/trip/stop matching health
- Feed staleness

Full technical write-up: Java source cross-references for every field, matching mechanics (block grouping and activity-window gating), naming concerns, and proposed follow-up metrics:

[Metrics Endpoint — Feature Spec](https://github.com/OneBusAway/maglev/wiki/Metrics%E2%80%90Endpoint)

## Changes

The implementation is split into three layered, independently reviewable commits:

1. **`gtfsdb`** — Adds `GetActiveTripBlockIDsForAgency` and `GetActiveLayoverBlockIDsForAgency` to count blocks that are active right now, either mid-trip or laying over. This matches Java's `BlockStatusServiceImpl#getActiveBlocksForAgency`.

2. **`internal/gtfs`** — Adds `Manager.GetMetrics` to match real-time records, trips, and stops against the static schedule. Matching is grouped by static block ID and gated by an activity window, matching Java's `GtfsRealtimeSource` / `GtfsRealtimeTripLibrary` grouping and `isTripActive` semantics.

3. **`internal/restapi`** — Adds the thin handler, model, and route registration, following the existing `config.json` / `current-time.json` pattern.

## Verification

Every field was cross-checked against a live production Java `metrics.json` response for the same upstream GTFS-RT feed during development, rather than relying only on isolated unit tests.

| Field | Maglev | Java |
|---|---:|---:|
| `scheduledTripsCount` | 69 | 70 |
| `realtimeRecordsTotal` | 69 | 70 |
| `realtimeTripCountsMatched` | 57 | 57 |
| `stopIDsMatchedCount` | 1701 | 1700 |

The remaining ±1 differences are expected live-feed polling/timing skew rather than implementation discrepancies. See the feature spec for the full before/after story, including the two rounds of debugging that brought these numbers this close.

## Known, Deliberate Deviations from Java

- **`timeSinceLastRealtimeUpdate`** returns `0` for an agency with no covering feed, instead of Java's raw current-epoch-seconds. This is a genuine Java bug: `MonitoredResult._lastUpdate` defaults to `0`, resulting in `(now - 0) / 1000`. This was confirmed by decoding a live production sample and isn't worth reproducing.

- **`realtimeTripCountsMatched` / `realtimeTripCountsUnmatched`** use static-ID resolution plus the same activity-window gate as Java, rather than reproducing Java's full schedule-deviation block-matching engine (`GtfsRealtimeTripLibrary#applyTripUpdatesToRecord`). This reproduces Java's counts exactly on the live data tested while avoiding disproportionate complexity for a monitoring endpoint. The full rationale is documented in the feature spec.

## Spec / Wiki Gap

`metrics.json` currently has no entry in `testdata/openapi.yml` or the upstream `maglev.wiki` either. I'm flagging this in line with `CONTRIBUTING.md` rather than blocking the implementation on it.

The feature spec above is the first step toward properly documenting the endpoint. I'll add the formal API documentation once we've agreed on the fields, naming, and behavior.

## Test Plan

- [x] Manual verification against live production Java `metrics.json` for the same upstream feed (see Verification above)

## Summary by CodeRabbit

* **New Features**
  * Added a metrics endpoint at `/api/where/metrics.json`.
  * Reports scheduled trip coverage, real-time matching, unmatched trips and stops, agency data, and feed update freshness.
  * Supports static schedule data alone or combined with real-time feeds.
  * Includes active trips and layover blocks, with deduplicated results.
  * Applies standard API-key validation and rate limiting.

* **Tests**
  * Added comprehensive coverage for metrics calculations, agency attribution, feed freshness, deduplication, and API responses.

### CODE REVIEW

**Comment by aaronbrethorst** [comment]
### Code review

Found 4 issues:

1. `timeSinceLastRealtimeUpdate` reports `0` (i.e. "just updated") for feeds that are actually dead, which inverts the signal this endpoint exists to provide. `snapshotRealtimeFeedState` enumerates feeds by iterating `manager.feedTrips`, and staleness is only recorded when `feedLastUpdate` has an entry; every agency without a recorded entry is then backfilled to `0`. Two concrete cases: (a) after `staleFeedThreshold` (5 min) of consecutive failures, `clearFeedData` does `delete(manager.feedLastUpdate, feedID)` (`internal/gtfs/gtfs_manager.go:97`), so staleness climbs during the outage and then snaps back to `0` exactly when the outage becomes severe; (b) a vehicle-positions-only feed never gets a `feedTrips` key at all (`internal/gtfs/realtime.go:317-318` only assigns it when `tripData != nil`), so it is invisible here and its agencies report `0` forever. The PR body's "deliberate deviation" note covers agencies with *no* covering feed, not these.

https://github.com/OneBusAway/maglev/blob/fc3dc7e6d61ba67a7da96519fc13436de32dee9f/internal/gtfs/metrics.go#L256-L261

2. `/api/where/metrics.json` does not exist in `testdata/openapi.yml` (confirmed: zero matches for "metrics" in the spec on `main`). CLAUDE.md says of that spec: "All API endpoints MUST behave identically to what is defined in this OpenAPI spec. This is the single source of truth for request parameters, response schemas, field names, types, and status codes." With no entry, the response shape here is unverifiable and `openapi_conformance_test.go` cannot cover it. The PR body flags this, which is the right thing to do per CONTRIBUTING.md — noting it explicitly so it's a conscious maintainer decision (and, since `make check-openapi` pins this file to upstream `sdk-config`, the spec entry has to land upstream, not in this repo).

https://github.com/OneBusAway/maglev/blob/fc3dc7e6d61ba67a7da96519fc13436de32dee9f/internal/restapi/routes.go#L88-L90

3. Dead struct fields: `feedMetrics.tripsUnmatched` and `feedMetrics.stopsUnmatched` are assigned in `computeFeedMetrics` (L334, L337) but never read anywhere — `applyFeedMetrics` only consumes `recordsTotal`, `tripsMatched`, and `stopsMatched`, and the unmatched counts are recomputed from the cross-feed dedup sets in `populateRealtimeMetrics`. CONTRIBUTING.md calls out "leftover dead code" as a common review finding; Go won't flag unused struct fields.

https://github.com/OneBusAway/maglev/blob/fc3dc7e6d61ba67a7da96519fc13436de32dee9f/internal/gtfs/metrics.go#L287-L293

4. PR size: +1494 lines across 10 files, against CONTRIBUTING.md's "Keep PRs as short as possible, ideally no more than 200 lines." Even excluding the sqlc-generated `gtfsdb/` churn and the 536-line test file, `internal/gtfs/metrics.go` alone is 627 lines of new logic in a single commit's worth of surface area. The three layers are genuinely coupled (the new queries are dead without their caller), so a clean split is not obvious — but this is worth an explicit call rather than a silent exception.

https://github.com/OneBusAway/maglev/blob/fc3dc7e6d61ba67a7da96519fc13436de32dee9f/internal/gtfs/metrics.go#L26-L28

**Review by aaronbrethorst — CHANGES_REQUESTED** [review]
Thanks for taking this on — it's a genuinely useful endpoint and there's a lot
of careful work here. Before the blockers, the things I checked that are right:

- The route goes through `rateLimitAndValidateAPIKey`.
- `snapshotRealtimeFeedState` takes only `realTimeMutex.RLock()` with an
  immediate `defer`, and releases before any DB work — so the documented
  `staticMutex → realTimeMutex` ordering is respected and there's no lock held
  across a query. That's the thing I most expected to go wrong in a PR that
  reads realtime state, and it doesn't.
- `r.Context()` is propagated into `GetMetrics` and every DB call.
- The response exposes agency IDs, counts and unmatched trip/stop IDs — no API
  keys, feed URLs, or config headers. Good.

**Blocker 1 — `timeSinceLastRealtimeUpdate` reports 0 ("fresh") for dead feeds.**

This is the one that has to change, because it inverts the endpoint's whole
purpose. Tracing it:

`snapshotRealtimeFeedState` enumerates feeds by iterating `manager.feedTrips`
and reads `lastUpdate, hasUpdate := manager.feedLastUpdate[feedID]`. Staleness
is only computed `if feed.hasUpdate`, and then the backfill loop sets
`TimeSinceLastRealtimeUpdate[agencyID] = 0` for anything untracked
(internal/gtfs/metrics.go, around the `if _, tracked := ...; !tracked` block).

Meanwhile `clearFeedData` (internal/gtfs/gtfs_manager.go:97) does
`delete(manager.feedLastUpdate, feedID)`, and the polling loop calls it once
`time.Since(lastSuccessfulFetch) > staleFeedThreshold` (internal/gtfs/realtime.go:802).

So the observable behavior is: staleness climbs during an outage, and then at
the exact moment the circuit breaker declares the feed dead, it snaps to 0. A
watchdog polling this endpoint sees the metric go healthy precisely when the
outage gets bad enough to trip protection. That's worse than not reporting it.

Related, same root cause: `manager.feedTrips[feedID]` is only assigned when
`tripData != nil && tripErr == nil` (internal/gtfs/realtime.go:317), so a feed
configured with only `vehicle-positions-url` never gets a `feedTrips` key and is
invisible to this endpoint entirely.

The fix is to drive feed enumeration off the configured feed list rather than
`feedTrips`, and to distinguish "never updated" and "cleared as stale" from
"updated 0 seconds ago" — a null, a `-1`, or an explicit `feedHealthy` boolean,
whichever you prefer, but not a 0 that reads as fresh.

**Blocker 2 — the endpoint isn't in the OpenAPI spec, and this one's on me.**

`testdata/openapi.yml` has zero occurrences of `metrics`, and CLAUDE.md says
that spec is the single source of truth for all endpoints. But `make
check-openapi` pins that file to upstream `sdk-config`, so you can't fix this in
this PR even if you wanted to — it has to land upstream first. I'm flagging it
so it's a deliberate decision rather than a silent gap; I'll take the upstream
change. Don't block on it, but please note the divergence in the PR description
per CONTRIBUTING.md's spec-discrepancy guidance.

**Blocker 3 — dead fields.** `feedMetrics.tripsUnmatched` and
`.stopsUnmatched` (metrics.go:289 and :292) are assigned at :334 and :337 and
never read anywhere. Go won't flag unused struct fields, so these would just sit
there. Please delete them, or wire them up if they were meant to be used.

**On size:** +1494 lines against CONTRIBUTING.md's ~200-line guidance. I'm not
going to insist you re-cut work that's already written, but this would have been
much easier to review — and would have gotten you feedback sooner — as three
PRs: the sqlc queries, the `internal/gtfs/metrics.go` computation with its unit
tests, and then the thin handler/model/route on top. Worth considering for the
next feature of this size.

Non-blocking, but worth a look while you're in here:

- `applyFeedMetrics` takes the *minimum* staleness across feeds covering an
  agency, so one healthy feed masks a stale sibling. Maximum is the safer
  choice for a monitoring metric.
- `StopIDsMatchedCount` is summed with `+=` across feeds while
  `StopIDsUnmatchedCount` is deduplicated through `addToAgencySet` — so for a
  multi-feed agency, a stop seen by two feeds counts twice on one side and once
  on the other. Pick one and apply it consistently.
- The `MetricsSnapshot` doc comment (metrics.go:25) references
  `countCombinedRecords`; the function is actually `countMatchedGroups`.
- Matched-trip activity is gated on `time.Now()` rather than `api.Clock`, which
  is why `TestMetricsHandlerWithRealTimeData` can only assert matched == 0 — the
  test comment says as much. Threading the clock through would let that test
  assert something real, and this is the endpoint where you most want that.
- The new agency-scoped queries filter on `routes.agency_id` with no supporting
  index (`idx_block_layover_route_service_time` leads with `route_id`), so
  that's four scans per agency per request on an uncached endpoint.
- Yesterday's service window uses a fixed `+24h` shift, which is off by an hour
  across DST boundaries.

One housekeeping note: the CodeRabbit summary appended to the PR body describes
changes that aren't in this diff (routes-for-location defaults, includeReferences
gating) — it looks like a stale cross-branch summary. Worth clearing so the next
reader isn't misled.

Happy to re-review as soon as blockers 1 and 3 are addressed.

**Comment by Ahmedhossamdev** [comment]
> ## [Quality Gate Passed](https://sonarcloud.io/dashboard?id=OneBusAway_maglev&pullRequest=1362) **Quality Gate passed**
> 
> Issues  [1 New issue](https://sonarcloud.io/project/issues?id=OneBusAway_maglev&pullRequest=1362&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [0 Accepted issues](https://sonarcloud.io/project/issues?id=OneBusAway_maglev&pullRequest=1362&issueStatuses=ACCEPTED)
> 
> Measures  [0 Security Hotspots](https://sonarcloud.io/project/security_hotspots?id=OneBusAway_maglev&pullRequest=1362&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [0.0% Coverage on New Code](https://sonarcloud.io/component_measures?id=OneBusAway_maglev&pullRequest=1362&metric=new_coverage&view=list)  [0.0% Duplication on New Code](https://sonarcloud.io/component_measures?id=OneBusAway_maglev&pullRequest=1362&metric=new_duplicated_lines_density&view=list)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=OneBusAway_maglev&pullRequest=1362)

This one's pre-existing on main, not introduced by this PR - trip_for_vehicle_handler.go isn't touched by any commit here (checked with git log main..feat/metrics-endpoint -- internal/restapi/trip_for_vehicle_handler.go, no hits). Worth a separate cleanup PR, but out of scope for this one.

File: internal/gtfs/metrics.go [hunk 1]
```diff
@@ -0,0 +1,666 @@
...
+}
+
+func (manager *Manager) activeTripsForAgency(ctx context.Context, now time.Time, agency gtfsdb.Agency) (int, error) {
+	loc, err := time.LoadLocation(agency.Timezone)
+	if err != nil {
+		loc = time.UTC
+	}
+	localNow := now.In(loc)
+	midnight := time.Date(localNow.Year(), localNow.Month(), localNow.Day(), 0, 0, 0, 0, loc)
+	sinceMidnight := max(localNow.Sub(midnight), 0)
```
burma-shave, line 143: `sinceMidnight` is computed with `localNow.Sub(midnight)` (real elapsed duration) instead of wall-clock math, reproducing the DST pitfall `internal/utils.CalculateSecondsSinceServiceDate`'s doc comment explicitly warns about: using `time.Sub` diverges from GTFS by 3600s during the DST fall-back ambiguous hour, causing wrong schedule offsets. During that hour, `countActiveBlocksAt` gets an `at` offset by up to 3600s from true GTFS wall-clock time-of-day, over/under-counting active blocks in `scheduledTripsCount`.

internal/gtfs/metrics.go
```diff
@@ -0,0 +1,666 @@
...
+// applyFeedMetrics accumulates a feed's per-agency totals into snapshot.
+// Matched/unmatched trip and stop IDs are handled separately by
+// populateRealtimeMetrics (via addToAgencySet), since they need
+// deduplicating across feeds that cover the same agency rather than summed
+// directly.
+func applyFeedMetrics(snapshot *MetricsSnapshot, agencyID string, metrics feedMetrics, hasUpdate bool, staleness int64) {
+	snapshot.RealtimeRecordsTotal[agencyID] += metrics.recordsTotal
+	snapshot.RealtimeTripCountsMatched[agencyID] += metrics.tripsMatched
+
+	if hasUpdate {
```
burma-shave, line 630: This only updates the agency's staleness from feeds with `hasUpdate==true`, so a feed that has never successfully updated is silently excluded from the "most-stale feed wins" comparison. Example: agency A covered by feed-1 (healthy, staleness=5s) and feed-2 (`hasUpdate=false`, never updated) ends up with `TimeSinceLastRealtimeUpdate[A]=5`, hiding feed-2's total failure — contradicting the adjacent comment's stated intent that the worst case should win.

internal/gtfs/metrics.go
```diff
@@ -0,0 +1,666 @@
...
+	unmatchedStopIDsByAgency := make(map[string]map[string]bool, len(snapshot.AgencyIDs))
+	coveredAgencies := make(map[string]bool, len(snapshot.AgencyIDs))
+
+	for _, feed := range manager.snapshotRealtimeFeedState() {
+		metrics, err := manager.computeFeedMetrics(ctx, feed.trips, now)
+		if err != nil {
+			return err
+		}
+
+		agencyIDs := feed.agencyFilter
```
burma-shave, line 274: When a feed has no configured `agency-ids` and its trips don't resolve to any static route, `resolvedAgencyIDs` comes back empty, so this loop never runs and `coveredAgencies` is never set for the intended agency. `TimeSinceLastRealtimeUpdate` then falls back to 0 ("no covering feed") instead of -1 (unknown/broken) — the exact bug class this PR's "Fix stale feeds reporting as fresh" commit was meant to close, still open for unfiltered feeds whose trips no longer match the static schedule.

internal/gtfs/metrics.go
```diff
@@ -0,0 +1,666 @@
...
+	found := false
+	for _, trip := range group {
+		if len(trip.StopTimeUpdates) == 0 {
+			continue
+		}
+		prediction := firstPredictionTime(trip.StopTimeUpdates[0])
+		if prediction == nil {
+			continue
+		}
+		if !found || prediction.Before(bestPrediction) {
```
burma-shave, line 544: Picking the trip with the earliest first-stop prediction as "the active leg" of a block can select a stale/just-finished trip update over the genuinely active one: if a just-finished trip's update still has all-past predictions, its (past) prediction sorts earlier than the active next trip's (future) prediction, so `representativeTrip` picks the finished trip. `isTripActive` then returns false and the whole block is undercounted as not active even though it has an active leg in the same poll.

internal/gtfs/metrics.go
burma-shave, line 274: Feed metrics get attributed to every agency ID in `feed.agencyFilter` without checking it's a real agency from `ListAgencies`/`snapshot.AgencyIDs`. A stale or misspelled `agency-ids` entry (e.g. an agency removed from static GTFS without updating the RT config) creates an orphan key in the per-agency response maps that a client iterating the documented `agencyIDs` field will never see.

burma-shave (changes requested):
Automated review (medium effort). Five findings below, all in `internal/gtfs/metrics.go`. The strongest is the DST issue at line 143: `sinceMidnight` is computed with `time.Sub`, reproducing a pitfall `internal/utils.CalculateSecondsSinceServiceDate` was explicitly written to avoid. Several others trace gaps in the "unknown vs. stale vs. fresh" freshness logic that this PR's own "Fix stale feeds reporting as fresh" commit was meant to close but doesn't fully cover. (A sixth finding about the raw -1/0 freshness sentinel was dropped as a duplicate of the existing CodeRabbit comment on line 49.)

burma-shave:
@Ahmedhossamdev checking in on this older PR. It is still blocked by the earlier requested changes and the open CodeRabbit/OpenAPI follow-up. Could you please address those review items and let us know when it is ready for re-review?

Ahmedhossamdev:
> @Ahmedhossamdev checking in on this older PR. It is still blocked by the earlier requested changes and the open CodeRabbit/OpenAPI follow-up. Could you please address those review items and let us know when it is ready for re-review?

Sure

internal/gtfs/metrics.go
```diff
@@ -0,0 +1,671 @@
...
+// knows about. Feed IDs are collected from feedTrips, feedVehicles,
+// feedAlerts, and feedLastUpdate together, not feedTrips alone: a feed
+// configured with only a vehicle-positions-url never gets a feedTrips entry
+// (see the trip-updates guard in updateFeedRealtime), so relying on
+// feedTrips alone would make it invisible to metrics entirely.
+func (manager *Manager) snapshotRealtimeFeedState() []realtimeFeedState {
+	manager.realTimeMutex.RLock()
+	defer manager.realTimeMutex.RUnlock()
+
+	feedIDs := make(map[string]bool, len(manager.feedTrips))
```
burma-shave, line 224: `snapshotRealtimeFeedState` still does not enumerate configured feed IDs from `feedAgencyFilter` (or another config-derived feed list). In production `InitGTFSManager` pre-populates `feedAgencyFilter` for configured feeds with `agency-ids`, but if such a feed has never successfully fetched then it may have no `feedTrips`, `feedVehicles`, `feedAlerts`, or `feedLastUpdate` entry. In that case the feed is absent from this snapshot and the agency later falls back to `TimeSinceLastRealtimeUpdate == 0`, which reads as no covering feed/fresh rather than `realtimeUpdateUnknown`. Please include configured feed IDs in the snapshot enumeration and add a regression test for a configured, agency-filtered feed that has never fetched any data.

internal/gtfs/metrics.go
```diff
@@ -0,0 +1,671 @@
...
+	matchedStopIDsByAgency := make(map[string]map[string]bool, len(snapshot.AgencyIDs))
+	unmatchedStopIDsByAgency := make(map[string]map[string]bool, len(snapshot.AgencyIDs))
+
+	for _, feed := range manager.snapshotRealtimeFeedState() {
+		metrics, err := manager.computeFeedMetrics(ctx, feed.trips, now)
+		if err != nil {
+			return err
+		}
+
+		agencyIDs := feed.agencyFilter
```
burma-shave, line 283: For an unfiltered feed, attribution still depends entirely on resolving route IDs from the current realtime trips. If the feed has no configured `agency-ids` and its trip IDs no longer resolve to static routes, `metrics.resolvedAgencyIDs` is empty, the feed contributes to no agency, and every agency falls back to `TimeSinceLastRealtimeUpdate == 0`. That is the same misleading freshness signal for a configured-but-broken feed. Either define unfiltered configured feeds as covering the static agency set for freshness purposes, or otherwise add an explicit unknown/unattributable feed path so this does not report as `0`; please include a regression test for the unfiltered unresolved-trip case.

burma-shave (changes requested):
Thanks for the follow-up fixes. Most of the previous metrics comments are addressed, but I still see two freshness cases that can report an agency as having no covering realtime feed (`0`) when a configured feed is actually present but broken/unattributable. I’m excluding the CodeRabbit outside-diff items here as requested.

aaronbrethorst:
### Code review

Found 3 issues:

1. `scheduledTripsCount` undercounts compared with Java by about 25% on King County Metro. The code says a block "counts as active ... while laying over between two of its trips", but the layover half comes from `block_layover`. The importer only writes a row there when the previous trip's last stop ID equals the next trip's first stop ID (`gtfsdb/helpers.go` L1439, `if lastStopCurrent.Stop.Id != firstStopNext.Stop.Id { continue }`). Java's `BlockIndexFactoryServiceImpl#createLayoverIndices` adds a layover for every consecutive trip pair in a block, with no stop check. On KCM, Java and this PR disagree on the same request (`/api/where/metrics.json?key=test`, fetched about 20 ms apart, twice). Java returned `{"1":125,"40":7}`, then `{"1":123,"40":7}`. This PR returned `{"1":98,"40":5}`, then `{"1":94,"40":5}`. A SQL check on the same DB that treats a block as active from its first trip's start to its last trip's end gives exactly 125/7 and 123/7. So the missing blocks are the ones laying over at a stop whose ID differs from the one they arrived at. For example, block `8220649` at 24.66h has trips ending at 24.53h and starting at 25.02h, and has no `block_layover` row for that gap.

https://github.com/OneBusAway/maglev/blob/4630f0cba3e763575515cd738771a8ef06c4e11e/internal/gtfs/metrics.go#L185-L202

https://github.com/OneBusAway/maglev/blob/4630f0cba3e763575515cd738771a8ef06c4e11e/gtfsdb/helpers.go#L1435-L1442

https://github.com/OneBusAway/onebusaway-application-modules/blob/095bf1ac3aeb7009b9f78e7f1cb4c65bd38b988a/onebusaway-transit-data-federation/src/main/java/org/onebusaway/transit_data_federation/impl/blocks/BlockIndexFactoryServiceImpl.java#L306-L318

2. A feed with no `agency-ids` has its counts attributed only to agencies found through its *matched* trips. `classifyTrips` adds a route ID only when the trip ID resolves. When a feed's trip IDs stop resolving (the classic "static GTFS changed, the RT feed didn't" outage), `resolvedAgencyIDs` is empty. `realtimeRecordsTotal`, `realtimeTripCountsUnmatched` and `realtimeTripIDsUnmatched` then all report 0 for every agency, and `TestGetMetrics_UnresolvableFeedIsNotAttributed` locks that in. I confirmed it with a scratch test: one agency, an unfiltered feed with two trip updates whose IDs don't resolve (valid `route_id`) gives `records=0 matched=0 unmatched=0 ids=[]`. Java does not work this way. `GtfsRealtimeSource#start` sets an unconfigured source's `agencyIds` to every agency, and `getTripDescriptorAsBlockDescriptor` calls `addUnmatchedTripId`, so Java reports those IDs as unmatched. The wiki's Metrics-Endpoint page doesn't cover attribution, and this isn't in the PR's deviations table. The same attribution rule also explains the live KCM run: the default feed covers every agency, yet agencies `1` and `23` report `timeSinceLastRealtimeUpdate: 0`, meaning no covering feed.

https://github.com/OneBusAway/maglev/blob/4630f0cba3e763575515cd738771a8ef06c4e11e/internal/gtfs/metrics.go#L282-L307

https://github.com/OneBusAway/maglev/blob/4630f0cba3e763575515cd738771a8ef06c4e11e/internal/gtfs/metrics.go#L457-L465

https://github.com/OneBusAway/onebusaway-application-modules/blob/095bf1ac3aeb7009b9f78e7f1cb4c65bd38b988a/onebusaway-transit-data-federation/src/main/java/org/onebusaway/transit_data_federation/impl/realtime/gtfs_realtime/GtfsRealtimeSource.java#L518-L530

3. Commit history breaks CONTRIBUTING.md's review rules. CONTRIBUTING.md says: "Do not write commit messages that refer indirectly to PR comments — e.g. 'apply review comments,' 'address PR feedback,'". It also says: "Add new commits when addressing requested changes — don't rewrite already-reviewed history." `b998ceec` is titled "Address CodeRabbit findings on metrics endpoint" and `1f0c47bf` is titled "Address review comments on metrics docs and tests". The branch was force-pushed on 2026-09-08 after the 08-19 and 08-28 reviews, and the commits those reviews covered (`fc3dc7e6`, `24405ab6`) are gone. The latest subject, "fix: cleanup by resetting real-time data and removing test database files", also breaks the 50-character and capitalization rules.

https://github.com/OneBusAway/maglev/commit/b998ceec094317d2a803dd7e2481d7723dd05a88

https://github.com/OneBusAway/maglev/commit/1f0c47bf0c032a272732f619022a36ea2f30fa65

aaronbrethorst (changes requested):
Thanks for the persistence on this one. Every blocker from my 8/19 review and burma-shave's 8/28 and 9/10 reviews is addressed:
- dead feeds now report `-1`
- DST math is fixed
- never-updated feeds are covered
- unmatched-trip freshness is covered
- the representative trip is gone
- no orphan agency keys remain

I also diffed `metrics.json` against the Java server on the same King County Metro data: keys, nesting, envelope, units, and auth behavior all match. The helper move into `internal/utils/maps.go` doesn't change trips-for-route or trips-for-location output.

Three things still need to change.

**1. `scheduledTripsCount` undercounts compared with Java by about 25%.** Two live comparisons about 20 ms apart:

| | Java | This PR |
|---|---|---|
| Run 1 | `{"1":125,"40":7}` | `{"1":98,"40":5}` |
| Run 2 | `{"1":123,"40":7}` | `{"1":94,"40":5}` |

The layover half of the active-block count comes from `block_layover`. `buildBlockLayoverIndex` only writes a row when the previous trip ends at the same stop the next trip starts from (`if lastStopCurrent.Stop.Id != firstStopNext.Stop.Id { continue }`). Java's `BlockIndexFactoryServiceImpl#createLayoverIndices` counts every gap between consecutive trips in a block.

Counting a block as active from its first trip's start to its last trip's end, on the same database, gives exactly Java's 125/7 and 123/7. Please compute active blocks that way inside the metrics query, not from `block_layover`. That table is correct for the endpoints that use it, so leave it as is.

**2. A feed without `agency-ids` hides its unmatched trips.** Agencies are resolved only from trips that matched static data. So when a feed's trip IDs stop matching, which is exactly the outage this endpoint exists to catch, records and unmatched trip IDs read 0 for every agency. `TestGetMetrics_UnresolvableFeedIsNotAttributed` currently asserts that behavior.

Java's `GtfsRealtimeSource#start` assigns a feed with no configured agencies to all agencies and reports those IDs as unmatched. Please do the same, and flip that test.

**3. Commit messages.** `b998ceec` ("Address CodeRabbit findings…") and `1f0c47bf` ("Address review comments…") refer to review feedback, which CONTRIBUTING asks you to avoid. The latest subject is over 50 characters and lowercase. The repo merges with merge commits, so these land on main as-is. Please give them real what-and-why messages.

Worth fixing alongside, not blocking:
- With several `agency-ids`, every listed agency gets the feed's full matched count. Java splits by trip ID prefix.
- `MetricsSnapshot`'s doc comment still says matched counts are "typically higher here than in Java."
- Matched-trip activity still uses `time.Now()` instead of `api.Clock`.
- The `Metrics‐Endpoint.md` wiki page still describes `representativeTrip`; updating the spec is on our side.

Happy to re-review.

Ahmedhossamdev:
Rebased onto current main and addressed the three items from the 9/14 review.

1. `scheduledTripsCount` undercount: 8483cf75. A block is now active from its first trip's start to its last trip's end, in one query that replaces both the trip-window and `block_layover` queries. `block_layover` is untouched. I only have RABA locally, so please re-run the KCM comparison against Java.
2. Feeds without `agency-ids`: a4547b4c. Such a feed now covers every agency, like `GtfsRealtimeSource#start`, so its records and unmatched trip IDs are reported when its trips stop resolving. `TestGetMetrics_UnresolvableFeedIsNotAttributed` is flipped to `TestGetMetrics_UnfilteredFeedCoversAllAgencies`.
3. Commit messages: reworded b998ceec, 1f0c47bf, the QueryInBatchesReserving commit and the latest one. The rebase onto main needed a force push anyway.

From the non-blocking list:
- Matched counts now go to each matched trip's own agency, the equivalent of Java's prefix split (a4547b4c).
- Removed the "typically higher than Java" line from the `MetricsSnapshot` doc (a4547b4c).
- Trip activity now uses `api.Clock` (39ce3c63). Staleness stays on the wall clock since `feedLastUpdate` is stamped with `time.Now()`.

Rebase notes: main's refactor of the stop-reference code in trips-for-route and trips-for-location conflicted with the helper move. I kept main's version and pointed it at `utils.QueryInBatches`.

Ready for another look.

gtfsdb/query.sql
```diff
@@ -1010,6 +1010,36 @@ SELECT COUNT(*) FROM stops;
...
+-- block_id is optional in GTFS, so trips.id is substituted for trips with no
+-- block_id, matching them 1:1 to their own "block" rather than dropping them.
+-- Callers run this once for today's service (at the current time-of-day) and
+-- once for yesterday's service (the same instant shifted +24h) to catch
+-- after-midnight trips, mirroring the today/yesterday pattern in
+-- trips_for_route_handler.go.
+-- The slice param must come last so the other params keep stable ?N
+-- numbers once sqlc expands it (see GetActiveLayoverBlockIDsForRoute). `at`
+-- is only needed in HAVING, after the slice, so it is bound up front in FROM.
+SELECT COALESCE(trips.block_id, trips.id) AS block_id
```
burma-shave: This fallback does not work for real blockless GTFS trips. The importer stores an absent block ID via `nulls.String(t.BlockID)`, which produces a valid empty string, not SQL `NULL`; therefore `COALESCE(trips.block_id, trips.id)` groups all blockless trips for an agency under `''` instead of counting each trip as its own block. Please normalize empty strings here (for example with `NULLIF(trips.block_id, '')` in both the SELECT and GROUP BY) or otherwise ensure imported blockless trips are stored as true NULLs, and add coverage that exercises the imported empty-string case.

burma-shave (changes requested):
Read-only follow-up review of pre-existing PR comments found most earlier metrics-specific feedback addressed, but a few actionable issues remain. Blocking inline finding below: active block counting still mishandles imported blockless trips. Also still unaddressed from earlier CodeRabbit comments, but outside the current commentable diff lines: `buildTripsForLocationEntries` still calls `GetStopTimesForTripIDs(ctx, validVehicleTrips)` and `GetStopsByIDs(ctx, allStopIDs)` directly instead of batching, and `applyScheduledTripPositionToStatus` still discards errors from `fetchStopCoordsForStopTimes`.

gtfsdb/query.sql
Ahmedhossamdev: Fixed in e644047d. The query now uses `COALESCE(NULLIF(trips.block_id, ''), trips.id)` in both the SELECT and the GROUP BY. `TestGetMetrics_ScheduledTripsCountTreatsImportedBlocklessTripsSeparately` stores `BlockID` via `nulls.String("")` the way the importer does, with two active trips. It fails with 1 on the old query and passes with 2.

Ahmedhossamdev:
On the two outside-diff CodeRabbit items: both are pre-existing on main and unchanged by this PR, which only renames the batching helper in those files.

- `buildTripsForLocationEntries` calling `GetStopTimesForTripIDs(ctx, validVehicleTrips)` and `GetStopsByIDs(ctx, allStopIDs)` unbatched: same lines on main (trips_for_location_handler.go:806, :859).
- `stopByID, _ := api.fetchStopCoordsForStopTimes(...)`: same on main (scheduled_block_helper.go:312, :409).

I'll fix them in a separate PR to keep this one scoped to the metrics endpoint.

Ahmedhossamdev:
Rebased onto current main (e70bd6ba) to pick up the latest merges. No conflicts and no code changes. The last reviewed commit, e644047d, is now 80e09c13.

Ahmedhossamdev:
Opened #1476 to track the two pre-existing items (unbatched trips-for-location queries, discarded `fetchStopCoordsForStopTimes` errors).

internal/gtfs/metrics.go
```diff
@@ -0,0 +1,650 @@
...
+	for feedID := range manager.feedVehicles {
+		feedIDs[feedID] = true
+	}
+	for feedID := range manager.feedAlerts {
+		feedIDs[feedID] = true
+	}
+	for feedID := range manager.feedLastUpdate {
+		feedIDs[feedID] = true
+	}
+	for feedID := range manager.feedAgencyFilter {
```
burma-shave: This loop currently treats every configured `agency-ids` filter as an active covering feed, but `feedAgencyFilter` is populated from all `config.RTFeeds`, including disabled feeds and feeds with no realtime URLs. Those feeds will never set `feedLastUpdate`, so a disabled/broken configuration can permanently force `timeSinceLastRealtimeUpdate` to `-1` for an agency even when another enabled feed for that agency is updating normally. The snapshot should only include feeds that are actually enabled/started, or otherwise carry enough config state to distinguish disabled/non-started feeds from enabled feeds that have never updated.

internal/gtfs/metrics.go
```diff
@@ -0,0 +1,650 @@
...
+	agencyFilter map[string]bool
+	lastUpdate   time.Time
+	hasUpdate    bool
+}
+
+// snapshotRealtimeFeedState enumerates every feed the manager knows about.
+// The union of all five per-feed maps is needed: any one alone can miss a
+// vehicle-positions-only feed (no feedTrips entry) or a configured feed
+// that has never fetched (only feedAgencyFilter entry).
+func (manager *Manager) snapshotRealtimeFeedState() []realtimeFeedState {
```
burma-shave: A configured unfiltered realtime feed (`agency-ids` omitted) that never successfully fetches is not represented by `feedAgencyFilter`, `feedTrips`, `feedVehicles`, `feedAlerts`, or `feedLastUpdate`. As a result this snapshot never sees it, and the final backfill reports `0` (no covering realtime feed) for every agency instead of `-1` (covering feed exists but freshness is unknown/dead). This contradicts the documented sentinel behavior and the intended Java-like "unfiltered feed covers all agencies" semantics.

internal/utils/maps.go
```diff
@@ -9,3 +14,54 @@ func MapValues[K comparable, V any](m map[K]V) []V {
...
+func QueryInBatchesReserving[T any](ctx context.Context, ids []string, reserved int,
+	query func(context.Context, []string) ([]T, error)) ([]T, error) {
+	if len(ids) == 0 {
+		return []T{}, nil
+	}
+	// A reserved budget at or above IDsPerBatchedQuery leaves no headroom for
+	// even one batched ID under the SQLite bind limit. Silently forcing
+	// batchSize to 1 would still push each statement past the limit; refuse
+	// instead and force the caller to batch the reserved dimension too.
+	if reserved >= IDsPerBatchedQuery {
```
burma-shave: Returning an error whenever `reserved >= IDsPerBatchedQuery` regresses callers that previously worked with one ID per batch under the actual bundled SQLite bind-variable limit. For example, the metrics/trips helpers that reserve active service IDs can now return HTTP 500 for agencies/days with roughly 900 active service IDs, even though each statement with one batched ID plus those reserved IDs would be accepted by SQLite builds with the 32766 variable limit used here. Please avoid rejecting this case globally; either fall back safely to one-ID batches when it fits the runtime limit, or batch the reserved dimension at the affected callers.

burma-shave (changes requested):
Claude Code's read-only review found three actionable issues. The prior active-block/blockless-trip blocker appears addressed by the latest change, and the outside-diff trips-for-location/scheduled-position items are being tracked separately in #1476, but these newly observed issues need attention before approval.

PR #1373, opened 2026-08-20
Order stop search results by combined stop ID

### LINKED ISSUE #1369: search-stop: order results by combined stop ID
## Summary

`search/stop.json` orders results by the raw stop ID instead of the combined `{agencyId}_{stopId}` ID the spec requires. On multi-agency feeds this changes the order of `data.list` and, because `maxCount` truncates on that order, which stops are returned at all.

## Current Behavior

`searchStopsByName` (`gtfsdb/fts_queries.go`) ends with `ORDER BY s.id`.

With two matching stops — `aaa_stop` served by a route of agency `999`, `zzz_stop` served by a route of agency `111` — the response is:

```
[999_aaa_stop, 111_zzz_stop]
```

With `maxCount=1`, the response keeps `999_aaa_stop`.

## Expected Behavior

Per [wiki spec](https://github.com/OneBusAway/maglev/wiki/search-stop) step 3, stops in the matching bucket are ordered by combined stop ID, ascending lexicographic:

```
[111_zzz_stop, 999_aaa_stop]
```

With `maxCount=1`, the response keeps `111_zzz_stop`.

## Acceptance Criteria

- `data.list` is ordered by combined stop ID ascending. Single-agency feeds are unaffected.
- `maxCount` truncation and the `limitExceeded` flag are computed after that ordering.
- Test uses a multi-agency fixture where raw and combined ordering disagree, asserting both the full-list order and the `maxCount`-truncated result.

## Notes

This cannot be fixed by changing the `ORDER BY` alone. `stops` has no `agency_id` column; the handler resolves each stop's agency from its serving routes after the FTS query returns. The current `LIMIT maxCount + 1` therefore truncates before the sort key is known.

ARCoder181105:
Closes #1369.

`search/stop.json` ordered results by the raw stop ID. The spec orders by the combined `{agencyId}_{stopId}` ID, and the deployed Puget Sound server agrees — it returns `19_1085` ahead of `1_21680`. On a multi-agency feed the raw ordering is wrong, and since `maxCount` truncates on that order, a capped query returns a different set of stops entirely.

Stacked on #1372, which adds the precomputed index this sorts by. **That PR should merge first**; only the last commit here is new. GitHub can't base a PR on a branch in a fork, hence both targeting `main`.

Handler discrepancy worth flagging: a stop no route serves resolves no agency and so has no combined ID to sort by. Those now sort last instead of interleaved, so they no longer consume `maxCount` slots ahead of real results — `TestSearchStopsHandlerRouteTypeExclusion` changed to match. Legacy interleaves them because every stop gets a bundle-level agency at ingestion; Maglev has no equivalent, the same structural gap the wiki already records for parent stations.

Verified with `make test` and `go vet` under both build tag sets. On the King County Metro feed, a broad prefix query goes from 3.3s with the agency resolved inline to 0.004s against the index.

## Summary by CodeRabbit

* **New Features**
  * Stop searches now include agency information and consistently order results by combined agency and stop IDs.
  * Stops without an associated agency appear after agency-linked results.
  * Foreign-key enforcement is enabled across all database connections.

* **Bug Fixes**
  * Improved stop-to-agency indexing when schedule data is incomplete or unjoinable.
  * Prevented duplicate parent-station results and corrected result-limit behavior.

* **Reliability**
  * Database connections now close correctly when setup or indexing fails.


Ahmedhossamdev (commented):
Looks good - verified the ordering against `StopBeanIdComparator` in the Java
reference and it matches, and the `maxCount`/`limitExceeded` interaction is
covered well by the new tests. One small optional note below on the
routes-fallback path, not a blocker.

internal/restapi/search_stops_handler.go
```diff
@@ -253,6 +250,23 @@ func (api *RestAPI) searchStopsHandler(w http.ResponseWriter, r *http.Request) {
...
 
+// agencyIDForSearchedStop returns the agency whose prefix the stop's combined ID carries:
+// the lexicographically lowest agency serving it - a stable, if not numeric-minimal,
+// choice for multi-agency stops.
+//
+// The search query resolves this from the precomputed index, and reusing that value keeps
+// each stop's combined ID identical to the key the results were sorted by. A stop added
+// since the last import has no index entry yet, so fall back to its serving routes, which
+// GetRoutesForStops has already ordered by agency ID.
+func agencyIDForSearchedStop(stop gtfsdb.SearchStopsByNameRow, combinedRouteIDs []string) string {
```
Ahmedhossamdev: Optional: this fallback can resolve a real agency ID for a stop that the SQL
`ORDER BY` (fts_queries.go) treated as agency-less because it has no
`stop_agencies` row, which would put that stop out of combined-ID order in the
response. Looks unreachable today since `stop_agencies` is always rebuilt
atomically with the data it derives from, so not blocking, just flagging in
case incremental updates get added later and this stops being purely
theoretical.

ARCoder181105: Removed the fallback rather than documenting the caveat. GetRoutesForStops and BuildStopAgencies run the same stop_times -> trips -> routes join, so any stop reaching this line already has an index row when the index is current - the branch could only ever fire on a stale index, and then it would disagree with the sort key exactly as you describe. Reading the indexed agency directly makes that divergence structurally impossible instead of just unlikely.

Separately, stop_agencies was reworked upstream to hold one row per (stop, agency) pair instead of a single collapsed row, so the query in this thread has also changed to resolve the agency with MIN() in a scalar subquery rather than a join, to avoid fanning a multi-agency stop out into duplicate result rows.

gtfsdb/models.go
```diff
@@ -159,6 +159,11 @@ type Stop struct {
 	ParentStation      sql.NullString
 }
 
+type StopAgency struct {
+	StopID   string
+	AgencyID string
+}
```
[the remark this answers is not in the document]
ARCoder181105: `gtfsdb/models.go` is sqlc-generated (`DO NOT EDIT` header) and gets overwritten by `make models`. A comment here wouldn't survive the next regen, and none of the other generated types in this file have one either. The table itself is already documented at the `stop_agencies` `CREATE TABLE` in `schema.sql`. Not making this change.

burma-shave:
## OBA API Change Review: `search-stop`

This change fixes `GET /api/where/search/stop.json` so stop-search matches are ordered by the public combined stop ID (`{agencyId}_{stopId}`) before `maxCount` truncation, rather than by raw GTFS `stop_id`.

Previously, a multi-agency feed could return a different order — and, with `maxCount`, a different set of stops — because the SQL query truncated before the handler knew the agency prefix. The new query resolves the stop's agency from the precomputed stop-agency index, sorts by the same combined ID that the API returns, then applies the cap.

**Summary:** Issue #1369 appears fully closed. The main intentional side effect is that matches with no resolvable agency sort after routed/agency-resolved matches before the zero-route filter removes them; this is tested and now documented in the wiki.

<details>
<summary>Detailed review</summary>

### Goal check

- ✅ `data.list` is ordered by combined stop ID ascending.
- ✅ `maxCount` truncation happens after that ordering.
- ✅ `limitExceeded` remains based on the pre-filter capped result.
- ✅ Tests cover a multi-agency fixture where raw stop ID order disagrees with combined ID order.
- ✅ Tests cover the capped result and the agency-less/zero-route interaction.

Overall: **fully closed**.

### Client impact

No response fields or types change; this is an order/content behavior change for capped stop-search queries.

| Behaviour | Wayfinder/SDK | iOS | Android |
|---|---|---|---|
| Stop-search `data.list` order changes to combined-ID order; capped searches may return a different first N stops. | SDK callers of `searchForStop` may observe order/content changes. Wayfinder currently uses `stops-for-location`, so no Wayfinder impact found. | No direct `search/stop` method found. | No direct `search/stop` Retrofit method found; current search uses `stops-for-location`. |
| Agency-less/unserved FTS matches sort after served stops before filtering. | Observable only to SDK consumers of `searchForStop`; type-compatible. | No impact found. | No impact found. |

### Spec check

- **Combined-ID ordering:** consistent with the `search-stop` spec.
- **`maxCount` truncation after ordering:** consistent with the spec.
- **Agency-less matches sorting last:** intentional Maglev behavior. This has been documented in `search-stop.md` and the multi-agency deviation tracker.

Overall: **spec-consistent after the wiki update**.

### Verification

Targeted tests passed locally:

```text
go test -tags "sqlite_fts5 sqlite_math_functions" ./gtfsdb ./internal/restapi -run 'TestSearchStopsByName|TestSearchStopsHandlerOrdersByCombinedID|TestSearchStopsHandlerRouteTypeExclusion|TestSearchStopsHandlerParentStationReferences'
```

</details>

burma-shave (changes requested):
Code review findings. Note: these concern `gtfsdb/client.go` and `gtfsdb/schema.sql`, which aren't in this PR's diff directly (that code landed via #1372) — since #1373 is stacked on top of it, flagging here rather than opening a separate issue against `main`.

### 1. `gtfsdb/client.go:44` — `NewClient` leaks the opened `*sql.DB` when `backfillStopAgencyIndex` fails

It returns `nil, err` without closing `db`. If `backfillStopAgencyIndex` errors (e.g. disk full during the rebuild transaction, or a permission error on the DB file), the caller has no `*Client` to close, so the underlying `*sql.DB` (WAL files, connection pool) leaks. On a retry/supervisor-restart loop this exhausts file descriptors and can leave stale WAL locks blocking a later successful open. This is a new failure path — previously `NewClient` never returned an error after `db` was successfully opened.

### 2. `gtfsdb/client.go:60` — `PRAGMA foreign_keys = ON` is connection-scoped, not database-scoped, so pooled connections never get it

It's executed once via `db.ExecContext` during schema migration, but SQLite's `foreign_keys` pragma applies only to the connection it ran on. `createDB` opens the DB with plain `sql.Open(DriverName, config.DBPath)` (no `_foreign_keys=on` in the DSN), and `configureConnectionPool` sets `SetMaxOpenConns(25)` for file-backed DBs. Under concurrent load, `database/sql` opens additional connections that never ran the one-time pragma, so FK enforcement is OFF on those connections. If `recreateStopAgenciesTable`'s rebuild (or any other FK-dependent insert) lands on such a connection, a dangling `agency_id` reference silently succeeds instead of failing and rolling back the transaction — contradicting the doc comment ("a failure partway through never leaves stop_agencies dropped or half-populated") and the guarantee `TestBackfillStopAgencyIndex_RollsBackOnFailure` asserts. The test itself can't catch this because the test env forces `:memory:` with `SetMaxOpenConns(1)`, so it always reuses the one connection that ran the pragma.

### 3. `gtfsdb/client.go:66` — Backfill skip check can't detect a permanently-empty index

The skip check only verifies `stop_times` has rows (`served`), not that they actually join to `trips`/`routes`. If `stop_times` contains rows whose `trip_id` doesn't resolve through `trips`/`routes` (reachable via the FK-enforcement gap above, not just manual corruption), `BuildStopAgencies`'s join inserts 0 rows, so `indexed` stays 0. The skip condition `indexed > 0 || served == 0` never becomes true (`served` stays 1), so every subsequent `NewClient` call re-triggers the full "rebuilding stop agency index" transaction and table scan indefinitely instead of settling after one successful build.

### 4. `gtfsdb/schema.sql:172` — Comment references a nonexistent method `Client.rebuildLegacyStopAgenciesTable`

The actual implementation is `backfillStopAgencyIndex`/`recreateStopAgenciesTable` in `gtfsdb/client.go`. A maintainer grepping for `rebuildLegacyStopAgenciesTable` from this comment to trace the legacy-migration path will find nothing.

ARCoder181105:
Addressed the four findings on the `stop_agencies` startup path. Four commits, one per concern.

**Foreign keys weren't actually enforced.** `foreign_keys` is connection-scoped and
non-persistent, so the schema's `PRAGMA foreign_keys = ON` only applied to the one
connection that ran migrations — every other connection in the pool (up to 25 for
file-backed DBs) had it off. Checked with 5 concurrent connections on both drivers:

| DSN | cgo | purego |
|---|---|---|
| plain path (before) | `[0 0 0 0 0]` | `[0 0 0 0 0]` |
| with DSN param (after) | `[1 1 1 1 1]` | `[1 1 1 1 1]` |

Moved enforcement into the DSN. The two drivers need different syntax, so `DSN(path)`
lives in each build-tagged driver file next to its existing `DriverName`.

**`NewClient` leaked the `*sql.DB`** when the backfill failed — the caller gets no
`*Client`, so nothing could close the pool. `createDB` had the same defect on its own
two post-open failure paths; fixed all three.

**The backfill skip check couldn't settle.** It only asked whether `stop_times` had
rows, not whether they joined through to `routes`. A DB where nothing joins would
rebuild on every `NewClient` forever. Now checks the same join `BuildStopAgencies`
uses, via `EXISTS` so it stays an index seek.

**Fixed a stale comment** in `schema.sql` naming a method that was renamed before merge.

Risk checks on the FK change: RABA import still passes (import order is parent→child),
`clearAllGTFSDataWithQueries` deletes child→parent so the clear isn't rejected, and the
existing rollback test still passes. No latent FK violations surfaced.

New tests: FK enforcement across simultaneously-held pooled connections (the existing
`:memory:` + `MaxOpenConns(1)` setup can't catch this), and the backfill settling
instead of rebuilding. `go vet` and `make test` green under both build tag sets.

burma-shave (approved):

PR #1378, opened 2026-08-21
Fix swallowed duplicated trip lookup errors

### LINKED ISSUE #1366: trips-for-route: DUPLICATED trip lookup failures are swallowed
Depends on #1316 — `resolveDuplicatedBaseTrip` is introduced there, not on `main`.

**Current**: `resolveDuplicatedBaseTrip` (`trips_for_route_handler.go:977`) logs any non-`sql.ErrNoRows` error from `GetTrip` and returns a zeroed `gtfsdb.Trip`. Both lookups do this — the direct ID (`:982`) and the suffix-stripped one (`:993`). The caller at `:443` cannot tell a real DB failure from "no static counterpart", so the handler continues and answers 200 with a trip whose service date silently falls back to the query day.

**Spec**: per CONTRIBUTING, only `errors.Is(err, sql.ErrNoRows)` is a not-found path; every other DB error belongs in `serverErrorResponse` as a 500.

**Fix**: return the error alongside the trip. Continue the suffix fallback only for `ErrNoRows`; propagate anything else to the call site and route it through `serverErrorResponse`.

**Test**: cover both lookup-failure paths — the direct ID and the stripped ID — asserting a 500 rather than a partial 200.

priyanshu7739410:
## Description

Fixes #1366.

When resolving a `DUPLICATED` trip, `sql.ErrNoRows` is expected and allows the handler to fall back to the stripped trip ID. However, other database errors were being treated like a not-found result.

This change propagates unexpected database errors from both the direct and fallback trip lookups so the endpoint returns HTTP 500 instead of silently continuing.

### Tests

- Added regression coverage for direct lookup database errors.
- Added regression coverage for fallback lookup database errors.
- Added coverage confirming legitimate suffix-stripped fallback still succeeds.
- `go test -tags "purego" ./internal/restapi/...`
- `git diff --check`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of duplicated trips when resolving their base-trip information.
  - Database lookup failures now return a clear server error instead of silently continuing with incomplete results.
  - Duplicated trips remain available when resolution is unsuccessful, preserving their original identifiers.
  - Added support for successful fallback resolution when duplicated trip identifiers use numeric suffixes.
  - Improved response consistency for duplicated-trip lookup failures and fallback scenarios.


burma-shave (approved):

burma-shave (changes requested):
merge conflicts

priyanshu7739410:
Thanks, I am working on resolving the merge conflicts now. I will push the updated changes once they are resolved.

priyanshu7739410:
Merge conflicts have been resolved and all CI checks are passing. I also verified the changes locally.

burma-shave (approved):

PR #1384, opened 2026-08-21
Fix schedule-for-stop references for route-less stops and no-service dates

### LINKED ISSUE #1376: schedule-for-stop: queried stop references omitted when it has no routes
## Background

`schedule-for-stop` should include the queried stop and its agency in
`data.references` whenever `includeReferences` is enabled (the default).

## Finding

When a valid stop has no associated routes, the handler returns empty
references immediately:

- `data.entry.stopId` contains the queried stop ID.
- `data.entry.stopRouteSchedules` is empty, as expected.
- `data.references.stops` and `data.references.agencies` are also empty.

This happens because the no-routes path returns `NewEmptyReferences()` before
the normal reference-building logic runs.

For example, a request for an isolated stop returns:

```json
{
  "entry": {
    "stopId": "1_isolated_stop_no_routes",
    "stopRouteSchedules": []
  },
  "references": {
    "agencies": [],
    "routes": [],
    "stops": []
  }
}
```

That means the stop ID in `data.entry.stopId` has no corresponding full stop
record in `data.references.stops`, despite references being enabled.

## Specification

The [Minimal Guarantees](https://github.com/OneBusAway/maglev/wiki/schedule-for-stop#minimal-guarantees) require every entity ID in `data.entry` to have a full record in `data.references` when `includeReferences` is enabled. The [Success Guarantees](https://github.com/OneBusAway/maglev/wiki/schedule-for-stop#success-guarantees) specifically require the queried stop to appear in `data.references.stops`. [The `includeReferences` parameter](https://github.com/OneBusAway/maglev/wiki/schedule-for-stop#request-parameters) defaults to `true`; only an explicit `false` returns empty reference arrays.

## Expected behavior

When `includeReferences` is omitted or `true`, an isolated valid stop should
return:

- the queried stop in `references.stops`
- its agency in `references.agencies`
- an empty `references.routes` list

When `includeReferences=false`, all reference lists should remain empty.

## Ask

Update the no-routes path to build references when they are enabled, and add
coverage for a valid stop with no associated routes.

### LINKED ISSUE #1381: schedule-for-stop: route and agency references empty on dates with no active service
### Summary

`references.routes` and `references.agencies` are built from the queried date's schedule rows, so they come back empty whenever no trips are active on that date. Deployed OBA keeps them populated from the routes serving the stop.

### Current Behaviour

`buildScheduleForStopReferences` calls `collectRouteAndAgencyIDs(scheduleRows)` (`internal/restapi/schedule_for_stop_handler.go:243`), so no schedule rows means no route references.

```
GET /api/where/schedule-for-stop/1_1010.json?date=2030-01-01   (outside feed validity)

OBA:     stopRouteSchedules []   references: agencies 1, routes 2, stops 1
maglev:  stopRouteSchedules []   references: agencies 1, routes 0, stops 1
```

This affects every stop queried on a holiday, an out-of-window date, or with an expired feed — not just stops without routes (#1376).

### Spec Requirement

Verified against `api.pugetsound.onebusaway.org`: route and agency references are derived from the stop's route list, independent of whether any trip is active on the queried date. Agencies are derived from those routes, not from the queried stop, so a stop with no routes returns an empty `agencies` list.

### Acceptance Criteria

- `references.routes` is built from `routesForStop` (`GetRoutesForStop`) rather than from `scheduleRows`.
- `references.agencies` is derived from those routes; the `seedAgency` seeding in `fetchAgencyRefs` is removed.
- `includeReferences=false` still returns every reference list empty.
- Test covers a valid stop on a date with no active service, asserting routes and agencies are populated.

ARCoder181105:
Fixes #1376
Fixes #1381

## What changed

Two defects in the references block of `schedule-for-stop`, three lines apart in the same handler.

**Stops with no routes returned empty references.** The handler bailed out with an empty references block as soon as a stop had no routes, before the `includeReferences` check further down. So a valid stop answered with its own ID in `entry.stopId` and no matching record in `references.stops`. The early return was only an optimization — `GetScheduleForStopOnDate` expands an empty route ID slice to `IN (NULL)` and returns no rows — so deleting it lets the normal path handle the case.

**Route and agency references were built from the queried day's schedule rows.** Both lists emptied out whenever no service was active: a holiday, a date outside the feed's validity period, or an expired feed. They now come from the routes serving the stop. `GetRoutesForStop` already selects every column the route reference needs, so the `GetRoutesByIDs` round trip goes away, and agencies derive from those routes instead of being seeded with the queried stop's agency.

## Spec

[Success Guarantees](https://github.com/OneBusAway/maglev/wiki/schedule-for-stop#success-guarantees) require the queried stop to appear in `data.references.stops`. Verified against `api.pugetsound.onebusaway.org`:

```
GET schedule-for-stop/40_C03.json           (Westlake Link station, no routes)
  references: stops 1, routes 0, agencies 0

GET schedule-for-stop/1_1010.json?date=2030-01-01   (outside feed validity)
  stopRouteSchedules []   references: stops 1, routes 2, agencies 1
```

Note that agencies are route-derived, so a stop with no routes returns an empty `agencies` list — #1376 asked for the stop's own agency there, which the reference server does not do.

## Tests

`TestScheduleForStopHandlerReferencesWithoutSchedule` covers a route-less stop, the same stop with `includeReferences=false`, and a served stop on a date with no active service.

`TestScheduleForStopHandlerEmptyRoutes` was replaced: it claimed to cover "stop with no routes" but used `mustGetStops`, which only returns stops that have stop times, so it never entered the path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stop schedule responses when no routes or scheduled service are available.
  * Preserved relevant stop, route, and agency references even when schedules are empty.
  * Respecting reference inclusion settings continues to provide only the requested data.

### DIFF
--- a/internal/restapi/http_test.go
+++ b/internal/restapi/http_test.go
@@ -204,6 +204,18 @@ func mustGetStops(t testing.TB, api *RestAPI) []gtfsdb.Stop {
 	return stops
 }
 
+// mustGetStopWithoutRoutes returns the ID of a stop no trip calls at, which is
+// therefore served by no routes. The RABA fixture has 21 of them.
+func mustGetStopWithoutRoutes(t testing.TB, api *RestAPI) string {
+	t.Helper()
+	var stopID string
+	err := api.GtfsManager.GtfsDB.DB.QueryRowContext(context.Background(),
+		`SELECT id FROM stops WHERE id NOT IN (SELECT DISTINCT stop_id FROM stop_times) LIMIT 1`,
+	).Scan(&stopID)
+	require.NoError(t, err, "test data should contain at least one stop with no stop times")
+	return stopID
+}
+
 // mustGetStop return an active stop from the DB (stop with stop times)
 func mustGetStop(t testing.TB, api *RestAPI) gtfsdb.Stop {
 	t.Helper()
--- a/internal/restapi/schedule_for_stop_handler.go
+++ b/internal/restapi/schedule_for_stop_handler.go
@@ -95,15 +95,8 @@ func (api *RestAPI) scheduleForStopHandler(w http.ResponseWriter, r *http.Reques
 		routeIDs = append(routeIDs, rt.ID)
 	}
 
-	if len(routeIDs) == 0 {
-		api.sendResponse(w, r, models.NewEntryResponse(
-			models.NewScheduleForStopEntry(utils.FormCombinedID(agencyID, stopID), responseDate, nil),
-			*models.NewEmptyReferences(),
-			api.Clock,
-		))
-		return
-	}
-
+	// A stop with no routes falls through the normal path: GetScheduleForStopOnDate
+	// expands an empty route ID slice to IN (NULL) and returns no rows.
 	params := gtfsdb.GetScheduleForStopOnDateParams{
 		StopID:     stopID,
 		TargetDate: targetDate,
--- a/internal/restapi/schedule_for_stop_handler.go
+++ b/internal/restapi/schedule_for_stop_handler.go
@@ -211,7 +211,7 @@ func (api *RestAPI) scheduleForStopHandler(w http.ResponseWriter, r *http.Reques
 
 	references := models.NewEmptyReferences()
 	if ShouldIncludeReferences(r) {
-		references, err = api.buildScheduleForStopReferences(ctx, agencyID, agency, stop, scheduleRows, routeIDs)
+		references, err = api.buildScheduleForStopReferences(ctx, agencyID, stop, routesForStop, routeIDs)
 		if err != nil {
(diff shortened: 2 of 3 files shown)


ARCoder181105:
Good catch — applied in the latest commit.

The no-service case asserted only that `routes` and `agencies` came back non-empty, which would pass just as well if the handler referenced the wrong routes. It now fetches the stop's routes up front, requires them non-empty, and compares both reference lists against exact ID sets derived from them — route IDs combined with the agency prefix, agency IDs taken from the serving routes themselves rather than from the request namespace.

The two route-less cases assert against empty expected sets through the same comparison, so all three rows use one code path.

burma-shave (approved):

PR #1386, opened 2026-08-21
Report headway-based service in schedule-for-stop frequencies

ARCoder181105:
Addresses #1031 (the `exact_times=0` half; `exact_times=1` expansion follows in a separate PR).

Stacked on #1385 — its three commits show up in this diff until it merges. Review the last commit here.

## What changed

The handler passed `nil` for every direction group's frequencies, so `scheduleFrequencies` was always empty and headway-based trips appeared in `scheduleStopTimes` as if each ran once a day.

Each trip's frequency windows are now batch-loaded through the existing `GetFrequenciesForTrips` query, and a trip that has them is reported in `scheduleFrequencies` — one entry per window, sorted by start time — instead of `scheduleStopTimes`. A trip may declare several windows in a day, and a direction group may hold both kinds of service at once.

Two follow-on changes:

- **Headsign weighting.** A frequency trip's headsign now casts one vote per run it is expected to make in its windows, rather than one vote total, so a frequent route no longer loses the representative-headsign vote to a single scheduled trip.
- **One accumulator instead of three parallel maps.** The route/direction stop times and the headsign vote counts were carried in separate nested maps; frequencies would have added a third. They are now one `directionScheduleGroup` per direction.

## Spec

[Success Guarantees](https://github.com/OneBusAway/maglev/wiki/schedule-for-stop#success-guarantees): "Frequency-based service, where present in the feed, is represented separately from fixed-schedule stop times." Extension 5a specifies such trips contribute to `scheduleFrequencies` *instead of* `scheduleStopTimes`, and step 7 specifies the run-count weighting for headsigns.

No deployed OBA instance was available to verify against — neither the Puget Sound feed nor any other instance checked carries a `frequencies.txt` — so this follows the wiki's Java line references rather than observed behaviour.

`ScheduleFrequency` emits `stopHeadsign`, `arrivalEnabled` and `departureEnabled` as the wiki defines them. `testdata/openapi.yml` lists only the other six fields but does not forbid extras, so the conformance test passes on the superset.

## Tests

`TestScheduleForStopHandlerFrequencies` runs end to end against a feed with headway service in both directions plus a direction mixing headway and fixed trips. The RABA fixture has no `frequencies.txt`, so the test writes its own feed and loads it through a new `createTestApiWithFeed` helper — the shared fixture is untouched, and no existing test changes behaviour.

`TestGroupScheduleRowsByRouteAndDirection` gains a case for window ordering, the stop-time/frequency split, and the vote weighting.

## Summary by CodeRabbit

* **New Features**
  * Added support for displaying headway-based service frequencies in stop schedules.
  * Frequency entries include service windows, headway values, trip and service identifiers, and arrival/departure details.
  * Fixed-schedule and frequency-based trips can now appear together in the same direction, while maintaining separate data for opposing directions.

* **Bug Fixes**
  * Route references now use each route’s associated agency, improving route identification in multi-agency data.


Ahmedhossamdev (commented):
Hey @ARCoder181105 ,

The refactor is well-structured: replacing the three parallel maps with a single `directionScheduleGroup` accumulator makes the new frequency path easy to add and the existing headsign-vote logic easier to follow. The frequency spec steps are correctly implemented. References staying populated on no-service dates matches Java (live-verified on Unitrans). Two optional defensive notes inline; everything else looks good.

internal/restapi/schedule_for_stop_handler.go
```diff
@@ -46,12 +56,10 @@ func (api *RestAPI) scheduleForStopHandler(w http.ResponseWriter, r *http.Reques
 
 	if dateParam != "" {
 		var err error
+		// The format was validated above, so this only fails on an unusable agency timezone.
```
Ahmedhossamdev: Optional nit. The comment says "only fails on an unusable agency timezone," but
`loadAgencyLocation` (earlier in the handler) already returns a 500 if the timezone is invalid, and `ValidateServiceDate` just confirmed the date format is valid. By the time`ParseDate` is called here, both preconditions hold and the branch can never fire. If you want to make the invariant explicit: `panic("unreachable")`. If you'd rather drop the
dead check: removing the `if err != nil` block is also fine. Up to you.

internal/restapi/schedule_for_stop_handler.go
```diff
@@ -367,39 +329,128 @@ type scheduleRowContext struct {
...
+		}
+	}
+}
+
+// estimatedRunCount is how many times a trip is expected to serve the stop across its
+// frequency windows, used to weight its headsign vote against fixed-schedule trips.
+func estimatedRunCount(frequencies []gtfsdb.Frequency) int {
+	runs := 0
+	for _, frequency := range frequencies {
+		headway := time.Duration(frequency.HeadwaySecs) * time.Second
```
Ahmedhossamdev, line 448: Optional. The DB schema enforces `headway_secs > 0` so this can't panic in production, but the function doesn't document that assumption and a unit test passing
`gtfsdb.Frequency{HeadwaySecs: 0}` directly would still hit it. A guard makes the
invariant explicit:

```suggestion
		headway := time.Duration(frequency.HeadwaySecs) * time.Second
		window := time.Duration(frequency.EndTime - frequency.StartTime)
		if headway == 0 {
			continue
		}
		runs += int(window / headway)
```

Not blocking.

internal/restapi/schedule_for_stop_handler.go
```diff
@@ -260,75 +247,50 @@ func (api *RestAPI) buildScheduleForStopReferences(
...
+// buildRouteRefs converts the stop's routes into a combined-ID-keyed reference map,
+// alongside the distinct agency IDs those routes belong to.
+func buildRouteRefs(agencyID string, routesForStop []gtfsdb.GetRoutesForStopRow) (map[string]models.Route, []string) {
+	routeRefs := make(map[string]models.Route, len(routesForStop))
+	agencyIDs := make([]string, 0, len(routesForStop))
+	seenAgencies := make(map[string]bool, len(routesForStop))
 
-	for _, route := range fetchedRoutes {
+	for _, route := range routesForStop {
 		combinedRouteID := utils.FormCombinedID(agencyID, route.ID)
```
Ahmedhossamdev, line 258: Carry-forward note (pre-existing in the removed `fetchRouteRefs`): CLAUDE.md says to use
`route.AgencyID` rather than the stop's agency when forming combined route IDs, since a
stop can be served by routes from other agencies. `route.AgencyID` is available in the
loop. For single-agency feeds this makes no difference, so not blocking now, but worth
fixing before a multi-agency feed is added:

```suggestion
		combinedRouteID := utils.FormCombinedID(route.AgencyID, route.ID)
```

burma-shave (approved):

PR #1398, opened 2026-08-23
fix(restapi): account for extra query binds

### LINKED ISSUE #1367: Batched queries do not count their non-batched bind variables
**Current**: `queryInBatches` sizes batches by ID count alone (`reference_utils.go:520`), but several batched statements carry additional binds beyond the batched slice. `inServiceTripIDs` (`trips_for_location_handler.go:348`) caps `StopIds` at `idsPerBatchedQuery = 900` while the same statement also binds `len(day.serviceIDs) + 1` variables, so a feed with more than ~98 active service IDs on a day still crosses the 999 bound the constant's doc comment targets (`reference_utils.go:511-516`).

Not specific to that call site — `GetTripsByBlockIDs` and `GetActiveTripsWithNullBlockForRoute` on `main` have the same shape, an unbatched `sqlc.slice('service_ids')` alongside a batched one.

**Fix**: account for the fixed and secondary-slice binds when sizing the batch, rather than assuming the batched IDs are the whole statement. Bounding the service-ID list too is the alternative.

**Test**: a service-ID set large enough to overflow alongside a full stop batch, asserting the query still succeeds.

Raised by @aaronbrethorst as non-blocking on #1317 and explicitly out of that PR's scope.

soumajitgh:
## Summary

- Reserve SQLite bind capacity for fixed parameters and secondary sqlc.slice values.
- Batch block-trip and active-route queries to stay within the 999-variable compatibility limit.
- Add regression coverage for a full ID batch with service-ID and fixed binds.

Closes #1367

## Testing

- CGO_ENABLED=1 go test -tags "sqlite_fts5 sqlite_math_functions" ./internal/restapi

### DIFF
--- a/internal/restapi/reference_utils.go
+++ b/internal/restapi/reference_utils.go
@@ -2,6 +2,7 @@ package restapi
 
 import (
 	"context"
+	"database/sql"
 	"fmt"
 	"net/http"
 	"strconv"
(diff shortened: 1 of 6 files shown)


soumajitgh:
Sorry, I didn’t notice that this depends on #1317. Closing this for now.

PR #1401, opened 2026-08-25
Get logger from context, standardize duration logging to milliseconds

harinik:
## Summary by CodeRabbit

- **Logging Improvements**
  - Improved request-specific logging across transit searches, API requests, data imports, retries, and error handling.
  - Standardized elapsed-time and retry-delay values in logs as milliseconds.
  - Added clearer component context to diagnostic messages for easier troubleshooting.
  - Improved consistency for panic, validation, timeout, cancellation, and rate-limit diagnostics.
  - Preserved existing search, transit, and error-handling behavior.


internal/restapi/recovery_middleware.go
```diff
@@ -39,17 +39,22 @@ func NewRecoveryMiddleware(logger *slog.Logger, c clock.Clock) func(http.Handler
...
 					if e, ok := rec.(error); ok {
 						err = e
 					} else {
 						err = fmt.Errorf("%v", rec)
 					}
-					logging.LogError(logger, "handler panic recovered", err,
+					// try to get the request logger from the context. If exists, use it, otherwise, use
+					// the configured logger.
+					log := logger
+					reqLogger := logging.ForComponent(r.Context(), "http_server")
```
burma-shave, line 51: Panic-recovery logging silently drops the app's configured logger and the `request_id`, because `logging.ForComponent(r.Context(), ...)` can never see the request-scoped logger set by the inner middlewares.

`cmd/api/app.go` wires `NewRecoveryMiddleware(coreApp.Logger, ...)(sizeLimitMiddleware(RequestIDMiddleware(requestLogMiddleware(metricsHandler))))` — Recovery wraps *outside* `RequestIDMiddleware`/`requestLogMiddleware`. Those inner middlewares call `r = r.WithContext(ctx)` on their own local `r`, which never propagates back to the `r` captured in this deferred closure. So `logging.ForComponent(r.Context(), "http_server")` here always resolves `FromContext`'s fallback `slog.Default()`, never `coreApp.Logger`. Since `FromContext`/`ForComponent` never return nil, `if reqLogger != nil { log = reqLogger }` is dead code, and the `logger` parameter passed into `NewRecoveryMiddleware` is now unused. Any handler panic in production is logged via Go's default text logger instead of the app's configured JSON/level logger, and the `request_id` field (previously captured explicitly via `r.Context().Value(RequestIDKey)`) is gone entirely — panics become much harder to correlate with the triggering request.

gtfsdb/helpers.go
```diff
@@ -208,10 +208,10 @@ func (c *Client) StoreGtfsData(ctx context.Context, data *GtfsData) (bool, error
 	defer func() {
 		endTime := time.Now()
 
-		c.importRuntime = endTime.Sub(startTime)
+		c.importRuntime = time.Duration(endTime.Sub(startTime))
 
 		logging.LogOperation(logger, "gtfs_data_import_completed",
-			slog.Duration("duration", c.importRuntime),
+			slog.Float64("duration_ms", float64(c.importRuntime)/float64(time.Millisecond)),
```
burma-shave, line 214: `LogOperation`'s zero-duration suppression filter (`internal/logging/structured_logging.go:48`) is hardcoded to attr key `"duration"`, but this line (and `internal/gtfs/direction_precomputer.go:240`) renames the duration attr passed to `LogOperation` to `"duration_ms"` as a `Float64` instead of `slog.Duration("duration", ...)`. That silently disables the filter at both call sites: a zero-duration import/precompute run (e.g. an empty stop set) that previously suppressed a meaningless `duration: 0` log line now always emits `duration_ms: 0`, reintroducing the log clutter the filter was written to prevent.

gtfsdb/helpers.go
```diff
@@ -208,10 +208,10 @@ func (c *Client) StoreGtfsData(ctx context.Context, data *GtfsData) (bool, error
 	defer func() {
 		endTime := time.Now()
 
-		c.importRuntime = endTime.Sub(startTime)
+		c.importRuntime = time.Duration(endTime.Sub(startTime))
```
burma-shave, line 211: Nit: `endTime.Sub(startTime)` already returns `time.Duration`; wrapping it in `time.Duration(...)` is a redundant no-op cast.

internal/restapi/trips_for_route_handler.go
```diff
@@ -843,6 +844,7 @@ func (s *tripReferenceSets) collectTripIDsFromEntries(entries []models.TripsForR
 
 // fillMissingTrips loads the trips that were noted by ID but never fetched.
 func (api *RestAPI) fillMissingTrips(ctx context.Context, sets *tripReferenceSets) {
+	reqLogger := logging.ForComponent(ctx, "http_server")
```
burma-shave, line 847: Nit: `reqLogger` is built unconditionally before the early-return guard (`if len(sets.missing) == 0 { return }`), so every call with no missing trips (the common path) pays for a `slog.Logger.With(...)` allocation that's never used. Moving the guard above the logger construction avoids the wasted work.

burma-shave (changes requested):
Automated code review found two functional/observability regressions worth addressing before merge, plus two minor cleanups.

burma-shave (changes requested):
Thanks!

Be sure to look at our CONTRIBUTING.md for guidelines on commit messages. The commits will need to be amended to comply with that. We don't squash merges so the commit history needs to be clean.

Also there are some merge conlficts.

internal/restapi/trips_for_route_handler.go
harinik: Fixed.

gtfsdb/helpers.go
harinik: Fixed.

gtfsdb/helpers.go
harinik: I updated the LogOperation to look for prefix of duration and also the callers to pass in a duration type. Let me know if this addresses the concern.

internal/restapi/recovery_middleware.go
harinik: I reverted the recovery middleware to the earlier behavior.

burma-shave (changes requested):
The commits on the latest messages are much better, but can you also fix the messages on the previous commits?

PR #1404, opened 2026-08-27
fix: format readableTime in primary agency's timezone

### LINKED ISSUE #1403: readableTime in current-time endpoint always returns UTC instead of agency local time
# Problem

The `readableTime` field in `/api/where/current-time.json` is always formatted in UTC, ignoring the agency's timezone.

## Current behavior

```json
{
  "entry": {
    "readableTime": "2026-08-26T14:30:00Z",
    "time": 1756225800000
  }
}
```

## Expected behavior

The Java OBA reference server formats `readableTime` using the primary agency's local timezone.

For a Pacific-time agency:

```json
{
  "entry": {
    "readableTime": "2026-08-26T07:30:00-07:00",
    "time": 1756225800000
  }
}
```

## Why this matters

Consumers parsing `readableTime` expect a human-readable local time. Returning UTC forces clients to independently resolve the agency's timezone, which they may not have.

## Fix

* Resolve the primary agency's IANA timezone from GTFS data.
* Format `readableTime` using the resolved agency timezone instead of UTC.
* Fall back to UTC if the timezone cannot be loaded.

## Files affected

* `internal/models/current_time.go`
* `internal/restapi/current_time_handler.go`

Ahmedhossamdev:
Fixes: #1403.

The /api/where/current-time.json endpoint previously always formatted readableTime in UTC. The Java OBA reference server formats it in the agency's local timezone with correct UTC offset.

- Add NewCurrentTimeDataInLocation() to models for timezone-aware formatting
- Add agencyTimezone() helper to resolve the primary agency's IANA timezone
- Update handler to use agency timezone, falling back to UTC
- Update tests to validate RFC3339 correctness rather than exact string match

## Summary by CodeRabbit

* **Enhancements**
  * Current time displays now reflect the primary agency’s local timezone.
  * Time values continue using consistent RFC3339 formatting.
  * When no valid agency timezone is available, displays default to UTC.

* **Bug Fixes**
  * Corrected time formatting for agencies operating outside UTC.
  * Added reliable UTC fallback when agency data is unavailable or timezone settings are invalid.

### DIFF
--- a/internal/restapi/current_time_handler.go
+++ b/internal/restapi/current_time_handler.go
@@ -29,8 +29,9 @@ func agencyTimezone(api *RestAPI, r *http.Request) *time.Location {
 	if err != nil || len(agencies) == 0 {
 		return time.UTC
 	}
-	loc, err := time.LoadLocation(agencies[0].Timezone)
+	loc, err := loadAgencyLocation(agencies[0].ID, agencies[0].Timezone)
 	if err != nil {
+		api.Logger.Warn("failed to load agency timezone", "agencyID", agencies[0].ID, "error", err)
 		return time.UTC
 	}
 	return loc
--- a/internal/restapi/current_time_handler.go
+++ b/internal/restapi/current_time_handler.go
@@ -2,20 +2,36 @@ package restapi
 
 import (
 	"net/http"
+	"time"
 
 	"maglev.onebusaway.org/internal/models"
 )
 
 // currentTimeHandler returns the server's current time as a JSON response.
+// readableTime is formatted in the primary agency's local timezone.
 func (api *RestAPI) currentTimeHandler(w http.ResponseWriter, r *http.Request) {
-	// Health Check: fail if GTFS data is invalid
 	if !api.GtfsManager.IsReady() {
 		http.Error(w, "Service Unavailable: GTFS data invalid", http.StatusServiceUnavailable)
 		return
 	}
 
-	timeData := models.NewCurrentTimeData(api.Clock.Now())
+	loc := agencyTimezone(api, r)
+	timeData := models.NewCurrentTimeDataInLocation(api.Clock.Now(), loc)
 	response := models.NewOKResponse(timeData, api.Clock)
 
 	api.sendResponse(w, r, response)
+}
+
+// agencyTimezone returns the primary agency's IANA timezone location.
+// Falls back to UTC if the timezone cannot be loaded.
+func agencyTimezone(api *RestAPI, r *http.Request) *time.Location {
+	agencies, err := api.GtfsManager.GetAgencies(r.Context())
+	if err != nil || len(agencies) == 0 {
+		return time.UTC
+	}
+	loc, err := time.LoadLocation(agencies[0].Timezone)
+	if err != nil {
+		return time.UTC
+	}
+	return loc
 }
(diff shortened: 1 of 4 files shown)


internal/restapi/current_time_handler.go
```diff
@@ -2,20 +2,36 @@ package restapi
...
 }
+
+// agencyTimezone returns the primary agency's IANA timezone location.
+// Falls back to UTC if the timezone cannot be loaded.
+func agencyTimezone(api *RestAPI, r *http.Request) *time.Location {
+	agencies, err := api.GtfsManager.GetAgencies(r.Context())
+	if err != nil || len(agencies) == 0 {
+		return time.UTC
+	}
+	loc, err := time.LoadLocation(agencies[0].Timezone)
```
burma-shave: `agencyTimezone()` reimplements `time.LoadLocation` here instead of reusing the existing `loadAgencyLocation(agencyID, timezone)` helper in `timezone_helper.go`, which every other handler that resolves an agency's timezone already calls (`stops_for_location_handler.go`, `trips_for_route_handler.go`, `schedule_for_route_handler.go`, `vehicles_for_agency_handler.go`, etc.).

More importantly, those call sites log a `Warn` with the agency ID and error when the load fails (see `stops_for_location_handler.go:268-273`), giving operators visibility into a misconfigured `agency.txt` timezone. Here the error is silently swallowed and falls back to UTC with no log output, so the same misconfiguration is invisible specifically on `current-time.json`.

Worth reusing `loadAgencyLocation` and logging on the fallback path for consistency with the rest of the codebase (see CONTRIBUTING.md's Code Reuse section).

internal/restapi/current_time_handler_test.go
```diff
@@ -85,7 +86,9 @@ func TestCurrentTimeHandler_DeterministicTime(t *testing.T) {
...
 
-	// Readable time should match
-	expectedReadable := fixedTime.Format(time.RFC3339)
-	assert.Equal(t, expectedReadable, entry["readableTime"], "Readable time should match mock clock")
+	// readableTime must be formatted in the agency's local timezone (America/Los_Angeles).
+	agencyLoc, err := time.LoadLocation("America/Los_Angeles")
+	require.NoError(t, err)
+	expectedReadable := fixedTime.In(agencyLoc).Format(time.RFC3339)
+	assert.Equal(t, expectedReadable, entry["readableTime"], "readableTime should use agency timezone")
 }
```
burma-shave, line 99: `agencyTimezone()` has three fallback branches — `GetAgencies` error, empty agency list, and an invalid timezone string — but only the happy path (valid `America/Los_Angeles`) is covered here. Per CONTRIBUTING.md, "every new branch or condition introduced by a PR needs test coverage." Worth adding cases for the fallback branches (e.g. via a mocked/empty agency list) so a future change to the fallback behavior doesn't ship silently broken.

burma-shave (commented):
Code review findings

Ahmedhossamdev:
Thanks for the review, @burma-shave. I'll work on the finding today.

Ahmedhossamdev:
Hey @burma-shave, I think this is ready to merge. I’ve resolved your findings, and I’ll merge it now since I need this fix to be in `main` as soon as possible. If there are any other findings, I’ll open another PR.

PR #1407, opened 2026-08-28
Extract per-stop arrivals core into shared functions

ARCoder181105:
### Summary

Pure refactor. Moves the arrivals pipeline out of the `arrivals-and-departures-for-stop` handler into functions a second endpoint can call. No behavior change — the test suite passes untouched.

### What changed

- New `internal/restapi/arrivals_core.go` holding `arrivalsForStop` and `buildArrivalsReferences`.
- `arrivals_and_departures_for_stop_handler.go` drops from 744 to 235 lines and now just parses, calls the core, and builds the envelope.
- `arrivalsAccumulator` gathers the routes, trips, stops and situations that arrivals reference. This is what makes the code reusable across stops: a caller loops over many stops sharing one accumulator and gets a single deduplicated references block.

### Why

The pipeline — the ±1 day service window scan, batch route/trip resolution, per-row arrival construction, reference assembly — was entirely inline and not callable from anywhere. Implementing `arrivals-and-departures-for-location` without this means duplicating all of it.

### Two things worth a look

- `BuildTripStatus` now receives the vehicle the caller already looked up instead of `nil`. **Not a fix** — it falls back to the same `GetVehicleForTrip` call when handed `nil`, so the result is identical. It just drops one redundant lookup per arrival row.
- The extracted stop reference keeps its inline literal rather than calling `buildStopModel`. That helper defaults `Code` to the stop ID when `stops.code` is NULL, where this endpoint emits an empty string, so adopting it would be a response change rather than a refactor. Happy to switch it in a follow-up if the `buildStopModel` behavior is the intended one.

### Testing

`make test` passes with **zero test file changes** — that is the whole correctness argument for this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved arrivals and departures responses when no services match the requested time window.
  * Enhanced handling of arrivals spanning adjacent service days.
  * Improved reliability of real-time predictions, trip status, and reference data.
  * Added route-type filtering support for arrival results.
  * Improved cancellation and error handling during arrival searches.

* **Performance**
  * Reduced unnecessary alert, nearby-stop, and reference lookups when no arrivals are available.
  * Improved efficiency when processing arrival results and related trip information.


ARCoder181105:
Addressed both review points in f32e903. Quality gate is now green (0 issues).

**Cognitive complexity** — four functions were over the threshold of 15. Split each into its distinct sub-tasks:

| Function | Was | Extracted |
|---|---|---|
| `activeStopTimesForWindow` | 21 | `stopTimesForServiceDay` |
| `appendStopReferences` | 18 | `loadStopReferenceData`, `collectStopRoutes` |
| `buildArrival` | 17 | `combinedVehicleID`, `tripStatusForArrival` |
| `batchArrivalEntities` | 16 | `uniqueRouteAndTripIDs`, `tripStopCounts` |

One thing to be careful of if you review that extraction: the per-day loop has two different failure modes. Failing to resolve a day's active services is fatal on day 0 but tolerable on ±1, while failing to read that day's `stop_times` is tolerable on *every* day. Collapsing those into a single error return would have made a day-0 stop-times failure fatal. `stopTimesForServiceDay` keeps them separate.

**`GetStopsByIDs` error** — agreed, fixed. It was setting the result to nil and continuing, which dropped every stop reference and still answered 200, so the entry named stops the client had no way to resolve. It now returns the error and the handler surfaces a 500. `GetRoutesForStops` keeps its warn-and-continue, since losing it only costs each stop its `routeIds`.

Note that this second fix does make the PR no longer *strictly* behaviour-preserving — it only fires on a DB failure, so the suite still passes with no test changes, but the description's "no behavior change" is slightly overstated now. Happy to reword it if you'd prefer.

burma-shave (changes requested):
## Review summary

Automated review (medium effort) plus manual verification of the merge state against current `main`. Two correctness/behavior issues should block merge; the rest are cleanup findings worth addressing in follow-up commits.

### 1. Frequency support is missing, and this PR conflicts with main's frequency work — `internal/restapi/arrivals_core.go`

This branch predates main's frequency-support work (`BuildTripStatus` gained a `freqMap` parameter and callers now populate `arrival.Frequency`). Verified directly:

- `git merge main` into this branch produces a real `CONFLICT (content)` in `arrivals_and_departures_for_stop_handler.go`, spanning ~383 lines — this PR's side is a 3-line stub, main's side is the full old inline block that includes all frequency handling.
- On the clean PR branch (no merge in progress), `grep -n "freq\|Freq"` returns **zero matches** in `arrivals_core.go`, `arrivals_and_departures_for_stop_handler.go`, and `arrival_and_departure_for_stop_handler.go`.

So today, any stop served by a frequencies.txt-based (headway) trip gets `"frequency": null` in `/api/where/arrivals-and-departures-for-stop/{id}` responses instead of the populated block clients use for "every N minutes" countdown UIs. This isn't something git resolves automatically — whoever merges has to manually re-port frequency support into the new accumulator-based structure. Please rebase onto main and reintroduce frequency handling in `arrivals_core.go` before merge.

### 2. `GetStopsByIDs` failure now returns a hard 500 instead of degrading gracefully — `internal/restapi/arrivals_core.go` (`loadStopReferenceData`)

The old handler logged a warning on `GetStopsByIDs` error and continued with `batchStops = nil`, still returning 200 with arrivals populated but `references.stops` incomplete. The new code does:

```go
if err != nil {
    return nil, nil, fmt.Errorf("batch fetch stop references: %w", err)
}
```

which propagates all the way to `api.serverErrorResponse(w, r, err)` — a 500. A transient DB blip on the stop-reference batch query (unrelated to the arrivals already computed) now fails the whole request. This is an undisclosed behavior change; the PR description states "no behavior change." Since this sits in the exact region that conflicts with main (see #1), it's worth fixing while resolving that conflict rather than as a separate patch.

### 3. Dead code: `RouteTypes` / `isRouteTypeAllowed` filter is never wired up

`stopArrivalsInput.RouteTypes` is never set by the only caller (`handler_new.go`'s construction of `stopArrivalsInput`), so `isRouteTypeAllowed` is always called with `allowed == nil` and always returns `true`. This adds an unreachable branch to the hot per-row loop and a predicate function that looks like live functionality but is untested and unused. Either wire it up or drop it until it's needed.

### 4. Dead code: `arrivalsReferencesInput.stopAgencies` is never populated

Same pattern — the only call site never sets `stopAgencies`, so the per-stop agency lookup in `appendStopReferences` can never hit. A future contributor could reasonably assume per-stop agency overrides are already implemented and tested.

### 5. Reuse: route/stop reference building bypasses existing helpers in `reference_utils.go`

- `appendRouteReferences` hand-builds `models.Route` via `models.NewRoute` instead of reusing `buildRouteModels`, the documented single source of truth for `gtfsdb.Route -> models.Route` mapping.
- `collectStopRoutes` manually copies a `GetRoutesForStopsRow` into a fresh `gtfsdb.Route` instead of reusing `routeReferenceFromStopRow`/`routeReferencesForStops`.

Per CONTRIBUTING.md's Code Reuse guidance, these should call the existing helpers rather than duplicating the conversion logic — otherwise a future schema/null-handling fix applied to the shared helpers won't apply here.

### 6. Duplication: the singular arrival-and-departure handler still has its own inline reference-building logic

`internal/restapi/arrival_and_departure_for_stop_handler.go` hand-rolls the same ~110-line stop/route reference-building logic this PR just extracted into `arrivals_core.go`, leaving two parallel implementations that will drift over time. Worth a follow-up to migrate the singular handler onto the new shared helpers.

### 7. Design note: `arrivalsAccumulator.alertAgencyID` is a single scalar despite being documented as multi-stop/multi-agency

If a future arrivals-for-location endpoint loops `arrivalsForStop` across stops from different agencies while sharing one accumulator (the PR's stated purpose for this type), `acc.alertAgencyID` will lock onto whichever agency was set first, and later stops' alerts get namespaced under the wrong agency ID. Worth flagging now since it'll be harder to fix once a second caller depends on it.

---

Requesting changes primarily on #1 and #2 — the rest are good candidates for a follow-up commit or PR.

ARCoder181105:
Both blockers fixed, branch is up to date with `main`.

Commits: `a412b6e` merge, `09d1771` frequency, `9066093` the 500, `3f97961` comment.

Merged `main` rather than rebasing — CONTRIBUTING says not to rewrite reviewed history. Say the word if you want a rebase instead.

**1. Frequency — fixed.** Re-ported into `arrivals_core.go`: `batchArrivalEntities` returns `freqMap` via `fetchFrequenciesForTrips` (fatal on error), `tripStatusForArrival` passes it as `BuildTripStatus`'s 6th arg, and a new `applyFrequency` sets `arrival.Frequency` (guarded, since `selectFrequency` panics on an empty slice). `TestArrivalsAndDeparturesWithFrequency` passes unedited — no test file in this PR is modified. #1408 gets frequency for free via the shared core.

**2. `GetStopsByIDs` 500 — reverted** to warn-and-continue. Note that 500 came from CodeRabbit's review earlier in this PR, so the two reviews pulled opposite ways. Yours wins.

**3 & 4. `RouteTypes` / `stopAgencies` — not dead.** Both are consumed by #1408, which stacks on this branch: `arrivals_and_departures_for_location_handler.go:212` and `:163`. Standalone they do look unreachable. I'll strip them and move them to #1408 if you prefer this PR self-contained.

**5. Reference helpers — I don't think these are safe swaps:**

- `buildRouteModels` labels routes with its `agencyID` *parameter*; `appendRouteReferences` uses each route's own `route.AgencyID`. Multi-agency feeds diverge on `id`, `agencyId`, and `nullSafeShortName`.
- `routeReferencesForStops` returns `utils.MapValues(...)` — map order. That would make each stop's `routeIds` non-deterministic.
- `routeReferenceFromStopRow` returns `models.Route`; `collectStopRoutes` needs a `gtfsdb.Route` for the accumulator.

Same call as the `buildStopModel` note in the description. If `buildRouteModels`' agency semantics are intended, that's a separate bug worth fixing.

**6. Singular handler duplication — agreed**, separate PR.

**7. `alertAgencyID` — comment fixed.** The field is read only by the single-stop handler (`arrivals_and_departures_for_stop_handler.go:172`); #1408 passes a per-stop `agencyID` to `situations.add` directly (`:219`), so the mixup can't happen. The comment claimed multi-stop use — reworded.

`go vet` (both tags), `make test`, `go fmt` clean. No conflicts with `main`.

burma-shave (approved):

PR #1408, opened 2026-08-28
Add arrivals-and-departures-for-location endpoint

### LINKED ISSUE #799: Implement arrivals-and-departures-for-location endpoint
**Overview**
The `arrivals-and-departures-for-location` endpoint is currently fully functional in the legacy Java OneBusAway codebase but has not yet been implemented in the new Go `maglev` architecture. This issue tracks the work required to port the endpoint and achieve full core feature parity.

**Endpoint Details**
* **Method:** `GET`
* **Path:** `/api/where/arrivals-and-departures-for-location.json`
* **Java Handler:** https://github.com/OneBusAway/onebusaway-application-modules/blob/main/onebusaway-api-webapp/src/main/java/org/onebusaway/api/actions/api/where/ArrivalsAndDeparturesForLocationAction.java

**Expected Behavior**
The new `maglev` implementation should match the legacy Java API's core behavior. This includes:
1. Accepting the same spatial and temporal query parameters (bounding box, radius, time windows).
2. Processing active stop times over a 3-day window to support overnight trips.
3. Returning the identical JSON schema, correctly deduplicating `Agencies`, `Routes`, `Stops`, `Trips`, and `Situations` into the `references` block.
4. Returning nearby stops (`StopWithDistance`) for the searched radius.

*Note on Sorting:* The legacy Java application sometimes injected a custom `primarySortAgency` via Spring `@Autowired` configurations (e.g., for Puget Sound/Sound Transit). This is presentation-layer logic and should **not** be ported to Maglev. The references block should remain unsorted to match the existing `arrivals-and-departures-for-stop` Maglev endpoint.

**Legacy Output Reference**
To assist with schema parity and testing, a sample of the expected JSON output generated by the legacy Java codebase can be found here:
* **Java JSON Output Link:** https://api.pugetsound.onebusaway.org/api/where/arrivals-and-departures-for-location.json?key=TEST&lat=47.653&lon=-122.307&radius=1000  

**Acceptance Criteria**
- [ ] Implement parameter parsing and spatial validation.
- [ ] Implement core GTFS data retrieval and processing logic (avoiding N+1 queries).
- [ ] Ensure the JSON response payload strictly matches the legacy schema.
- [ ] Add unit and integration tests (including context cancellation).

ARCoder181105:
Closes #799. Replaces #787.

> **Stacked on #1407.** The diff below includes that PR's commit until it merges — review only the last three commits here, or wait for #1407 to land and this will narrow on its own.

### Summary

Implements `GET /api/where/arrivals-and-departures-for-location.json` — arrivals for every stop in a bounding box or radius. Built on the arrivals core extracted in #1407, so the handler is a loop over the spatial query's stops plus response assembly.

### What changed

| | |
|---|---|
| `arrivals_and_departures_for_location_handler.go` | New handler |
| `models/stops.go`, `models/response.go`, `models/arrival_and_departure.go` | `StopWithDistance`, entry type, response constructor |
| `utils/api.go`, `models/constants.go` | `maxCount` ceiling is now per-endpoint; this one needs 1000, the shared default is 250 |
| `gtfsdb/query.sql` | `ORDER BY` on `GetAgenciesForStops` — multi-stop agency resolution was otherwise nondeterministic |
| `routes.go` | Route registration |

### Parameters

Per the OpenAPI spec: `lat`/`lon` required, `radius` takes precedence over `latSpan`/`lonSpan`, `minutesBefore` 5, `minutesAfter` 35, `maxCount` 250 clamped to 1000, plus `routeType` and `emptyReturnsNotFound`.

### `nearbyStopIds`

The union of the stops within 100 m of each matched stop, each excluding itself, measured from the centre of the search area and ordered nearest first.

Deliberately **not** "every stop in the box" — a matched stop with no neighbour within 100 m does not appear, while a stop just outside the box does if it neighbours one inside.

### Verified against the live deployed server

Ran locally against the King County Metro feed and compared to `api.pugetsound.onebusaway.org` on the same query. Puget Sound aggregates several agencies where a maglev instance loads one static feed, so the comparison is restricted to agency `1`:

| | maglev | pugetsound (agency-1 subset) |
|---|---|---|
| `stopIds` | 18 | 18 — **identical set** |
| `nearbyStopIds` | 19 | 19 — **identical set**, distances agree to 0.05 m |
| response keys | 102 shared | — |

Parameter behaviour matches on `maxCount` (truncates all three lists, sets `limitExceeded`), `maxCount` clamping, `routeType`, `radius` vs span precedence, and the empty-area case. The two response-key differences are both explained: `tripStatus.scheduled` is a pre-existing maglev field on the existing for-stop endpoint too, and the situation keys are absent only because the sole alert in that area belongs to agency `40`, which this config does not load — alerts resolve correctly elsewhere in the feed.

### Divergences from the deployed server — please review these

Five behaviors differ, each because the deployed server contradicts the OpenAPI spec:

1. **`stopIds` is deduplicated.** The Java service adds every matched stop to the list twice (`StopWithArrivalsAndDeparturesBeanServiceImpl` lines 115 and 122; its own comment says 115 shouldn't be there). Live response: 44 entries, 22 unique.
2. **An empty result keeps the `entry`/`references` envelope.** The Java empty path skips `BeanFactoryV2` and emits a different shape (`{stops, nearbyStops, situations, timeZone}`, no `entry`, no `references`). The spec marks both required.
3. **`emptyReturnsNotFound=true` returns 404.** Java sets the 404 and then overwrites it with a 200, so it never ships.
4. **`lat`/`lon` alone search the default 600 m radius.** The Java action never configures a default radius, so it builds a zero-area box and always returns empty.
5. **Arrivals are sorted by arrival time.** The Java sort compares a distance field that is never populated, so it is a no-op and the order is hash order.

`routeType` filtering the arrivals and pruning nearby stops but **not** the stop search is preserved as-is — spec and deployed server agree, and changing it would be the bigger break. Confirmed live: `routeType=99` returns the full `stopIds` list with zero arrivals and zero nearby stops.

### Testing

13 tests covering parameter validation, `radius` vs span precedence, the default-radius fallback, `maxCount` truncating all three lists, `routeType` filtering arrivals but not stops, `emptyReturnsNotFound`, `stopIds` dedup, `nearbyStopIds` ordering and reference resolution, and `includeReferences=false`. Added to `TestOpenAPIConformance_LocationEndpoints`, which validates the response against the spec schema.

## Summary by CodeRabbit

* **New Features**
  * Added a location-based arrivals and departures endpoint supporting coordinates, time windows, route filters, and result limits.
  * Results include nearby stops, distances, service situations, references, and configurable empty-result handling.
  * Added endpoint-specific result ceilings and shared limit validation.
  * Nearby stops and arrivals are ordered consistently.
* **Bug Fixes**
  * Agency and stop results now use deterministic ordering.
  * Improved handling of time-window and arrival parameters, including defaults and validation.
* **Tests**
  * Added comprehensive coverage for validation, filtering, sorting, limits, empty areas, references, and OpenAPI conformance.


ARCoder181105:
Cleared the SonarCloud findings in c5489a0. Quality gate is green (0 issues).

Three functions were over the cognitive complexity threshold of 15. Split each into named steps:

| Function | Was | Extracted |
|---|---|---|
| `arrivalsAndDeparturesForLocationHandler` | 19 | `sendArrivalsForLocationError`, `truncateLocationLists`, `combinedStopIDs`, `registerReferencedStops`, `locationReferences` |
| `agenciesForStops` | 17 | `agencyLocationOrUTC`, `mostCommonAgency` |
| `parseArrivalsForLocationParams` | 16 | `parseRequiredLocation`, `parseEpochMillisParam`, `parseOptionalBoolParam`, `forwardFieldErrors` |

Two of these removed real duplication rather than just moving lines: the cancelled-versus-failed error branch was written out twice, and the field-error funnelling loop was written out twice.

No behaviour change — the handler tests and the full suite pass unchanged.

Also rebased onto the latest #1407 after its review fixes. Once that merges, the diff here drops to the endpoint alone.

burma-shave:
I'll generate a spec for this, hopefully next week. We'll hold off on merging until this can be reviewed against a spec like the other endpoints.

aaronbrethorst:
### Code review

Found 3 issues:

1. Request cost grows linearly with the number of stops in the search area, so radii the endpoint accepts time out. `collectArrivalsForStops` runs the whole per-stop pipeline once per stop. For each stop that means 3x `GetActiveServiceIDsForDate` (the same three dates every time), 3x `GetStopTimesForStopInWindow`, route, trip and stop-count batches, and per-arrival trip status. `nearbyStopsForLocation` then runs one spatial query plus `GetAgenciesForStops` per stop. Measured on the King County Metro feed at `lat=47.6097&lon=-122.3331&time=1789570800000`: `radius=500` takes 1.7 s (Java: 0.09 s), `radius=5000` takes 8.3 s, and `radius=5000&minutesAfter=120` takes 19 s (Java: 0.13 s). `radius=20000`, the clamp ceiling, takes 28 s (Java: 0.23 s). The last two go past the server's 10 s `WriteTimeout`, so the client gets an empty reply (curl exit 52). The handler also keeps computing after the write deadline passes. Fetching service IDs once per service day and batching stop_times and nearby lookups across all stops would fix this.

https://github.com/OneBusAway/maglev/blob/cb1075bfd58c5a738479b8f0931fa71678b99272/internal/restapi/arrivals_and_departures_for_location_handler.go#L196-L221

https://github.com/OneBusAway/maglev/blob/cb1075bfd58c5a738479b8f0931fa71678b99272/internal/restapi/arrivals_and_departures_for_location_handler.go#L238-L245

2. `maxCount` trims the three entry lists but not the references. The accumulator collects trips, routes and stops for every arrival and every matched stop before truncation. `registerReferencedStops` also registers all matched stops, not the truncated list. Java builds references from the already-trimmed beans (`StopWithArrivalsAndDeparturesBeanServiceImpl` trims before `BeanFactoryV2.getResponse` runs). For `lat=47.6097&lon=-122.3331&radius=500&maxCount=5&time=1789570800000`:
   - Java returns 5 arrivals with 6 trips and 14 stops in references (18 KB).
   - This PR returns 5 arrivals, which use 5 trips and 21 stop IDs, but references hold 234 trips and 175 stops (140 KB).

https://github.com/OneBusAway/maglev/blob/cb1075bfd58c5a738479b8f0931fa71678b99272/internal/restapi/arrivals_and_departures_for_location_handler.go#L66-L82

https://github.com/OneBusAway/onebusaway-application-modules/blob/095bf1ac3aeb7009b9f78e7f1cb4c65bd38b988a/onebusaway-transit-data-federation/src/main/java/org/onebusaway/transit_data_federation/impl/beans/StopWithArrivalsAndDeparturesBeanServiceImpl.java#L162-L195

3. The handler file adds its own copies of parsers that already exist (CONTRIBUTING.md says "Before writing new logic, check whether it already exists" and "If what you write is reusable ... add it to the file matching its category rather than leaving it local to your handler"):
   - `parseOptionalBoolParam` duplicates `utils.ParseBoolParam`.
   - `parseRequiredLocation` hand-rolls the required-field check that `utils.ParseRequiredFloatParam` already does.
   - `parseMinutesParam` and `parseEpochMillisParam` re-implement the `minutesBefore`/`minutesAfter`/`time` parsing in `parseArrivalsAndDeparturesParams`.

https://github.com/OneBusAway/maglev/blob/cb1075bfd58c5a738479b8f0931fa71678b99272/internal/restapi/arrivals_and_departures_for_location_handler.go#L450-L531

https://github.com/OneBusAway/maglev/blob/cb1075bfd58c5a738479b8f0931fa71678b99272/internal/utils/api.go#L396-L416

https://github.com/OneBusAway/maglev/blob/cb1075bfd58c5a738479b8f0931fa71678b99272/internal/restapi/arrivals_and_departures_for_stop_handler.go#L46-L79

aaronbrethorst (changes requested):
Thanks for the work here. Building on the arrivals core from #1407 was the right approach, and results line up well with Java. With `maxCount=1000` over three King County Metro areas (a downtown 500 m radius, a downtown span box, and a sparser Rainier Valley radius), the unique stop sets and arrival sets are identical to Java's. Your arrival objects are byte-identical to what main's for-stop endpoint returns. The five Java differences listed in the description all hold up live, and I appreciate them being written down.

Three things need to change before this can merge.

**1. Large or busy areas time out.** Each stop goes through the full per-stop pipeline:
- three `GetActiveServiceIDsForDate` calls for the same three dates every time
- three `GetStopTimesForStopInWindow` calls
- the route/trip/stop-count batches
- per-arrival trip status
- plus a spatial query and `GetAgenciesForStops` per stop for nearby stops

Downtown Seattle, compared with Java:

| Request | This PR | Java |
|---|---|---|
| `radius=500` | 1.7s | 0.09s |
| `radius=2000` | 4.5s | 0.17s |
| `radius=5000&minutesAfter=120` | 19s, empty reply (past the 10s write deadline) | 0.13s |
| `radius=20000` (your clamp ceiling) | 28s, server logged a 500 | 0.23s |

Please:
- Hoist the per-request work (active service IDs per date, and the agency map) out of the stop loop.
- Batch the stop-times lookups across stops.
- Do the nearby-stops lookup as one query.
- Honor `ctx` so the handler stops working once the deadline has passed.

**2. `maxCount` doesn't trim references.** Everything is collected into the accumulator before `truncateLocationLists` runs. At `radius=500&maxCount=5` you return 5 arrivals with 234 trips and 175 stops in references (140 KB). Java returns 6 trips and 14 stops (18 KB), because it builds references from the trimmed results. Please build references from the truncated lists.

**3. Reuse the existing parsers.** CONTRIBUTING's Code Reuse section asks for this:
- `parseOptionalBoolParam` duplicates `utils.ParseBoolParam`.
- `parseRequiredLocation` reimplements `utils.ParseRequiredFloatParam`.
- `parseMinutesParam` and `parseEpochMillisParam` copy `parseArrivalsAndDeparturesParams`.

Smaller things worth fixing while you're in there:
- `NewArrivalsAndDeparturesForLocationResponse(nil, refs, nil, nil, nil, false, clock)` is the unlabeled positional call CONTRIBUTING warns about.
- Nearby stops served only by routes not running on the query date are kept (for example `1_4213`/`1_4214` on the seasonal Waterfront Shuttle), where Java drops them.
- A failure in the nearby lookup still returns 200 with nearby stops silently missing.

Separately, burma-shave's request to hold this until the endpoint has a wiki spec still stands. Given the size (966 lines), splitting the nearby-stops piece into a follow-up PR would also make the next review much easier.

ARCoder181105:
All review feedback is addressed in six scoped commits (reviewed history untouched):

**Performance**
- Arrivals now load in one batched pass: active service IDs resolve once per date, stop_times load once per agency per day via the new `GetStopTimesForStopsInWindow` query, and route/trip/frequency entities resolve once per request instead of per stop.
- Nearby stops are a single expanded-box query instead of one spatial + agency lookup per stop. `ctx` is honored throughout so cancelled requests stop working.
- Fixed a real bug found while batching: the new query mixed `sqlc.slice` with numbered `?2`/`?3` params, which silently bound stop IDs as the window (it worked for exactly 1 stop by accident). Window predicates now come first so numbering is stable.

**References**
- References are rebuilt from the truncated entry lists only, so a `maxCount=5` reply no longer carries hundreds of trips and stops. Situation IDs stay global by design: alerts are few and per-trip tracking would be invasive.

**Nearby correctness**
- Lookup failures propagate instead of returning a silent partial 200.
- Nearby stops served only by routes not running on the query date are dropped (Java parity, e.g. the seasonal shuttle).
- Each nearby stop records its own agency, so `nearbyStopIds` and `references.stops` agree on cross-agency feeds.

**Parsers**
- Reused `ParseRequiredFloatParam`, `ParseBoolParam`, and one shared minutes/epoch-millis helper across both arrivals handlers. Sharing also fixes a duration overflow where huge `minutes` values wrapped negative instead of clamping to the one-day window.
- The empty envelope uses a named `NewEmptyArrivalsAndDeparturesForLocationResponse` factory instead of the unlabeled positional call.

**Tests**
- Unit-level 1000-item clamp test (RABA has 375 stops, so the ceiling can't be hit in integration), radius-vs-spans precedence test, overflow test, reference-trimming assertions under `maxCount=1`, and a conformance test pinned to a fixture-known time that requires a non-empty arrivals list before schema validation.

Verification: `go vet` (both tag sets), `gofmt`, and `make test` all pass.

Two things left for you: the wiki-spec hold still stands, and I kept everything in this PR rather than splitting nearby-stops out since those fixes are entangled with the batch flow — happy to split if you'd still prefer that.

Issue #1377, opened 2026-08-20
Investigate unprefixed situation IDs on route/location/search endpoints


burma-shave:
## Background

OneBusAway situation (service alert) IDs are supposed to be agency-prefixed
combined IDs (`{agencyId}_{entityId}`), matching every other OBA entity ID and
matching legacy Java's `AgencyAndIdLibrary` convention (`AgencyAndId`, always
qualified — see `AgencyAndIdLibrary.convertFromString`).

Maglev's codebase already has this convention implemented for situations —
`situationID()` / `agencyIDForAlert()` in `internal/restapi/reference_utils.go`
— but only two endpoints actually use it.

## Finding

Two conventions currently coexist:

- **`trip-details` and `trip-for-vehicle`** call `tripSituationsFor` /
  `TripSituations` → `situationRefsFromAlerts` → `situationID()`, which
  produces agency-prefixed situation IDs (idempotently — it skips prefixing
  when the raw alert ID is already agency-prefixed by the feed itself).
- **Six other endpoints** call `BuildSituationReferences(alerts)` directly,
  which stamps `ID: alert.ID` — the **raw, unprefixed** ID — with no agency
  qualification at all:
  - `internal/restapi/route_handler.go:66`
  - `internal/restapi/vehicles_for_agency_handler.go:253`
  - `internal/restapi/routes_for_location_handler.go:84`
  - `internal/restapi/stops_for_location_handler.go:263`
  - `internal/restapi/search_stops_handler.go:237`
  - `internal/restapi/route_search_handler.go:82`

So today, the same real-world GTFS-RT alert can be published under two
different situation IDs depending on which endpoint returns it — e.g.
`"alert123"` from `stops-for-location` vs. `"1_alert123"` from
`trip-details` — whenever the backing feed doesn't already self-prefix its
alert IDs. (Puget Sound's production feed happens to self-prefix, which is
why this hasn't been visibly hit yet — `situationID()` is a no-op in that
case and both paths agree.)

None of these six endpoints' wiki pages currently document this at all —
they just describe the `situations` reference array generically, with no
mention that its IDs are unprefixed or that this differs from Java.

## How this was found

Surfaced while reviewing PR #804, which adds `GET /api/where/situation/{id}`.
That new endpoint's own ID-matching behavior is being addressed directly in
PR #804's review — this issue is only about the six pre-existing endpoints
above, which are outside that PR's diff.

## Ask

[...]

PR #1411, opened 2026-08-29
Prefix situation IDs on the remaining endpoints

omlahore:
Migrates the five endpoints that still stamp raw feed IDs onto the agency-prefixed convention `trip-details` and `trip-for-vehicle` already use.

The archaeology #1377 asks for is in [my comment there](https://github.com/OneBusAway/maglev/issues/1377#issuecomment-5461499185). Short version: `BuildSituationReferences` has stamped `ID: alert.ID` since `0c4e274` on 2025-12-16, when raw IDs were the only convention that existed. `situationID()` and `agencyIDForAlert()` arrived eight months later in `0af3b41`, `9185418` and `6033fa4`, applied only where the trip work needed them. Nothing in the history argues for keeping raw IDs at the other callers, so this reads as drift rather than a decision.

## What changed

One helper, `situationReferencesForAlerts`, composing the two existing functions, and five call sites moved onto it.

| Endpoint | Fallback agency |
|---|---|
| `route_handler.go` | the requested agency |
| `vehicles_for_agency_handler.go` | the requested agency |
| `search_stops_handler.go` | none |
| `route_search_handler.go` | none |
| `stops_for_location_handler.go` | none |

`agencyIDForAlert` prefers the alert's own informed entity and only consults the fallback when the alert names no agency. So the three multi-agency endpoints, which can return results from several agencies in one response and therefore have no honest single agency to use, still prefix any alert that names its own, and leave the rest exactly as they are today.

**Note the issue lists six endpoints and only five remain.** `routes_for_location_handler.go` stopped calling `BuildSituationReferences` at some point and now leaves `references.situations` empty entirely, with a comment at line 71 saying as much. That looks like a separate bug and is deliberately not touched here.

## Blast radius

Exactly **three existing assertions changed**, all at the two single-agency endpoints: `TestRouteHandlerWithSituations` and the two `TestVehiclesForAgencyHandler_*SituationsPopulatedInReferences`. Nothing in the multi-agency tests moved, which is the evidence that the empty fallback is a no-op there rather than an assertion I quietly relaxed.

Added `TestSituationReferencesForAlertsNoFallbackAgency` covering the empty-fallback path that no handler test reaches: alert names its own agency and is prefixed, alert names none and keeps its raw ID, feed ID already self-prefixed and is not prefixed twice.

Since Puget Sound's feed self-prefixes its alert IDs, `situationID` is idempotent on production data and published IDs do not move for the deployment this actually hits.

## One behaviour change worth flagging

`situationRefsFromAlerts` skips alerts with an empty ID, which `BuildSituationReferences` included. An alert with no ID cannot be referenced by any entry's `situationIds`, so it was already unreachable, but it does mean such an alert now disappears from `references.situations` instead of appearing as a blank-ID entry. Say the word if you would rather preserve it.

## Deliberately left out

The residual gap on the three multi-agency endpoints. Closing it properly means threading the matched stop or route back to each alert so the alert's agency can be derived from what it matched, and `collectAlertsForStops` discards that association today. That is a larger change than this issue calls for, and it is a design question rather than a fix, so it belongs to you rather than to this PR.

The wiki updates in the issue's third bullet are held until the migration question is settled, since the text depends on the answer.

## Checks

`go vet` under both `sqlite_fts5 sqlite_math_functions` and `purego`, `go fmt`, and `make test` all pass locally.

Refs #1377

## Summary by CodeRabbit

* **Bug Fixes**
  * Situation references now include the correct agency context across routes, vehicles, stops, and searches.
  * Alerts without an agency continue using their original identifiers when results span multiple agencies.
  * Prevented agency prefixes from being duplicated on already-prefixed alert IDs.

### DIFF
--- a/internal/restapi/reference_utils.go
+++ b/internal/restapi/reference_utils.go
@@ -717,6 +717,18 @@ func (api *RestAPI) situationReferences(refs []situationRef) []models.Situation 
 	return situations
 }
 
+// situationReferencesForAlerts builds a situation reference block carrying the
+// agency-prefixed IDs that trip-details and trip-for-vehicle already publish,
+// rather than the raw feed IDs BuildSituationReferences stamps.
+//
+// fallbackAgencyID is consulted only for an alert that names no agency of its
+// own. Callers whose result set spans agencies pass "", because there is no
+// single agency the alert can honestly be scoped to; such an alert keeps its
+// raw ID, as it did before.
+func (api *RestAPI) situationReferencesForAlerts(alerts []gtfs.Alert, fallbackAgencyID string) []models.Situation {
+	return api.situationReferences(situationRefsFromAlerts(alerts, fallbackAgencyID))
+}
+
 func (rb *referenceBuilder) getAgenciesList() []models.AgencyReference {
 	agencies := make([]models.AgencyReference, 0, len(rb.presentAgencies))
 	for _, agency := range rb.presentAgencies {
--- a/internal/restapi/route_handler.go
+++ b/internal/restapi/route_handler.go
@@ -63,7 +63,7 @@ func (api *RestAPI) routeHandler(w http.ResponseWriter, r *http.Request) {
 
 	// Populate situation references for alerts affecting this route
 	alerts := api.GtfsManager.GetAlertsForRoute(routeID)
-	situations := api.BuildSituationReferences(alerts)
+	situations := api.situationReferencesForAlerts(alerts, agencyID)
 	references.Situations = append(references.Situations, situations...)
 
 	response := models.NewEntryResponse(routeData, *references, api.Clock)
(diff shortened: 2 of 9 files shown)


burma-shave:
Closing this PR in favor of #1421.

Since this PR was opened, we clarified the situation-reference contract in the Maglev GitHub wiki: `references.situations` should only contain situation objects that resolve explicit `situationIds` values in the same response. Endpoints without `situationIds` should not include ambient situation references.

That changes the direction here. Rather than normalizing IDs for the ambient situation references on these endpoints, we should remove those ambient references from endpoints that do not expose `situationIds`, and handle any endpoint that should expose true situation references separately.

Thanks for the archaeology and for identifying the inconsistency — that work directly led to the clearer rule captured in #1421.

PR #1416, opened 2026-09-02
Remove Logger from Application struct

harinik:
After the changes in !1424, all the RestAPI server handlers will retrieve the logger from the request context. Removing the Logger field from Application enforces that api.Logger is not used directly and encourages getting the logger from the request context.

Application's Logger was only being used by the RestAPI handlers so removing it does not affect any other components. Verified that all unit tests pass.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved request-scoped logging for API requests, warnings, errors, and panic recovery.
  - Preserved relevant request context in server logs, making troubleshooting clearer and more consistent.
  - Resolved logging-related build issues affecting API components.

- **Refactor**
  - Standardized server lifecycle, metrics, request, and error logging through explicitly configured logging.


aaronbrethorst:
### Code review

aaronbrethorst (dismissed):
Thanks. The refactor itself is correct. Against current main it's really 7 files: `cmd/api/*`, `internal/app/application.go`, and two test files. Production behavior is unchanged: `main.go` sets the default logger and passes the same one in, every request path already uses `logging.FromContext` (which falls back to `slog.Default()`), and request logs still carry `request_id`.

It needs a rebase before it can merge. The branch was cut from #1424's history before that PR was squashed, so it conflicts with main in five files. Some things to watch for:

- **Two files merge "cleanly" but won't compile.** In `internal/restapi/arrivals_core.go` and `current_time_handler.go`, git keeps both copies of the moved `reqLogger :=` lines, which gives seven "no new variables on left side of :=" errors.
- **Take main's side in `internal/logging/structured_logging.go`.** This branch still has the old `attr.Key == "duration" || ...` check. Main has the `KindDuration` check requested on #1424, and resolving in this branch's favor would undo it.
- **Rebase rather than merge main in.** A merge would bring the pre-squash #1424 commits ("fixes", "review", "merge from main") onto main a second time. Rebasing just your Logger-removal commit onto main avoids all of the above.

While you're there, `BuildApplication`'s doc comment in `cmd/api/app.go` still says it creates the logger.

Once it's rebased and green, no re-review is needed unless the resolution changes behavior.

burma-shave:
looks good! I just did a quick rebase to get rid of the old commits

PR #1388, opened 2026-08-21
Cover schedule-for-stop response shape guarantees

ARCoder181105:
Fixes #1034

Stacked on #1387 — its commits show up in this diff until it merges. Review the last commit here.

## What changed

Re-read the endpoint against the wiki and covered what the suite was missing. The existing tests already assert natural route sorting, alphabetical direction sorting, block-boundary flags, both `date` formats, invalid dates, and malformed IDs, so this adds only the gaps:

- The envelope's own fields: `code`, `text`, `version`, `currentTime`.
- Every reference list is present, and `trips`, `stopTimes` and `situations` stay empty — the entry references trips by ID only, which #1032 established.
- Every route named in `stopRouteSchedules` has a record in `references.routes`.
- Stop times carry combined `tripId` and `serviceId`, always include `stopHeadsign`, and come back sorted by departure time.

Departure-time ordering was previously guaranteed only by the query's `ORDER BY`, with nothing asserting it end to end. That matters more now that #1387 appends synthesized stop times after the ordered ones.

## Spec

[Success Guarantees](https://github.com/OneBusAway/maglev/wiki/schedule-for-stop#success-guarantees) and the [Response Structure](https://github.com/OneBusAway/maglev/wiki/schedule-for-stop) section, which documents the envelope and the `scheduleStopTimes[]` field shape.

Coverage of the remaining spec clauses lands with the PRs that implement them: route-less stops and no-service dates in #1384, date-validation ordering in #1385, frequencies in #1386, `exact_times=1` expansion in #1387.

## Summary by CodeRabbit

* **Bug Fixes**
  * Stop schedules now display departures in chronological order for more consistent results.
  * Exact-time trips are now represented as individual scheduled stop times rather than headway windows.

PR #1418, opened 2026-09-03
Return 500 when an entity lookup fails

omlahore:
## The rule this follows

`CONTRIBUTING.md` states it directly, and names the reference implementation:

> When a DB lookup fails, distinguish "not found" from "the query itself failed": check `errors.Is(err, sql.ErrNoRows)` and return 404 via `sendNotFound` only for that case; any other error should go through `serverErrorResponse` (500). Collapsing every error into a 404 hides real outages behind a "not found" response — `internal/restapi/route_handler.go` is a good reference for this pattern.

Three lookups do not follow it, so a busy database, an I/O error or a cancelled context is reported to the client as a missing entity.

## The change

Two of the three take their id from the client, so they get the `route_handler.go:21-30` form exactly:

- `shapes_handler.go:19`, `GetAgency(ctx, agencyID)`
- `schedule_for_route_handler.go:25` and `:31`, `GetRoute` and `GetAgency`

`trip_handler.go:32` is different and goes straight to `serverErrorResponse` with no `ErrNoRows` branch. Its argument is `route.AgencyID`, read out of the database after `GetTrip` and `GetRoute` have both already succeeded. A miss there cannot mean the client asked for something absent; it would mean a route referencing an agency row that is gone. `trip_details_handler.go:158-162` already does exactly this with the same value.

22 production lines across three files.

## Verification

New test at `internal/restapi/db_error_status_test.go`. It closes a database and asserts the handler answers 500 rather than 404, which exercises the real `ErrNoRows`-versus-other distinction instead of asserting against a stub.

It deliberately does **not** use `createTestApi`: that helper returns a package-level GTFS manager built once through `sync.Once` and shared by every test in the package, so closing its database would break the whole suite. The test builds its own in-memory manager, which is the pattern `current_time_handler_test.go:139` already uses.

```
gofmt -l ./internal                                             clean
go vet -tags "sqlite_fts5 sqlite_math_functions" ./...          clean
go vet -tags purego ./...                                       clean
make test                                                       14 packages ok, 0 failures
```

## Scope, deliberately

**`tripHandler` is not covered by the new test, and I would rather say so than imply it is.** Its `GetTrip` lookup runs first and still maps every error to 404, so a closed database never reaches the agency lookup this PR fixes. That blanket 404 is shared with `trip_details_handler.go:146`, so correcting it is a separate two-file change rather than something to smuggle in here.

**Two more sites have the same defect and are left out because they are contested.** `schedule_for_stop_handler.go:33` is touched by #1385, #1386, #1387, #1388 and #1401, and `stops_for_route_handler.go:31` by #1380 and #1401. Fixing them here would conflict with all of those.

**No helper was extracted.** The repeated block is four lines, and a new indirection for three call sites is more review surface than the fix, so the sites stay inline.

## Summary by CodeRabbit

- **Bug Fixes**
  - Database failures in shapes and route-schedule endpoints now return HTTP 500 instead of being incorrectly reported as “Not Found.”
  - Missing routes, agencies, and records continue to return HTTP 404 when appropriate.
  - Trip-related database lookup failures now return a server error response rather than a misleading 404.

- **Tests**
  - Added coverage verifying correct error responses when the database is unavailable.


omlahore:
Windows failure here is `TestSearchStopsHandlerParentStationCrossAgencyReference`, same test that failed on main on Sep 2: https://github.com/OneBusAway/maglev/actions/runs/33676638140

This PR doesn't touch search_stops, and ubuntu-latest is green.

omlahore:
Rebased onto main. The Windows failure was not from this change: the branch predated 031bf7b6, so it was still running the old `search_stops` cross-agency fixture, which fails under the `purego` SQLite build that the Windows job uses. Reproduced it locally with `-tags=purego`, and `TestSearchStopsHandlerParentStationCrossAgencyReference` passes after the rebase. Full package is green under both tag sets.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Thanks. This follows the `route_handler.go` pattern exactly. I checked that every not-found still returns the same 404 body: `schedule-for-route`, `shape`, and `trip` with unknown IDs and unknown agencies match both main and the Java server byte for byte. The trip handler's agency lookup going straight to 500 is sound, since `routes.agency_id` has a foreign key and that branch can't be a legitimate not-found. The new test also fails with main's handlers restored, which is what I want to see.

PR #1424, opened 2026-09-05
Logging improvements: Get logger from context, standardize duration logging to milliseconds

harinik:
* Improved request-specific logging across http handlers.
* Added component context to log messages for easier troubleshooting.
* Removed remaining uses of api.Logger and replaced with request logger, where appropriate.

NOTE: This PR is meant to replace !1401

## Summary by CodeRabbit

- **Improvements**
  - Standardized operational logs across API, GTFS, and realtime processing with component and request context.
  - Request identifiers now carry through request logs, improving traceability during troubleshooting.
  - Operation durations, retry intervals, and realtime timing metrics are reported consistently in milliseconds.
  - Unified logging now covers API errors, warnings, imports, and realtime updates without changing response behavior or processing logic.


omlahore (commented):
In `LogOperation` the new branch calls `attr.Value.Duration()` for anything keyed `duration` or `duration_ms`, and `slog.Value.Duration()` panics when the kind is not a duration. Every call site today passes `slog.Duration` so nothing is broken right now, but the key says milliseconds, so `slog.Int64("duration_ms", 1500)` is the natural thing for someone to write next and that panics with `Value kind is Int64, not Duration`. I ran it to check.

I guess a kind check before the conversion would cover it, so only rewriting when `attr.Value.Kind() == slog.KindDuration` and leaving an already numeric value alone. The `slog.Duration("duration_ms", ...)` in `direction_precomputer.go` and `gtfsdb/helpers.go` reads a little odd for the same reason, since the unit is in the key but the value is still a `time.Duration`.

burma-shave (changes requested):
The previous comments regarding type checking the value passed for the duration_ms key seem worth fixing. Also, I think we can drop support for the `duration` key as it only seems to be used in tests.

harinik:
Addressed comments, PTAL.

burma-shave (approved):

burma-shave:
Thanks @harinik!

PR #1425, opened 2026-09-05
Remove ambient situation references from endpoints without situationIds

### LINKED ISSUE #1421: Remove ambient references.situations from endpoints without situationIds
## Background

We have standardized the OBA API situation-reference contract in the Maglev GitHub wiki:

> `data.references.situations` contains exactly the service-alert situation objects needed to resolve `situationIds` values that appear somewhere in the response data. If the response data has no `situationIds` values, `data.references.situations` is empty.

See the Maglev GitHub wiki page: [Situation-References](https://github.com/OneBusAway/maglev/wiki/Situation-References).

This means `references.situations` is a true reference pool, not an ambient alert-enrichment channel. A situation should not appear merely because it may affect a route, stop, agency, or trip returned by the endpoint. Without a corresponding `situationIds` field, clients have no reliable context for associating the situation with the rest of the response.

## Current problem

Some Maglev endpoints still populate `data.references.situations` even though their response data contains no `situationIds` field.

Known affected endpoints/code paths:

- `GET /api/where/route/{id}.json`
  - `internal/restapi/route_handler.go`
- `GET /api/where/search/route.json`
  - `internal/restapi/route_search_handler.go`
- `GET /api/where/search/stop.json`
  - `internal/restapi/search_stops_handler.go`
- `GET /api/where/stops-for-location.json`
  - `internal/restapi/stops_for_location_handler.go`

These situation objects are ambient: they are not referenced by `entry.situationIds`, `list[].situationIds`, `tripStatus.situationIds`, or any other field in the response data.

This is inconsistent with the documented endpoint specs in the Maglev GitHub wiki, especially:

- `Situation-References`
- `route`
- `search-route`
- `search-stop`
- `stops-for-location`

## Desired behavior

For endpoints with no `situationIds` fields, `data.references.situations` should be an empty array, even when matching service alerts exist for the queried/returned routes or stops.

Specifically:

- `route` should include the owning agency reference, but no situation references.
- `search/route` should include agency references for returned routes, but no situation references.
- `search/stop` should include route/agency/parent-stop references as applicable, but no situation references.
- `stops-for-location` should include route/agency/stop references as applicable, but no situation references.

Endpoints that already expose explicit `situationIds` should continue to include matching situation references.

## Implementation notes

Remove ambient situation population from the affected handlers. Likely removals include calls to:

- `api.GtfsManager.GetAlertsForRoute(...)`
- `api.collectAlertsForRoutes(...)`
- `api.collectAlertsForStops(...)`
- `api.situationReferencesForAlerts(...)`
- `api.BuildSituationReferences(...)` when used only to populate unreferenced `references.situations`

Be careful not to remove situation-reference logic from endpoints where `situationIds` are actually emitted.

## Testing

Add or update tests that seed applicable alerts and assert:

- affected endpoints still return normal route/stop/search data;
- `data.references.situations` is empty;
- no unrelated reference arrays regress.

Suggested endpoint tests:

- `route` with a route alert → `references.situations == []`
- `search/route` with a returned route alert → `references.situations == []`
- `search/stop` with a returned stop alert → `references.situations == []`
- `stops-for-location` with a returned stop alert → `references.situations == []`

soumajitgh:
Four endpoints return service-alert situations even though their response data contains no `situationIds`, leaving clients without a way to associate those alerts with response objects. Remove this ambient alert population from `route`, `search/route`, `search/stop`, and `stops-for-location` so `references.situations` remains `[]`, including when applicable alerts exist.

Agency, route, and parent-stop references are preserved. Endpoints that emit explicit `situationIds` retain their existing situation-reference logic. This follows the contract clarified in #1421 and the maintainer's closing comment on #1411.

Regression coverage seeds matching route/stop alerts for all four endpoints, checks empty arrays rather than null, and verifies normal response data and other references. Updated the existing tests that required ambient situations, including the `stops-for-location` includeReferences cases. All four endpoint regressions failed against the original handlers before the fix.

Validation passed:

- `go test -tags 'sqlite_fts5 sqlite_math_functions' ./...`
- `go vet -tags 'sqlite_fts5 sqlite_math_functions' ./...`
- `go test -tags purego ./internal/restapi`
- `git diff --check`

Closes #1421

## Summary by CodeRabbit

* **Behavior Changes**
  * Route, stop search, and nearby-stop responses no longer include ambient situation references.
  * Situation references remain empty even when matching alerts exist for returned routes or stops.
  * Agency and route references continue to be provided where applicable.
  * Responses remain consistent across empty and populated result sets.

* **Bug Fixes**
  * Prevented unrelated alerts from appearing in route and stop search results.

### DIFF
--- a/internal/restapi/route_handler.go
+++ b/internal/restapi/route_handler.go
@@ -43,6 +43,7 @@ func (api *RestAPI) routeHandler(w http.ResponseWriter, r *http.Request) {
 		route.Color.String,
 		route.TextColor.String)
 
+	// This response has no situationIds, so situation references remain empty.
 	references := models.NewEmptyReferences()
 
 	includeReferences := ShouldIncludeReferences(r)
@@ -60,11 +61,6 @@ func (api *RestAPI) routeHandler(w http.ResponseWriter, r *http.Request) {
 		// Use the existing helper to map the database row to the model
 		references.Agencies = append(references.Agencies, models.AgencyReferenceFromDatabase(&agency))
 	}
-
-	// Populate situation references for alerts affecting this route
-	alerts := api.GtfsManager.GetAlertsForRoute(routeID)
-	situations := api.BuildSituationReferences(alerts)
-	references.Situations = append(references.Situations, situations...)
 
 	response := models.NewEntryResponse(routeData, *references, api.Clock)
 	api.sendResponse(w, r, response)
--- a/internal/restapi/route_handler_test.go
+++ b/internal/restapi/route_handler_test.go
@@ -105,7 +105,8 @@ func TestRouteHandler_EntityIDWithUnderscores(t *testing.T) {
 	}
 }
 
-// Route responses have no situationIds, even when a matching alert exists.
+// TestRouteHandlerOmitsAmbientSituations verifies that matching alerts do not
+// populate situation references and the owning agency remains referenced.
 func TestRouteHandlerOmitsAmbientSituations(t *testing.T) {
 	api := createTestApi(t)
 	defer api.Shutdown()
--- a/internal/restapi/route_handler_test.go
+++ b/internal/restapi/route_handler_test.go
@@ -105,9 +105,8 @@ func TestRouteHandler_EntityIDWithUnderscores(t *testing.T) {
 	}
 }
 
-// TestRouteHandlerWithSituations verifies that a real-time alert informing a
-// route shows up in references.situations for that route's response.
-func TestRouteHandlerWithSituations(t *testing.T) {
+// Route responses have no situationIds, even when a matching alert exists.
+func TestRouteHandlerOmitsAmbientSituations(t *testing.T) {
 	api := createTestApi(t)
 	defer api.Shutdown()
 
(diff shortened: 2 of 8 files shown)


omlahore (commented):
`vehicles-for-agency` looks like it has the same shape and it is not in this PR. It still does the ambient populate in `vehicles_for_agency_handler.go`, and the `TripStatus` there is built with `models.NewTripStatus()` rather than `BuildTripStatus`, so `situationIds` never gets filled. Seeding a route alert against the RABA fixture gives `references.situations` of 1 while the vehicle's `tripStatus.situationIds` is `[]`, which I think is the case #1421 is trying to remove. Its two situation tests assert the current behaviour so they would have to move with it.

Smaller thing, `collectAlertsForStops` has no callers left once this lands. `collectAlertsForRoutes` is still used by the vehicles handler so that one stays either way.

aaronbrethorst:
### Code review

Found 1 issue:

1. `collectAlertsForStops` has no callers once this lands. `search/stop` and `stops-for-location` were its only two users, and this PR removes both calls, so the helper becomes dead code. `collectAlertsForRoutes` is still used by `vehicles_for_agency_handler.go` and should stay. This was also raised in an earlier review comment on this PR and hasn't been addressed. (CONTRIBUTING.md says "untested code paths and leftover dead code are common review findings")

https://github.com/OneBusAway/maglev/blob/f4b812a4fc17090511bdb5993685556176b5eb9f/internal/restapi/reference_utils.go#L429-L439

aaronbrethorst (dismissed):
Thanks for this. The behavior is right, and it matches both the spec and the Java server:

- `Situation-References.md` says situations only appear when something in the response points at them with `situationIds`.
- In Java 2.7.1, `RouteAction`, `search/RouteAction`, `search/StopAction`, and `StopsForLocationAction` all build responses through `BeanFactoryV2` paths that never add situations to references.
- I compared all four endpoints on King County Metro data against main and the Java server. Agency, route, and stop references are unchanged; only the ambient situations are gone.
- The updated tests fail with the old handlers, so they're doing real work.

One thing to fix before merge. With `search/stop` and `stops-for-location` no longer calling it, `collectAlertsForStops` in `internal/restapi/reference_utils.go` has no callers. CONTRIBUTING calls out leftover dead code specifically, and no linter here catches unused unexported methods, so please delete it. (omlahore raised the same point on 9/9.)

Once #1430 lands, `collectAlertsForRoutes` and `deduplicateAlerts` lose their last production callers too. If #1430 merges first, please remove those in the same commit.

After that this is ready to go, and I'm happy to re-review quickly.

PR #1164, opened 2026-07-10
fix: correct FTS stop search ordering and limitExceeded boundary logic

ARCoder181105:
### Description
This PR resolves two specification compliance issues with the `/api/where/search/stop` endpoint:

1. **Ordering:** The search results are now correctly sorted by the combined stop ID in ascending lexicographic order, rather than alphabetically by stop name.
2. **LimitExceeded Flag:** Fixed a false-positive bug where `limitExceeded` was set to `true` when the number of matches exactly equaled `maxCount`. The handler now queries for `limit + 1` rows to accurately verify if additional results exist before truncating the array.

### Changes
* **`gtfsdb/fts_queries.go`**: Updated the `searchStopsByName` SQL query to `ORDER BY s.id`.
* **`internal/restapi/search_stops_handler.go`**: Modified the database call to request `limit + 1` rows, properly evaluate `isLimitExceeded`, and truncate the final slice.
* **`gtfsdb/fts_queries_test.go`**: Updated the existing alphabetical ordering test to assert against the new lexicographic ID sorting.
* **`internal/restapi/search_stops_handler_test.go`**: Added `TestSearchStopsHandlerLimitExceeded` to explicitly verify the `limitExceeded` boundary logic.

closes #1154 

## Summary by CodeRabbit

* **Bug Fixes**
  * Search results are now consistently sorted by stop ID (lexicographic), ensuring stable ordering across searches.
  * Improved `maxCount`/pagination behavior so the `limitExceeded` response flag correctly reflects when additional matching stops are available, while still returning only the requested number of stops.

PR #1426, opened 2026-09-07
fix :- return 404 when no stops match prefix query in search-stop

### LINKED ISSUE #1162: Final Spec Verification & Test Coverage Completion for Search Stops Endpoint
**Objective**
Execute the final specification verification for the `/api/where/search/stop.json` endpoint to ensure exhaustive test coverage and strict long-term parity. This requires correcting non-conformant baseline tests and eliminating all remaining coverage gaps.

**Coverage Gaps to Address**

* **Filter States:** Differentiate between pre-filter `404 Not Found` (no matches) and post-filter `200 OK` (empty list after exclusions).
* **Route Exclusion:** Verify the exclusion of shuttle/school bus route types (`711`, `712`, `713`, `714`) and `0`-route stops, while retaining multi-route stops.
* **Parameter Toggles:** Validate default `maxCount=20` behavior, exact-boundary `limitExceeded` triggers, and empty references when `includeReferences=false`.
* **Parent Resolution:** Ensure `parent` station IDs use the `agencyId_parentId` format and correctly populate `references.stops`.
* **Deterministic Sorting:** Guarantee alphabetical sorting for `references.routes` and `references.agencies`.

**Acceptance Criteria**

* Existing end-to-end and sanitization tests (e.g., `TestSearchStopsHandlerNoResults`) are updated to assert fully spec-compliant HTTP status codes and envelopes.
* Comprehensive unit and integration tests are added for all identified coverage gaps.
* The endpoint demonstrates 100% verified test parity against the finalized specification across all parameter combinations and error boundaries.

Aaravanand00:
This PR completes the remaining spec parity requirements for the `/api/where/search/stop.json` endpoint mentioned in #1162.

### Changes included :- 
- **`internal/restapi/search_stops_handler.go`**: Updated the handler to properly return an HTTP `404 Not Found` response if the database query returns 0 stops (pre-filter empty state). Previously, it fell through to return `200 OK` with an empty list.
- **`internal/restapi/search_stops_handler_test.go`**: Updated the `TestSearchStopsHandlerNoResults` test to assert `http.StatusNotFound` instead of `http.StatusOK`, aligning the test with the actual required spec behavior.

*Note: All other requirements from #1162 (such as route-type exclusions, `includeReferences`, exact-boundary `limitExceeded`, parent station resolution, and sorting) are already covered by existing tests merged in previous PRs (like #1164 and #1284).*

Fixes #1162

## Summary by CodeRabbit

- **Bug Fixes**
  - Stop searches with no matching results now return a 404 Not Found response instead of a successful response containing an empty list.
  - Queries containing only whitespace, punctuation, or unsupported search terms now receive a clear 404 response.
  - The response includes the appropriate 404 status code for consistent client handling.
  - Searches that find matching stops continue to return successful responses with results.

### DIFF
--- a/internal/restapi/search_stops_handler.go
+++ b/internal/restapi/search_stops_handler.go
@@ -86,8 +86,7 @@ func (api *RestAPI) searchStopsHandler(w http.ResponseWriter, r *http.Request) {
 	terms := extractFTS5Terms(sanitizedQuery)
 
 	if len(terms) == 0 {
-		response := models.NewListResponseWithRange([]models.Stop{}, *models.NewEmptyReferences(), false, api.Clock, false)
-		api.sendResponse(w, r, response)
+		api.sendNotFound(w, r)
 		return
 	}
 
@@ -141,6 +140,11 @@ func (api *RestAPI) searchStopsHandler(w http.ResponseWriter, r *http.Request) {
 			)
 			return
 		}
+	}
+
+	if len(stops) == 0 {
+		api.sendNotFound(w, r)
+		return
 	}
 
 	stops, isLimitExceeded := utils.PaginateSlice(stops, 0, limit)
--- a/internal/restapi/search_stops_handler_test.go
+++ b/internal/restapi/search_stops_handler_test.go
@@ -105,8 +105,8 @@ func TestSearchStopsHandlerNoResults(t *testing.T) {
 
 	resp, stopsResp := callAPIHandler[StopsResponse](t, api, searchStopsURL(url.Values{"input": {"NonExistentStopName12345"}}))
 
-	assert.Equal(t, http.StatusOK, resp.StatusCode)
-	assert.Empty(t, stopsResp.Data.List)
+	assert.Equal(t, http.StatusNotFound, resp.StatusCode)
+	assert.Equal(t, http.StatusNotFound, stopsResp.Code)
 }
 
 func TestSearchStopsHandlerMaxCount(t *testing.T) {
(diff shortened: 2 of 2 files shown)


omlahore (commented):
I think the pre-filter rule only lands on one of the two paths that can return zero matches. `sanitizeFTS5Query` strips the input before the query is built, so `input="   "` or `input='*()"'` or `input="!!!"` all bail out at the `len(terms) == 0` return further up with a 200 and an empty list, and never reach the new check.

On the RABA fixture that gives `"NonExistentStopName12345"` a 404 but `"!!!"` a 200, and both of those matched nothing before any filtering. I am not sure whether #1162 means unparseable input to be its own case, but if it does I guess it is worth a line in the test so the split is deliberate rather than just where the early return happens to sit.

ARCoder181105:
I was thinking , to wait before merging this , there is already a sibling sub issue PR is opened , #1162 is acting as a final closing sub issue of the spec verification for this endpoint , when all the sibling sub issues are closed , i prefer u then run a final checks like test coverage etc. so that if there is anything caught an eye it should be either fixed or documented , 

Summary : u made the PR a bit early as i planned in my workflow , no worries just wanted to give the heads up

omlahore:
I guess that makes sense. I'm still sort of unsure about the zero match split that I mentioned. #1162 is the parity issue, so I guess that's where it belongs. Thanks 👍

Aaravanand00:
@omlahore ptal..

ARCoder181105:
Thank you for submitting this PR. 

Unfortunately, I will be closing it as the proposed changes are inaccurate when compared to the [current wiki spec](https://github.com/OneBusAway/maglev/wiki/search-stop). 

The spec explicitly requires returning an HTTP 200 with an empty list for no-match queries, rather than the HTTP 404 implemented here.

Please also ensure you do not pick up issues that are already assigned to someone else. 

That being said, it is my bad that I haven't updated the issue itself—a lot of things have changed in the documentation since the issue was originally created.

Next time, please ask in the issue comments before starting work or making a PR. This allows me to verify that the issue isn't stale and that the requirements still align with the latest spec before you spend your time on it.

Issue #1018, opened 2026-06-04
spec-gap: Fix BlockSequence indexing to continuously increase across trips


ARCoder181105:
**Description:**

In `block_handler.go`, `BlockSequence` is populated using `int(stop.StopSequence - 1)`. Since `StopSequence` comes directly from the GTFS `stop_times` table, it resets to `0` at the beginning of each new trip.

The wiki specifies that `BlockSequence` should be a "zero-based sequence index across all stops in the entire block configuration". Therefore, it should continuously increase across all trips in the block (e.g., from `0` to `N-1` for the whole block).

**Acceptance Criteria:**
- Calculate a continuous, 0-indexed `BlockSequence` across all trips in a block configuration.
- Add test coverage to ensure `BlockSequence` correctly continues incrementing between trips.

ARCoder181105:
Closing as we're dropping the block endpoint from maglev.

PR #1428, opened 2026-09-08
fix(block): reset distanceAlongBlock between configurations

### LINKED ISSUE #1427: blockDistance leaks across service configurations in block handler
# blockDistance leaks across service configurations in block handler

## Summary

The `/api/where/block/{id}.json` endpoint carries the `blockDistance`
accumulator from one `BlockConfiguration` into the next.

In `internal/restapi/block_handler.go`, `blockDistance` is declared
outside the loop over `serviceID`:

```go
var blockDistance float64

for _, serviceID := range serviceIDs {
    ...
}
```

As a result, when a block has trips associated with multiple
`service_id`s, the distance calculation for a later configuration
starts with the cumulative distance from the previous configuration.

## Reproduction

Using a block with two trips assigned to different `service_id`s, the
handler produces output equivalent to:

```text
Configuration 0:
  Stop 0: DistanceAlongBlock = 0
  Stop 1: DistanceAlongBlock = 111195.10

Configuration 1:
  Stop 0: DistanceAlongBlock = 111195.10  <-- unexpected
  Stop 1: DistanceAlongBlock = 222390.20
```

The first stop of the second configuration inherits the accumulated
distance from the first configuration.

## Expected behavior

Each `BlockConfiguration` should calculate its distances independently:

```text
Configuration 0:
  Stop 0: DistanceAlongBlock = 0
  Stop 1: DistanceAlongBlock = 111195.10

Configuration 1:
  Stop 0: DistanceAlongBlock = 0
  Stop 1: DistanceAlongBlock = 111195.10
```

Distance should still accumulate normally within each configuration.

## Root cause

`blockDistance` is scoped outside the `serviceID` loop, so its value
persists between configurations.

Moving the declaration inside the loop would reset the accumulator for
each configuration:

```go
for _, serviceID := range serviceIDs {
    var blockDistance float64
    ...
}
```

## Impact

This produces incorrect `distanceAlongBlock` values in block API
responses whenever multiple service configurations are present.

Clients using these values to determine position along a block can
receive offsets that include distance accumulated from a different
service configuration.

## Proposed fix

Scope `blockDistance` to the individual `BlockConfiguration`.

A regression test should verify that:

1. Each configuration starts with `DistanceAlongBlock == 0`.
2. Distance continues accumulating correctly within each configuration.

## Scope

This issue is specifically about `blockDistance` leaking between
service configurations. It does not address the separate `BlockSequence`
behavior tracked in #1018.

priyanshu7739410:
### Summary
Fixes #1427. Resolves a bug in `transformBlockToEntry` where the `blockDistance` accumulator was being erroneously carried across entirely distinct calendar configurations. This ensures that every new `BlockConfiguration` returned by the `/api/where/block/{id}.json` endpoint accurately tracks its own geometric offsets beginning from `0.0`.

### Root Cause
The `var blockDistance float64` accumulator was previously declared outside the iteration block over `serviceIDs`, meaning its value was never reset between independent service days.

### Fix
Scoped `var blockDistance float64` directly into the individual `serviceIDs` iteration loop.

### Tests
Added `TestBlockHandlerCrossConfigurationDistanceReset` to verify that `DistanceAlongBlock` is strictly `0.0` at the first stop of independent configurations belonging to the same block entity, without harming the distance accumulations on contiguous trips within that block.

### Scope
This PR is narrowly focused to **only** address the cross-configuration `blockDistance` leak described in #1427. It intentionally does **not** address the broader trip ordering mechanisms or the distinct within-trip `BlockSequence` incrementing behavior flagged in #1018.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed distance calculations so each service configuration starts at zero while continuing to accumulate correctly within that configuration.
  - Prevented distance values from carrying over between separate service configurations.

- **Tests**
  - Added coverage verifying distance resets and accumulation across multiple service configurations.

### DIFF
--- a/internal/restapi/block_handler.go
+++ b/internal/restapi/block_handler.go
@@ -78,9 +78,9 @@ func transformBlockToEntry(block []gtfsdb.GetBlockDetailsRow, blockID, agencyID 
 
 	configurations := make([]models.BlockConfiguration, 0, len(serviceGroups))
 
-	var blockDistance float64
-
 	for _, serviceID := range serviceIDs {
+		var blockDistance float64
+
 		serviceStops := serviceGroups[serviceID]
 
 		config := &models.BlockConfiguration{
--- a/internal/restapi/block_handler_test.go
+++ b/internal/restapi/block_handler_test.go
@@ -222,41 +222,28 @@ func TestBlockHandlerContextCancellation(t *testing.T) {
 }
 
 func TestBlockHandlerCrossConfigurationDistanceReset(t *testing.T) {
-	// Construct a synthetic block with 2 distinct service IDs
-	rows := []gtfsdb.GetBlockDetailsRow{
-		{
-			ServiceID: "weekday", TripID: "trip1", StopSequence: 1, StopID: "stopA",
-			Lat: 0.0, Lon: 0.0,
-		},
-		{
-			ServiceID: "weekday", TripID: "trip1", StopSequence: 2, StopID: "stopB",
-			Lat: 1.0, Lon: 0.0,
-		},
-		{
-			ServiceID: "weekend", TripID: "trip2", StopSequence: 1, StopID: "stopC",
-			Lat: 0.0, Lon: 0.0,
-		},
-		{
-			ServiceID: "weekend", TripID: "trip2", StopSequence: 2, StopID: "stopD",
-			Lat: 1.0, Lon: 0.0,
-		},
-	}
+	api := createTestApi(t)
+	defer api.Shutdown()
 
-	entry := transformBlockToEntry(rows, "block_1", "agency")
+	resp, model := callAPIHandler[BlockEntryResponse](t, api, blockURL("25_1"))
 
-	require.Len(t, entry.Configurations, 2, "expected exactly 2 configurations")
+	assert.Equal(t, http.StatusOK, resp.StatusCode)
+
+	entry := model.Data.Entry
+	require.GreaterOrEqual(t, len(entry.Configurations), 2, "expected at least 2 configurations")
 
 	config0 := entry.Configurations[0]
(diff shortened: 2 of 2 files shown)


omlahore (commented):
Fix looks right to me. I ran the new test against the merge base and it fails there on the config 1 assertion, so it is a real regression test and not just a passing one.

One thing I noticed while checking it. The bug is visible through the fixture already, and on `25_1`, which the tests above it are using. On the merge base `/api/where/block/25_1` returns `DistanceAlongBlock` of 0 for the first configuration and 433750.77 for the first stop of the second one, and both go to 0 with your change. So I think an assertion through `callAPIHandler` would cover the same thing while also going through the real `GetBlockDetails` row shape, which the hand built rows kinda skip since they leave the arrival and departure fields at zero.

priyanshu7739410:
Thanks for the review! I’ve updated the regression test to use the existing 25_1 fixture and exercise the API through callAPIHandler, as suggested. It now verifies both the reset between configurations and accumulation within each configuration. The update is pushed to the PR.

omlahore:
Test reads right to me now. I dropped the unused import locally and ran it against the merge base and it fails there with `expected: 0, actual: 433750.7712378005` on the config 1 assertion, so it still catches the thing it is meant to catch.

The build is red though. `gtfsdb` is still imported at the top of `block_handler_test.go` and nothing uses it now that the synthetic rows are gone, so `go vet` fails under both tag sets and both Test jobs are failing on it. SonarCloud going green threw me for a second, but it does not compile the Go.

priyanshu7739410:
Thanks! I removed the unused gtfsdb import and pushed the cleanup in 55309631.

The updated regression test passes, and all CI checks are now green.

PR #1430, opened 2026-09-11
Populate situationIds on vehicles-for-agency

### LINKED ISSUE #1429: vehicles-for-agency returns ambient references.situations, same contract break as #1421
## Summary

`GET /api/where/vehicles-for-agency/{id}.json` populates `data.references.situations` with alerts that nothing in the response resolves. This is the same contract break as #1421, which lists four endpoints and does not include this one.

The wiki contract, quoted from #1421:

> `data.references.situations` contains exactly the service-alert situation objects needed to resolve `situationIds` values that appear somewhere in the response data. If the response data has no `situationIds` values, `data.references.situations` is empty.

## Why it was missed

This endpoint looks compliant from the response shape. Each vehicle carries a `tripStatus`, and `models.TripStatus` has a `situationIds` field, so a search for endpoints without one does not turn it up.

The field is never filled. `vehicles_for_agency_handler.go` builds the status with `models.NewTripStatus()` at lines 139 and 222 rather than `api.BuildTripStatus`, and `BuildTripStatus` is the only thing that sets `SituationIDs` (`trips_helper.go:67`). `NewTripStatus` leaves it as an empty slice. Meanwhile lines 250 to 253 still add both route-scoped and agency-scoped alerts to `references.situations`:

```go
alerts := deduplicateAlerts(
    api.collectAlertsForRoutes(routeIDs),
    api.GtfsManager.GetAlertsByIDs("", "", id),
)
references.Situations = append(references.Situations, api.BuildSituationReferences(alerts)...)
```

So every situation this endpoint returns is ambient, on both the route path and the agency path.

## Reproduction

Seed an agency-scoped alert, request the endpoint, then collect every `situationIds` in the body and compare against `references.situations`. Using the existing helpers in `situation_references_test.go`:

| endpoint | refs.situations | situationIds found |
|---|---|---|
| vehicles-for-agency | 1 | 0 |
| trips-for-location | 1 | 2 |
| trip-details | 1 | 1 |
| trip-for-vehicle | 1 | 2 |

The other three resolve their references. This one does not. A route-scoped alert behaves the same way.

## Suggested fix

Either drop the ambient population the way #1425 does for the other four, or populate `tripStatus.situationIds` so the references have something pointing at them. Dropping it looks more consistent with #1421, but the second option is also defensible here since the field exists on the model.

Worth noting that `TestVehiclesForAgencyHandler_SituationsPopulatedInReferences` and `TestVehiclesForAgencyHandler_AgencySituationsPopulatedInReferences` both assert the current behaviour, so they would need to move with whichever way this goes.

## Note on coverage

`TestSituationIDsResolveToReferences` covers the forward direction, that every emitted `situationIds` resolves to a reference. There is no test for the reverse, that every reference is resolved by some `situationIds`, which is what #1421 is actually about. A table-driven test over the endpoints in that direction would have caught this and would stop the four in #1421 from regressing later.

omlahore:
Fixes #1429.

The wiki documents tripStatus.situationIds for this endpoint, and TripStatus in the OpenAPI spec has it too. So this isn't one of the #1421 endpoints that have no situationIds field at all. The bug was both halves. The ambient route and agency sweep put situations in references that nothing pointed at, and the field itself was never filled in.

The handler now fills situationIds from the trip's alerts, and it builds references.situations from the same collector trips-for-route uses. The ambient sweep is gone.

I added two more commits after the first review. The alert lookup takes the route and agency from the routes already batched for the response, so it doesn't run its own GetTrip and GetRoute for every vehicle. And an alert informed about the nominal trip stays on the entry when the vehicle gets interlined onto a different trip.

I added vehicles-for-agency to TestSituationIDsResolveToReferences, and both situation tests check the seeded alert is visible before the request. The interlining test is new and it fails without the last commit. make test and make lint pass.

## Summary by CodeRabbit

- **Bug Fixes**
  - Vehicle-for-agency responses now include situation references associated with each active or nominal trip.
  - Situation IDs are consistently propagated to vehicle trip statuses and resolve to the corresponding situation references.
  - Alerts remain available when trips are interlined.
  - Duplicate situation references are removed when the same alert applies through multiple trips.
  - References no longer include unrelated ambient route- or agency-level situations.


internal/restapi/vehicles_for_agency_handler.go
```diff
@@ -245,12 +245,6 @@ func (api *RestAPI) vehiclesForAgencyHandler(w http.ResponseWriter, r *http.Requ
 		references.Agencies = []models.AgencyReference{models.AgencyReferenceFromDatabase(agency)}
 		references.Routes = routeRefList
 		references.Trips = tripRefList
```
aaronbrethorst, line 263: The premise here doesn't hold up against the wiki contract, and this fix goes the wrong direction.

`vehicles-for-agency` is **not** an endpoint without `situationIds`. The wiki page for this endpoint documents the field explicitly:

> **`data.list[].tripStatus.situationIds`** — IDs of active service alerts affecting this trip. Each ID corresponds to a situation in `references.situations`; see Situation-References.

and its response-structure block lists `situationIds` in the `data.list[].tripStatus` schema. `TripStatus` in `testdata/openapi.yml` (the schema `VehicleStatus.tripStatus` points at) carries `situationIds` too. The field is also on the wire on every response today: `models.NewTripStatus()` initialises `SituationIDs: []string{}` with no `omitempty`, so the endpoint already emits `"situationIds": []` for every vehicle. I confirmed that against this branch.

That puts the endpoint on the other side of the line #1421 draws. #1421 scopes itself to endpoints "even though their response data contains no `situationIds` field," and burma-shave's closing comment on #1411 splits it the same way: "remove those ambient references from endpoints that do not expose `situationIds`, and handle any endpoint that **should expose true situation references separately**." This is one of the latter. #1429 spotted both options and called the populate path "defensible... since the field exists on the model" — given the wiki documents the field, it's the required one.

Concrete regression: seed a route- or agency-scoped alert for the route a tracked vehicle is running, then GET `/api/where/vehicles-for-agency/{id}.json`. On this branch the response carries no alert information anywhere — `tripStatus.situationIds: []` and `references.situations: []`. A rider tracking that bus can no longer see that their route is disrupted. That contradicts this page's own Success Guarantee ("Trips, routes, agencies, stops, and **service alerts** referenced in the list are populated in the references block") and Main Success Scenario step 6.

The populate direction is small and uses helpers that already exist. `api.TripSituations(ctx, tripID)` returns both halves from one lookup — I ran it against the RABA fixture with a route-scoped alert seeded and it returns `["25_zz-review-alert"]` plus the matching `models.Situation`. Roughly: in the `vehicle.Trip != nil` branch, set `tripStatus.SituationIDs` from it for `activeTripID`, accumulate the returned situations in a dedup map alongside `routeRefs`/`tripRefs`, and assign that map's values to `references.Situations` here. That also drops the ambient agency-wide `GetAlertsByIDs("", "", id)` sweep, which is the part #1421 is actually complaining about, while keeping alerts that genuinely affect a returned trip.

One more thing worth folding in: `TestSituationIDsResolveToReferences` (`internal/restapi/situation_references_test.go:66`) documents itself as covering "every endpoint that emits situationIds," and `vehicles-for-agency` is missing from its table — precisely because the field was never populated. If you take the populate direction, adding a case there is the natural regression coverage, and it closes the gap #1429's own "Note on coverage" section describes.

internal/restapi/vehicles_for_agency_handler_test.go
```diff
@@ -198,20 +196,10 @@ func TestVehiclesForAgencyHandler_SituationsPopulatedInReferences(t *testing.T)
...
-	require.NotEmpty(t, model.Data.References.Situations, "expected at least one situation in references")
-	found := false
-	for _, sit := range model.Data.References.Situations {
-		if sit.ID == alertID {
-			found = true
-			break
-		}
-	}
-	assert.True(t, found, "expected situation with id %q in references.situations", alertID)
+	assert.Equal(t, []models.Situation{}, model.Data.References.Situations)
```
aaronbrethorst: Independent of which direction the fix goes, this assertion (and the matching one at line 222) can't fail for the reason the test name claims.

The old body anchored the seeded alert with `require.NotEmpty(t, model.Data.References.Situations, ...)` before looking for `alertID`. With that gone, nothing in the test verifies the alert was ever seeded or that it matches the vehicle's route. If `MockAddAlert` stops firing `rebuildMergedRealtimeLocked`, or `mustGetTrip` starts returning a trip on a route the informed entity doesn't match, or the route-ID form drifts, `references.situations` is empty for a reason that has nothing to do with ambient dropping — and the test still passes green. The eight lines of alert setup above become decoration.

A cheap anchor keeps it honest: assert `api.GtfsManager.GetAlertsForRoute(rawRouteID)` (and `GetAlertsByIDs("", "", agencyID)` in the agency case) is non-empty before the request, so the test proves the handler dropped an alert it could actually see rather than that there was never one there.

Smaller: both renamed tests lost their doc comments. The surrounding file comments its situation tests, and the new names describe behaviour that's the opposite of the obvious default, so a one-liner saying why empty is correct here is worth keeping.

internal/restapi/vehicles_for_agency_handler.go
omlahore: Thanks, you're right, I missed that the wiki documents situationIds for this endpoint. It now populates them from the trip's alerts using the same collector trips-for-route uses, and only the agency-wide sweep is gone. I also added vehicles-for-agency to TestSituationIDsResolveToReferences, and that case fails on main.

internal/restapi/vehicles_for_agency_handler_test.go
omlahore: You're right about the tests too. Both now check that the seeded alert is visible before the request, and the doc comments are back.

internal/restapi/vehicles_for_agency_handler.go
```diff
@@ -187,6 +188,7 @@ func (api *RestAPI) vehiclesForAgencyHandler(w http.ResponseWriter, r *http.Requ
 				vehicleStatus.OccupancyStatus = occupancy
 			}
 
+			tripStatus.SituationIDs = situations.addRefs(api.situationRefsForTrip(ctx, activeTripID))
```
aaronbrethorst: `situationRefsForTrip` runs `GetTrip(activeTripID)` + `GetRoute(routeID)` for **every vehicle**, inside a loop that was deliberately written to avoid exactly this — lines 77-96 collect the route IDs and batch them through `GetRoutesByIDs` precisely so the per-vehicle path does no route lookups. `vehicles-for-agency` is the real-time polling endpoint (15% of the k6 load profile), and a large agency returns thousands of vehicles, so this adds ~2N SQLite round-trips per poll with no cache in front of them.

The data is already in hand for the common case: `routeID` is `vehicle.Trip.ID.RouteID` and the agency is `routeByID[routeID].AgencyID`, which makes the lookup free:

```go
routeID := vehicle.Trip.ID.RouteID
agencyID := ""
if route, ok := routeByID[routeID]; ok {
    agencyID = route.AgencyID
}
tripStatus.SituationIDs = situations.addRefs(
    situationRefsFromAlerts(api.GtfsManager.GetAlertsByIDs(activeTripID, routeID, agencyID), agencyID),
)
```

`trips_for_route_handler.go` already has `tripSituationRefs(ctx, tripID, tripsByID, routeAgencyMap)` for this shape — resolve from preloaded maps, fall back to `situationRefsForTrip` only when the trip isn't indexed. Worth reusing rather than paying the per-vehicle queries.

aaronbrethorst: Alerts are resolved from `activeTripID` only, so when `resolveActiveTripID` interlines the vehicle onto a different trip, any alert informed about the *nominal* trip or its route is silently dropped — even though the response still reports `tripId` = nominal trip and emits a route reference for the nominal route.

Concrete case: vehicle V is reported on trip A (route R1); the block resolves the active trip to B (route R2). A GTFS-RT alert informed about route R1 was included in `references.situations` on `main` (via `collectAlertsForRoutes(routeIDs)`, which was keyed off `vehicle.Trip.ID.RouteID`). After this change it appears nowhere in the response, so a client rendering the vehicle's `tripId`/route has no way to surface it.

Feeding both IDs through the collector keeps the nominal path covered and costs nothing extra, since the dedup map already collapses the overlap:

```go
tripStatus.SituationIDs = situations.addRefs(api.situationRefsForTrip(ctx, vehicle.Trip.ID.ID))
if activeTripID != vehicle.Trip.ID.ID {
    tripStatus.SituationIDs = append(tripStatus.SituationIDs,
        situations.addRefs(api.situationRefsForTrip(ctx, activeTripID))...)
}
```

If dropping the nominal trip's alerts is the deliberate call, that's defensible — but it's a behavior change from `main` that isn't mentioned in the PR description and has no test pinning it.

aaronbrethorst (changes requested):
Thanks for turning this around so quickly — `28c1b31` takes the fix in the right direction. Populating `tripStatus.situationIds` from `situationRefsForTrip` and building `references.situations` from the same refs is exactly right: it closes both halves of #1429 at once instead of trading one for the other. Reusing `situationCollector` / `situationReferences` rather than hand-rolling, restoring the original test names and doc comments, adding the pre-flight `require`s so the assertions can't pass vacuously, and registering `vehicles-for-agency` in `TestSituationIDsResolveToReferences` are all the right calls. This is the shape I want.

Two things left before it lands, both on the new `situationRefsForTrip` call — details inline:

1. **N+1 on a hot endpoint.** `situationRefsForTrip` does its own `GetTrip` + `GetRoute` per vehicle, inside the loop that lines 77-96 deliberately batch with `GetRoutesByIDs` to avoid exactly that. `vehicles-for-agency` is ~15% of the k6 profile and gets polled hard; on a large agency this is roughly 2N extra SQLite round-trips per poll. The route ID and agency are already in hand (`vehicle.Trip.ID.RouteID`, `routeByID`), and `tripSituationRefs(ctx, tripID, tripsByID, routeAgencyMap)` in `trips_for_route_handler.go` already exists for this case.

2. **Nominal trip's alerts get dropped under interlining.** Resolving alerts from `activeTripID` alone means an alert informed about the *nominal* trip or route is lost, even though the response still reports `tripId` as the nominal trip and emits a reference for its route. `main` picked those up via `collectAlertsForRoutes(routeIDs)`, so this is a behavior change. Dedup makes feeding both IDs through the collector essentially free; if you'd rather keep it to the active trip, that's arguable, but it needs to be deliberate — a comment saying so and a test pinning the behavior.

Please add these as new commits rather than amending — the first two have been reviewed and I'd rather not re-read them.

One more note for the description when you update it: the original framing of #1429 as "the same contract break as #1421" isn't right, and it's worth correcting so the next person reading this history doesn't repeat it. #1421 covers endpoints with no `situationIds` field at all, where dropping the references is the only coherent option. This endpoint documents the field — see `data.list[].tripStatus.situationIds` on the `vehicles-for-agency` wiki page, and `TripStatus` in the OpenAPI spec — so the dangling references were half the bug and the never-populated field was the other half. Your second commit is the fix; the PR title and body still describe the first one.

internal/restapi/vehicles_for_agency_handler.go
omlahore: Yeah, you're right, that was undoing the batching. It takes the route and agency from routeByID now. The per-trip lookup only kicks in when the route isn't in the batch, so the common path doesn't do any queries.

omlahore: No, that wasn't deliberate. Both trip IDs go through the collector now, and they're deduped by situation ID because an alert can inform both. I added a test with an alert scoped to the nominal trip only, and it fails on the commit before it. Thanks.

internal/restapi/vehicles_helper.go
```diff
@@ -301,3 +301,29 @@ func projectVehicleDistanceOnShape(
...
+// response still carries the nominal trip and its route, so alerts informed
+// about either trip belong on the entry.
+func (api *RestAPI) vehicleSituationRefs(ctx context.Context, nominalTripID, activeTripID, routeID string, routeByID map[string]gtfsdb.Route) []situationRef {
+	refsForTrip := func(tripID string) []situationRef {
+		route, indexed := routeByID[routeID]
+		if !indexed {
+			return api.situationRefsForTrip(ctx, tripID)
+		}
+
+		return situationRefsFromAlerts(api.GtfsManager.GetAlertsByIDs(tripID, routeID, route.AgencyID), route.AgencyID)
```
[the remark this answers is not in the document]
omlahore: Yeah, that's a real gap. I fixed it in 3856c9b0, so each trip now gets looked up with its own route. The handler passes the active trip's route down since it already resolves it for the references, so there aren't any extra queries. No block in the RABA fixture spans two routes, so I put the coverage in a unit test on the helper and it fails without the change.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Thanks for working through the last round. Everything I asked for is in:

- Agencies now come from the batched `routeByID`, and a route fetched for an active trip is cached back into it, so there's no per-vehicle N+1.
- The nominal trip's alerts are kept under interlining, looked up against their own route, deduplicated, and pinned by a test.
- The description now frames #1429 correctly relative to #1421.

This also lines up with Java. `BeanFactoryV2.getTripStatus` adds each situation ID to `tripStatus.situationIds` and the same bean to references, so references hold exactly the referenced situations. The nominal-trip union is the one intentional difference, and I asked for it.

A heads-up for whoever lands second between this and #1425: once both are in, `collectAlertsForRoutes`, `collectAlertsForStops`, and `deduplicateAlerts` have no production callers and should be deleted.

PR #1431, opened 2026-09-11
Restore the DUPLICATED base trip lookup in trips-for-route

omlahore:
The #1286 merge dropped the resolveDuplicatedBaseTrip call from trips-for-route. The handler went back to an inline lookup, and that undid two earlier fixes.

#1316 made it keep a DUPLICATED trip ID when the stripped ID doesn't resolve either. The inline copy adopts the stripped ID before checking it, so a feed ID like no-such-trip.00060 comes back as no-such-trip.

#1378 (for #1366) made a real database error return 500 instead of being swallowed. The inline copy only logs it and carries on.

This calls the helper again, the same way the handler did before the merge. The helper was still there, only a test was calling it.

I added a test for the first case. On main it fails with activeTripId tfr-agency_no-such-trip instead of tfr-agency_no-such-trip.00060, and it passes with this. The existing DuplicatedTripLookupFailures test passes on main too, so it doesn't catch the missing call. The rest of internal/restapi passes with make test's tags.

cc @ARCoder181105

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of duplicated trips when the associated base trip cannot be resolved.
  - Unresolved duplicated trips remain visible with their original feed trip ID.
  - The route trips endpoint continues returning successfully when a duplicated trip cannot be matched.
  - Trip status information consistently preserves the unresolved active trip ID.
  - Related trip date and overlap processing is skipped when no matching base trip is available.


aaronbrethorst:
### Code review

Found 1 issue:

1. The fix commit's subject line is 58 characters, and the commit has no body. Since this PR is merged with a merge commit, the commit lands on main as-is, and none of the description's reasoning (which fixes #1286 undid, and why) is in the commit. (CONTRIBUTING.md says "Limit the subject line to 50 characters", "Use the body to explain what and why", and "It should be possible to delete the PR and have all necessary context for each change live in the commit itself.")

https://github.com/OneBusAway/maglev/commit/8c3960ad6ff83814ded9e2ac0c30cd5d1501937b

https://github.com/OneBusAway/maglev/blob/9face1335db12a3ac2fd206a95a3b15cf69efa66/CONTRIBUTING.md#L16-L25

omlahore:
I reworded the commit to "Restore the DUPLICATED base trip lookup" and put the reasoning in the body. The diff itself didn't change.

aaronbrethorst (dismissed):
Thanks for tracking this down. The code is right. I compared it against the pre-#1286 handler, and the helper call, the 500 on a real lookup error, and the `baseTrip.ID != ""` check are all back. #1286's past-midnight service-date logic is untouched. Your new `DuplicatedTripKeepsUnresolvedFeedID` test fails against main's handler, and trips-for-route on King County Metro data is byte-identical to main.

Two small things before merge:

1. **Commit message.** `8c3960ad` has a 58-character subject and no body. CONTRIBUTING asks for a subject of 50 characters or fewer and a body that explains what and why. Since this is restoring something a merge silently dropped, the body is exactly where that story should live (#1286 dropped the call and undid #1316 and #1378). No human has reviewed this yet, so rewording it is fine.

2. **Make the lookup-failure test catch the 500 regression.** As you noted, `TestTripsForRouteHandler_DuplicatedTripLookupFailures` passes on main too. It only gets a 500 because `includeSchedule` defaults to true and a later schedule lookup hits the injected error. With `includeSchedule=false` added to its URLs, both database-error subtests fail on main and pass here. That's the test that would have caught #1286, so it's worth having.

Happy to re-review right away.

omlahore:
The reword is already in 50c9673f. I added includeSchedule=false to both database-error subtests in ffe7869e, and now they fail on the old handler 🙌

PR #1437, opened 2026-09-12
Skip the bulk-insert time budget on Windows

omlahore:
TestBulkInsertPerformance asserts that inserting 10k stop_times finishes in under five seconds. On #1420 the Windows runner took 5.0675s and failed the build over that, which is a 1.4% overshoot on a wall-clock budget on a shared runner.

cmd/api/lock_safety_test.go and internal/gtfs/reload_test.go already skip on Windows, and both say the same thing, that its SQLite file I/O is too slow for a CI timeout. I gated only the timing assertion here, so the insert and the row-count check still run on Windows.

make test and make lint pass.

## Summary by CodeRabbit

* **Tests**
  * Adjusted bulk-insert performance checks so Windows validates insertion results without enforcing the five-second timing threshold.

### DIFF
--- a/gtfsdb/bulk_insert_test.go
+++ b/gtfsdb/bulk_insert_test.go
@@ -3,6 +3,7 @@ package gtfsdb
 import (
 	"context"
 	"database/sql"
+	"runtime"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Thanks for this. Gating only the timing assertion is the right call: the insert and the row-count check still run on Windows, and the CI history backs you up. This test's Windows timings jump between roughly 2s and 4s across main commits that don't touch `gtfsdb`, so the 5.07s failure on #1420 was the runner, not a regression.

One small thing for a follow-up if you feel like it: the new comment attributes the slowness to SQLite file I/O, but this test runs against `:memory:`. The real cause is that Windows CI uses the pure-Go driver (`-tags=purego`), which is far slower than CGO for bulk inserts. Not worth holding the merge over.

PR #1451, opened 2026-09-15
perf: optimize VehiclesForAgencyID with realtime index (#762)

### LINKED ISSUE #762: Optimize VehiclesForAgencyID with route-based realtime index

DsThakurRawat:
## Summary

Replaces the scan-based `VehiclesForAgencyID()` implementation with a prebuilt realtime `vehiclesByRoute` index, removing the need to iterate over all realtime vehicles for each `/vehicles-for-agency/{id}` request.

Closes #762. 
*(Re-implements the approach from #763 on top of the new `mergedRealtime` snapshot architecture)*

## Changes

    - Updates `VehiclesForAgencyID()` to union prebuilt route buckets instead of scanning `realTimeVehicles`.
- `internal/gtfs/realtime.go`
    - Adds `vehiclesByRoute map[string][]gtfs.Vehicle` to `mergedRealtime`.
    - Builds `vehiclesByRoute` inside `rebuildMergedRealtimeLocked()`.
    - Indexes only vehicles with `Trip != nil` and non-empty `Trip.ID.RouteID`, matching the old filtering behavior.
    - Updates mock vehicle helpers (`MockAddVehicle`, `MockAddVehicleWithOptions`, `MockAddDuplicatedVehicle`) to keep `vehiclesByRoute` in sync during testing.
- `internal/gtfs/gtfs_manager_test.go`
    - Adds manager-level unit test `TestManager_RebuildMergedRealtime_VehiclesByRoute` covering route bucket union, nil/empty handling, and stale-feed clearing.

## Note on Vehicle Ordering

The returned vehicle slice now reflects map-iteration order across the agency's routes rather than the global `allVehicles` insertion order. This adheres to the API specification (the endpoint returns a vehicle array without ordering guarantees, and handlers build references via keyed maps), but is called out here for reviewers.

## Why

`routesByAgencyID` is already available from static GTFS data. Pairing it with a realtime `route -> vehicles` index moves the filtering work out of the request path and into the existing realtime rebuild path, which aligns with how `alertIdx` and other realtime structures are maintained.

## Validation Passed

- `make test` (all packages passing)
- `make lint`
- `make fmt`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vehicle retrieval for agency views by using route-based vehicle associations.
  * Improved consistency when displaying vehicles associated with routes, including data from multiple realtime feeds.
  * Corrected handling of vehicles without trips or route IDs so they are excluded from route-based results.
  * Ensured outdated route vehicle data is removed when realtime information is rebuilt.

* **Performance**
  * Reduced unnecessary filtering during vehicle retrieval, improving efficiency for agency and route lookups.


Ahmedhossamdev (commented):
Looks correct - the refactor preserves the vehicle set that VehiclesForAgencyID returns and reuses the same snapshot pattern as duplicatedVehicleByRoute.

 Two things worth addressing before merge: 
 
 (a) the PR description says "adds robust manager-level tests" but the commit doesn't add any. 
 
 (b) the new vehiclesByRoute build path in rebuildMergedRealtimeLocked has no direct test - the existing handler tests exercise the mock path, not the rebuild path. 
 
 A couple of small nits inline. MockAddDuplicatedVehicle was not updated and now diverges from the production rebuild; not a blocker because nothing currently depends on it, but worth fixing while you're here.

internal/gtfs/gtfs_manager.go
```diff
@@ -613,14 +613,12 @@ func (manager *Manager) VehiclesForAgencyID(ctx context.Context, agencyID string
 		routeIDs[route.ID] = true
 	}
 
-	// Step 2: Acquire real-time lock independently to read vehicles.
-	rtVehicles := manager.GetRealTimeVehicles()
+	// Step 2: Acquire real-time snapshot independently to read vehicles.
+	merged := manager.mergedRealtime()
```
Ahmedhossamdev: The "Step 2" wording is leftover from the two-step version. There's just one step now.

```suggestion
	// Read vehicles for those routes from the current realtime snapshot.
	merged := manager.mergedRealtime()
```

internal/gtfs/gtfs_manager.go
```diff
@@ -613,14 +613,12 @@ func (manager *Manager) VehiclesForAgencyID(ctx context.Context, agencyID string
...
 
 	var vehicles []gtfs.Vehicle
-	for _, v := range rtVehicles {
-		if v.Trip != nil && routeIDs[v.Trip.ID.RouteID] {
-			vehicles = append(vehicles, v)
-		}
+	for routeID := range routeIDs {
+		vehicles = append(vehicles, merged.vehiclesByRoute[routeID]...)
 	}
```
Ahmedhossamdev, line 622: Worth calling out in the PR description that this changes the returned slice's order. The old scan preserved `allVehicles` order; iterating `routeIDs` gives map-iteration order across routes. It doesn't violate the spec and the handler already builds references from maps, but a reader assuming the previous order would be surprised.

internal/gtfs/realtime.go
```diff
@@ -701,10 +707,14 @@ func (manager *Manager) rebuildMergedRealtimeLocked() {
...
 	vehicleLookupByVehicle := make(map[string]int, len(allVehicles))
 	duplicatedVehicleByRoute := make(map[string][]gtfs.Vehicle)
+	vehiclesByRoute := make(map[string][]gtfs.Vehicle)
 	for i, vehicle := range allVehicles {
 		if vehicle.Trip != nil && vehicle.Trip.ID.ID != "" {
 			vehicleLookupByTrip[vehicle.Trip.ID.ID] = i
 		}
+		if vehicle.Trip != nil && vehicle.Trip.ID.RouteID != "" {
+			vehiclesByRoute[vehicle.Trip.ID.RouteID] = append(vehiclesByRoute[vehicle.Trip.ID.RouteID], vehicle)
+		}
```
Ahmedhossamdev, line 717: There's no direct test for this new index. `VehiclesForAgencyID` is covered end-to-end via `MockAddVehicleWithOptions`, but that path bypasses `rebuildMergedRealtimeLocked`. A small manager-level test that seeds two `feedVehicles` entries, calls the rebuild, and asserts `vehiclesByRoute` would cover the new lines (nil-Trip skip, empty-RouteID skip, multi-feed union, stale-feed clear).

internal/gtfs/gtfs_manager_mock.go
```diff
@@ -56,13 +56,21 @@ func (m *Manager) MockAddVehicle(vehicleID, tripID, routeID string) {
...
 	merged.vehicleLookupByVehicle[vehicleID] = idx
 	if tripID != "" {
 		merged.vehicleLookupByTrip[tripID] = idx
 	}
+	if routeID != "" {
+		if merged.vehiclesByRoute == nil {
+			merged.vehiclesByRoute = make(map[string][]gtfs.Vehicle)
+		}
+		merged.vehiclesByRoute[routeID] = append(merged.vehiclesByRoute[routeID], v)
+	}
```
Ahmedhossamdev, line 69: `clone()` at `realtime.go:105-107` already guarantees `vehiclesByRoute` is non-nil, so the guard here is dead code. Same at lines 146-148 in `MockAddVehicleWithOptions`. Drop the guard here (and there) or drop it in `clone()`, not both.

```suggestion
	if routeID != "" {
		merged.vehiclesByRoute[routeID] = append(merged.vehiclesByRoute[routeID], v)
	}
```

internal/gtfs/gtfs_manager_mock.go
```diff
@@ -134,6 +142,13 @@ func (m *Manager) MockAddVehicleWithOptions(vehicleID, tripID, routeID string, o
 	if tripID != "" && !opts.NoTrip {
 		merged.vehicleLookupByTrip[tripID] = idx
 	}
+	if !opts.NoTrip && routeID != "" {
+		if merged.vehiclesByRoute == nil {
+			merged.vehiclesByRoute = make(map[string][]gtfs.Vehicle)
+		}
+		merged.vehiclesByRoute[routeID] = append(merged.vehiclesByRoute[routeID], v)
```
Ahmedhossamdev, line 143: While you're updating the vehicle mocks, `MockAddDuplicatedVehicle` (lines 168-204) needs the same treatment. Production `rebuildMergedRealtimeLocked` files DUPLICATED vehicles into both `duplicatedVehicleByRoute` and `vehiclesByRoute` (the route-bucket append at `realtime.go:715` runs before the DUPLICATED short-circuit at `:721`). This helper only appends to `duplicatedVehicleByRoute`, so a test that added a duplicated vehicle used to see it in `VehiclesForAgencyID` (old code scanned `merged.vehicles`) and now would not. No test currently depends on this, but it's the same class of mock/production drift you're fixing here.

internal/gtfs/gtfs_manager.go
DsThakurRawat: Updated comment wording in 1e2e934.

internal/gtfs/gtfs_manager.go
DsThakurRawat: Good call  added a note about vehicle slice ordering across routes in the PR description.

internal/gtfs/realtime.go
DsThakurRawat: Added TestManager_RebuildMergedRealtime_VehiclesByRoute in gtfs_manager_test.go covering nil-Trip, empty-RouteID, multi-feed union, and stale-feed removal in 1e2e934.

internal/gtfs/gtfs_manager_mock.go
DsThakurRawat: Dropped the redundant nil checks in both MockAddVehicle and MockAddVehicleWithOptions in 1e2e934.

internal/gtfs/gtfs_manager_mock.go
DsThakurRawat: Updated MockAddDuplicatedVehicle to also populate vehiclesByRoute, keeping it aligned with rebuildMergedRealtimeLocked in 1e2e934.

PR #1452, opened 2026-09-17
Fix trips-for-route layover handling

### LINKED ISSUE #1444: trips-for-route: blocks between trips (layover) are dropped
## Summary

`trips-for-route` drops blocks whose vehicle is between two trips on the route at the query time: its last trip has ended and the next hasn't started. Java returns one entry for each of those blocks.

## Current behaviour

`trips-for-route/1_100275.json?time=1789398000000` (08:00 local; window 07:30–08:10). Java returns these blocks and Maglev omits them:

| Block | Route trips around the query time | Java's entry |
|---|---|---|
| 8230889 | 07:06–07:49, then 08:12–08:57 | `851648161` (the trip that just ended) |
| 8230885 | 07:00–07:42, then 08:06–08:54 | `851648641` (the next trip) |
| 8230894 | 07:12–07:54, then 08:18–09:06 | `851648351` (the next trip) |

On `trips-for-route/1_100001.json` at the same time, block `8222244` (07:17–07:51, then 08:10–08:27) is missing the same way.

Each of these blocks has a trip on the route inside the window, so they pass step 2 of the spec. The likely cause, not yet confirmed in code:
- `GetActiveTripInBlockAtTime` finds no trip "in progress" during the gap, so the block is skipped.
- The `block_layover` fallback only has rows when a trip's last stop is the next trip's first stop, so layovers at different terminals never qualify. This is the same table limitation found while reviewing #1362.

## Spec requirement

[trips-for-route](https://github.com/OneBusAway/maglev/wiki/trips-for-route):

> The response contains one entry per active block whose scheduled trips overlap the query window

> **The result set is not open-ended.** … Client code (e.g. Wayfinder's vehicle-marker sweep, Android's trip-extrapolation layer) treats an entry's presence in `data.list` as "this vehicle is currently active" and removes or ages out anything missing.

So a vehicle laying over at a terminal disappears from the map for the length of its layover.

## Acceptance criteria

- [ ] A block with a route trip in the window is returned while its vehicle is between trips, whether or not the two trips share a terminal stop
- [ ] The active trip chosen during a gap matches Java's rule (`TripStatusBeanServiceImpl` / block location). The table above shows Java picking the previous trip in one case and the next trip in others, so confirm the rule before implementing.
- [ ] The three route 8 blocks and block `8222244` above are returned
- [ ] A test covers a layover at the same terminal and one at different terminals

## Scope note

The opposite problem, blocks with no route trips leaking in, is #1443.

Spec review: #1050

soumajitgh:
## Summary

- retain active blocks while vehicles are between scheduled trips
- use the existing Java-compatible block snapshot to select the previous or next trip
- handle layovers at both shared and different terminal stops
- preserve overnight, interlining, and reused-block behavior

## Testing

- `go test -tags 'sqlite_fts5 sqlite_math_functions' ./internal/restapi -count=1`

Closes #1444

## Summary by CodeRabbit

- **Bug Fixes**
  - Route trip results now continue to show blocks during layovers between consecutive trips.
  - Improved trip selection during layovers, including same-terminal and different-terminal transitions.
  - Route results now more accurately identify active trips and service days across multiple blocks.
  - Prevented unrelated blocks from appearing when trips share indexing information.
  - Improved handling of reused block identifiers across service days and route evidence.
  - Interlined trips can now correctly appear as the active portion of a requested route block.


omlahore:
Both new tests fail on main, 0 entries in one and 1 in the other, and they pass here, so I think the layover part is right. I also ran main's whole trips-for-route file against this branch and nothing else moved.

internal/restapi/trips_for_route_handler.go
```diff
@@ -166,11 +173,20 @@ func (api *RestAPI) tripsForRouteHandler(w http.ResponseWriter, r *http.Request)
...
 				}
 			}
 		}
+
+		// A block discovered through today's queried-route trip can be running
+		// an interlined trip from yesterday's service (including across agency
+		// timezones). Keep those blocks eligible for previous-day resolution.
+		// The reverse is intentionally not done: a yesterday-only candidate must
+		// not resolve an unrelated reuse of its block ID in today's service.
+		for blockID := range currentLinkedBlocks {
```
burma-shave: This copies every current-day linked block into the previous-day candidate set. That preserves cross-service interlining, but it also lets a today-only queried-route block resolve an unrelated previous-day reuse of the same `block_id`. In an asymmetric case where today has route R in block B but yesterday's B is active only on another route, previous-day resolution can emit yesterday's other-route active trip, and `buildBlockTripForRoute` (keyed only by block ID across both service-day sets) can associate it with today's queried-route trip. Please keep the previous-day candidates tied to service-day evidence for the queried route, or add a regression test for that asymmetric reused-block scenario.

internal/restapi/trips_for_route_handler.go
```diff
@@ -575,6 +565,145 @@ func (api *RestAPI) tripsForRouteHandler(w http.ResponseWriter, r *http.Request)
...
+// day's candidate blocks.
+func (api *RestAPI) tripsForRouteBlockSpans(
+	ctx context.Context,
+	sd tripsForRouteServiceDay,
+) ([]gtfsdb.GetTripSpansForBlocksRow, error) {
+	return queryInBatchesReserving(ctx, sd.blockIDs, len(sd.serviceIDs),
+		func(ctx context.Context, batch []string) ([]gtfsdb.GetTripSpansForBlocksRow, error) {
+			nullableBatch := make([]sql.NullString, len(batch))
+			for i, blockID := range batch {
+				nullableBatch[i] = sql.NullString{String: blockID, Valid: true}
```
burma-shave: This duplicates the nullable-string helper already used by the near-identical `trips_for_location_handler.go` conversion. Please use `nulls.String(blockID)` here and keep the existing helper/import instead of hand-rolling `sql.NullString`.

burma-shave (changes requested):
Requesting changes based on the reused-block service-day edge case below. The nullable-string helper note is minor cleanup, but included while this area is being touched.

soumajitgh:
Resolved the merge conflicts with the latest main. All REST API and database tests pass.

burma-shave (approved):

PR #1453, opened 2026-09-17
Fix trips-for-route block index leakage

### LINKED ISSUE #1443: trips-for-route: blocks with no trips on the route leak in via block_trip_index
## Summary

`trips-for-route` returns blocks that have no trips on the requested route. Blocks are selected through `block_trip_index`, and one index can mix blocks from many routes. `GetBlocksForBlockTripIndexIDs` then returns every block in those indices that has *any* trip overlapping the time window, whatever route that trip is on.

## Current behaviour

`trips-for-route/1_100001.json?time=1789398000000` (08:00 local; window 07:30–08:10). Java returns 7 entries and Maglev returns 15. Eight of Maglev's entries come from blocks with **zero** trips on route 1 that day:

| Block | Trips on route `100001` | Active trip Maglev reports |
|---|---|---|
| 8222240 | 0 | `788759691`, route 13, 07:36–08:09 |
| 8222275 | 0 | `788759211`, route 13, 07:46–08:03 |
| 8230854 | 0 | `840044161`, route 62, 07:53–09:05 |
| 8230855 | 0 | `800855061`, route 62 |
| 8230860 | 0 | `800855001`, route 62 |
| 8230861 | 0 | `800855081`, route 62 |
| 8230925 | 0 | `721367971`, route 62, 07:15–08:20 |
| 8230973 | 0 | `800855091`, route 62 |
| 8232336 | 0 | `801594011`, route 70, 07:29–08:07 |

The cause, reproduced directly against the Maglev database:
- `GetBlockTripIndexIDsForRoute('100001', active services)` returns 3 index IDs.
- One of them, index `520` (key `151056|1530`), contains 54 blocks serving routes 1, 2, 13, 62, 70 and `100219`.
- `GetBlocksForBlockTripIndexIDs` filters on `t.max_departure_time`/`t.min_arrival_time` for *any* trip in those indices, so every block above is selected (18 blocks in total).

Route 8 (`1_100275`) shows the same pattern: block `8230879`, which only runs route 48, appears in Maglev but not Java.

## Spec requirement

[trips-for-route](https://github.com/OneBusAway/maglev/wiki/trips-for-route), Main Success Scenario step 2:

> The server resolves the route ID and looks up all vehicle blocks that have at least one trip on that route active within the window spanning 30 minutes before and 10 minutes after the query time.

In Java, [`BlockStatusServiceImpl.java#L182-L188`](https://github.com/OneBusAway/onebusaway-application-modules/blob/095bf1ac3aeb7009b9f78e7f1cb4c65bd38b988a/onebusaway-transit-data-federation/src/main/java/org/onebusaway/transit_data_federation/impl/blocks/BlockStatusServiceImpl.java#L182-L188) and Java's `BlockTripIndex` only group blocks that share the same trip sequence, so the time check runs against trips on the route.

## Acceptance criteria

- [ ] A block is selected only if it has a trip on the requested route that's active in the window
- [ ] The example above returns no entries for the eight blocks listed
- [ ] Interlined blocks, whose route trip is in the window but whose *active* trip is on another route, are still returned. That's a documented, intentional behaviour.
- [ ] A test with two routes sharing one `block_trip_index` asserts the other route's blocks don't leak in

## Scope note

The opposite problem, blocks Java returns and Maglev drops, has a different cause and is tracked in #1444. Check whether `GetTripsByBlockTripIndexIDs` in other handlers has the same leak.

Spec review: #1050

soumajitgh:
## Summary

- require a requested-route trip to overlap the query window before selecting its block
- preserve interlined blocks whose active trip belongs to another route
- add regression coverage for unrelated routes sharing one `block_trip_index`

A block is excluded when all of its trips on the requested route fall outside the query window, even if another trip in that block is active. This matches Java's route-derived index intervals.

The existing cross-day fixture now keeps its requested-route trip inside the documented window, so it continues testing service-day resolution without depending on the leaked-block behavior.

## Testing

- `make test`
- `go fmt ./...`

Closes #1443

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed route-based trip searches returning unrelated blocks that shared the same block-trip index.
  - Trips-for-route results now include only blocks containing trips for the requested route, across current and previous service days.

- **Tests**
  - Added coverage for shared block-trip indexes and cross-day block reuse scenarios.


soumajitgh:
@coderabbitai review

omlahore:
I checked main's test file against this branch and TestTripsForRouteHandler_ReusedBlockIDAcrossDays comes back with 0 entries, which I guess is why tfr-yest-a moved into the window. So a block gets dropped now when its queried-route trip sits outside the window, even if another trip in it is active. I'd guess that's what Java does since the index intervals come from the route's trips, but it'd be worth a line in the description.

PR #1454, opened 2026-09-17
Fix arrival departure stop sequence

### LINKED ISSUE #1440: arrival-and-departure-for-stop: stopSequence is matched as GTFS stop_sequence, not position
## Summary

`arrival-and-departure-for-stop` treats the `stopSequence` parameter as the GTFS `stop_sequence` value. The wiki, OpenAPI, and Java all define it as the 0-based position of the stop in the trip's stop list, with a search that expands outward from that position. On any feed whose `stop_sequence` isn't exactly 0-based and contiguous, Maglev returns 404 for requests Java answers.

## Current behaviour

Trip `1_1137791` visits stop `1_42010` twice, at GTFS `stop_sequence` 1 and 8 (positions 0 and 7 in an 8-stop trip).

`/api/where/arrival-and-departure-for-stop/1_42010.json?tripId=1_1137791&serviceDate=1789369200000&stopSequence=N`

| `stopSequence` | Java: code, `entry.stopSequence` | Maglev: code, `entry.stopSequence` |
|---|---|---|
| 0 | 200, 0 | **404** |
| 1 | 200, 0 | 200, 0 |
| 7 | 200, 7 | **404** |
| 8 | 200, 7 | 200, 7 |

Maglev already *emits* `entry.stopSequence` as the 0-based position. It just doesn't accept that same value as input, so a client can't round-trip the value from a previous response.

Related observation, not verified as a defect yet: with no `stopSequence`, Java picked the second visit (7) and Maglev picked the first (0). The wiki says the visit closest to the query time wins. Worth checking while fixing this.

## Spec requirement

[arrival-and-departure-for-stop](https://github.com/OneBusAway/maglev/wiki/arrival-and-departure-for-stop), Request Parameters:

> **`stopSequence`** — Optional. The 0-indexed position of the stop in the trip's stop time list as maintained internally by OBA. This is **not** the GTFS `stop_sequence` value. When provided, the server starts its search at this position and expands outward one step at a time; a mismatch of a few positions (caused by stops being added or removed from a trip) is tolerated.

Java: [`ArrivalAndDepartureServiceImpl.java#L1124`](https://github.com/OneBusAway/onebusaway-application-modules/blob/095bf1ac3aeb7009b9f78e7f1cb4c65bd38b988a/onebusaway-transit-data-federation/src/main/java/org/onebusaway/transit_data_federation/impl/ArrivalAndDepartureServiceImpl.java#L1124)

## Acceptance criteria

- [ ] `stopSequence` is interpreted as the 0-based position in the trip's stop times
- [ ] The lookup starts at that position and expands outward until the stop is found, so a small offset still resolves
- [ ] `stopSequence=0` and `stopSequence=7` in the example above return the first and second visit
- [ ] A test covers a loop trip, a feed whose `stop_sequence` starts at 1, and an offset of one or two positions

## Scope note

This is the request parameter only. GTFS-RT StopTimeUpdate matching by `stop_sequence` is #1414 / #1415, which is correctly about the GTFS value.

Spec review: #1264

Eldax23:
## Summary

Fixes the `stopSequence` query parameter on `arrival-and-departure-for-stop`
to match the OBA API spec: a 0-indexed position in the trip's stop list,
not the raw GTFS `stop_sequence` column value. The old exact-match lookup
against that column only worked on feeds where `stop_sequence` is 0-based
and contiguous, so it 404'd on feeds starting at 1 (e.g. King County Metro)
and on loop trips where a stop is visited more than once.

Fixes #1440

## What changed

- `findStopTimeForTripStop` fetches a trip's full ordered `stop_times` once
  and treats the requested index as a position in that list. When the exact
  position doesn't hold the target stop, it expands outward one step at a
  time (checking the lower index before the higher), matching Java's
  `getBlockStopTime`. This tolerates position drift from stops added or
  removed since a client last saw the trip's layout.
- The no-`stopSequence` case now picks the occurrence closest to the query
  time among multiple visits of a stop, matching Java's existing tie-break,
  instead of always returning the first occurrence.

## Pre-existing test changes (flagging for review)

Two existing tests encoded assumptions that only held under the old,
incorrect exact-match behavior — both are addressed in a separate commit
from the fix itself:

- `TestArrivalAndDepartureForStopHandler_LoopRouteStopSequence` passed raw
  GTFS `stop_sequence` values as the query parameter; updated to pass the
  correct 0-based positions.
- `TestArrivalAndDepartureForStopHandlerWithWrongStopSequence` asserted
  that an offset of +100 from a valid position would 404. Once position
  drift is tolerated by design, the outward search can legitimately walk
  back to a valid position given enough distance — there's no upper bound
  on search distance, matching Java's own unbounded expansion. I rewrote
  it to request a stop ID absent from the trip entirely, which still 404s
  regardless of position. Flagging this as a judgment call in case
  maintainers want a different approach (e.g. capping search distance).

## Testing

Added three tests covering the acceptance criteria from #1440: a feed
whose `stop_sequence` starts at 1, a requested position offset by one or
two from a stop's actual position, and the no-`stopSequence` path on a
loop trip.

Ran `go vet` under both `sqlite_fts5 sqlite_math_functions` and `purego`
build tags, `make test`, and `go fmt` per CONTRIBUTING.md.

## Size note

This PR is larger than the suggested 200-line guideline — the fix itself
is small, but the new tests (~280 lines) cover three distinct scenarios
named explicitly in the issue's acceptance criteria and didn't seem
separable from the fix without losing that traceability. Happy to split
further if preferred.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved arrival and departure lookups for stops on trips, including loop routes and feeds with different stop-sequence numbering.
  * Stop requests now resolve more reliably when sequence positions are slightly offset, including zero-based positions.
  * When no sequence is provided, the closest visit to the requested time is selected.
  * Invalid stop requests return a not-found response instead of matching an unrelated stop.
  * Improved handling of repeated stops and trips with varying sequence conventions.


internal/restapi/arrival_and_departure_for_stop_handler.go
```diff
@@ -296,6 +252,30 @@ func (api *RestAPI) arrivalAndDepartureForStopHandler(w http.ResponseWriter, r *
...
+	}
+	targetStopTime := struct {
+		ArrivalTime   int64
+		DepartureTime int64
+		StopSequence  int64
+		StopHeadsign  string
+	}{
+		ArrivalTime:   matchedStopTime.ArrivalTime,
+		DepartureTime: matchedStopTime.DepartureTime,
+		StopSequence:  matchedStopTime.StopSequence,
```
burma-shave, line 276: This still carries the raw GTFS `stop_sequence` forward as the only sequence value, and the response later passes `int(targetStopTime.StopSequence)-1` to `NewArrivalAndDeparture`. That means a trip with raw sequences like `10, 20, 30` resolves the requested position correctly, but returns `entry.stopSequence = 19` for the second stop instead of the required zero-based position `1`.

Please return the matched stop-time row together with its zero-based index in `orderedStopTimes`, use that index for the API response `entry.stopSequence`, and keep `matchedStopTime.StopSequence` only for logic that really needs the raw GTFS value, such as GTFS-RT matching.

internal/restapi/arrival_and_departure_for_stop_handler.go
```diff
@@ -776,3 +754,80 @@ func (api *RestAPI) getPredictedTimes(
...
+	return best, found
+}
+
+func findStopTimeByPosition(stopTimes []gtfsdb.StopTime, stopCode string, requestedIndex int) (gtfsdb.StopTime, bool) {
+	n := len(stopTimes)
+	idx := requestedIndex
+
+	// Distance to the farther edge of the slice , past this, there's
+	// nothing left on either side to check.
+	maxDist := max(idx, n-1-idx)
```
burma-shave, line 809: `maxDist` is based on the client-supplied `stopSequence`, so an out-of-range value makes this loop run proportional to the request parameter rather than the trip length. For example, `stopSequence=50000000` against a one-stop trip still iterates about 50 million times before returning not found.

Please bound the search work to the stop-time slice size, or reject/clamp unreasonable positions before entering this function, so malformed requests cannot pin a handler goroutine.

burma-shave (changes requested):
Requesting changes based on the verified `arrival-and-departure-for-stop` issues below.

Additional cleanup finding that cannot be attached inline because the files are not in this PR diff: `GetTargetStopTimeWithTotalStops` and `GetTargetStopTimeWithTotalStopsBySequence` are now unused after this handler refactor, but still remain in `gtfsdb/query.sql`, generated `gtfsdb/query.sql.go`, and prepared statements in `gtfsdb/db.go`. Please remove the dead sqlc queries and regenerate models if they are no longer needed.

Eldax23:
Thank you for the review!

- `entry.stopSequence` now reports the matched row's actual 0-based index in `orderedStopTimes`, not `rawStopSequence - 1`. `matchedStopTime.StopSequence` is kept only where the raw GTFS value is actually needed (GTFS-RT `StopTimeUpdate` matching).
- `findStopTimeByPosition` no longer expands outward from the requested index — it's a single O(n) pass over the trip's own `stop_times`, so cost is bounded by trip length regardless of how far out of range `stopSequence` is.
- Also removed the now-unused `GetTargetStopTimeWithTotalStops`/`...BySequence` queries from `gtfsdb/query.sql` and regenerated via `make models`, per your note.

internal/restapi/arrival_and_departure_for_stop_handler.go
```diff
@@ -773,3 +751,103 @@ func (api *RestAPI) getPredictedTimes(
...
+// over stopTimes, so a wildly out-of-range requestedIndex costs no more
+// than a valid one.
+func findStopTimeByPosition(stopTimes []gtfsdb.StopTime, stopCode string, requestedIndex int) (gtfsdb.StopTime, int, bool) {
+	n := len(stopTimes)
+	if n == 0 {
+		return gtfsdb.StopTime{}, 0, false
+	}
+	idx := requestedIndex
+
+	if idx < 0 {
```
burma-shave, line 814: Could we add explicit coverage for both clamp branches here? This PR adds behavior for negative and too-large `stopSequence` values, but the current tests cover the matching/tie cases without directly exercising `idx < 0` or `idx > n-1`. A regression that clamps to the wrong endpoint, or panics on an out-of-range endpoint, would not be caught.

Comment ID: CC-01

internal/restapi/arrival_and_departure_for_stop_handler.go
```diff
@@ -773,3 +751,103 @@ func (api *RestAPI) getPredictedTimes(
...
+		}
+	}
+
+	if !found {
+		return gtfsdb.StopTime{}, 0, false
+	}
+	return stopTimes[bestIdx], bestIdx, true
+}
+
+func absInt(v int) int {
```
burma-shave, line 848: Minor cleanup: this duplicates the existing `absInt64` helper in `internal/restapi/trip_updates_helper.go`. Since this package already has an absolute-value helper, could we reuse or consolidate it instead of adding a second near-identical implementation?

Comment ID: CC-02

gtfsdb/query.sql
```diff
@@ -1467,42 +1467,6 @@ SELECT CAST((
 
 -- Optimized queries using SQLite window functions
```
burma-shave, line 1468: With the target-stop-time window queries removed, this section header now applies only to `GetBlockTripSequence`, and sqlc carries it into that query's generated doc comment. Could we remove/reword this header, or move it, so the generated docs do not read as if there are multiple optimized queries in this section?

Comment ID: CC-03

burma-shave (changes requested):
re-review found a few remaining test/cleanup items after the latest updates. The prior stopSequence/index behavior and dead sqlc-query cleanup appear addressed; the comments below are the remaining actionable follow-ups.

Eldax23:
Hey , so thanks again for your review!

Addressed all three:

- Both clamp branches (`idx < 0`, `idx > n-1`) now have explicit test coverage — `stopSequence=-1` clamps to position 0, `stopSequence=<stop count>` clamps to the last position, each asserting the correct stop is returned.
- Removed the duplicate `absInt` — `findStopTimeByPosition` now reuses the existing `absInt64` from `trip_updates_helper.go`.
- Removed the stale `-- Optimized queries using SQLite window functions` header; `GetBlockTripSequence`'s own doc comment already stands on its own without it.

internal/restapi/arrival_and_departure_for_stop_handler.go
```diff
@@ -773,3 +751,96 @@ func (api *RestAPI) getPredictedTimes(
...
+// If requestedIndex is non-nil, it is treated as the 0-based position of the
+// stop within the trip's stop list (per the OBA API spec — NOT the raw GTFS
+// stop_sequence value), We return whichever visit sits closest to that position, so a slightly
+// stale position (a stop added/removed since the client last saw this
+// trip) still resolves instead of 404ing.
+func findStopTimeForTripStop(stopTimes []gtfsdb.StopTime, stopCode string, requestedIndex *int, queryOffset int64) (gtfsdb.StopTime, int, bool) {
+	// when no position is given: a stop can appear twice on a loop trip, so pick
+	// whichever visit's arrival/departure is closest to the query time
+	// (matches Java's behavior , not just "first match").
+	if requestedIndex == nil {
```
Ahmedhossamdev: Java's `getBlockStopTime` only takes the positional path when `stopSequence > -1`, anything negative goes to the closest-by-time branch (`-1` is the query bean default). On a loop trip at the end of its run, `stopSequence=-1` gets the second visit from Java and the first visit here.

```suggestion
	if requestedIndex == nil || *requestedIndex < 0 {
```

internal/restapi/arrival_and_departure_for_stop_handler.go
```diff
@@ -773,3 +751,96 @@ func (api *RestAPI) getPredictedTimes(
...
+// over stopTimes, so a wildly out-of-range requestedIndex costs no more
+// than a valid one.
+func findStopTimeByPosition(stopTimes []gtfsdb.StopTime, stopCode string, requestedIndex int) (gtfsdb.StopTime, int, bool) {
+	n := len(stopTimes)
+	if n == 0 {
+		return gtfsdb.StopTime{}, 0, false
+	}
+	idx := requestedIndex
+
+	if idx < 0 {
```
Ahmedhossamdev: With negatives handled by the closest-time path, this branch can't be reached anymore and can go. Only the upper clamp is needed.

internal/restapi/arrival_and_departure_for_stop_handler_test.go
```diff
@@ -905,6 +1021,448 @@ func TestArrivalAndDepartureForStop_PositiveUTCOffset_ServiceDateRegression(t *t
...
+	endpoint = fmt.Sprintf(baseEndpoint, utils.FormCombinedID(agencyID, stopCID)) + "&stopSequence=2"
+	resp, model = callAPIHandler[ArrivalAndDepartureResponse](t, api, endpoint)
+	require.Equal(t, http.StatusOK, resp.StatusCode)
+	assert.Equal(t, 2, model.Data.Entry.StopSequence)
+}
+
+// Regression test for findStopTimeByPosition's negative-index clamp:
+// a stopSequence below 0 must clamp to position 0, not panic or select
+// the wrong stop.
+func TestArrivalAndDepartureForStopHandler_NegativeStopSequenceClampsToFirst(t *testing.T) {
```
Ahmedhossamdev: This trip visits stopA once, so it passes whether negatives clamp or fall back to time. A loop trip with a query time near the second visit would pin down the Java behavior. It could also be a row in a table test on `findStopTimeForTripStop`, since the function takes a plain slice and needs none of the agency/route/calendar setup.

internal/restapi/arrival_and_departure_for_stop_handler.go
```diff
@@ -296,6 +252,30 @@ func (api *RestAPI) arrivalAndDepartureForStopHandler(w http.ResponseWriter, r *
...
+	}
+
+	queryOffset := int64(currentTime.Sub(serviceMidnight))
+
+	matchedStopTime, matchedIdx, found := findStopTimeForTripStop(orderedStopTimes, stopCode, params.StopSequence, queryOffset)
+	if !found {
+		api.sendNotFound(w, r)
+		return
+	}
+	targetStopTime := struct {
```
Ahmedhossamdev: This struct is now just a copy of `matchedStopTime` (and `StopHeadsign` is never set). Using `matchedStopTime` directly below would drop it.

Ahmedhossamdev:
Nice work @Eldax23, and thanks for working through the earlier rounds. The positional search matches Java now, including loop trips, 1-based feeds, offsets and the tie-break. One behavior difference is left, plus the test size.

Java only uses the positional search when `stopSequence > -1`, and `-1` is its default for "not set". So a negative value falls back to the closest-by-time visit instead of clamping to position 0. On a loop trip that returns a different visit. Separately, most of the new test code is the same fixture setup repeated six times. Since `findStopTimeByPosition` and `findClosestStopTime` are pure functions, one table test over `[]gtfsdb.StopTime` would cover these cases in a fraction of the lines and bring the PR much closer to the size guideline.

Eldax23:
Thanks for the review. I went through the points and made the changes.

- I initially missed the Java spec detail around the stop sequence behavior, so I corrected that and added the relevant coverage.

- I consolidated the repeated stop-sequence resolution cases into a table-driven test against the pure resolver functions, covering out-of-range clamping, tie-breaking, and closest-by-time behavior.

- I left `TestArrivalAndDepartureForStopHandler_StopSequenceStartsAtOne` unchanged because it specifically verifies the API's 0-based `stopSequence` against GTFS data whose `stop_sequence` starts at 1.

- I also kept `TestArrivalAndDepartureForStopHandler_NoStopSequence_PicksClosestToQueryTime` because the table test covers the resolver logic directly, while this test verifies the HTTP behavior when `stopSequence` is omitted and ensures the handler parses that case correctly.

Let me know if you think either of these HTTP-level tests would still be redundant.

PR #1455, opened 2026-09-17
fix: preserve absent arrival prediction values

### LINKED ISSUE #1441: arrival-and-departure-for-stop: predicted times default to scheduled, lastUpdateTime missing
## Summary

With no real-time data for a trip, `arrival-and-departure-for-stop` sets `predictedArrivalTime` and `predictedDepartureTime` to the scheduled times. Java and the wiki return `0`. `lastUpdateTime` is also missing from arrival entries on both `arrival-and-departure-for-stop` and `arrivals-and-departures-for-stop`.

All comparisons below ran on 2026-09-14 against Java OneBusAway 2.7.1 and Maglev `main` at `d89b6161`. Both loaded the same gtfstidy-processed King County Metro feed (`FAL26-161.0`), so data differences are ruled out. (No GTFS-RT feed is configured on either server.)

## Current behaviour

`arrival-and-departure-for-stop/1_75403.json?tripId=<first arrival's trip>&serviceDate=1789369200000`:

| | `predicted` | `scheduledArrivalTime` | `predictedArrivalTime` | `predictedDepartureTime` | `lastUpdateTime` present |
|---|---|---|---|---|---|
| Java | false | 1789397820000 | **0** | **0** | yes (`0`) |
| Maglev | false | 1789397820000 | **1789397820000** | **1789397820000** | **no** |

The plural endpoint (`arrivals-and-departures-for-stop/1_75403.json?time=1789398000000`) already returns `predictedArrivalTime: 0` for the same arrival, so the two endpoints disagree with each other too. It's also missing `lastUpdateTime`.

A client checking `predictedArrivalTime > 0`, which is the common OBA idiom, will read a schedule-only arrival on the singular endpoint as a real-time prediction.

## Spec requirement

[arrival-and-departure-for-stop](https://github.com/OneBusAway/maglev/wiki/arrival-and-departure-for-stop):

> `predicted` is `false`. `predictedArrivalTime` and `predictedDepartureTime` are `0` (not absent).

> **`data.entry.lastUpdateTime`** — Unix milliseconds of the most recent real-time position update for this vehicle. `0` when no real-time data is available.

## Acceptance criteria

- [ ] With no real-time data, `predictedArrivalTime` and `predictedDepartureTime` are `0` on `arrival-and-departure-for-stop`
- [ ] Arrival entries on both endpoints include `lastUpdateTime`, `0` when there's no real-time data
- [ ] Real-time predictions are unchanged
- [ ] Tests assert both fields for a schedule-only arrival on both endpoints

Spec reviews: #1264, #1263

soumajitgh:
## Summary

- return `0` for predicted arrival and departure times when no real-time prediction exists
- always include `lastUpdateTime`, using `0` when real-time data is unavailable
- add schedule-only regression coverage for both arrival endpoints

## Testing

- `go test -tags "sqlite_fts5 sqlite_math_functions" ./internal/models ./internal/restapi`

Closes #1441

## Summary by CodeRabbit

- **Bug Fixes**
  - When real-time data is unavailable, predicted arrival and departure times now remain empty rather than displaying scheduled times as predictions.
  - Trips without real-time updates are correctly marked as not predicted.
  - Arrival and departure responses now consistently include a zero-valued last update time when no update is available.
- **Tests**
  - Added coverage for single-stop and multi-arrival responses without real-time data, including validation of zero-valued prediction fields.


Ahmedhossamdev:
LGTM. I compared against Tampa's hosted Java on the same HART feed for stop `1_1013`, trip `1_2008465`: both endpoints on maglev now return `predicted=false, predictedArrivalTime=0, predictedDepartureTime=0, lastUpdateTime=0`, which matches Java exactly. The Java behavior comes from `ArrivalsAndDeparturesBeanServiceImpl.applyBlockLocationToBean` gating the `setPredicted*Time` calls on `isPredictedArrivalTimeSet()`, so removing the scheduled-time pre-seed in the singular handler is the right fix.

One optional nit inline on each new test. No blockers.

Separately, I noticed a related issue in the opposite direction (`predicted=false` on maglev when Java says `true` and maglev's own `tripStatus.scheduleDeviation` shows the trip is being tracked). Filed as #1456 — not something to fold into this PR.

internal/restapi/arrival_and_departure_for_stop_handler_test.go
```diff
@@ -401,6 +401,35 @@ func TestGetPredictedTimes_NoRealTimeData(t *testing.T) {
 	assert.False(t, predicted)
 }
 
+func TestArrivalAndDepartureForStop_NoRealTimeDataUsesZeroPredictionTimes(t *testing.T) {
+	mockClock := clock.NewMockClock(time.Date(2010, 1, 1, 8, 2, 0, 0, time.UTC))
+	api := createTestApiWithClock(t, mockClock)
+	defer api.Shutdown()
+	api.GtfsManager.MockResetRealTimeData()
```
Ahmedhossamdev: Optional nit. `setupDelayPropTestData` does not add any real-time data, so the pre-setup `MockResetRealTimeData()` on line 408 is a no-op. The `t.Cleanup(...)` on the next line is enough on its own.

```suggestion
	defer api.Shutdown()
	t.Cleanup(api.GtfsManager.MockResetRealTimeData)
```

internal/restapi/arrivals_and_departures_for_stop_handler_test.go
```diff
@@ -842,6 +842,31 @@ func TestPluralArrivals_NoMatchingOrPriorStop(t *testing.T) {
 	assert.True(t, found, "expected to find arrival for trip %s", expectedTripID)
 }
 
+func TestPluralArrivals_NoRealTimeDataUsesZeroPredictionTimes(t *testing.T) {
+	mockClock := clock.NewMockClock(time.Date(2010, 1, 1, 8, 2, 0, 0, time.UTC))
+	api := createTestApiWithClock(t, mockClock)
+	defer api.Shutdown()
+	api.GtfsManager.MockResetRealTimeData()
```
Ahmedhossamdev: Same optional nit as the sibling test. `setupDelayPropTestData` does not add any real-time data, so the pre-setup `MockResetRealTimeData()` on line 849 is a no-op. The `t.Cleanup(...)` on the next line is enough.

```suggestion
	defer api.Shutdown()
	t.Cleanup(api.GtfsManager.MockResetRealTimeData)
```

omlahore:
I ran these on main. The singular prediction test and the models one both fail there, so I think this does what it says.

The plural one passes on main though. LastUpdateTime.IsZero() is true whether the field comes back as 0 or gets left out, so it doesn't really cover the lastUpdateTime half of #1441. Asserting on the JSON the way you did in the models test would catch it.

PR #1460, opened 2026-09-19
Fall back to tripStatus deviation for arrival prediction

### LINKED ISSUE #1456: arrival-and-departure endpoints: predicted flag stays false when trip has real-time data
## Summary

On both `arrivals-and-departures-for-stop` and `arrival-and-departure-for-stop`, Maglev returns `predicted: false` and `predictedArrivalTime: 0` for arrivals whose trip is being tracked in real time, when the TripUpdate does not carry a per-stop update matching (or earlier than) the queried stop. Java returns `predicted: true` for the same arrival.

Maglev's own `tripStatus.predicted` and `tripStatus.scheduleDeviation` on the same response show that Maglev has the real-time data. It just doesn't propagate it to the arrival-level `predicted` field.

All comparisons below ran on 2026-09-17 against Java OneBusAway at `api.tampa.onebusawaycloud.com` and Maglev `main` at `6924cc75`. Both loaded the HART feed.

## Current behaviour

`arrivals-and-departures-for-stop/1_1013.json?minutesBefore=40&minutesAfter=120`:

| Trip | Field | Java | Maglev |
|---|---|---|---|
| `1_2008464` | `predicted` | true | **false** |
| `1_2008464` | `predictedArrivalTime` | 1789666092000 (schedArr + 174s) | **0** |
| `1_2008464` | `predictedDepartureTime` | 1789666092000 | **0** |
| `1_2008464` | `tripStatus.predicted` | true | true |
| `1_2008464` | `tripStatus.scheduleDeviation` | 174 | 174 |
| `1_2008464` | `vehicleId` | `1_2141` | `1_2141` |
| `1_2008464` | `lastUpdateTime` | 1789666937000 | 1789666937000 |

Same divergence on `1_2008467` (Maglev deviation 71s) and `1_2008468` (Maglev deviation 174s) at the same stop.

A client checking `arrival.predicted` will treat these RT-tracked buses as schedule-only, and `predictedArrivalTime == 0` will keep delay-aware UI code from applying the deviation Maglev has already computed on the same object.

## Why

Maglev uses two independent code paths for "is this predicted":

- `tripStatus.predicted` and `tripStatus.scheduleDeviation` come from `BuildTripStatus`, which considers a trip predicted whenever any real-time source exists for it.
- The arrival-level `predicted` flag and `predictedArrivalTime` come from `getPredictedTimes` in `internal/restapi/arrival_and_departure_for_stop_handler.go:626`, which returns `false` unless the TripUpdate has (a) a stop-time update matching the queried stop, (b) a per-stop update for a prior stop (propagated delay), or (c) a trip-level `Delay`.

For trips where the TripUpdate only carries updates for stops past the queried one (and no trip-level delay), `getPredictedTimes` finds nothing and returns false, even though `BuildTripStatus` on the very same request has already computed a schedule deviation from the vehicle position.

Java uses one flag (`blockLocation.isPredicted()`, see `ArrivalsAndDeparturesBeanServiceImpl.applyBlockLocationToBean` at line 450) for both trip-status and arrival-level predictedness, so the two stay in sync.

## Expected behaviour

When Maglev has real-time data for a trip (`tripStatus.predicted=true` or a valid `scheduleDeviation`), the arrival-level `predicted` field should also be `true`, and `predictedArrivalTime` / `predictedDepartureTime` should be `scheduledArrivalTime + scheduleDeviation` / `scheduledDepartureTime + scheduleDeviation` when no per-stop prediction is available.

## Acceptance criteria

- [ ] When `tripStatus.predicted` is true for an arrival, the top-level `predicted` field is also true.
- [ ] When only trip-level schedule deviation is known (no per-stop update for the queried stop or an earlier one), `predictedArrivalTime` and `predictedDepartureTime` are `scheduledArrivalTime + scheduleDeviation` and `scheduledDepartureTime + scheduleDeviation`.
- [ ] The singular endpoint (`arrival-and-departure-for-stop`) behaves identically.
- [ ] The existing test `TestPluralArrivals_VehiclePositionAloneDoesNotPredict` is updated or removed to reflect the new behavior.
- [ ] Tests assert this behavior for a schedule-deviation-only arrival on both endpoints.

Related: #1441

Ahmedhossamdev:
Fixes: #1456

## Bug

The same response could disagree with itself. `tripStatus.predicted` said `true` and `tripStatus.scheduleDeviation` was populated, but the arrival's own `predicted` field said `false` with `predictedArrivalTime=0`. Java doesn't do this: it uses one flag for both.

Concrete case from the issue, stop `1_1013`, trip `1_2008464`:

| Field | Java | Maglev before |
|---|---|---|
| `tripStatus.predicted` | true | true |
| `tripStatus.scheduleDeviation` | 174 | 174 |
| `predicted` (top-level) | true | **false** |
| `predictedArrivalTime` | schedArr + 174s | **0** |

A client filtering on `arrival.predicted` treated a real-time-tracked bus as schedule-only. A client relying on `predictedArrivalTime > 0` fell back to the scheduled time even though the deviation was right there on the same object.

## Why

Two independent code paths were computing "is this predicted":

- `getPredictedTimes` (per-stop): true only if the TripUpdate has an STU for this stop, an earlier stop, or a trip-level `Delay`.
- `BuildTripStatus` (whole-trip): true whenever any real-time signal exists (fresh vehicle OR any StopTimeUpdate anywhere in the block).

For HART trips whose TripUpdates only carry per-stop entries for stops past the queried one, `BuildTripStatus` still resolved a `ScheduleDeviation` via its closest-in-time STU fallback, but `getPredictedTimes` returned false. The response then contradicted itself.

Java uses one signal for both: `blockLocation.isPredicted()` drives the arrival's `predicted` flag, and `setPredictedTimesFromScheduleDeviation` fills the predicted times as `scheduled + deviation` when the per-stop path (`setPredictedTimesFromTimepointPredictionRecords`) fails. See `ArrivalsAndDeparturesBeanServiceImpl.applyBlockLocationToBean` and `ArrivalAndDepartureServiceImpl.setPredictedTimesFromScheduleDeviation`.

## Fix

Add a fallback in both handlers: when `getPredictedTimes` returns `false` but `tripStatus.Predicted` is `true`, mark the arrival predicted and set predicted times to `scheduled + tripStatus.ScheduleDeviation`.

- New helper `predictedTimesFromTripStatus` in `trips_helper.go`.
- Wired into `buildArrival` in `arrivals_core.go` (plural handler) and the singular handler.
- Two existing tests codified the old strict behavior (`TestPluralArrivals_NoMatchingOrPriorStop`, `TestPluralArrivals_VehiclePositionAloneDoesNotPredict`); flipped both to assert the new behavior.
- Added two `_DownstreamOnlySTUPredictsFromDeviation` tests (one per handler) that reproduce the exact issue scenario.

## Verification

Live diff against Tampa hosted Java (same HART feed) for stop `1_1013`:

- Before: `predicted=false, predictedArrivalTime=0` on the three tracked trips at that stop.
- After (locally rebuilt): `predicted=true, predictedArrivalTime=scheduled+deviation`, matching Java.

## Summary by CodeRabbit

* **Bug Fixes**
  * Arrival and departure predictions now remain available when real-time updates only affect downstream stops.
  * Schedule deviations are applied consistently to calculate predicted times.
  * Vehicle-position updates can now provide schedule-based predictions when no explicit prediction is available.
  * Existing arrival and departure values are preserved when no valid fallback prediction can be calculated.


omlahore (commented):
I pulled this and ran make test and go vet, both passed. The part that worried me at first was dropping the if !predicted zeroing in arrivals_core, but I think it's fine since every false path in getPredictedTimes already returns zero times anyway. The one I'm not totally sure about is the vehicle position only case, where predicted comes back true and predictedArrivalTime is just the scheduled time because the deviation is unknown, not a real zero, so a client would read it as on time, I guess Java does the same so it's probably on purpose but is that what you want here?

PR #1463, opened 2026-09-20
Return JSON current-time when GTFS is unready

### LINKED ISSUE #1457: current-time: unready GTFS returns plain-text 503; envelope currentTime and entry.time can disagree
## Summary

`GET /api/where/current-time.json` diverges from its wiki spec in two places, both in `internal/restapi/current_time_handler.go`. No other endpoint work is attached to this; `/healthz` already owns readiness.

## 1. Unready GTFS returns plain-text HTTP 503

Wiki (Minimal Guarantees): the response always carries HTTP 200 and a JSON envelope with `version`, `code`, `text`, and `currentTime`.

Maglev:

```go
if !api.GtfsManager.IsReady() {
    http.Error(w, "Service Unavailable: GTFS data invalid", http.StatusServiceUnavailable)
    return
}
```

That is a non-JSON body and a status the OpenAPI spec does not document for this path. `/healthz` already returns JSON 503 while GTFS is indexing. This endpoint only needs the server clock; `agencyTimezone` already falls back to UTC when agencies cannot be loaded.

A client that always parses the OBA envelope will fail during startup instead of getting a usable server time.

## 2. Envelope `currentTime` and `data.entry.time` are sampled separately

Wiki suspected defect 2: Java samples `data.entry.time` in the action and `currentTime` later when the envelope is built, so the two "now" fields can disagree by a few milliseconds.

Maglev still does this:

```go
timeData := models.NewCurrentTimeDataInLocation(api.Clock.Now(), loc)
response := models.NewOKResponse(timeData, api.Clock) // samples Clock.Now() again
```

Maglev already fixed suspected defect 1 (`readableTime` uses the agency timezone). Defect 2 is still present.

## Acceptance criteria

- [ ] When GTFS is not ready, `current-time` still returns HTTP 200 and the standard JSON envelope (clock / UTC fallback). Readiness stays on `/healthz`.
- [ ] `currentTime` and `data.entry.time` are the same instant
- [ ] `readableTime` timezone behaviour is unchanged
- [ ] Tests cover the unready path and the single-sample invariant

tejasva-vardhan:
## Summary
- `current-time` no longer returns a plain-text HTTP 503 while GTFS is indexing. The server clock does not depend on a loaded feed, so this path always returns HTTP 200 and the OBA JSON envelope. Readiness stays on `/healthz`.
- Envelope `currentTime` and `data.entry.time` are now stamped from the same clock sample, so they cannot disagree by a few milliseconds.

## Spec discrepancy
Most `/api/where` handlers fail closed when GTFS is unready. This endpoint is the exception on purpose: the wiki's current-time guarantees are HTTP 200 + JSON envelope. `/healthz` still reports 503 during indexing.

## Test plan
- [ ] `GET /api/where/current-time.json` while GTFS is unready returns HTTP 200 JSON, not 503
- [ ] `currentTime` equals `data.entry.time`
- [ ] `readableTime` timezone behaviour is unchanged
- [ ] `/healthz` still returns 503 while GTFS is indexing

Fixes #1457

## Summary by CodeRabbit

* **Bug Fixes**
  * Current-time responses are now returned even when transit data is unavailable.
  * Current-time responses use a consistent timestamp across the response and its data.
  * Readable time values now fall back to UTC when transit or database information is unavailable.
* **Tests**
  * Added coverage for unavailable transit data, UTC fallback behavior, and consistent timestamp handling.


Ahmedhossamdev:
LGTM @tejasva-vardhan. Both defects reproduce on `main`, the fix is minimal, and the `advancingClock` test with `clk.n == 1` is the right way to pin the single-sample invariant. Nice work. Just address the coderabbot reviews and I'll merge.

tejasva-vardhan:
@Ahmedhossamdev thanks for the review. i have addressed coderabbit reviews . looking forward for your check.

Ahmedhossamdev:
LGTM @tejasva-vardhan. Handler matches Java (no readiness check) and the wiki (single sample for `currentTime` and `entry.time`), and the follow-up commits cover the nil-GTFS branches in `agencyTimezone`. Ready to merge.

PR #1466, opened 2026-09-20
fix: preserve actual HTTP response status in middleware

### LINKED ISSUE #1465: fix: preserve actual HTTP response status in middleware
## Problem

The HTTP response wrappers used by request logging and metrics can record a status code that the client did not actually receive.

Both `responseWriter.WriteHeader` and `metricsResponseWriter.WriteHeader` overwrite their stored status code every time they are called.

For example:

```go
w.WriteHeader(http.StatusNotFound)
w.WriteHeader(http.StatusInternalServerError)

### PR DESCRIPTION
**Pull request by hardik-agarwal18**
## What

Fix HTTP response status tracking in the request logging and metrics middleware.

Both response-writer wrappers previously overwrote the recorded status code when
`WriteHeader` was called more than once, even though Go only honors the first
status.

They also did not explicitly record the implicit `200 OK` committed by the
first `Write`.

## Changes

- Preserve the first HTTP status code in `responseWriter`
- Preserve the first HTTP status code in `metricsResponseWriter`
- Record implicit `200 OK` when the response body is written first
- Add regression tests for both middleware wrappers

## Validation

- `go test -tags "sqlite_fts5 sqlite_math_functions" ./internal/restapi` ✅

Closes #1465

### CODE REVIEW

**Review by Ahmedhossamdev — COMMENTED** [review]
Great work @hardik-agarwal18. The fix seems correct to me.
 
Only two things worth fixing before merging:

The two wrappers are near-identical after this change so extracting a shared statusCapturingWriter and having both middlewares embed it would prevent them drifting again. And the four new tests have the same structure, could be one table-driven test per file, or one total if you pull the wrapper out.

File: internal/restapi/metrics_middleware.go [hunk 1]
```diff
@@ -48,10 +48,24 @@ func MetricsHandler(m *metrics.Metrics) func(http.Handler) http.Handler {
 // metricsResponseWriter wraps http.ResponseWriter to capture status code.
 type metricsResponseWriter struct {
 	http.ResponseWriter
-	statusCode int
+	statusCode  int
+	wroteHeader bool
```
Ahmedhossamdev: This and `responseWriter` at `request_logging_middleware.go:12` are identical after this change. A shared `statusCapturingWriter` embedded by both middlewares would keep them from drifting the next time either one gets touched.

internal/restapi/metrics_middleware_test.go
```diff
@@ -168,3 +168,35 @@ func TestMetricsHandler_VariousStatusCodes(t *testing.T) {
 		})
 	}
 }
+
+func TestMetricsResponseWriter_PreservesFirstStatus(t *testing.T) {
```
Ahmedhossamdev: Same shape as the two tests in `request_logging_test.go`. Table-driven would tighten this up, especially if the wrapper gets extracted.

hardik-agarwal18:
Thanks for the review @Ahmedhossamdev !

Agreed. I'll extract the shared status-capturing writer and simplify the
duplicated tests in a follow-up commit.

hardik-agarwal18:
Thanks for the CodeRabbit review! I’ve addressed the actionable feedback.

### Updates

* Added a shared `statusCapturingWriter` to avoid duplicated response-writer implementations.
* Preserved the first final HTTP status code.
* Added handling for informational `1xx` responses, including `103 Early Hints`, while treating `101 Switching Protocols` as final.
* Added regression tests for:

  * explicit status followed by `Write`
  * `103 Early Hints` followed by a final status
  * `103 Early Hints` followed by `Write`
  * implicit `200 OK`
* Added documentation comments for the new helper functions.

### Validation

```text
go test -tags "sqlite_fts5 sqlite_math_functions" ./internal/restapi
✅ PASS

go test -tags "sqlite_fts5 sqlite_math_functions" ./...
✅ PASS
```

Thanks again for the detailed review!

internal/restapi/status_capturing_writer.go
```diff
@@ -0,0 +1,44 @@
...
+
+// WriteHeader records the first final HTTP status while allowing informational
+// 1xx responses to be forwarded without finalizing the response.
+func (w *statusCapturingWriter) WriteHeader(code int) {
+	if w.wroteHeader {
+		return
+	}
+
+	if code >= http.StatusContinue && code < http.StatusOK &&
+		code != http.StatusSwitchingProtocols {
```
burma-shave, line 27: Please add a regression test for `WriteHeader(http.StatusSwitchingProtocols)`. The implementation intentionally excludes `101 Switching Protocols` from the non-final informational-status pass-through, so 101 is treated as the first final status. Current tests cover `103 Early Hints`, but they do not assert this 101 behavior. Without that test, a future simplification of this branch to treat all 1xx statuses as non-final would pass the suite while regressing the intended 101 handling. A small test should assert that `WriteHeader(http.StatusSwitchingProtocols)` captures/emits 101 and prevents a later final status or `Write` from changing it.

burma-shave (changes requested):
Requesting changes for one small regression-test coverage gap found in final review. The shared status-capturing writer behavior appears correct, and the prior human/CodeRabbit feedback appears addressed, but the new special-case branch for 101 Switching Protocols should be covered before merge.

Ahmedhossamdev (commented):
Thanks @hardik-agarwal18, the shared writer and the table-driven test cover what I asked for, and the 101 case is in. LGTM.

burma-shave (approved):

